PR #654, opened 2013-07-10
add XMLHttpRequest support (and cookie passing support)

dai-shi:
reference: https://github.com/tmpvar/jsdom/issues/380


chicoxyzzy:
is it enabled by default? if yes how to set xhr to no-op?

Sebmaster:
Implement a [custom resource loader](https://github.com/tmpvar/jsdom#custom-external-resource-loader).

PR #963, opened 2014-11-24
Add partial support for HTMLElement.dataset

jeffcarp:
As brought up by issue https://github.com/tmpvar/jsdom/issues/961.

This PR supports getting and setting existing `data-*` attributes. From a brief conversation on IRC it seems like implementing support for adding new `data-*` attributes would be nearly impossible until `Object.observe` is available. Not sure if a partial feature is merge-able, but submitting the PR anyway to open the conversation. 

Supporting HTMLElement.dataset also requires an implementation of DOMStringMap, which is included here. I was not sure where to put that implementation, so just let me know where the most natural place is for it.

### DIFF
--- a/lib/jsdom/browser/index.js
+++ b/lib/jsdom/browser/index.js
@@ -21,6 +21,22 @@ function matchesDontThrow(el, selector) {
   }
 }
 
+var DOMStringMap = function (attributes) {
+
+  Object.keys(attributes).forEach(function (key) {
+    Object.defineProperty(this, key, {
+      get: function() {
+        return attributes[key]._nodeValue;
+      },
+      set: function(value) {
+        attributes[key]._nodeValue = value;
+        return value;
+      },
+      enumerable: true
+    });
+  }, this);
+};
+
 /**
  * Creates a window having a document. The document can be passed as option,
  * if omitted, a new document will be created.
@@ -117,6 +133,9 @@ exports.createWindow = function(dom, options) {
     this.raise = function(){
       dom.Node.prototype.raise.apply(window.document, arguments);
     };
+
+    // Not sure if this is the right place for this
+    this.DOMStringMap = DOMStringMap;
 
     this.setTimeout = function (fn, ms) { return startTimer(setTimeout, clearTimeout, fn, ms); };
     this.setInterval = function (fn, ms) { return startTimer(setInterval, clearInterval, fn, ms); };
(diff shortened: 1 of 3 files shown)


dsheiko:
Yep without Proxy there is no way to intercept access to arbitrary properties and Proxy is still not available in `nodejs` (v4.2.0). However I need it to unit-test CommonJS modules of in-browser app. So here a workaround which works for data-attributes already set it HTML, but of course ignore any new added as properties of dataset object. The polyfill:
```
 /**
     * Convert data-attr into key
     * data-foo-bar -> fooBar
     * @param {String} val
     * @returns {String}
     */
var _attrToDataKey = function( val ){
      var out = val.substr( 5 );
      return out.split( "-" ).map(function( part, inx ){
        if ( !inx ) {
          return part;
        }
        return part.charAt( 0 ).toUpperCase() + part.substr( 1 );
      }).join( "" );
    },
    /**
     * Produce dataset object emulating behavior of el.dataset
     * @param {Element} el
     * @returns {Object}
     */
    _getNodeDataAttrs = function( el ){
      var i = 0,
          atts = el.attributes,
          len = atts.length,
          map = [],
          proxy = Object.create({}),
          attr,
          datakey;
        for ( ; i < len; i++ ){
          attr = atts[ i ].nodeName;
          if ( attr.indexOf( "data-" ) === 0 ) {
            datakey = _attrToDataKey( attr );
            map[ datakey ] = atts[ i ].nodeValue;
            Object.defineProperty( proxy, datakey, {
              get: function() {
                return map[ datakey ];
              },
              set: function ( val ) {
                el.setAttribute( attr, val );
              }
            });
          }
        }
        return proxy;
    };

Object.defineProperty( global.window.Element.prototype, "dataset", {
  get: function() {
    return _getNodeDataAttrs( this );
  }
});
```
Just include it in the callback when the DOM is created. Usage example:
```
jsdom.env({
      html: html,
      done: function ( err, window ) {
        global.window = window;
        require( "./polyfill-jsdom" ) );

        // in the HTML <body data-bind="old value">
        var el = window.document.body;
        console.log( el.dataset.bind ); // old value
        el.dataset.bind = "new value";
        console.log( el.dataset.bind ); // new value, <body data-bind="new value">
        console.log( el.getAttribute( "data-bind" ) ); // new value

      }
    });
```

jaywcjlove:
@dsheiko  According to your method of errors occurred.

```
"before all" hook:
     TypeError: Cannot redefine property: dataset
```

dsheiko:
@jaywcjlove sorry it was buggy, now is claimed fixed (see the changes in my previous comment)
tested with Node v4.2.0 and jsdom@7.0.1

jaywcjlove:
@dsheiko `global.window.Element` is not this `prototype` method.I have tested in Node v4.2.0 and jsdom@7.0.1.

dsheiko:
That's just weird. Pls, check on your side, when you receive `window` object from jsdom (as a parameter of done callback) there must be `window.Element` interface defined

* Please download the GIST https://gist.github.com/dsheiko/37e4673cce20d5896510
* unpack the files to an arbitrary directory
* run `npm i`
* run `node test.js`

On Linux (Ubuntu) it works fine. I will check later on MacOS and Windows, but supposed to be no difference.

jaywcjlove:
@dsheiko Thank you, By your method,I solved the problem.

PR #990, opened 2014-12-28
LOTS of spec-compliance updates

domenic:
This all started by wanting to do something simple by updating DocumentType to the modern DOM Standard ... little did I know how deep the rabbit hole went. Putting this up before going to sleep in case you want to take a look, @Sebmaster. Help appreciated with the four failing tests also...

- Updated `DOMImplementation` to mostly work per-spec, including removing/neutering public feature-related methods and fixing up the createDocument and createHTMLDocument methods
- Re-did the `CharacterData` implementation to follow the algorithms in the DOM Standard; this notably removes a few exceptions
- Re-did `Comment`, `Text`, and `ProcessingInstruction` to follow the DOM Standard and derive from `CharacterData`
- Re-did `DocumentType` to follow the DOM Standard and be much simpler
- Fixed a variety of accessors on `Node` and `Element`; some were removed that were nonstandard (especially setters); others were updated to reflect the spec; etc.
- Removed all traces of entities, entity types, notations, default attributes, and CDATA sections
- Re-did name/qname validation to work with the xml-name-validator package and some centralized algorithms
- Made the XML parser at least somewhat aware of processing instructions

As part of this a *lot* of tests had to be updated. A few common themes:

- Try to have documents be parsed via the public API, instead of creating them using internal constructors
- Properly test XML as XML, in the XML parsing mode
- Remove a bunch of tests related to obsolete things
- Try to update attribute-related tests to reference `name` instead of `nodeName`, since eventually we want to get rid of `nodeName` for attributes

Still to-do: four somewhat-mysterious test failures


domenic:
@Sebmaster this is ready for another review... I amended some stuff into the original commit. Also I updated the submodule and it's working locally but failing on Travis? Submodules :-/.

PR #991, opened 2015-01-01
Add virtual console

jeffcarp:
(via #971)

#### Adds:

- `jsdom.getVirtualConsole`, for fetching an internal console emitter
- `virtualConsole.sendTo`, for forwarding internal output to any console

#### Design decisions:

1. `sendTo` requires its argument to only have at least one valid console method. I did this so that you could pick and choose which methods you wanted to proxy. Let me know if you think it should be fully compliant.
1. Kept the `window.raise` calls in the original console handlers (it looks like `'error'` is [the only method that has a side effect](https://github.com/jeffcarp/jsdom/blob/virtual-console/lib/jsdom/level1/core.js#L969)  in `raise`).

#### Questions:

1. The way I accomplished keeping console events separate between windows doesn't feel super elegant. If anything comes to mind in this area I'd love to hear how to improve it.
1. This PR doesn't replicate any console functionality (like `console.count` having internal state). Is that a needed feature?

~~[edit] jshint fixes forthcoming...~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,6 +330,30 @@ serializeDocument(doc) === "<!DOCTYPE html><html><head></head><body>hello</body>
 doc.documentElement.outerHTML === "<html><head></head><body>hello</body></html>";
 ```
 
+### Capturing Console Output
+
+#### Forward a window's console output to the Node.js console
+
+```js
+var jsdom = require("jsdom");
+var window = jsdom.jsdom().parentWindow;
+
+jsdom.getVirtualConsole(window).sendTo(console);
+```
+
+#### Get an event emitter for a window's console
+
+```js
+var jsdom = require("jsdom");
+var window = jsdom.jsdom().parentWindow;
+
+var virtualConsole = jsdom.getVirtualConsole(window);
+
+virtualConsole.on("log", function (message) {
+  console.log("console.log called ->", message);
+});
+```
+
 ## What Standards Does jsdom Support, Exactly?
 
 Our mission is to get something very close to a headless browser, with emphasis more on the DOM/HTML side of things than the CSS side. As such, our primary goals are supporting [The DOM Standard](http://dom.spec.whatwg.org/) and [The HTML Standard](http://www.whatwg.org/specs/web-apps/current-work/multipage/). We only support some subset of these so far; in particular we have the subset covered by the outdated DOM 2 spec family down pretty well. We're slowly including more and more from the modern DOM and HTML specs, including some `Node` APIs, `querySelector(All)`, attribute semantics, the history and URL APIs, and the HTML parsing algorithm.
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -92,11 +92,7 @@ function Window(options) {
     clear: wrapConsoleMethod("clear"),
     count: wrapConsoleMethod("count"),
     debug: wrapConsoleMethod("debug"),
-    error: function (message) {
-      var args = Array.prototype.slice.call(arguments);
-      window._virtualConsole.emit.apply(window._virtualConsole, ["error"].concat(args));
-      window.raise("error", message);
-    },
+    error: wrapConsoleMethod("error"),
     group: wrapConsoleMethod("group"),
     groupCollapse: wrapConsoleMethod("groupCollapse"),
     groupEnd: wrapConsoleMethod("groupEnd"),
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -9,6 +9,7 @@ var createFrom = require("../utils").createFrom;
 var Location = require("./location");
 var History = require("./history");
 var toFileUrl = require("../utils").toFileUrl;
+var VirtualConsole = require("../virtual-console");
 
 var cssSelectorSplitRE = /((?:[^,"']|"[^"]*"|'[^']*')+)/;
 
@@ -77,34 +78,35 @@ function Window(options) {
   this.clearTimeout = stopTimer.bind(this, window);
   this.__stopAllTimers = stopAllTimers.bind(this, window);
 
+  this._virtualConsole = new VirtualConsole();
+
+  function wrapConsoleMethod(method) {
+    return function () {
+      var args = Array.prototype.slice.call(arguments);
+      window._virtualConsole.emit.apply(window._virtualConsole, [method].concat(args));
+    };
+  }
+
   this.console = {
-    assert: function (expression, message) {
-      if (!expression) {
-        window.raise("error", "Assertion failed: " + message);
-      }
-    },
-    clear: function () {},
-    count: function () {},
-    debug: function () {},
+    assert: wrapConsoleMethod("assert"),
+    clear: wrapConsoleMethod("clear"),
+    count: wrapConsoleMethod("count"),
+    debug: wrapConsoleMethod("debug"),
     error: function (message) {
+      var args = Array.prototype.slice.call(arguments);
+      window._virtualConsole.emit.apply(window._virtualConsole, ["error"].concat(args));
       window.raise("error", message);
     },
-    group: function () {},
-    groupCollapse: function () {},
-    groupEnd: function () {},
-    info: function (message) {
-      window.raise("info", message);
-    },
-    log: function (message) {
-      window.raise("log", message);
-    },
-    table: function () {},
-    time: function () {},
-    timeEnd: function () {},
-    trace: function () {},
-    warn: function (message) {
-      window.raise("warn", message);
-    }
+    group: wrapConsoleMethod("group"),
+    groupCollapse: wrapConsoleMethod("groupCollapse"),
+    groupEnd: wrapConsoleMethod("groupEnd"),
+    info: wrapConsoleMethod("info"),
+    log: wrapConsoleMethod("log"),
+    table: wrapConsoleMethod("table"),
+    time: wrapConsoleMethod("time"),
+    timeEnd: wrapConsoleMethod("timeEnd"),
+    trace: wrapConsoleMethod("trace"),
+    warn: wrapConsoleMethod("warn")
   };
 
   this.XMLHttpRequest = function () {
--- a/lib/jsdom/virtual-console.js
+++ b/lib/jsdom/virtual-console.js
@@ -1,0 +1,24 @@
+"use strict";
+
+var EventEmitter = require("events").EventEmitter;
+var utils = require("./utils");
+
+function VirtualConsole() {
+  // If "error" event has no listeners,
+  // EventEmitter throws an exception
+  this.on("error", function () {});
+}
+
+utils.inheritFrom(EventEmitter, VirtualConsole, {
+  sendTo: function (anyConsole) {
+    Object.keys(anyConsole).forEach(function (method) {
+      if (typeof anyConsole[method] === "function") {
+        this.on(method, function () {
+          anyConsole[method].apply(anyConsole, arguments);
+        });
+      }
+    }, this);
+  }
+});
+
+module.exports = VirtualConsole;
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -330,6 +330,30 @@ serializeDocument(doc) === "<!DOCTYPE html><html><head></head><body>hello</body>
 doc.documentElement.outerHTML === "<html><head></head><body>hello</body></html>";
 ```
 
+### Capturing Console Output
+
+##### Forward a window's console output to the NodeJS console
```
domenic: Too many `#`s. "Node.js", not "NodeJS".

lib/jsdom/virtual-console.js
```diff
@@ -0,0 +1,80 @@
...
+var vConsole = {};
+
+var emitters = {};
+var windows = [];
+
+// Returns an EventEmitter that emits on any console events
+vConsole.getVirtualConsole = function(window) {
+  var virtualConsole = new EventEmitter();
+
+  saveEmitterForWindow(virtualConsole, window);
```
domenic: The design here is way too clever, and without weak maps will leak memory. Just do `window._virtualConsole = virtualConsole`. Probably if it already exists, return it, instead of creating a new one.

lib/jsdom/virtual-console.js
```diff
@@ -0,0 +1,80 @@
...
+
+  virtualConsole.sendTo = function(anyConsole) {
```
domenic: This should be a prototype method on a `VirtualConsole` class (that derives from `EventEmitter`) instead of having it be re-created every time.

lib/jsdom/virtual-console.js
```diff
@@ -0,0 +1,80 @@
...
+
+    if (!isValidConsole(anyConsole, window)) {
```
domenic: No need for this test IMO. If it doesn't have any console methods they'll get errors later, closer to the source.

lib/jsdom/virtual-console.js
```diff
@@ -0,0 +1,80 @@
...
+      var msg = "sendTo was not passed a compliant console";
+      throw new Error(msg);
+    }
+
+    var vc = vConsole.getVirtualConsole(window);
```
domenic: You can just use `this` instead, when this becomes a method on `VirtualConsole`.

lib/jsdom/virtual-console.js
```diff
@@ -0,0 +1,80 @@
...
+    Object.keys(anyConsole).forEach(function(method) {
+      if (typeof anyConsole[method] === "function") {
+        vc.on(method, anyConsole[method]);
```
domenic: Node.js console methods are functions, i.e. don't require `this`-binding. But in general that's not a good assumption. So instead, `vc.on(method, function () { anyConsole[method].apply(anyConsole, arguments); })`

lib/jsdom/virtual-console.js
```diff
@@ -0,0 +1,80 @@
...
+      }
+    });
+  };
+
+  return virtualConsole;
+};
+
+// Window uses this to send events to subscribed emitters
+vConsole.fireEvent = function(window, method, args) {
+  args = [method].concat(args);
```
domenic: If there's only one virtual console per window this should be easier. `window._virtualConsole.emit.apply([method].concat(args))`

lib/jsdom/browser/Window.js
```diff
@@ -77,32 +78,43 @@ function Window(options) {
   this.clearTimeout = stopTimer.bind(this, window);
   this.__stopAllTimers = stopAllTimers.bind(this, window);
 
+  var wrapConsoleMethod = function(method) {
```
domenic: Use a function declaration not a function expression

lib/jsdom/browser/Window.js
```diff
@@ -77,32 +78,43 @@ function Window(options) {
...
+    return function() {
+      var args = Array.prototype.slice.call(arguments);
+      vConsole.fireEvent(window, method, args);
+    };
+  };
+
   this.console = {
```
domenic: No need to keep the `raise`s. I'd like to get rid of `raise` entirely, actually.

jeffcarp:
Thanks for taking a look at this! I corrected all issues brought up in the above comments, plus:

- Got rid of all `raise`s except the one in `console.error`, which looks like it needs to be there for the [tests to pass](https://github.com/tmpvar/jsdom/blob/master/test/window/console.js#L36).
- Also about `console.error`: it looks like you can't `.emit("error"`, which is keeping me from testing `console.error`. What do you think the behavior should be? This sounds janky, but `console.error` could `.emit("consoleError"`. (Not advocating for that, just can't think of a better idea).
- It wasn't in jshint but I added a space after the `function` keyword in this diff, since it looks like that's the convention

domenic:
> Got rid of all raises except the one in console.error, which looks like it needs to be there for the tests to pass.

We should remove all those tests, and replace them with ones that use the new virtual console API.

> Also about console.error: it looks like you can't .emit("error", which is keeping me from testing console.error. What do you think the behavior should be? This sounds janky, but console.error could .emit("consoleError". (Not advocating for that, just can't think of a better idea).

We can fix this by putting a dummy listener for `"error"` on all newly-created event emitters for this purpose. Probably in the `VirtualConsole` constructor.

> [Edit] test is failing due to possibly faulty assumption - console.error shouldn't actually throw an exception, right?

This is probably because by default EventEmitters will re-throw errors if you `emit("error", ...)` and there are no listeners.

lib/jsdom/browser/Window.js
```diff
@@ -77,34 +78,35 @@ function Window(options) {
   this.clearTimeout = stopTimer.bind(this, window);
   this.__stopAllTimers = stopAllTimers.bind(this, window);
 
+  this._virtualConsole = getVirtualConsole(window);
```
domenic: OK, now that I see how this panned out, a new suggestion to be a bit simpler: have `virtual-console.js` default-export `VirtualConsole`, and then you can just do `this._virtualConsole = new VirtualConsole()`. Then you can inline `exports.getVirtualConsole = function (window) { return window._virtualConsole; }`.

lib/jsdom/virtual-console.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+var VirtualConsole = require("events").EventEmitter;
```
domenic: This doesn't work, because that means on the next line you're adding a new method to all EventEmitter instances. This could be most easily done with the `inheritFrom` util in `jsdom/utils.js`:

```js
function VirtualConsole() {}
inheritFrom(EventEmitter, VirtualConsole, {
  sendTo: function (anyConsole) { ... }
});
```

test/jsdom/virtual-console.js
```diff
@@ -0,0 +1,111 @@
+"use strict";
+
+var jsdom = require("../..");
+var getVirtualConsole = require("../../lib/jsdom/virtual-console");
+var EventEmitter = require("events").EventEmitter;
+
+var consoleMethods = [
+  "assert",
```
domenic: You can add `"error"` to these once we get the dummy listener in place.

domenic:
Starting to look quite nice!

jeffcarp:
Great! It's looking much cleaner now. The [previously failing tests in question](https://github.com/tmpvar/jsdom/blob/master/test/window/console.js) testing `console.error` pass now (I'm still delegating to `window.raise` for `console.error`). It looks like those are testing `window.console` from the outside and shouldn't be updated to use the virtual console interface.

lib/jsdom/browser/Window.js
```diff
@@ -77,34 +78,35 @@ function Window(options) {
   this.clearTimeout = stopTimer.bind(this, window);
   this.__stopAllTimers = stopAllTimers.bind(this, window);
 
+  this._virtualConsole = new VirtualConsole(window);
```
domenic: No params needed

lib/jsdom/virtual-console.js
```diff
@@ -0,0 +1,24 @@
+"use strict";
+
+var EventEmitter = require("events").EventEmitter;
+var utils = require("./utils");
+
+function VirtualConsole () {
```
domenic: No need for space after the function name. My rule is `function<space><name>()`, where if `<name>` is the empty string, that becomes `function<space>()`

domenic:
Sweet! Added some nit comments. The other thing is: I'm planning to kill `raise` entirely. So let's just remove any use of `raise` from `error`, and remove those tests that test `window.document.errors`.

lib/jsdom/browser/Window.js
```diff
@@ -77,34 +78,31 @@ function Window(options) {
...
+    group: wrapConsoleMethod("group"),
+    groupCollapse: wrapConsoleMethod("groupCollapse"),
+    groupEnd: wrapConsoleMethod("groupEnd"),
+    info: wrapConsoleMethod("info"),
+    log: wrapConsoleMethod("log"),
+    table: wrapConsoleMethod("table"),
+    time: wrapConsoleMethod("time"),
+    timeEnd: wrapConsoleMethod("timeEnd"),
+    trace: wrapConsoleMethod("trace"),
+    warn: wrapConsoleMethod("warn")
```
domenic: I think it's fine for now. Eventually when I move this stuff off of the instance and on to the prototype I can consider re-organizing.

domenic:
This is lovely. Amazing pull request, thanks for sticking with it through all the feedback. Docs, tests, the whole works. I will merge it along with #990 this weekend.

jeffcarp:
Awesome! Thanks for working through it with me.

domenic:
Merged!!

jeffcarp:
:fireworks: :tada: :confetti_ball:

reggi:
I can't get this to work and it's driving me nuts:

```
var jsdom = require("jsdom")
var window = jsdom.jsdom('<html><body><script type="text/javascript">console.log(\"hello world\")</script></body></body>').defaultView;
jsdom.getVirtualConsole(window).sendTo(console);
```

http://stackoverflow.com/questions/28788624/jsdoms-getvirtualconsole-example

domenic:
I am pretty sure the event is already sent to the virtual console by the time you call getVirtualConsole in that case.

reggi:
Thanks @domenic I just figured that out :)

reggi:
No support for script tags added to the dom?

```
var argv = require('minimist')(process.argv.slice(2));
var file = argv._[0]

var jsdom = require("jsdom")
var doc = jsdom.jsdom('<html><body></body></body>')
var window = doc.defaultView;

jsdom.getVirtualConsole(window).sendTo(console);

var scriptEl = window.document.createElement("script");
scriptEl.src = file;
window.document.body.appendChild(scriptEl);

var stringDoc = jsdom.serializeDocument(doc)
console.log(stringDoc)
```

How am I supposed to add content?

Sebmaster:
```js
var doc = jsdom.jsdom('<html><body></body></body>')
var window = doc.defaultView;

jsdom.getVirtualConsole(window).sendTo(console);

var scriptEl = window.document.createElement("script");
scriptEl.text = "console.log('abc')";
window.document.body.appendChild(scriptEl);

var stringDoc = jsdom.serializeDocument(doc)
console.log(stringDoc)
```
prints
```
abc
<html><head></head><body><script>console.log('abc')</script></body></html>
```

so looks good to me?

domenic:
RIght, `src = "index.js"` will not work since resolving `"index.js"` relative to `"about:blank"` doesn't work so well.

reggi:
It works!

Yay

```
var fs = require("fs")
var jsdom = require("jsdom")
var argv = require('minimist')(process.argv.slice(2));
var file = argv._[0]
var contents = fs.readFileSync(file)

var doc = jsdom.jsdom('<html><body></body></body>')
var window = doc.defaultView;

jsdom.getVirtualConsole(window).sendTo(console);

var scriptEl = window.document.createElement("script");
scriptEl.text = contents;
window.document.body.appendChild(scriptEl);

var stringDoc = jsdom.serializeDocument(doc)
console.log(stringDoc)
```

PR #993, opened 2015-01-03
Need to append doctype for createHTMLDocument; should appear as first ch...

brettz9:
...ild node per step 3 of https://dom.spec.whatwg.org/#dom-domimplementation-createhtmldocument


domenic:
createHTMLDocument and doctype are being re-done as part of #990. In any case, you can't make a change like this without tests.

PR #1001, opened 2015-01-18
[WIP] Re-do initialization and contextification

domenic:
Using the new vm module in io.js, we can contextify pretty easily and without any native modules.

This also re-does the initialization lifecycle to get rid of the horrible globally-shared modifications.

To-do:

- [ ] Tests run much slower now, possibly because before Window objects were lazily created and now they are created every time. We could make them lazy again.
- [x] A single failing test in window/script.js that has me scratching my head
- [x] A lot of tests related to frames are broken because I re-jiggered some window stuff but didn't bother yet to fix how frames work.
- [x] Fix read-me messaging to prominently point out that we are io.js-only.
- [x] Consider removing parentWindow (nonstandard) in favor of defaultView (standard).
- [x] Make Travis use iojs instead of Node
- [x] <del>Consider while we're at it getting rid of `window.raise` for script errors, in favor of `window.onerror` or similar.</del> postponed to #1039 I think
- [x] Browser tests are failing

Putting this up for review and help fixing the tests if people have spare time.


Sebmaster:
Oh yeah, I had that failing test case in window/script.js in my branch some time ago too, no idea where that's coming from exactly.

Anyways, finally rewriting that init logic.

Re: Lazy generation:
Imo we should think about implementing "browser"-windows properly, which are long living and then create script-context windows, which then load the document. Lazily creating the window and just returning the document kind of seems to run counter to that, although I'm not sure what your plans regarding that are.

domenic:
> Imo we should (long-term) think about implementing "browser"-windows properly, which are long living and then create script-context windows, which then load the document.

Hmm, I kind of have been leaving this to zombie.js so far. I think the distinction here is largely between Window and WindowProxy.

> Lazily creating the window and just returning the document kind of seems to run counter to that, although I'm not sure what your plans regarding that are.

Well my first thought in this PR was that it was silly to have a document without a window, so I should just create the window first and then return the document since that's the API that jsdom is used to exposing. But it looks likely that jsdom is document-first for performance reasons. That is, if you don't run any script, we don't even have to create a global object, which is a nice boost. Unsure how often it applies in the real world though...

Sebmaster:
> Hmm, I kind of have been leaving this to zombie.js so far. I think the distinction here is largely between Window and WindowProxy.

It's kind of problematic that we just don't support `location.href` for example, since that's used a lot, or "clicking" on links, submitting forms, etc.

Re: window:
I haven't looked at the jsdom codebase for some time now, but is creating a window really that expensive? Shouldn't that be just the creation of a few objects? Maybe we're losing a lot of perf there for work we shouldn't actually do at creation time?

domenic:
More investigation is needed. I just noticed the tests running a lot slower after the change. But that could be anything; my guess of it being window creation might be totally wrong.

Sebmaster:
Fixed the frame-related errors in level2/html, however the window/frame testsuite still fails completely.

Sebmaster:
Frame tests now run successfully on my machine (Windows), no idea why they still fail on Travis.

There's an additional failure in jsdom/index and something else still crashes the test runner there. The vm context failure is also still there, I have no idea where to even start with that one.

domenic:
Repro so far for contextify failure:

```js
"use strict";

var vm = require("vm");

function Window() {
    vm.createContext(this);

    this._document = new Document({ defaultView: this });
}

Window.prototype = {
    get document() {
        return this._document;
    }
};

function Document(options) {
    this._defaultView = options.defaultView;
}

Document.prototype = {
    get defaultView() {
        return this._defaultView;
    }
}

var window = new Window();

vm.runInContext("var result = document.defaultView === this;", window);

console.log(window);
```

`window.result` logs as `false`.

Sebmaster:
Even simpler:
```js
"use strict";

var vm = require("vm");

var window = {};
window.window = window;
window.second = { window: window }; // another object boundary
vm.createContext(window);
window.document = { defaultView: window }; // another object boundary

vm.runInContext("var res1 = document.defaultView === this;\
var res2 = window === this;\
var res3 = window.window === this;\
var res4 = window.second.window === this", window);

console.log(window);
```

Seems it occurs if the access happens through another object boundary.

Sebmaster:
I tried switching our own url resolution out in 1f6f4c5, however it looks like nothing in npm really perfectly fits the browser url resolution.

domenic:
It could be that our exhaustive tests (added a while ago) are over-strict. But, I'd prefer to just do the minimal fix necessary for now, and overhaul the URL parser at a later date. Hopefully I can figure out the problem sometime soon...

Sebmaster:
LGTM!

PR #1002, opened 2015-01-19
handle empty string for html

fluffybunnies:
Noticed a script of mine using jsdom was occasionally hanging. The problem was when the explicit html input happened to be empty and created/loaded/done were not calling back.

The unit test I created is kinda crappy, feel free to update it or just remove.

Cheers!


domenic:
Good catch, thanks! Will tweak and merge tomorrow.

domenic:
Merged as de26d5ac0378b4e9cbfc28bae9abbc23df583190, thanks!

PR #1003, opened 2015-01-19
handle null bytes in html

fluffybunnies:
While building a sitemap generator using your lib, I found it not uncommon to receive null bytes in page content. fs.readFile will fail without an error code in this case.


domenic:
This is fixed in io.js to give ENOENT, but I will merge this for the 3.0.x branch and then revert it for 4.0.0x (which is io.js only)

domenic:
Merged as 2a951e43ac1959d76496528f68e2e799c2fbaae4, with a new release soon!

PR #1007, opened 2015-01-21
Only uncheck radios in same form

zpao:
When selecting another radio element, only other radios in the same form should be unchecked.

See http://jsbin.com/daqoge/1/ as an example. This is taken pretty much directly from a test we have in React where we're ensuring we maintain browser behavior. In order to make it pass in jest we actually "fix" this in our (ancient) jsdom env by overridding the setter for `HTMLInputElement.checked` (https://github.com/facebook/jest/blob/master/src/lib/jsdom-compat.js#L245-L284).

I think this is the right fix but there's a lot of code I didn't read. I also didn't write a test - I'll look into that process but if anybody more familiar wants to jump in and help, that would be appreciated.

### DIFF
--- a/lib/jsdom/level2/html.js
+++ b/lib/jsdom/level2/html.js
@@ -1036,7 +1036,8 @@ define('HTMLInputElement', {
       var radios = new core.HTMLCollection(this, core.mapper(root, function(el) {
         return el.type === 'radio' &&
                el.name &&
-               el.name.toLowerCase() === name;
+               el.name.toLowerCase() === name &&
+               el._radioButtonGroupRoot === root;
       }));
 
       radios._toArray().forEach(function(radio) {
--- a/test/level2/html.js
+++ b/test/level2/html.js
@@ -19832,6 +19832,31 @@ exports.tests = {
     });
   },
 
+  radio_group_with_same_name_outside_form: function(test) {
+    // NOTE: this is virtually the same as the radio_group_with_same_name_in_several_forms_work test,
+    // however this test moves the first radio group outside of a form so they are siblings
+    // of the form containing the other radio group.
+    var html = '<div>' +
+        '<input type="radio" name="group1" value="3" checked="checked" id="form1-input1" />' +
+        '<input type="radio" name="group1" value="2" id="form1-input2" />' +
+        '<form>' +
+        '<input type="radio" name="group1" value="1" checked="checked" id="form2-input1" />' +
+        '<input type="radio" name="group1" value="5" id="form2-input2" /></form>' +
+        '</div>';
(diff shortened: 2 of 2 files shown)


domenic:
I am almost certain this was fixed in jsdom 1.4.0. As such I'd appreciate a failing test to show that this is still an issue.

domenic:
Actually it might have been fixed in 1.0.0-pre.6.

zpao:
https://github.com/tmpvar/jsdom/blob/master/test/level2/html.js#L19814-L19833 would lead me to believe you're correct. However that's just testing when you have 2 sibling forms. If I copy & modify that test to match the jsbin, then it fails (will add that to PR in a moment)

zpao:
Note: I did confirm that the new test fails in master without the rest of my patch.

domenic:
Sweet!! We'd better merge this then :D.

Joris-van-der-Wel:
Here are a few quotes from [WHATWG](https://html.spec.whatwg.org/)

__radio button group definition__
> The radio button group that contains an input element _a_ also contains all the other input elements _b_ that fulfill all of the following conditions:
>
> * The input element _b_'s type attribute is in the Radio Button state.
> * Either _a_ and _b_ have the same form owner, or they both have no form owner.
> * Both _a_ and _b_ are in the same home subtree.
> * They both have a name attribute, their name attributes are not empty, and the value of _a_'s name attribute is a compatibility caseless match for the value of _b_'s name attribute.
>
> A document must not contain an input element whose radio button group contains only that element.


[html.js:1036](https://github.com/zpao/jsdom/blob/f7ddb1611b9beb14b92530042e753ecfe42e05f6/lib/jsdom/level2/html.js#L1036) loops over only the descendants of the radio button group (or at least, that was my intention :) ). But it does not strictly follow the spec with nested forms. 
So a check for the form owner should be added. The form owner can be changed by setting the IDL. Read here: https://html.spec.whatwg.org/multipage/forms.html#form-owner

domenic:
Well it would be *lovely* if you wanted to make it follow the spec exactly. Let me know if you're up for that and if so I'll wait on merging.

Joris-van-der-Wel:
Sure, but probably not before this weekend.

It looks pretty involved though, iirc jsdom does not implement setting the form attribute. Following the spec strictly would require tracking form owners explicitly (instead of looping over parentNode).

domenic:
Oh I see, yeah that sounds like a pain. Not worth tying up this PR for.

domenic:
Merged as 3a755591130f9666f2ee19837d3cc86019915836, with 3.0.2 going out shortly!

zpao:
Awesome turnaround, thanks all!

PR #1008, opened 2015-01-23
Add config option for custom resource loader.

### LINKED ISSUE #939: Providing a request field in config to circumvent requests
Essentially I'm thinking jsdom could allow you to provide your own custom request function to replace require("request") (which I think it uses internally but I could very well be wrong?). The benefit of this would be to allow you to use faster or different machinery to fetch resources. For example, on a server you could avoid network chatter by just grabbing resources directly from your database, or a stubbed out cache (this is my intended use). In testing, you could use it to mock without having to monkey-patch. Something like this:

jsdom.env(html, { request: ... });

tobie:
Closes #939.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@ jsdom.env(config);
   - `cookieDomain`: a cookie domain for the manually set cookie; defaults to `127.0.0.1`.
 - `config.headers`: an object giving any headers that will be used while loading the HTML from `config.url`, if applicable
 - `config.features`: see Flexibility section below. **Note**: the default feature set for `jsdom.env` does _not_ include fetching remote JavaScript and executing it. This is something that you will need to _carefully_ enable yourself.
+- `config.resourceLoader`: a function that intercepts subresource requests and allows you to respond with your own content. More below.
 - `config.done`, `config.loaded`, `config.created`: see below.
 
 Note that at least one of the callbacks (`done`, `loaded`, or `created`) is required, as is one of `html`, `file`, or `url`.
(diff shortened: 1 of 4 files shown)


domenic:
This seems pretty reasonable. I'm worried though about the fact that now we need to document and commit to the public API for the until-now-private resourceLoader. It probably would need some cleanup before that's ready---the `fetch` you made seems like a nice start.

Also, I don't like using `this` as an implicit zeroth argument, so we'd want to pass that as the first argument.

tobie:
I don't think there's a lot of use-cases for exposing the API beyond the fetch method itself, tbh.

What about passing the equivalent of a bound `.fetch()` method only, and optionally grouping `url`, `cookie` and `referrer` in an vanilla obj which can be passed to `fetch`, e.g.:

```js
require("jsdom").env({
  url: "http://example.com/",
  resourceLoader: function(fetch, resource, callback) {
    fetch(resource, function(err, body) {
      if (/\.js$/.test(resource.url.href)) {
        body = '"use strict";\n' + body;
      }
      callback(err, body)
    });
  },
  features: // …
});
```

Or even making `fetch` a method of the resource object itself:

```js
require("jsdom").env({
  url: "http://example.com/",
  resourceLoader: function(resource, callback) {
    resource.fetch(function(err, body) {
      if (/\.js$/.test(resource.url.href)) {
        body = '"use strict";\n' + body;
      }
      callback(err, body)
    });
  },
  features: // …
});
```

domenic:
I like the latter a lot. Maybe just call it `defaultFetch` and we're good to go.

domenic:
Also it would be nice to get a test similar to your last example in there.

tobie:
Working on this now, it seems the [whole](https://github.com/tmpvar/jsdom/blob/3a75559/lib/jsdom/level2/html.js#L29) [cookie string](https://github.com/tmpvar/jsdom/blob/3a75559/lib/jsdom/level2/html.js#L111) is sent to the server instead of just the `key=value` pair. This seems like a violation of [RFC6265](http://tools.ietf.org/html/rfc6265#section-5.4).

The fix is simple (replace `document._cookie` by `document.cookie` in [line 29](https://github.com/tmpvar/jsdom/blob/3a75559/lib/jsdom/level2/html.js#L29)). Are you OK with me fixing it in this pull request or would you prefer a separate PR?

tobie:
Went ahead and fixed that here. LMK if that's an issue for you.

domenic:
Lovely, with a test too! Time to merge :D

domenic:
Merged as bea2c9ed012624ec2aa0e3ba04705cfb454414e2 and 87e8d1f46166bb4e60744b80aa6133e1838b870c, and released as version 3.1.0!

Issue #1010, opened 2015-01-26
Byte-order-marks breaking browserify.


sterpe:
There are extraneous byte-order marks at the beginning of lib/jsdom/browser/Window.js and lib/jsdom/browser/documentAdapter

Browserify's bom removal seems to catch the bom in documentAdapter but for some reason does not catch the bom in Window.js

https://github.com/substack/node-browserify/issues/1095

domenic:
Fixed in 3.0.3.

PR #1011, opened 2015-01-26
remove byte-order-marks

sterpe:
Fixes issue #1010, by removing the BOMS.


domenic:
Well, I mean, this is definitely a browserify bug. But then again, I hate BOMs with a passion, and am angry they managed to sneak their way into our source files. So, might as well fix; thanks!

PR #1012, opened 2015-01-27
Use travis container system for faster test startup

Sebmaster:
Only enable container structure for node builds, because it currently fails with browser builds due to errors with the travis tunnel.


domenic:
Good stuff, merged as fbcdc3a39e01bd65425c9cb862cfd7a332cd1328.

PR #1025, opened 2015-02-09
Fix node-canvas integration by extending setAttribute from Element.

robertjustjones:
Using the node-canvas integration with jsdom in a [simple example](https://gist.github.com/robertjustjones/9af52b2eccfbf6187ae1) will issue the following error on toDataURL():

```[Error: the surface type is not appropriate for the operation]```

The root cause is that the canvas width/height do not get set by the setAttribute methods of Element.  The PR bridges and extends certain setters for Element so that both htmltodom and later calls to setAttribute now work. Although admittedly hacky, the fix is in the "transitional" code and applies only when node-canvas is used.  There could be other setters that need to be extended. If there's a suggestion for a cleaner or more complete fix, I'd be happy to test that. You can imagine it took a fair amount of spelunking to get to this version ;-).

### CODE REVIEW

**Comment by domenic** [comment]
So I'm pretty sure we have ways of making this work better ... e.g. adding `width` and `height` getters to the canvas that return the attribute values. But the most important thing is, if you could add a test, then we can play around with different strategies that make the test pass :)

**Comment by robertjustjones** [comment]
Thanks @domenic. I considered trying to add one but since the test requires that node-canvas be installed, I wasn't sure how you would want to structure the "with" and "without" tests. Ie, doing an npm (un)installs in the test scripts.  Suggestions?

In case you missed it, the "simple example" gist in the first post does have a failing case. I could put that in a self-contained repo like https://github.com/robertjustjones/jcanvas-in-node.  In fact, that's two commented lines away from an example case itself.

**Comment by Sebmaster** [comment]
@robertjustjones Just add it to `devDependencies`, those are only installed on full-package installs and not in npm release mode.

**Comment by robertjustjones** [comment]
@domenic I've added two tests that fail as follows without the fix. Let me know if there's a better location for these.

```
AssertionError: A.width
AssertionError: A.height
Error: the surface type is not appropriate for the operation
```

**Comment by robertjustjones** [comment]
To run the tests using io.js requires a recent fix to node-canvas:
``` npm install git+https://github.com/Automattic/node-canvas.git```

**Comment by robertjustjones** [comment]
Hey @domenic do you need me to take this further here for now? The fix could/should go into 3.x and 4.x, but Travis isn't passing because it's not pulling the latest master branch version on node-canvas needed for io.js.

**Comment by domenic** [comment]
Well, we'll need to get a version of node-canvas published to npm first that supports io.js before we can consider taking this.

**Comment by robertjustjones** [comment]
Hmm, ok, but it works now in node.js with your 3.1.x using node-canvas 1.2.1. Having two nodes is really going to suck.

**Comment by domenic** [comment]
node-canvas 1.2.2 was released with support for io.js!

**Comment by robertjustjones** [comment]
@domenic I've updated to node-canvas@1.2.2 now that the npm registry is updated ;-).

Can you help me with why the build fails here? It's working in my environment for iojs-v1.8.1.
```
$ node --version
v1.8.1
$ npm --version
2.8.3
$ nvm --version
0.23.3
```

domenic:
Uh because of the merge commit there are so many changes I can't really diagnose anything. Can you do a rebase on master instead?

domenic:
Oh I see if I go to "files changed" it's more readable. Still a rebase would be great.

Anyway my guess is that you need to preinstall the binary dependencies. So, add stuff like https://github.com/Automattic/node-canvas/wiki/Installation---Ubuntu-and-other-Debian-based-systems to .travis.yml (probably around line 13 so it only is done for the io.js tests)

domenic:
A better fix for this BTW might be to add an _attrModified to HTMLCanvasElement.

domenic:
OK, I managed to fix this! Next release (6.2.1 or 6.3.0) will have the fixes.

robertjustjones:
Awesome, thanks!

PR #1023, opened 2015-02-12
Browserified jsdom cannot execute <script> tags

Sebmaster:
Our contextify-shim.js doesn't even have a run function. I tried to add one, just

```js
  o.run = function (code, filename) {
    o.eval(code + "\n//# sourceURL=" + filename);
  };
```

but this of course doesn't work since it's running the code in the global scope (e.g. in a worker), where variables like `window` and `document` don't work right.

### DIFF
--- a/lib/jsdom/contextify-shim.js
+++ b/lib/jsdom/contextify-shim.js
@@ -1,5 +1,53 @@
+"use strict";
+
+var acorn = require("acorn");
+var findGlobals = require("acorn-globals");
+var escodegen = require("escodegen");
+
 module.exports = function (o) {
   o.getGlobal = function () {
     return o;
+  };
+
+  o.run = function (code, filename) {
+    var comments = [], tokens = [];
+    var ast = acorn.parse(code, {
+      ecmaVersion: 6,
+      allowReturnOutsideFunction: true,
+      ranges: true,
+      // collect comments in Esprima's format 
+      onComment: comments,
+      // collect token ranges 
+      onToken: tokens
+    });
+
+    // make sure we keep comments
+    escodegen.attachComments(ast, comments, tokens);
+
+    var globals = findGlobals(ast);
+    for (var i = 0; i < globals.length; ++i) {
+      if (globals[i].name === "window") {
+        continue;
+      }
+
+      var nodes = globals[i].nodes;
+      for (var j = 0; j < nodes.length; ++j) {
+        var type = nodes[j].type;
+        var name = nodes[j].name;
+        nodes[j].type = "MemberExpression";
+        nodes[j].property = {
+          name: name,
+          type: type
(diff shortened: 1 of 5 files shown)


Sebmaster:
I did a thing, it works, pls comment.

See https://travis-ci.org/tmpvar/jsdom/jobs/50514037 for the proper build with browser test.

domenic:
Wow, looks pretty awesome :D. Maybe add a test for async access within some nested scopes?

lib/jsdom/contextify-shim.js
```diff
@@ -1,5 +1,49 @@
...
+      // collect token ranges 
+      onToken: tokens
+    });
+
+    // make sure we keep comments
+    escodegen.attachComments(ast, comments, tokens);
+
+    var globals = findGlobals(ast);
+    for (var i = 0; i < globals.length; ++i) {
+      if (globals[i].name === "window") continue;
```
domenic: curlies and newline. JSHint should be catching this? Can we JSHint this file now?

package.json
```diff
@@ -65,7 +65,10 @@
     "parse5": ">= 1.3.1 < 2.0.0",
     "request": ">= 2.44.0 < 3.0.0",
     "xml-name-validator": "^1.0.0",
-    "xmlhttprequest": ">= 1.6.0 < 2.0.0"
+    "xmlhttprequest": ">= 1.6.0 < 2.0.0",
+    "acorn-globals": "^1.0.2",
+    "acorn": "~0.11.0",
```
domenic: Pin to 0.11.0 exactly until https://github.com/marijnh/acorn/issues/209 is fixed

test/worker-runner.js
```diff
@@ -22,7 +22,7 @@ self.onmessage = function (e) {
     "level1/html": require("../test/level1/html"), // ok
     "level2/core": require("../test/level2/core"), // ok
     "level2/html": require("../test/level2/html"), // 6/708
```
domenic: Curious if level2/html got any better? Wonder what makes it fail so hard when level1/html works.

Sebmaster: Tried it, but the test run crashes without finishing. Haven't looked that much into it yet. We still need to rework the browser test runner somewhat.

domenic: OK cool, was just curious. And yeah agreed it is kind of crazy.

Sebmaster:
Fixed the nitpicks, but need to get some sleep before I add a test. Will do later tonight though.

Sebmaster:
Added the async test.

domenic:
Fixed, at least somewhat, with 0ec0d0edd9df4747903bfe143d82681f009a4809. Will try to track down more cases as they pop up.

PR #1031, opened 2015-02-18
Start refactoring of events

Sebmaster:
See commit for implementation details. Events are a somewhat easy target because the seperate parts are pretty self-contained.

Since we now have a somewhat recent version of v8 (\o/ finally), we can start refactoring our stuff really nicely with "real" private variables and we should also start to remove the level thing and get some proper prototype-chains in there, since that's currently severly lacking and/or just wrong.

The structure is just some initial draft which we'll probably have to improve incrementally. Comments?

(Since this is based on #1001, the same tests fail, however the refactor didn't break anything... yet.)


lib/jsdom/browser/Window.js
```diff
@@ -339,7 +336,11 @@ module.exports = function Window(options) {
       });
     }
   });
-};
+}
+
+inherits(dom.EventTarget, Window, dom.EventTarget.prototype);
```
domenic: Is the third arg here necessary?

lib/jsdom/browser/Window.js
```diff
@@ -339,7 +336,11 @@ module.exports = function Window(options) {
...
+
+module.exports = Window;
```
domenic: I like this line at the top

lib/jsdom/events/EventException.js
```diff
@@ -0,0 +1,27 @@
...
+    Error.captureStackTrace(this, EventException);
+  }
+}
+
+inheritFrom(Error, EventException, {
+  UNSPECIFIED_EVENT_TYPE_ERR: 0,
+  get code() { return this._code; }
+});
+
+module.exports.EventException = EventException;
```
domenic: `module.exports = EventException;` (no extra property), and at the top

lib/jsdom/browser/Window.js
Sebmaster: Yes, for some reason our inherits doesn't include the methods otherwise, we should maybe have a look at that too.

lib/jsdom/events/EventException.js
```diff
@@ -0,0 +1,27 @@
+﻿"use strict";
```
domenic: This file needs a TODO that this whole concept should probably be killed, as doesn't exist in the spec.

lib/jsdom/events/EventTarget.js
```diff
@@ -0,0 +1,188 @@
...
+
+      if (implementation._hasFeature("ProcessExternalResources", "script")) {
+        if (listeners.indexOf(traditionalHandler) === -1) {
+          listeners.push(traditionalHandler);
+        }
+      }
+    }
+  }
+  return listeners;
+}
```
domenic: Add a newline at end of file, here and elsewhere. (There should be an option in your IDE to always do this.)

lib/jsdom/events/EventTarget.js
```diff
@@ -0,0 +1,188 @@
...
+    dispatchPhase(event, iterator);
+  }
+
+  event._currentTarget = null;
+  event._eventPhase = event.NONE;
+
+  return !event._preventDefault;
+};
+
+module.exports.EventTarget = EventTarget;
```
domenic: Hopefully with the WeakMap -> Symbol move, addAlias can go away, and this can become just `module.exports = EventTarget`

lib/jsdom/level2/events.js
```diff
@@ -9,26 +9,7 @@ var core = require("../level1/core"),
...
-    Error.call(this, this._message);
-    if (Error.captureStackTrace) {
-        Error.captureStackTrace(this, core.EventException);
-    }
-};
-inheritFrom(Error, core.EventException, {
-  UNSPECIFIED_EVENT_TYPE_ERR : 0,
-  get code() { return this._code;}
-});
+core.EventException = require("../events/EventException").EventException;
```
domenic: Any reason this needs to be exposed on `core`?

lib/jsdom/events/EventException.js
domenic: Actually can just make it into a single line `module.exports = function EventException() {`

Sebmaster: We should probably have a unified approach across all objects for this, in case "private" (jsdom-exclusive) methods will be added later. I'd have the object exported as exports[objectName], then all private methods can go on the exports too.

Another approach would be having a $private-Symbol, which we can require everywhere and just set on the constructor with an object which includes our private methods.

lib/jsdom/MultihandleWeakMap.js
```diff
@@ -0,0 +1,49 @@
+﻿"use strict";
+
+var instances = new WeakMap();
+
+function MultihandleWeakMap() {
```
domenic: As discussed over IM we should probably use symbols instead of weak maps for now.

lib/jsdom/level2/events.js
Sebmaster: `core` is `dom` which is how we currently build our window, could move that out while we're at it already but we don't have an infrastructure for that currently. Maybe build some kind of imports-define object and walk through it in the Window constructor?

lib/jsdom/events/EventException.js
domenic: I would rather stay with one class per file in general. If we need specific exceptions we can deal with those (and name them appropriately, e.g. `event-target.js` instead of `EventTarget.js`).

lib/jsdom/level2/events.js
domenic: Right but there is no `window.EventException` in my browser, so it probably shouldn't exist on core/dom. It should just be an internal type to jsdom (until we kill it).

Sebmaster:
Done.

With Symbols we once again can't totally remove public properties since Symbol access apparently differs from property access in v8 and the vm module only forwards property access. However we can reduce all public properties to a single "empty" map.

domenic:
> the vm module only forwards property access.

AAAAARGH :( more vm bugs :( :( :(

Sebmaster:
@domenic Pls fix :p

lib/jsdom/events/EventTarget.js
```diff
@@ -0,0 +1,189 @@
+﻿"use strict";
+
+var EventException = require("./EventException");
+
+var propEvents = Symbol();
+
+module.exports = EventTarget;
+
+function EventTarget() {
+  if (!this.__jsdom) {
```
domenic: Wait you need this even for non-`Window`s? I would just put the symbols directly on `this`...

Sebmaster: No, this is only necessary for windows, but Window inherits from EventTarget. How'd you special case Window?

domenic: Oh I see, I forgot about Window inheriting from EventTarget :-/. Needs a comment then, with a link to the io.js bug.

Sebmaster: Fixed.

Sebmaster:
We have another vm bug/incompat:

`this` in non-strict contexts when undefined, should refer to window. This is currently not the case for example in EventTarget (after removing use strict) and crashes the w3c tests since they require this behaviour in their test runner.

domenic:
My interpretation of what's happening here is that the W3C tests use

```js
addEventListener("error", ...);
```

in their global `<script>` context. The global object inside of `<script>` is `window`, i.e. the window proxy. The window proxy doesn't have a `addEventListener` method, but the vm module will fall back to the sandbox object (a `Window` instance), which does. When it calls `Window.prototype.addEventListener`, it does so with the window proxy as `this`, instead of the `Window` instance. Since the window proxy doesn't have any internal state---i.e., it doesn't have the `__jsdom` property---this fails.

This seems OK, in that I think it's how browsers work too. The question is, what do browsers do to make this work? I can think of a couple alternatives:

- Add the private state to the window proxy.
- Make private state transparently redirect so that when it tries to get it from the global proxy, it instead gets it from the window.

I will ask around on public-script-coord. I also might be missing something entirely.

domenic:
Oh wait, I think I see the problem. `addEventListener` should reference the global, not the global proxy. Hmm..... that is not how the vm module behaves.

domenic:
Hmm but why does `window.addEventListener(...)` work, but `addEventListener` fail?? Given that `window` is a window proxy?

Maybe there is a bug in vm here that does not correctly set the `this` of calls in the global scope?

domenic:
Oh, I see. Yes.

> Maybe there is a bug in vm here that does not correctly set the this of calls in the global scope?

is correct. It's always using `undefined`.

Sebmaster:
Looks like saucelabs is broke, but we have passing node tests \o/

Sebmaster:
Oh, our rewriting is buggy, need to fix that :(

Sebmaster:
There we go, also fixed.

Sebmaster:
Okay, the event target migration is done, I think.

There's a failing test in there, but I think since it's accessing private APIs the test is wrong; I'm not sure how to rewrite it to match public APIs or just remove it. @domenic?

domenic:
I think we should not use symbols until we fix vm.

That test looks removable, yeah.

domenic:
I'm actually going to implement my suggestions and then create a new PR with them :). No need to jump on this quite yet.

PR #1032, opened 2015-02-19
Location.prototype.reload leaks a global "errors"

Sebmaster:
Hey,

I'm not precisely sure why, but calling `window.location.reload` leaks `errors` to the global scope:

```
> require("jsdom").jsdom().parentWindow.location.reload()
undefined
> global.errors
[ { type: 'error',
    message: 'NOT IMPLEMENTED: location.reload',
    data: null } ]
```

Adding some sort of an automatic globals checking to your test suite can catch this quickly in the future.

### DIFF
--- a/lib/jsdom/browser/utils.js
+++ b/lib/jsdom/browser/utils.js
@@ -18,6 +18,6 @@ exports.NOT_IMPLEMENTED = function (target, nameForErrorMessage) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


Sebmaster:
Added a PR here for the target fix, seems like the tests pass, however we seem to have broken something in frames already in 3.1.1 (or earlier).

PR #1037, opened 2015-02-20
Fix url resolution

Sebmaster:
This fixes our url resolution which was broken since ea8f5bf0b5d29fa32376e14f2e7dd6fb4b02e63a.

I plan to implement the [url resolution algorithm](https://url.spec.whatwg.org/#concept-basic-url-parser) sometime soon(tm), but we need a fix until then, especially for the 3.x release cycle before the window rework.

This should make #1009 closeable.

### DIFF
--- a/lib/jsdom/utils.js
+++ b/lib/jsdom/utils.js
@@ -161,9 +161,26 @@ exports.memoizeQuery = function memoizeQuery(fn) {
 * A slightly-more-compliant version of `url.resolve`, taking care of a few Node bugs.
 */
 exports.resolveHref = function resolveHref(baseUrl, href) {
-  // If href is "about:blank", Node tries to be too clever.
-  if (href === "about:blank") {
+  var about = "about:blank";
+
+  // if we're redirecting to another site on about:blank, just let it through
+  if (href.substring(0, about.length) === about) {
     return href;
+  }
+
+  // if we have to resolve from about:blank we need a bit of special logic
+  if (baseUrl.substring(0, about.length) === about) {
+    if (href.search(/^([A-Za-z0-9]+):/) === 0) { // we have an absolute url
+      baseUrl = href;
+      href = "";
+    } else if (href[0] === "#") { // we have a location hash on about:blank
+      return baseUrl.split(/#/)[0] + href;
+    } else if (!href) { // we just have some base url
+      return baseUrl;
+    } else { // must be a file url
+      baseUrl = "file://" + href;
+      href = "";
+    }
   }
 
   if (baseUrl === resolveHref.memoizedUrl && resolveHref.cache && resolveHref.cache[href]) {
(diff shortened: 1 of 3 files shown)


Sebmaster:
While this fixes the bug for now, long term we should think about properly resolving URLs. There's a module for this: URIjs.

There's a problem however: The file urls we are given may, or may not be fully qualified, which makes working with raw URIjs impossible. We have to a fixed schema for our whole public API like the `scripts` parameter. We then have to strip the file:// prefix before reading the resource.
This'd again break back-compat so maybe we should already pull that into 4.0?

domenic:
Thank you so much! Will try to land soon.

> URIjs

I'm not 100% sure this follows the URL Living Standard. It references a lot of other outdated specs in its docs. Regardless, we want to make sure whichever we choose passes the test suite in web-platform-tests. See also the work @rubys has been doing on the test suite and making it runnable in browsers: https://url.spec.whatwg.org/interop/test-results/. He has a reference implementation too, although it is PEGjs based so might be significantly slower than a hand-coded one. Thankfully URIjs has a good list of other implementations in its readme. We'll also want to check out @petkaantonov's https://github.com/petkaantonov/urlparser which is tuned for speed.

> There's a problem however: The file urls we are given may, or may not be fully qualified, which makes working with raw URIjs impossible. We have to a fixed schema for our whole public API like the scripts parameter. We then have to strip the file:// prefix before reading the resource.

I don't quite understand the problem you're describing, although it does remind me of some problematic things I've seen myself. Can you give an example?

lib/jsdom/utils.js
```diff
@@ -161,11 +161,28 @@ exports.memoizeQuery = function memoizeQuery(fn) {
 * A slightly-more-compliant version of `url.resolve`, taking care of a few Node bugs.
 */
 exports.resolveHref = function resolveHref(baseUrl, href) {
-  // If href is "about:blank", Node tries to be too clever.
-  if (href === "about:blank") {
+  var about = "about:blank";
+
+  // if we're redirecting to another site on about:blank, just let it through
+  if (href.substring(0, about.length) === about) {
```
domenic: In io.js I think we can just use startsWith

test/living-html/location.js
```diff
@@ -17,7 +17,9 @@ exports["window.location.port for an about:blank window"] = function (t) {
 };
 
 exports["window.location.hash is manipulable"] = function (t) {
-  var window = jsdom.jsdom().parentWindow;
+  var window = jsdom.jsdom("", {
```
domenic: These tests don't work with about:blank?

Sebmaster: Changing hash specifically works with the patch, however changing search for example still breaks so I kinda changed for uniformity.

lib/jsdom/utils.js
Sebmaster: Since 3.x is the last version we support node on I wanted to make this compatible so we can issue a last bugfix release for it.

PR #1052, opened 2015-03-02
replace _eventType to use _type in events

lovebear:
_eventType has 2 lines of 'dangling' code (never gets used for anything). 

Also things like `new Event('click')` or `new MouseEvent('click')` don't work because it was being set to _eventType but on dispatch it only checks _type to see if a type is set. 
https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/level2/events.js#L256

But it'll be always null since it was being set to _eventType but that never gets used.
https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/level2/events.js#L35

I decoded to just remove _eventType and use _type entirely instead of adding checks for _eventType because http://www.w3.org/TR/2000/REC-DOM-Level-2-Events-20001113/events.html#Events-Event-type only has type and no eventType attribute.


lovebear:
The test failing is:
https://github.com/tmpvar/jsdom/blob/master/test/level2/events.js#L104

Not sure if ok to delete, events are initialized now when using the Event constructor.
Follows browser behaviour: http://jsfiddle.net/ypzs2x5s/

domenic:
createEvent still needs to work, so this isn't OK as-is.

It is a good catch that `eventType` doesn't exist anymore. The correct spec is https://dom.spec.whatwg.org/#interface-event, though, per our CONTRIBUTING.md.

My guess is that eventType (maybe _eventType, maybe eventType) is still used somewhere, causing the test failure. So we can't delete it until we've fixed the test failure.

lovebear:
Delete the eventType being passed, as per the unchanged code it was set to _eventType (that didn't do anything), so the result is the same.

Now mimics the browser doing `new Event('blah')` having it's type set to 'blah' and being able to dispatch.
And `createEvent()` has no type and cannot dispatch right away until initialized.

lovebear:
Not sure if the error message looks ok to you

lib/jsdom/level2/events.js
```diff
@@ -267,12 +265,33 @@ core.AttributeList.prototype.$setNode = attrNodeInterceptor(core.AttributeList.p
...
+      'touchevent':       core.Event, // 'TouchEvent',
+
+      // old, not part of spec anymore
+      'mutationevents':   core.MutationEvent
     }
-    return new core.Event(eventType);
+
+    Constructor = interfaceTable[typeLower] || null;
+    if (!Constructor) {
+      this.raise('error', 'The provided event type (' + type + ') is invalid.', {error: 'NotSupportedError' });
```
domenic: Message looks good. But instead of using raise, do `throw new core.DOMException(core.DOMException.NOT_SUPPORTED_ERR, ...)`

lib/jsdom/level2/events.js
```diff
@@ -267,12 +265,33 @@ core.AttributeList.prototype.$setNode = attrNodeInterceptor(core.AttributeList.p
...
-core.Document.prototype.createEvent = function(eventType) {
-    switch (eventType) {
-        case "MutationEvents": return new core.MutationEvent(eventType);
-        case "UIEvents": return new core.UIEvent(eventType);
-        case "MouseEvents": return new core.MouseEvent(eventType);
-        case "HTMLEvents": return new core.Event(eventType);
+core.Document.prototype.createEvent = function(type) {
+    var Constructor = null;
+    var typeLower = type.toLowerCase();
+    var interfaceTable = {
```
domenic: We can create this table outside the function so that callers don't have to create it every time.

domenic:
Thanks so much! Merged as c00024db6248d2940d59254a80108088b04b9c22, with a few minor code style tweaks. Really happy to have your contribution!

PR #1053, opened 2015-03-02
Refactor events take 2

domenic:
Building on #1031. Right now I have to go to lunch but after I get back I can fix a few failing tests of DOMException hopefully. Then we'll have a new DOMException that we can use without requiring all of level1/core, which means that EventTarget can throw a DOMException without causing a circular dependency weirdness.


domenic:
OK cool, got this working and ready for a review, @Sebmaster.

Sebmaster:
What are we going to do with `window.addEventListener.call(myNode, ...)` now? We should probably also add a test for that.

domenic:
I think we can fix that, probably using your technique, in a follow-up. But for now we're basically doing what we've always done and shadowing the prototype methods.

Sebmaster:
Something is broken with the browser tests, need to fix that, but I haven't looked into it.

domenic:
Oh, it's the let/const usage, cuz Sauce doesn't have Chrome 41 yet :-/. Hmm.

PR #1054, opened 2015-03-02
update README to not be silly

tjwebb:
cc @emilyrose @mikeal

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A JavaScript implementation of the WHATWG DOM and HTML standards, for use with [
 $ npm install jsdom
 ```
 
-Note that as of our 4.0.0 release, jsdom no longer works with Node.js™, and instead requires io.js. You are still welcome to install a release in [the 3.x series](https://github.com/tmpvar/jsdom/tree/3.x) if you are stuck on legacy technology like Node.js™.
+Note that as of our 4.0.0 release, jsdom no longer works with Node.js™, and instead requires io.js. Node.js users are still welcome to install a release in [the 3.x series](https://github.com/tmpvar/jsdom/tree/3.x).
 
 ## Human contact
 


mikeal:
Nobody is going to merge a PR with the word "dickishness" in the commit log :(

tmpvar:
agree.

tjwebb:
Someone already asked nicely: https://github.com/tmpvar/jsdom/pull/1046.

tjwebb:
I would like the original PR to be taken seriously and evaluated for its merit, which includes not purposefully alienating and excluding the vast majority of node.js users, and not summarily dismissed for @domenic's stated reason of "No Thanks".

mikeal:
Being that the "nice one" wasn't merged why did you think one with a more hostile comment in it *would* be taken seriously? I might agree with the sentiment behind the change but this particular PR can *only* be seen as trolling given the content. This PR doesn't actually advance the conversation about the README language, it only continues to polarize it.

tjwebb:
I am trolling, you're absolutely right. The polarization event has already occurred. I'm perfectly happy to bring attention to it using all kinds of means, since I doubt most of the other 136 contributors of this project, upstream dependent modules, and the vast majority of other developers who depend on this module, have any idea that this has even occurred.

A cynical person might look at the 5-line `vm` change in 4.0.0 and think that maybe he created that incompatibility on purpose. @domenic is ostracizing the majority of developers in the nodejs community, and guess what? I'm feeling a little bit ostracized right now.

When you break people's stuff to suit your own selfish interests and then shove it in their faces on the README, I think some unpleasantness is called for.

mikeal:
This kind of behavior might make you feel better personally but it actually makes the situation worse. I guarantee this unpleasantness weighs more on me than anyone with commit access to this repository.

tjwebb:
> This kind of behavior

It's funny how suppression of dissent tends to have the reverse effect. I'm being ostentatious in attempt to bring attention to an issue that was immediately **closed and locked** with no discussion. I am the Squeaky Wheel.

> but it actually makes the situation worse.

How do you define "worse"? Do you mean it's now less likely that the rhetoric is restored to a less antagonistic phrasing? Or that you're now personally inconvenienced by this situation? If the former, and you can make a good case for why, I'll consider altering course. But if you really expected to just Move Right Along without being inconvenienced or hassled for a blatant act of concerted divisiveness, I'm not sure how we move constructively forward with that as our starting point.

mikeal:
> How do you define "worse"?

Your behavior makes it *less* likely that the language will be changed.

tjwebb:
> Your behavior makes it less likely that the language will be changed.

What is your case for this? Spitefulness?

mikeal:
@tjwebb I actually agree with changing the language. However, that opinion is less likely to be considered the more that blatant trolling occurs carrying that same message. Anyone with a successful project is used to filtering out trolls and this behavior puts your message along with everyone that agrees with you in the filter.

tjwebb:
Then the correct course of action, short of changing the language, is to close and lock *this* thread and re-open https://github.com/tmpvar/jsdom/pull/1046. Is there anything stopping a jsdom maintainer from effecting this immediately?

mikeal:
> Is there anything stopping a jsdom maintainer from effecting course of action immediately?

Yes, the fact that they would be responding to trolling :)

If someone gave me commit access tomorrow I would close and lock this issue and then wait long enough that any change couldn't be associated with this behavior.

domenic:
I'm sorry I let this troll thread go on so long; I was at an appointment.

PR #1055, opened 2015-03-03
Various upgrades to deps and browser CI infrastructure

domenic:
- Upgrade local selenium-standalone package to the latest version. This now requires a two-step installation/start process.
- Use get-saucelabs-browsers to ensure we are always testing the latest stable Chrome version. This is apparently necessary since Sauce Labs does not have any way of doing so with a simple parameter like "stable".
- Upgrade all other packages to their latest versions, being sure to pin < 1.0.0 packages unless I've received confirmation from the author that they will not be breaking back-compat on patch versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,22 +66,22 @@
     "xml-name-validator": ">= 2.0.1 < 3.0.0",
     "xmlhttprequest": ">= 1.6.0 < 2.0.0",
     "acorn-globals": "^1.0.2",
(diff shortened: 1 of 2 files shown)


Sebmaster:
`get-saucelabs-browsers` shouldn't be needed anymore now, since saucelabs switched to using the latest Chrome when (not) providing an empty version string.

PR #1058, opened 2015-03-05
Upgrading NodeUnit reveals failures in w3c tests?

domenic:
Please help @Sebmaster! It looks like the first 9 w3c tests pass and the last 9 do not? I tried rearranging them and the same thing happens. My best guess is something wrong with w3ctest.js?

### DIFF
--- a/test/worker-runner.js
+++ b/test/worker-runner.js
@@ -32,6 +32,7 @@ self.onmessage = function (e) {
 
     "living-dom/attributes.js": require("../test/living-dom/attributes.js"), // 1/11
     "living-dom/compare-document-position.js": require("../test/living-dom/compare-document-position.js"), // 0/20
+    "living-dom/event-target.js": require("../test/living-dom/event-target.js"), // ok
     "living-dom/node-contains.js": require("../test/living-dom/node-contains.js"), // 0/20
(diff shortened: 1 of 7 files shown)


test/worker-runner.js
```diff
@@ -32,6 +32,7 @@ self.onmessage = function (e) {
 
     "living-dom/attributes.js": require("../test/living-dom/attributes.js"), // 1/11
     "living-dom/compare-document-position.js": require("../test/living-dom/compare-document-position.js"), // 0/20
+    "living-dom/event-target.js": require("../test/living-dom/event-target.js"), // 0/20
```
domenic: 0/20? :P

domenic:
LGTM, with nice atomic commits too. Merge at will, or I will do so when I get a chance. Great stuff, and thanks for saving the day :D

Sebmaster:
There might actually be another problem at the event fireing order in regards to legacy events although I think it's correct. We should at least add another test for that in any case.

domenic:
You mean like `.onload = ...`? Yeah I think we have an open issue for that. https://github.com/tmpvar/jsdom/issues/696

test/worker-runner.js
Sebmaster: Fixed

Sebmaster:
Oh yeah, right. We're good then.

PR #1059, opened 2015-03-07
Improve the performance of define by about ~40%

Sebmaster:
Since define is used to copy over all of our dom properties (which are a lot), it has to be pretty performant.
This patch does two things:
* Removes the need to allocate a new anonymous function by replacing the forEach with a standard for loop
* If the property to copy is a simple defined property (not defined with defineProperty), avoid the call to defineProperty ourselves too

Especially the second point is important. It shaves off about 40% of time in define, and about 20% of overall initialization time, which reduces the time for 10000 instances from about 8 seconds to 6 seconds on my laptop.

### DIFF
--- a/lib/jsdom/utils.js
+++ b/lib/jsdom/utils.js
@@ -100,12 +100,20 @@ exports.inheritFrom = function inheritFrom(Superclass, Subclass, properties) {
  *
  * - `object` {Object} the target object
  * - `properties` {Object} the source from which to copy property descriptors
+ * 
+ * This function is performance sensitive, it's used to copy the DOM into every window.
  */
 exports.define = function define(object, properties) {
-  Object.getOwnPropertyNames(properties).forEach(function (name) {
+  var keys = Object.getOwnPropertyNames(properties);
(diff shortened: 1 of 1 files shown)


domenic:
LGTM!!

domenic:
Actually, maybe add a comment explaining that this is a performance-sensitive function so that future people don't get confused and try to refactor it back.

Sebmaster:
Done. Before you merge, can you check on your side too if this really does improve perf, I'm not sure I did the tests correctly, since those are always kind of fluctuating.

domenic:
Hmm, using this script I am having a hard time getting a difference:

```js
'use strict';
var jsdom = require('./').jsdom;
var date = Date.now();

for (var i = 0; i < 1000; ++i) {
  jsdom();
}

console.log(Date.now() - date);
```

After my CPU warmed up (on battery power, the first few runs were slower) it seemed to settle around 33s for both master and performance/define-patch.

Leaving out of my current batch of merges for now, maybe investigate more in detail later...

PR #1060, opened 2015-03-07
Defer the initialization of vm

Sebmaster:
This reduces init times by about 600+% if scripting on the page is not needed, so gives a nice performance boost.

This breaks backwards compat somewhat!
Enabling scripting after already having received a window (i.e. adding ProcessExternalResources to document.implementation manually, not providing it in config/options) results in a changing document.defaultView.
If we need to increase major probably mainly depends on if document.implementation._addFeature is considered public API.

### DIFF
--- a/lib/jsdom/level1/core.js
+++ b/lib/jsdom/level1/core.js
@@ -281,6 +281,14 @@ core.DOMImplementation.prototype = {
       } else {
         this._features[feature].push(version);
       }
+
+      if (feature === "processexternalresources" &&
+          (version === "script" || (version.indexOf && version.indexOf("script") !== -1)) &&
+          !vm.isContext(this._ownerDocument._global)) {
+        vm.createContext(this._ownerDocument._global);
+        this._ownerDocument._defaultView._globalProxy = vm.runInContext("this", this._ownerDocument._global);
+        this._ownerDocument._defaultView = this._ownerDocument._defaultView._globalProxy;
+      }
     }
   },
 
--- a/lib/jsdom.js
+++ b/lib/jsdom.js
@@ -235,10 +235,6 @@ function processHTML(config) {
     return;
   }
 
-  window.document.implementation._addFeature('FetchExternalResources', ['script']);
-  window.document.implementation._addFeature('ProcessExternalResources', ['script']);
-  window.document.implementation._addFeature('MutationEvents', ['2.0']);
-
   function scriptComplete() {
     docsLoaded++;
 
@@ -273,6 +269,10 @@ function processHTML(config) {
   }
 
   if (config.scripts.length > 0 || config.src.length > 0) {
+    window.document.implementation._addFeature('FetchExternalResources', ['script']);
+    window.document.implementation._addFeature('ProcessExternalResources', ['script']);
+    window.document.implementation._addFeature('MutationEvents', ['2.0']);
+
     config.scripts.forEach(function (scriptSrc) {
       var script = window.document.createElement('script');
       script.className = 'jsdom';
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -1,6 +1,5 @@
 "use strict";
 
-var vm = require("vm");
 var URL = require("url");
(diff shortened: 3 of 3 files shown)


lib/jsdom/level1/core.js
```diff
@@ -281,6 +281,14 @@ core.DOMImplementation.prototype = {
       } else {
         this._features[feature].push(version);
       }
+
+      if (feature === "processexternalresources" &&
+          (version === "script" || (version.indexOf && version.indexOf("script") !== -1)) &&
+           !vm.isContext(this._ownerDocument._defaultView)) {
```
domenic: Should be checking isContext on _global, for consistency

domenic:
Is _addFeature called lazily---is that why this works?

This doesn't count as back-compat breaking, since _addFeature is not public (it has an underscore).

Sebmaster:
> Is _addFeature called lazily---is that why this works?

Yes, I missed a modification in processHTML before to really actually defer it, so fixed that up too now.

Sebmaster:
Oh yeah, testscript is a simple:
```js
var jsdom = require("./lib/jsdom");

function openWindow() {
  return new Promise(function (resolve, reject) {
    jsdom.env({
      html: "<!DOCTYPE html><html><head><title>Hello World!</title></head><body></body></html>",
        if (err) return reject(err[0]);
        resolve();
        window.close();
      }
    });
  });
}

global.startTest = function startTest() {
  var promise = Promise.resolve();

  for (var i = 0; i < 100; ++i) {
    promise = promise.then(openWindow);
  }

  return promise;
}

var start = Date.now();
if (console.profile) console.profile("jsdom init");

global.startTest().then(function () {
  if (console.profile) console.profileEnd("jsdom init");
  console.log(Date.now() - start);
}, function (e) {
  console.error(e, e.stack.toString());
});
```

Output here is:
```bash
$ iojs index
394

$ git checkout master
Switched to branch 'master'

$ iojs index
7130
```

domenic:
Still unclear how this works... What if someone initially creates a page with no scripts, but then latter adds a `<script>` element dynamically? Right now it only looks like the vm setup happens correctly if `config.scripts` is set? But will not happen if a `<script>` is inserted dynamically? How do we not have tests that catch that case?? I must be missing something....

Sebmaster:
All uses of scripts only work if you enable the `ProcessExternalResources` feature (which is done implicitly for the scripts provided during initialization and disabled afterwards). If someone wanted to insert scripts later dynamically, they had to enable that feature already, either by providing the feature during initialization, or calling _addFeature later themselves.

domenic:
> (which is done implicitly for the scripts provided during initialization and disabled afterwards)

Hmm, where is the "disabled afterwards" code? Right now it looks like anything that calls in to `processHTML` will automatically get the feature enabled.

Sebmaster:
See the reset [here](https://github.com/tmpvar/jsdom/blob/master/lib/jsdom.js#L246).

domenic:
Ahhh, I see.

OK. So what this patch *actually* does, is:

- If the author is using `jsdom.env`, and doesn't explicitly opt-in to script support, and doesn't implicitly opt-in to script support by adding `scripts:` or `src:` options, *then* we don't initialize vm.

It's ironic that this only applies to `jsdom.env`, which is more awkward to use than `jsdom.jsdom`, and usually is only worth using if you're going to use external resources like scripts! Makes me wonder about switching the default on jsdom.jsdom so they can get this benefit too. (I guess they can, if they explicitly disable script support.)

Sebmaster:
> If the author is using jsdom.env, and doesn't explicitly opt-in to script support, and doesn't implicitly opt-in to script support by adding scripts: or src: options, then we don't initialize vm.

Heh, that's a much better summary.

> Makes me wonder about switching the default on jsdom.jsdom so they can get this benefit too.

Ewww, I thought we switched that ages ago, but I guess that was .env only :(

domenic:
OK cool, I will update the commit message and merge into 4.0.3!

We should consider switching the default for 5.0.0.

PR #1062, opened 2015-03-13
Assure accurate childNodes state

medikoo:
`childNodes` properties reflect inaccurate state in some scenarios. This update fixes that

This bug caused [replaceContent](https://github.com/medikoo/dom-ext/blob/master/node/%23/_replace-content.js) module logic not working properly. While it works without issues in all major browsers

### DIFF
--- a/lib/jsdom/level1/core.js
+++ b/lib/jsdom/level1/core.js
@@ -567,6 +567,9 @@ core.Node.prototype = {
     if (this._childrenList) {
       this._childrenList._update();
     }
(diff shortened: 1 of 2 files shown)


domenic:
LGTM, great catch. Will be able to merge and do a release some time in the next day or two.

domenic:
Merged as 06710a9f0cc7ebd5b24eda9192008e66a4970db8, 4.0.3 coming shortly.

medikoo:
Thank you!

PR #1009, opened 2015-03-16
Pathname of anchor tag with relative `href`

Sebmaster:
I was working with JSDom and ran into some behavior with `myAnchorTag.pathname` that I thought was unexpected. Let me show an example:

```js
global.jsdom = require('jsdom').jsdom;
global.document = jsdom('<html><head><script></script></head><body></body></html>');
global.window = global.document.parentWindow;
global.navigator = window.navigator = { userAgent: 'JSDom', appVersion: '' };

var a = document.createElement('a');

a.href = "uri/what/pls";
a.pathname;
// /what/pls

a.href = "/uri/what/pls";
a.pathname;
// /uri/what/pls
```

Is it unexpected that the pathname always strips the first part of the path when it is relative + no leading slash? Or am I missing something obvious here? :open_mouth:

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -1,6 +1,5 @@
 "use strict";
 
-var URL = require("url");
 var CSSStyleDeclaration = require("cssstyle").CSSStyleDeclaration;
 var XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest;
 var notImplemented = require("./not-implemented");
@@ -8,6 +7,7 @@ var History = require("./history");
 var VirtualConsole = require("../virtual-console");
 var define = require("../utils").define;
 var inherits = require("../utils").inheritFrom;
+var resolveHref = require("../utils").resolveHref;
 var EventTarget = require("../events/EventTarget");
 var namedPropertiesWindow = require("../living/named-properties-window");
 var cssom = require("cssom");
@@ -151,7 +151,7 @@ function Window(options) {
     var lastUrl = "";
     xhr._open = xhr.open;
     xhr.open = function (method, url, async, user, password) {
-      url = URL.resolve(window.document.URL, url);
+      url = resolveHref(window.document.URL, url);
       lastUrl = url;
       return xhr._open(method, url, async, user, password);
     };
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -188,6 +188,19 @@ function Window(options) {
       xhr.addEventListener("readystatechange", setReceivedCookies);
       return xhr._send(data);
     };
+    Object.defineProperty(xhr, "response", {
+      get: function () {
+        if (this.responseType === "text") {
+          return this.responseText;
+        } else if (this.responseType === "json") {
+          return JSON.parse(this.responseText);
+        } else {
+          return null; // emulate failed requeest
+        }
+      },
+      enumerable: true,
+      configurable: true
+    });
     return xhr;
   };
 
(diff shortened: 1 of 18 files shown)


Sebmaster:
Appended a PR here which fixes this issue and should pass all tests to get the discussion here started.

I'm still not completely sure about urlmaster, I'll maybe experiment a bit with different url parsers, maybe [urlparser](https://github.com/petkaantonov/urlparser) (as per @domenic's [comment](https://github.com/tmpvar/jsdom/pull/1037#issuecomment-75272058)

Sebmaster:
Okay I switched to URIjs again. There's an issue there with the percent encoding which I reported there (medialize/URI.js#198), but I think it's the most standard conformant one (although it seems we still have to handle our about:blanks manually).

I think the only other failing test is one where the expect cases are wrong, which are actually more right with URIjs. The test cases are auto generated by urlmaster, however it seems it doesn't actually resolve the testcases itself well enough. We'll need to either convert them to explicit tests or remove them

lib/jsdom.js
```diff
@@ -361,6 +364,12 @@ function getConfigFromArguments(args, callback) {
     config.url = toFileUrl(config.file);
   }
 
+  if (config.url) {
+    // backslashes were originally not allowed in URLs, so this is error-handling only
+    // comparable to chrome which converts all backslashes to foward slashes
+    config.url = config.url.replace(/\\/g, '/');
```
domenic: Hmm shouldn't we just do config.url = url.resolve("about:blank", config.url)` which would normalize everything for us?

Sebmaster:
We should add the [URL tests](https://github.com/w3c/web-platform-tests/tree/master/url) from web-platform-tests too here.

Sebmaster:
Oooookay, I got some time to fix this up again. A few things I still know which are left (there might be a few more):

- [ ] Try to completely throw away `fixFileUrl` again so we don't break back-compat (also because I don't know if (or where) it's actually necessary)
- [ ] Run tests, fix them (remove urlmaster?)
- [ ] Either merge the two spec/* branches in jsdom/whatwg-url into master (which would make master deviate from the spec), try to get it into the spec somewhat quickly (or something else?).
- [ ] Check into the comments on [jsdom/whatwg-url#e8b2dbe](https://github.com/jsdom/whatwg-url/commit/e8b2dbefc36335fa9db6d689f424ec2eb2ca92d9).

Sebmaster:
- [ ] `removeAttribute` should reset the URL object somehow apparently (does in Chrome at least); `href` returns "" after.

Sebmaster:
Oh buy, fixing these tests is not fun. Windows fails because some dependency issues between tests (a previous tests kills the level2/style test), Linux/Travis fails on the filename_with_spaces test :(

domenic:
I'll take care of it :)

Sebmaster:
We have passing tests! @domenic please review.

domenic:
I am ready to merge this as soon as you give me the sign off btw. The nits and whatwg-url issues can be fixed in follow-ups or as I merge. Just want to make sure you're happy with it first.

Sebmaster:
Should be good to go for now.

domenic:
This hangs locally on w3c/index.js... going to try to track it down.

Sebmaster:
Windows? Might be our classic path resolving then. Might end up putting the patch in compat too...

domenic:
I bet it's the usage of things like .href and such

domenic:
Yeah resource.url.href is `c:/Users/Domenic/Dropbox/GitHub/tmpvar/jsdom/test/w3c/tests/resources/testharness.js` missing `file:///` and capitalization wrong. Working on a fix.

jmeas:
:tada:

domenic:
v6.0.0 coming up soon...

PR #1066, opened 2015-03-19
Bug in doctype parsing

Sebmaster:
JSDOM 3.0 started implementing DocumentType as per the spec and return the correct type when queried using document.doctype. However, there's a bug in the parsing code and simple doctypes of the form '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">' aren't getting properly parsed.

As a result there's no way for application to implement their own doc type parsing (which I used to do for 2.0) or get correct values from JSDOM.


Sebmaster:
Added PR which fixes it for our default config (parse5) at least.

domenic:
LGTM!

Sebmaster:
Just released 4.0.4 with this patch.

PR #1068, opened 2015-03-20
Add HTMLUnknownElement

Sebmaster:
All elements with unrecognized tags are created with HTMLUnknownElement instead of HTMLElement.
Also changed this to only occur on HTMLDocument since we need to start to split these up properly.

### DIFF
--- a/lib/jsdom/level2/html.js
+++ b/lib/jsdom/level2/html.js
@@ -402,7 +402,21 @@ core.HTMLDocument = function HTMLDocument(options) {
   this.implementation._addFeature('xml'   , '2.0');
 };
 
+var nonInheritedTags = new Set([
+  'article', 'section', 'nav', 'aside', 'hgroup', 'header', 'footer', 'address', 'dt',
+  'dd', 'figure', 'figcaption', 'main', 'em', 'strong', 'small', 's', 'cite', 'dfn', 'abbr',
+  'ruby', 'rt', 'rp', 'code', 'var', 'samp', 'kbd', 'i', 'b', 'u', 'mark', 'bdi', 'bdo', 'wbr'
+]);
+
 inheritFrom(core.Document, core.HTMLDocument, {
+  _defaultElementBuilder: function (document, tagName) {
+    if (nonInheritedTags.has(tagName.toLowerCase())) {
+      return new core.HTMLElement(document, tagName);
+    } else {
+      return new core.HTMLUnknownElement(document, tagName);
+    }
+  },
+
   _referrer : "",
   get referrer() {
     return this._referrer || '';
@@ -586,9 +600,9 @@ define('HTMLElement', {
   ]
 });
 
-core.Document.prototype._defaultElementBuilder = function(document, tagName) {
-  return new core.HTMLElement(document, tagName);
-};
+define('HTMLUnknownElement', {
+  // no additional properties & attributes
+});
 
 // http://www.whatwg.org/specs/web-apps/current-work/#category-listed
 var listedElements = /button|fieldset|input|keygen|object|select|textarea/i;
--- a/test/living-html/htmlelement.js
+++ b/test/living-html/htmlelement.js
@@ -1,0 +1,69 @@
+﻿"use strict";
+
+var jsdom = require("../..");
+
+var nonInheritedTags = [
+  "article", "section", "nav", "aside", "hgroup", "header", "footer", "address", "dt",
+  "dd", "figure", "figcaption", "main", "em", "strong", "small", "s", "cite", "abbr",
+  "code", "i", "b", "u"
+];
+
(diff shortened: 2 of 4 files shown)


Sebmaster:
This is the foundation to fix #1030.

domenic:
FYI in reality we need to merge HTMLDocument and Document. They are aliases. Not sure how that affects our short-term strategy though.

lib/jsdom/level2/html.js
```diff
@@ -586,9 +590,12 @@ define('HTMLElement', {
   ]
 });
 
-core.Document.prototype._defaultElementBuilder = function(document, tagName) {
-  return new core.HTMLElement(document, tagName);
-};
+define('HTMLUnknownElement', {
+  proto: {
```
domenic: Seems leftover

test/living-html/htmlelement.js
```diff
@@ -0,0 +1,27 @@
+﻿"use strict";
+
+var jsdom = require("../..");
+
+exports["unknown elements should return HTMLUnknownElement"] = function (t) {
```
domenic: Need tests that it works during parsing too.

Also need tests that elements without special classes, e.g. `<section>`, stay HTMLElement and not HTMLUnknownElement. Preferably an exhaustive test of such elements. And both parsing and explicit creation.

Sebmaster: Huh, not sure if that'll work with our current structure. Although I guess we could a put a list of all these special cases in an array or so. Are these documented explicitly somewhere?

lib/jsdom/level2/html.js
Sebmaster: Hmm? I moved _defaultElementBuilder later into HTMLDocument.

Sebmaster:
Re: HTMLDocument vs Document
Chrome gives
```
> XMLDocument.__proto__
function Document() { [native code] }
> HTMLDocument.__proto__
function Document() { [native code] }
```

FF:
```
document.implementation.createDocument("html", "a", document.implementation.createDocumentType("a", "a", "a")).__proto__.__proto__
DocumentPrototype { ... }
```

I just want to get the prototype chain correct here in medium term.

domenic:
Yeah browsers haven't caught up to spec's simplifications here yet, like for Attr.

test/living-html/htmlelement.js
domenic: Not sure, but there should be a list of all elements and their interfaces somewhere.

lib/jsdom/level2/html.js
domenic: The test: ohgod part

Sebmaster: Ohhh heh, true.

Sebmaster:
Does XmlDocument inherit from "HTMLDocument" then?

test/living-html/htmlelement.js
Sebmaster: Added.

domenic:
There is no separate document for XML and HTML. Just Document = HTMLDocument, which can be in XML or HTML parsing modes.

lib/jsdom/level2/html.js
```diff
@@ -402,7 +402,21 @@ core.HTMLDocument = function HTMLDocument(options) {
   this.implementation._addFeature('xml'   , '2.0');
 };
 
+var nonInheritedTags = [
```
domenic: Use Set

test/living-html/htmlelement.js
```diff
@@ -0,0 +1,46 @@
...
+};
+
+exports["non-inherited elements should have the HTMLElement type"] = function (t) {
+  t.expect(nonInheritedTags.length);
+
+  var doc = jsdom.jsdom();
+
+  for (var i = 0; i < nonInheritedTags.length; ++i) {
+    var el = doc.createElement(nonInheritedTags[i]);
+    t.ok(el.constructor === doc.defaultView.HTMLElement, "div element should inherit from HTMLElement too");
```
domenic: Assertion message is outdated

test/living-html/htmlelement.js
```diff
@@ -0,0 +1,46 @@
+﻿"use strict";
+
+var jsdom = require("../..");
+
+var nonInheritedTags = [
+  "article", "section", "nav", "aside", "hgroup", "header", "footer", "address", "dt",
+  "dd", "figure", "figcaption", "main", "em", "strong", "small", "s", "cite", "abbr",
+  "code", "i", "b", "u"
+];
+
```
domenic: Need parsing tests

Sebmaster:
Updated.

domenic:
LGTM, merge at will!

KimiSon:
What are y'all trying to do here? Hack into accounts that don't belong to u and finding pictures of someone's kid? 
Isn't that illegall??

PR #1069, opened 2015-03-22
Iframes are not cleaned up after the node DOMs are removed

Sebmaster:
Whenever an frame/iframe node element is created, there is a associated getter created for the parent window. However, the getter is not cleaned up when the node is removed. Also, `window.frames.length` or `window.length` doesn't change after iframes are removed.

The issue can be reproduced [here](https://gist.github.com/sylnp0201/1922dc0304998363a20d) with jsdom v3.1.2 and a simple Backbone view.

### DIFF
--- a/lib/jsdom/level2/html.js
+++ b/lib/jsdom/level2/html.js
@@ -2049,11 +2049,11 @@ function loadFrame (frame) {
 }
 
 function refreshAccessors(document) {
-  var window = document.defaultView;
+  var window = document._defaultView;
   var frames = document.querySelectorAll("iframe,frame");
 
   // delete accessors for all frames
-  for (var i = 0; i < window.length; ++i) {
+  for (var i = 0; i < window._length; ++i) {
     delete window[i];
   }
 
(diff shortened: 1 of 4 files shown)


Sebmaster:
Added PR here. I'm not sure why the situation got better with jsdom 4.0.4 though, the references shouldn't really have been cleaned up, not sure why the memory leak got fixed...

domenic:
Probably due to vm contexts using weak references whereas contextify requires manual disposal (which I assume we forgot to do).

Sebmaster:
I think the whole reference chain stayed intact though, since the parent still kept a reference to retrieve it via the name, or index accessor on "`.frames`".

test/window/frame.js
```diff
@@ -389,5 +389,31 @@ exports.tests = {
...
+    var window = document.defaultView;
+
+    test.strictEqual(window.length, 1, "frame should exist (.length)");
+    test.notEqual(window[0], undefined, "frame should exist (undefined check)");
+    test.strictEqual(window[0], window.foo, "index access and name access should return same reference");
+
+    test.done();
+  },
+
+  'remove frame': function (test) {
```
domenic: Would like an additional test with removing from the middle of frames, e.g. with three frames remove frame 1

lib/jsdom/level2/html.js
```diff
@@ -2056,7 +2056,28 @@ define('HTMLFrameElement', {
...
+        parent._frame(i, parent[i + 1]);
+      }
+
+      // delete last reference
+      delete parent[parent._length - 1];
+      parent._length--;
+    }
+
     this.addEventListener('DOMNodeInsertedIntoDocument', this._initInsertListener, false);
+    this.addEventListener('DOMNodeRemovedFromDocument', onRemove, false);
```
domenic: Can we use hooks (if we have one for node removed) instead of mutation events? We already use _atrrModified I noticed.

test/window/frame.js
```diff
@@ -389,5 +389,31 @@ exports.tests = {
...
+    var document = jsdom.jsdom("<!doctype html><iframe id='myFrame' name='foo'></iframe>");
+    var window = document.defaultView;
+
+    test.strictEqual(window.length, 1, "frame should exist (.length)");
+    test.notEqual(window[0], undefined, "frame should exist (undefined check)");
+    document.body.removeChild(document.getElementById("myFrame"));
+
+    test.strictEqual(window.length, 0, "frame shouldn't exist (.length)");
```
domenic: Can we test that '0' in window is false too?

test/window/frame.js
Sebmaster: Added, good catch, did indeed not work.

test/window/frame.js
Sebmaster: Added.

lib/jsdom/level2/html.js
Sebmaster: Switched to _detach. Which is also a bit later in the event cycle.

lib/jsdom/level2/html.js
```diff
@@ -2096,6 +2097,36 @@ define('HTMLFrameElement', {
...
+      }
+
+      // delete frame reference
+      if (self.getAttribute("name")) {
+        parent._frame(self.getAttribute("name"));
+      }
+
+      // delete index reference by moving all frames with a higher index one index down
+      for (var i = self._frameID; i < parent.length - 1; ++i) {
+        var next = parent.document.querySelectorAll("iframe,frame")[i];
```
Sebmaster: We need to rethink this, also with the frameIDs. What if frames are moved in the DOM, although this already would screw up the getters I think...

domenic: Moved = detach + attach, right? I think we should have enough hooks to make this work...

This does seem sketchy though, switching between the existing `parent` pseudo-array and `parent.document.querySelectorAll("iframe,frame")`; it's not clear which is the source of truth. Maybe just renumber everything from scratch (based on QSA results) whenever any frame is attached/detached? Probably simpler, and rare enough that perf shouldn't matter.

domenic: That said, if enough tests pass, I'm happy with whatever strategy :).

Sebmaster: We'd have to re-initialize frameID in that case.

Renumbering is probably the better choice, I'll add a test for moving frames and will fix that ... somehow.

Sebmaster:
Nice, should have everything fixed now, please give it another review @domenic.

lib/jsdom/level2/html.js
```diff
@@ -2048,6 +2048,17 @@ function loadFrame (frame) {
   }
 }
 
+function refreshAccessors(document) {
+  var window = document.defaultView;
+  var frames = Array.prototype.slice.call(document.querySelectorAll("iframe,frame"));
+  frames.forEach(function (frame, i) {
+    delete window[i];
+    defineGetter(window, i, function () {
```
domenic: Hmm is there a reason not to do `window[i] = frame.contentWindow`?

I guess if you navigated the frame then that would be incorrect, so getter is better. We should add a test for that case!

lib/jsdom/level2/html.js
```diff
@@ -2048,6 +2048,17 @@ function loadFrame (frame) {
   }
 }
 
+function refreshAccessors(document) {
+  var window = document.defaultView;
+  var frames = Array.prototype.slice.call(document.querySelectorAll("iframe,frame"));
+  frames.forEach(function (frame, i) {
```
domenic: Two things:

- Use `Array.prototype.forEach.call(frames` instead of slicing-then-forEaching, for efficiency.
- Perhaps we should set `window._length = frames.length` here and omit all other manual manipulation of `_length` below?

lib/jsdom/level2/html.js
```diff
@@ -2110,6 +2139,27 @@ define('HTMLFrameElement', {
...
+      if (self.contentWindow) {
+        self.contentWindow.close();
+      }
+
+      // delete frame reference
+      if (self.getAttribute("name")) {
+        parent._frame(self.getAttribute("name"));
+      }
+
+      parent._length--;
```
domenic: Can probably also move this to logic (including deletion) to refreshAccessors. Delete all accessors from 0 to `window._length`, then set `window._length` to `frames.length`, then add new accessors.

lib/jsdom/level2/html.js
```diff
@@ -2102,6 +2121,16 @@ define('HTMLFrameElement', {
               self.removeEventListener('DOMNodeInsertedIntoDocument', loader, false);
               this._waitingOnInsert = false;
               loadFrame(self);
+
+              parent._length++;
+              refreshAccessors(self._ownerDocument);
+
+              var name = self.getAttribute("name");
```
domenic: I'm confused why this is necessary here given the generic code above it, and why it doesn't just use `parent._frame` like the code above it does.

domenic:
Things are looking really good. Already a clear improvement over where we are currently, so at this point we might be reaching "just merge it already" status. There's a few minor cleanups you can do but that's probably enough.

lib/jsdom/level2/html.js
Sebmaster: We currently don't implement navigation, so not really a "problem" currently.

lib/jsdom/level2/html.js
Sebmaster: I checked, apparently the following two lines are different when executed at the same place
```js
defineGetter(window, i, function () { return frame.contentWindow; });
window._frame(i, frame);
```

No idea why that is, apparently the getter defined in _frame is never called. Maybe it's once again the proxy thing?

domenic:
@Sebmaster PTAL at newest commit

lib/jsdom/level2/html.js
```diff
@@ -2048,68 +2048,72 @@ function loadFrame (frame) {
...
-            this.addEventListener('DOMNodeInsertedIntoDocument', function loader () {
-              self.removeEventListener('DOMNodeInsertedIntoDocument', loader, false);
-              this._waitingOnInsert = false;
-              loadFrame(self);
-            }, false);
-            this._waitingOnInsert = true;
-          }
-        } else {
-          loadFrame(self);
+        if (this._parentNode) {
```
Sebmaster: Looks like this might cause problems in virtual dom constructions without appending to body. Maybe add a test for that?

domenic: I think the intent of all the crazy above this was to never load the frame until it was appended to body, so, should be equivalent? But yeah, worth checking to see if existing tests cover it and if not adding more and making sure the results are the same; will do.

Sebmaster: Yeah it is, but I think a Node can have `_parentNode` set, without it actually being appended to the document.

domenic: Oh, I see.

domenic: Fixed!

Sebmaster:
LGTM then. \o/

Issue #1027, opened 2015-02-16
Expired cookie is still present in document.cookie


rmurphey:
jsdom version: 3.1.1

I ran into a bug that is alluded to in #310 and referenced in ScottHamper/Cookies#37 and jmeas/backbone.simple-auth#1: A cookie set with an expiration date in the past -- as [recommended on MDN](https://developer.mozilla.org/en-US/docs/Web/API/document.cookie#Removing_a_cookie) as the way to delete a cookie -- is still present in `document.cookie` inside jsdom. 

jsdom test case:

```js
var jsdom = require('jsdom');

jsdom.env('<p>expired cookie bug</p>', function (err, window) {
  var doc = window.document;

  function create(name, value, days) {
    var date = new Date();
    date.setTime(date.getTime() + (days * 24 * 60 * 60 * 1000));
    var expires = days ? ';expires=' + date.toGMTString() : '';

    var c = name + '=' +
      encodeURIComponent(value) +
      expires +
      ';path=/';

    doc.cookie = c;
  }

  // set a cookie that expires in 10 days
  create('testcookie1', '1', 10);
  console.log('unexpired cookie is present', !!doc.cookie.match('testcookie1=1'));

  // set a cookie that has already expired
  create('testcookie2', '', -1);
  console.log('expired cookie is present', !!doc.cookie.match('testcookie2'));
});
```

This results in the following output:

```
unexpired cookie is present true
expired cookie is present true
```

Here is a browser test case:

```js
var doc = window.document;

function create(name, value, days) {
  var date = new Date();
  date.setTime(date.getTime() + (days * 24 * 60 * 60 * 1000));
  var expires = days ? ';expires=' + date.toGMTString() : '';

  var c = name + '=' +
    encodeURIComponent(value) +
    expires +
    ';path=/';

  doc.cookie = c;
}

// set a cookie that expires in 10 days
create('testcookie1', '1', 10);
console.log('unexpired cookie is present', !!doc.cookie.match('testcookie1=1'));

// set a cookie that has already expired
create('testcookie2', '', -1);
console.log('expired cookie is present', !!doc.cookie.match('testcookie2'));
```

The output is as expected:

```
unexpired cookie is present true
expired cookie is present false
```

inikulin:
[...]

Issue #846, opened 2014-07-30
cookieDomain should default to document.location.host ?


kapouer:
Is there a reason not to do
`window.document._cookieDomain = document.location.host` ?
if not i'll do a PR.

domenic:
Cookies are now handled much more realistically, so this should be taken care of now.

PR #1070, opened 2015-03-24
Properly implemented cookies, w3c tests fix and #1027 fix.

inikulin:
BTW, it introduces breaking change: `config.resourceLoader` doesn't receive `cookieDomain` in the arguments anymore. Because, it doesn't make sense with the current implementation.

### DIFF
--- a/lib/jsdom.js
+++ b/lib/jsdom.js
@@ -2,6 +2,7 @@
 var fs = require("fs");
 var path = require("path");
 var URL = require("url");
+var CookieJar = require('tough-cookie').CookieJar;
 
 var toFileUrl = require("./jsdom/utils").toFileUrl;
 var defineGetter = require("./jsdom/utils").defineGetter;
@@ -9,6 +10,8 @@ var defineSetter = require("./jsdom/utils").defineSetter;
 var features = require("./jsdom/browser/documentfeatures");
 var domToHtml = require("./jsdom/browser/domtohtml").domToHtml;
 var Window = require("./jsdom/browser/Window");
+
+require('./jsdom/living'); //Enable living standard features
 
 var canReadFilesFromFS = !!fs.readFile; // in a browserify environment, this isn"t present
 
@@ -21,6 +24,11 @@ function request() {
 exports.getVirtualConsole = function (window) {
   return window._virtualConsole;
 };
+
+exports.createCookieJar = function(){
+  return new CookieJar();
+};
+
 exports.debugMode = false;
 
 // Proxy feature functions to features module.
@@ -50,8 +58,8 @@ exports.jsdom = function (html, options) {
     parser: options.parser,
     url: options.url,
     referrer: options.referrer,
+    cookieJar: options.cookieJar,
     cookie: options.cookie,
-    cookieDomain: options.cookieDomain,
     resourceLoader: options.resourceLoader,
     deferClose: options.deferClose
   });
@@ -161,9 +169,11 @@ exports.env = exports.jsdom.env = function () {
       uri: config.url,
       encoding: config.encoding || "utf8",
       headers: config.headers || {},
-      proxy: config.proxy || null,
-      jar: config.jar !== undefined ? config.jar : true
+      proxy: config.proxy || null
     };
+
+    config.cookieJar = config.cookieJar || new CookieJar();
+    options.headers.cookie = options.headers.cookie || config.cookieJar.getCookieStringSync(config.url, { http: true });
 
     request(options, function (err, res, responseText) {
       if (err) {
@@ -181,11 +191,26 @@ exports.env = exports.jsdom.env = function () {
       config.html = responseText;
       config.url = res.request.uri.href;
 
+      var contentType = res.headers["content-type"];
+
       if (config.parsingMode === "auto" && (
-        res.headers["content-type"] === "application/xml" ||
-        res.headers["content-type"] === "text/xml" ||
-        res.headers["content-type"] === "application/xhtml+xml")) {
+        contentType === "application/xml" ||
+        contentType === "text/xml" ||
+        contentType === "application/xhtml+xml")) {
         config.parsingMode = "xml";
+      }
+
+      var cookies = res.headers["set-cookie"];
+
+      if (cookies) {
+        cookies = Array.isArray(cookies) ? cookies : [cookies];
+
+        cookies.forEach(function(cookieStr) {
+          config.cookieJar.setCookieSync(cookieStr, options.uri, {
+            http: true,
+            ignoreError: true
+          });
+        });
       }
 
       processHTML(config);
@@ -199,6 +224,7 @@ exports.serializeDocument = function (doc) {
 
 function processHTML(config) {
   var options = {
+    cookieJar: config.cookieJar,
     features: config.features,
     url: config.url,
     parser: config.parser,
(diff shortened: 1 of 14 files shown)


domenic:
> BTW, it introduces breaking change: config.resourceLoader doesn't receive cookieDomain in the arguments anymore. Because, it doesn't make sense with the current implementation.

So, I kind of forgot what cookieDomain does. Can you help me understand so I can write a better changelog? :)

inikulin:
`cookieDomain` was used to determine if we need to send Cookie header for the resource requests. It's semantics was broken overall:

1)If we have page loaded via URL resource loader still used `cookieDomain` to deside on Cookie header. So e.g. we had www.google.com loaded by jsdom, but it's cookies wasn't sent on resource requests because `cookieDomain` was set to `127.0.0.1` by default (see: #846)
2)This approach also broke specs compliance: by itself, domain information is insufficient to decide if we need to send cookies (we also need path, protocol, etc.)

Seems like this property was introduced to handle cases then you load page from the file system or using HTML source. So you can tell the page to act like it was loaded from the `cookieDomain`. I think the better way will be to completely get rid of `cookieDomain` and introduce `cookieUrl` or `aliasUrl` property. The latter looks better for me, because it has wider semantics, e.g. it will be useful to deal with the CORS (if we'll have it implemented one day).

Current PR uses `cookieDomain` in the following manner:
1)If it's set then `'https://' + cookieDomain` is used as document's `cookieUrl` (URL which is used to query `cookieJar` for the document's cookies).
2)If it's not set and page URL is provided then `document.URL` is used as document's `cookieUrl`.
3)If it's not set and page URL is not provided then `https://127.0.0.1` is used as document's `cookieUrl`.

Therefore, we don't need to provide `cookieDomain` in `config.resourceLoader `, because we provide `cookie` which was already obtained via querying `cookieJar` with `cookieUrl`.

domenic:
Hmm OK. Since it's going to be a breaking change anyway, let's remove cookieDomain entirely and just use `document.URL`. I don't think we want a fallback to localhost even; we can just use `about:blank` (i.e. the default `document.URL`).

I think people can already do "aliasUrl" by setting the `url` option in the `config` object.

inikulin:
OK. There is yet another thing that bothers me. I think we should need to cover the scenario when user wants to reuse cookie jar (e.g. like zombie.js does). Currently we don't have such API. We have a `config.jar` which is passed to the `request` lib, but this jar should be a [thin wrapper](https://github.com/request/request/blob/master/lib/cookies.js) around tough-cookie created by the [request.jar()](https://github.com/request/request/blob/master/lib/cookies.js) method. Therefore we can't reuse it on our side and end-user should additionaly `require('request')` to implement this functionality.
  
I suggest to get rid of `config.jar` property and let `config.document.cookie` property accept `CookieJar` object and also implement `jsdom.createCookieJar` method. So, end user will be able to use it this way:
```js
var cookieJar = jsdom.createCookieJar();

jsdom.env({
    url: 'http://google.com',
    document: {
        cookie: cookieJar
    },
        //...
    }
});

jsdom.env({
    url: 'http://code.google.com',
    document: {
        cookie: cookieJar
    },
        //...
    }
});
``` 
Please, let me know that do you guys think.

domenic:
> I suggest to get rid of config.jar property and let config.document.cookie property accept CookieJar object and also implement jsdom.createCookieJar method.

This seems like a good idea, although why not call it config.cookieJar?

inikulin:
`cookie` is more generic - we will be able to accept single cookie string, array of cookie strings or `CookieJar` object.

domenic:
Meh, dislike overloading, separate cookie and cookieJar options would be better.

inikulin:
Ok. BTW, I haven't implemented cookies for iframes yet. I saw you guys did some fixes in this area, so I decided to postpone it until the dust settles. I think it will be better to issue a separate PR for this.

inikulin:
@domenic @Sebmaster Guys, please review. I think we've done here. Regarding tough-cookie PR: we've just dealt with legal issues and is should be merged soon and we don't depend too much on it.

domenic:
Some style nits, but the big things to do are rebase and update docs. Otherwise ready to go!!

lib/jsdom.js
```diff
@@ -1,7 +1,9 @@
 var fs = require('fs');
 var path = require('path');
 var URL = require('url');
+var CookieJar = require('tough-cookie').CookieJar;
 
+var isArray = require('util').isArray;
```
domenic: prefer Array.isArray

lib/jsdom.js
```diff
@@ -185,13 +194,28 @@ exports.env = exports.jsdom.env = function () {
...
-        res.headers["content-type"] === "application/xhtml+xml")) {
+        contentType === "application/xml" ||
+        contentType === "text/xml" ||
+        contentType === "application/xhtml+xml")) {
         config.parsingMode = "xml";
       }
 
+      var cookies = res.headers["set-cookie"];
+
+      if(cookies) {
```
domenic: Style: space after " "

lib/jsdom/browser/Window.js
```diff
@@ -143,16 +143,35 @@ function Window(options) {
...
+      var cookieJar = window.document._cookieJar;
+      var cookieStr = cookieJar.getCookieStringSync(lastUrl, {http: true});
+
+      if (cookieStr) {
+        xhr.setDisableHeaderCheck(true);
+        xhr.setRequestHeader("cookie", cookieStr);
+        xhr.setDisableHeaderCheck(false);
       }
+
+      var setReceivedCookies = function() {
```
domenic: Style: space after function and if, a few places in this file.

lib/jsdom.js
inikulin: Arggh, forgot this one. Already replaced in other places.

lib/jsdom.js
inikulin: Really missing things like this in linter/[editor config](http://editorconfig.org/).

domenic:
Comment got lost, but I am definitely for adding jscs in addition to (or replacing, if it is truly a superset) JSHint.

inikulin:
@domenic Updated readme, please review.

domenic:
LGTM. Just needs a rebase.

inikulin:
@domenic Rebased

domenic:
Awesome, just waiting for the new tough-cookie release now right?

domenic:
Also why did we do

> Let options.document.cookie accept arrays.

?

inikulin:
> Awesome, just waiting for the new tough-cookie release now right?

Not really. Current version of tough cookie have some spec compilance issues, but in general they all was minor. So, we can release now with the current version. Since we have a tough-cookie dependency with `^` it will be automatically updated for new users/users who update jsdom once the new version of the tough-cookie will be released.

> Also why did we do

> Let options.document.cookie accept arrays.
> ?

Previous implementation allowed to set only one cookie string. E.g. I would like to set authentication cookie and locale cookie. With API it was impossible. Now you can pass two cookie strings in the array.

P.S. Need to rebase again =(

domenic:
Merged as bd5f563a628c2847cf9af8538d334bbbd4d62250 and 5.0.0 being released soon, woo!!

inikulin:
Hooray! :tada:

PR #1073, opened 2015-03-26
Support the eventInit parameter on core.Event

brandon-bethke-neudesic:
support the eventInit parameter on core.Event according to the specification here.
https://developer.mozilla.org/en-US/docs/Web/API/Event/Event

event = new Event(typeArg, eventInit);

eventInit - Optional. Is an EventInit dictionary, having the following fields:
"bubbles", optional and defaulting to false, of type Boolean, indicating if the event bubbles or not.
"cancelable", optional and defaulting to false, of type Boolean, indicating if the event can be canceled or not.

### DIFF
--- a/lib/jsdom/level2/events.js
+++ b/lib/jsdom/level2/events.js
@@ -25,9 +25,6 @@ core.Event = function(eventType, eventInit) {
       this._cancelable = false;
       if(eventInit.bubbles) this._bubbles = eventInit.bubbles;
       if(eventInit.cancelable) this._cancelable = eventInit.cancelable;
-
-      console.log(this._cancelable);
-      console.log(this._bubbles);
     }
 };
 core.Event.prototype = {
--- a/lib/jsdom/level2/events.js
+++ b/lib/jsdom/level2/events.js
@@ -9,7 +9,7 @@ var core = require("../level1/core"),
     defineSetter = utils.defineSetter,
     inheritFrom = utils.inheritFrom;
 
-core.Event = function(eventType) {
+core.Event = function(eventType, eventInit) {
     this._type = eventType || null;
     this._bubbles = null;
     this._cancelable = null;
@@ -19,6 +19,11 @@ core.Event = function(eventType) {
     this._timeStamp = null;
     this._preventDefault = false;
     this._stopPropagation = false;
+
+    if(arguments.length > 1 && eventInit) {
+      this._bubbles = eventInit.bubbles;
+      this._cancelable = eventInit.cancelable;
+    }
 };
 core.Event.prototype = {
     initEvent: function(type, bubbles, cancelable) {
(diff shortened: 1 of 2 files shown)


domenic:
Hi! I'm happy to have you contributing to jsdom :).

That said, could you please read CONTRIBUTING.md to get our guidance on how features are added? In particular, MDN is not a spec, and we definitely need tests in place for new features like this.

brandon-bethke-neudesic:
Sorry, this process is new to me. MDN is not the official standard, but it's accurate. In any case, here is the actual W3C DOM4 spec, showing the feature is supported. Section 4.2
http://www.w3.org/TR/dom/#event

event = new Event(type [, eventInitDict])
The optional eventInitDict argument allows for setting the bubbles and cancelable attributes via object members of the same name

brandon-bethke-neudesic:
The feature is also supported in the DOM Living Standard as referenced by jsdom's contributing guide. Section 3.2

https://dom.spec.whatwg.org/#interface-event

domenic:
Excellent! (And yeah, we use the original WHATWG DOM Standard, not the W3C fork.) So, let me walk you through a few minor tweaks that you can read off from the spec, now that you have it in hand:

- The `arguments.length > 1 && eventInit` test is not correct. Since EventInit is a dictionary, the code should instead follow the procedure at http://heycam.github.io/webidl/#es-dictionary. Implementing that whole procedure is probably overkill though. Instead I'd just do `if (eventInit === undefined || eventInit === null) { eventInit = {}; } if (typeof eventInit !== "object") { throw new TypeError("cannot convert eventInit argument to a dictionary"); }`.

- The better way to express the fact that those parameters are booleans is to just do `this._bubbles = Boolean(eventInit.bubbles)`, etc.

- It looks like our current behavior of setting `null` as the default isn't correct either. The type is `boolean`, not `boolean?`

So in the end I think this might match up better:

```js
// remove the current this._bubbles = null; this._cancelable = null;

if (eventInit === undefined || eventInit === null) {
  eventInit = {};
}
if (typeof eventInit !== "object") {
  throw new TypeError("cannot convert eventInit argument to a dictionary");
}

this._bubbles = Boolean(eventInit.bubbles);
this._cancelable = Boolean(eventInit.cancelable);
```

Make sense?

Tests look good, except we want to change the `null` one to not be null since that behavior is wrong. Also, it would be good to add a test or two for non-booleans, e.g. what happens if I pass `{ cancelable: 5, bubbles: 0 }`? (It should convert them to true/false.)

brandon-bethke-neudesic:
Thank you for doing the code review. I have implemented your changes and added more tests. I look forward to the PR being merged.

domenic:
Thank you so much for this! Merged as 9272d6d541e61f9e7b1990e19a7f0d71ab792581, with a 4.1.0 coming shortly.

PR #1075, opened 2015-03-26
Add engines advisory for npm

Sebmaster:
This is so npm outputs at least a warning when installing jsdom on nodejs instead of iojs.

@domenic I know I said I wouldn't, but it's so small, it's okaaaaay :S

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,9 @@
     "test-browser": "iojs ./test/browser-runner",
     "lint": "jshint lib/ && jshint test/"
   },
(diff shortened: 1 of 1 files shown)


domenic:
Engines is deprecated, I am pretty sure? npm will just produce a warning saying "don't use engines"?

Sebmaster:
That's engineStrict only, which would prevent the installation. The engine advice just prints a warning if the iojs version doesn't match.

PR #1082, opened 2015-04-09
1. Handling compression, 2. Using charset from headers

ssesha:
js resources are mostly served compressed from cdns eg. http://sd1.gemvara.net/static/bundle-bundle_site_head_39e0ef67d1c.js

### DIFF
--- a/lib/jsdom/level2/html.js
+++ b/lib/jsdom/level2/html.js
@@ -8,8 +8,7 @@ var core                  = require("../level1/core"),
     URL                   = require("url"),
     Path                  = require('path'),
     fs                    = require("fs"),
-    http                  = require('http'),
-    https                 = require('https');
+    request               = require('request');
 
 var isBrowser = Object.prototype.toString.call(process) !== "[object process]";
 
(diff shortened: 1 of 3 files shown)


domenic:
This isn't quite the right approach to this. We should just use the request package, instead of http/https.request. It already takes care of this stuff and we're already using it in a few places. It would then also handle redirects etc.

inikulin:
**FIXME**: request doesn't deal with the charset encoding. Seems like we need to use something like https://github.com/ashtuchkin/iconv-lite to deal with it + implement [this thing](http://www.w3.org/TR/html5/syntax.html#determining-the-character-encoding) in the parse5.

ssesha:
@domenic I have made the change to use the request module instead of http/https. I see that we get redirects and gzipped response handling taken care of. Haven't handled different charset encodings though. Can you review this?

lib/jsdom/level2/html.js
```diff
@@ -8,8 +8,8 @@ var core                  = require("../level1/core"),
     URL                   = require("url"),
     Path                  = require('path'),
     fs                    = require("fs"),
-    http                  = require('http'),
-    https                 = require('https');
+    zlib                  = require('zlib'),
```
domenic: No longer needed

domenic:
LGTM with one nit. Will merge (and leave a more detailed comment about encoding) when I get back to my computer for a few hours. Thanks!

ssesha:
removed the unused module

domenic:
Merged! Now to decide whether this is a bugfix or a feature...

ssesha:
Lets call it a feature, feature sounds cool

domenic:
Yeah, I went with a minor release https://github.com/tmpvar/jsdom/blob/master/Changelog.md#440 :)

PR #1087, opened 2015-04-17
Replace backticks with single ticks

framp:
The latest release is breaking the package on node v0.12.2:

    /.../jsdom/lib/jsdom.js:55
    throw new RangeError(`Invalid parsingMode option ${JSON.stringify(options.
                                         ^
    SyntaxError: Unexpected token ILLEGAL


inikulin:
Duplicate of the #1086

Issue #1026, opened 2015-02-13
submit event does not firing


wtfil:
Hi!

I am trying to use `jsdom` for running client-side tests. Everything seems alright, but I found than events does not working as expected (not like in browsers)

`form` does not triggering the `submit` event when  the `click` event on `button` occurred.

```html
<form action="">
	<button type="submit"></button>
</form>
```
I've created small repo to reproduce it
https://github.com/wtfil/jsdom-submit-bug

domenic:
@Joris-van-der-Wel thoughts? Not sure this fell under the purview of your changes but seems like you know the code pretty well by now :)

Joris-van-der-Wel:
haha ;)

The test uses  `<button type="submit">`, however the type attribute for `<button>` is not implemented in jsdom.
A `<input type="submit">` should work however.

Implementing both `<button type="submit">` and `<button type="reset">` should not be that hard, submitting and resetting is already implemented.

wtfil:
Thx for replay:)
I've checked `input[type=submit]` and it works pretty well!
These is not difference between `button` and `input` in my app, so I can easily switch to `input`. But I think, it would be nice to not depend on test environment :)

domenic:
Agreed! Glad to hear this should be a pretty easy fix. Maybe I can do it soon. (And yet, whenever I say that, it never seems to happen :-/)

marcandre:
I think this can be closed, as latest release seem to work fine.

PR #1088, opened 2015-04-17
Added click handler for submit button

rxgx:
Addresses #1026 where a form button with `type="submit"` should trigger a submit event when clicked.

cc: @Joris-van-der-Wel

### DIFF
--- a/lib/jsdom/level2/html.js
+++ b/lib/jsdom/level2/html.js
@@ -1077,6 +1077,17 @@ define('HTMLButtonElement', {
   proto: {
     get form() {
       return closest(this, 'FORM');
+    },
+    _eventDefaults: {
+      click: function (event) {
+        var target = event.target;
+        var form = target.form;
+        if (form) {
+          if (target.type === 'submit') {
+            form._dispatchSubmitEvent();
+          }
+        }
+      }
     }
   },
   attributes: [


lib/jsdom/level2/html.js
```diff
@@ -1270,6 +1270,14 @@ define('HTMLLegendElement', {
   proto: {
     get form() {
       return closest(this, 'FORM');
+    },
+    click: function() {
+      if (this.type && this.type === 'submit') {
```
Sebmaster: Do we need to check `this.type` explicitly? Shouldn't that be cought by the equality check?

lib/jsdom/level2/html.js
```diff
@@ -1270,6 +1270,14 @@ define('HTMLLegendElement', {
   proto: {
     get form() {
       return closest(this, 'FORM');
+    },
+    click: function() {
```
domenic: Why is this defined on HTMLLegendElement? You should be modifying the `click()` method of HTMLElement.

domenic: That will mean the below test will need to become something like `this.localName === "input" && this.type === "submit"`

domenic:
Needs tests!

lib/jsdom/level2/html.js
Sebmaster: Shouldn't it be either `HTMLButtonElement` or `HTMLInputElement`?

domenic: No, there is only one `click()` method in the HTML spec. Basically HTML uses spec-level polymorphism instead of JavaScript-level polymorphism :-/

Good point about button though; that means my above test is not quite right.

Joris-van-der-Wel:
Note that `<button>` is the same as `<button type="submit">`. submit is the [missing value default](https://html.spec.whatwg.org/#missing-value-default) (but not the invalid value default): https://html.spec.whatwg.org/#attr-button-type

Looks like jsdom already has a mechanism to deal with default actions for events, there's no need to touch the `click()` method. This is how the action is implemented for `<input>` elements. Look at line [371](https://github.com/tmpvar/jsdom/blob/c9c5a88069c5c4a656c6992de8e553f83e9d0e84/lib/jsdom/level2/html.js#L371), [402](https://github.com/tmpvar/jsdom/blob/c9c5a88069c5c4a656c6992de8e553f83e9d0e84/lib/jsdom/level2/html.js#L402), [972](https://github.com/tmpvar/jsdom/blob/c9c5a88069c5c4a656c6992de8e553f83e9d0e84/lib/jsdom/level2/html.js#L972)

So all you would have to do is add a `click` method to `_eventDefaults` for `HTMLButtonElement`. This also ensures that if your form is disabled, `click()` performs no action (per spec). And that your `click()` can be prevented using `preventDefault()`

domenic:
> Looks like jsdom already has a mechanism to deal with default actions for events, there's no need to touch the click() method

Gah, thanks for reminding us!!!

rxgx:
Thanks for the pointers to `_dispatchSubmitEvent`. I've seen a lot of tests where the count of the HTMLButton in the HTML file is duplicated. Am I able to edit these, in addition to the HTML, for level2 or have to create new files?

domenic:
Hmm I'm not sure what you mean? Is there something wrong with those tests that would mean you need to edit them?

I'd suggest adding a new test in https://github.com/rxgx/jsdom/tree/master/test/living-html, maybe called form-submission.js so we can use that file for other related stuff.

domenic:
Code LGTM now BTW

domenic:
Subsumed into #1154.

PR #1091, opened 2015-04-20
replaced grave accents with common single quotes

BenSower:
Replaced it with a "normal" quote and escaped the already existing quotes. (Imho it would be better to use single quotes for strings and double quotes for actual quotes in strings, but since all strings are declared using double quotes, I adapted the code style).


domenic:
Dupe of #1087, #1086.

Issue #317, opened 2011-09-21
Implement Ranges and Traversal


tmpvar:
<del>http://www.w3.org/TR/DOM-Level-2-Traversal-Range/</del>

- <ins>https://dom.spec.whatwg.org/#ranges</ins>
- <ins>https://dom.spec.whatwg.org/#traversal</ins>

We will need to find/write a test suite.

jeffdolle-sge:
+1 for this

domenic:
Spam +1 comments cause features to get moved to the back of the backlog. Please use the voting buttons instead. Deleting the recent +1s.

adrianheine:
I'm still there, btw, as is my half-finished code. I hope to get this done eventually.

acusti:
@adrianheine Does the TreeWalker implementation which landed help you get any closer to finishing your implementation?

philipp-spiess:
Any updates on this?

domenic:
https://twitter.com/slicknet/status/782274190451671040

caub:
I use sometimes `Range#createContextualFragment`, I don't really know why this useful method is there though

so I ended up doing this for tests
```js
global.document.createRange = () => ({
  createContextualFragment: str => JSDOM.fragment(str)
});
```

megamindbrian:
@domenic I think what @acusti was asking is what are the next steps to get this verified?

I am trying to use this:
https://github.com/PrismJS/prism/issues/896

Can someone recommend an alternative path?  I like the idea of having a highlighted diff and syntax highlighting with Prism.  Might even come out better than Github's highlighter.

tilgovi:
@adrianheine I might be interested in helping out if you share your WIP.

megamindbrian:
@adrianheine Is this relevant? http://tonsky.me/blog/chrome-intervention/

Fi1osof:
Got errors in jest tests.
```
TypeError: document.createRange is not a function

      at createRange (node_modules/popper.js/src/utils/findCommonOffsetParent.js:27:26)
      at findCommonOffsetParent (node_modules/popper.js/src/utils/getReferenceOffsets.js:16:85)
      at Popper.getReferenceOffsets (node_modules/popper.js/src/methods/update.js:29:28)
      at Popper.call (node_modules/popper.js/src/index.js:94:19)
      at node_modules/popper.js/dist/umd/popper.js:51:7
```

[...]

PR #1092, opened 2015-04-20
Implement NodeIterator

Joris-van-der-Wel:
Using https://dom.spec.whatwg.org/#interface-nodeiterator
Related to #317

I wrote my own test cases, I used istanbul to verify coverage (line and branch coverage is 100%).

The most tricky part about `NodeIterator` is its behavior when you remove nodes it iterates over. Each active `NodeIterator` must be notified about every node that is removed ("[removing steps](https://dom.spec.whatwg.org/#concept-node-remove-ext)").

The DOM specification assumes implementations have control over object lifecycles (destructors, weak references, etc). To implement it properly, without leaks, you would need to use weak references between the `Document` and `NodeIterator`s. We can not use such things in javascript (WeakMap and WeakSet are not adequate in this case) unless we would add a dependency on a compiled module (again :P). It looks like Ecma added this limitation on purpose to prevent side channel attacks.

Instead, I added an option that specifies the maximum number of `NodeIterator`s that removing steps will occur for. If this limit is reached, the oldest `NodeIterator` will no longer be tracked and it will start to throw whenever you try to access its state.

I tried to work out an alternative by asynchronously resolving the state of `NodeIterator`s. jsdom would store a lot of data about removals & insertions applied on nodes and the NodeIterator would use that data the next time you access it. This is pretty complex, but more importantly it would be  performance heavy because you would have to store a lot of state. Also the combination of iterating + modification would probably become something like `O(n^2)` instead of `O(n)`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ jsdom.env(config);
 - `config.features`: see Flexibility section below. **Note**: the default feature set for `jsdom.env` does _not_ include fetching remote JavaScript and executing it. This is something that you will need to _carefully_ enable yourself.
 - `config.resourceLoader`: a function that intercepts subresource requests and allows you to re-route them, modify, or outright replace them with your own content. More below.
 - `config.done`, `config.loaded`, `config.created`: see below.
+- `config.concurrentNodeIterators`: the maximum amount of `NodeIterator`s that you can use at the same time. The default is `10`; setting this to a high value will hurt performance.
 
 Note that at least one of the callbacks (`done`, `loaded`, or `created`) is required, as is one of `html`, `file`, or `url`.
 
--- a/lib/jsdom.js
+++ b/lib/jsdom.js
@@ -66,7 +66,8 @@ exports.jsdom = function (html, options) {
     cookieJar: options.cookieJar,
     cookie: options.cookie,
     resourceLoader: options.resourceLoader,
-    deferClose: options.deferClose
+    deferClose: options.deferClose,
+    concurrentNodeIterators: options.concurrentNodeIterators
   });
 
   if (options.created) {
@@ -242,7 +243,8 @@ function processHTML(config) {
     parser: config.parser,
     parsingMode: config.parsingMode,
     created: config.created,
-    resourceLoader: config.resourceLoader
+    resourceLoader: config.resourceLoader,
+    concurrentNodeIterators: config.concurrentNodeIterators
   };
 
   if (config.document) {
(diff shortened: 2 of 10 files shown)


domenic:
This is really amazing work. Left some nitpicky comments, but major <3 for implementing this. Very clean overall. And I think the max-node-iterators thingy is a pretty reasonable solution.

I asked in IRC if we should be trying to pass https://github.com/w3c/web-platform-tests/tree/13bff083fba249ed260966bca65319b1b35d3f34/dom/traversal/unfinished despite its "unfinished" name. Otherwise we should really consider submitting these test to web-platform-tests at some point.

Joris-van-der-Wel:
>> +- `config.concurrentNodeIterators`: the maximum amount of `NodeIterator`s that you can use at the same time. The default is `10`, setting this to a high value will hurt performance.

> Will it hurt performance or memory usage?

Strictly both. If `concurrentNodeIterators` was set to `Infinity`, each call to `createNodeIterator` would leak a little memory (until you stop using the entire `Document`). The performance is probably a bigger issue, since each removeChild() call has to go through all the `NodeIterator`s.

Joris-van-der-Wel:
> Why a single symbol instead of one per property?

While Symbol performance is good, it is still a little bit slower than regular properties.
I also think the code would look ugly with so many symbols, and annoying to type.

> Why use a well-known symbol instead of just Symbol()?

For debug-ability; It would be really annoying to have to use `obj[Object.getOwnPropertySymbols(obj)[0]]`. E.g. if I would want to log something internal to jsdom from within my own application.

domenic:
Hmm. I would rather not make it that easy to get at the internals. In browsers you can't get at them at all; ideally we'd use weak maps and hide them entirely, but they're too slow. So let's just use `Symbol()`.

Are you sure amortized one-time symbol + repeated prop lookup is slower than repeated symbol lookup?

Joris-van-der-Wel:
http://jsperf.com/symbol-vs-prop-vs-def-prop/13
(on my machine Symbol is 13% slower in chrome 42)

I did not test the "+ repeated prop lookup" part specifically though. Do you think it would be worth it to create a benchmark?

domenic:
I guess I don't mind too much. One day we'll figure out our overall story for private state and unify everything but in the meantime this is fine.

Joris-van-der-Wel:
Alright, i am done with the fixes.

I played around some more with Symbol. NodeIterator no longer exposes any jsdom specific properties. I added a private class called `NodeIteratorInternal` which contains all the interesting stuff (properties and methods). `NodeIterator` is now just a small class forwarding stuff to NodeIteratorInternal.

domenic:
Merged; thanks so much!

PR #1099, opened 2015-04-23
Implement ParentNode interface

### LINKED ISSUE #519: I want firstElementChild/lastElementChild and friends
They would be convenient. Recently had to use 

```js
var first = $(el).children().first()[0];
```

instead of

```js
var first = el.firstElementChild;
```

and felt super-sad.

Joris-van-der-Wel:
https://dom.spec.whatwg.org/#interface-parentnode
Fixes #519

* `children` should only be defined on `Document`, `Element` and `DocumentFragment`
* `children` should be a HTMLColleciton
* Implement firstElementChild
* Implement lastElementChild
* Implement childElementCount

There's more stuff in ParentNode, but as far as I can tell there is not yet a single browser which implements those. So I would rather wait (perhaps the api is not yet stable)

### DIFF
--- a/lib/jsdom/level1/core.js
+++ b/lib/jsdom/level1/core.js
@@ -390,18 +390,6 @@ core.Node.prototype = {
   DOCUMENT_FRAGMENT_NODE      : DOCUMENT_FRAGMENT_NODE,
   NOTATION_NODE               : NOTATION_NODE,
 
-  get children() {
-    if (!this._childrenList) {
-      var self = this;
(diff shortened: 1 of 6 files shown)


domenic:
I would like to implement the other things too, just for completeness :). Doesn't have to be in this PR though.

Joris-van-der-Wel:
I would rather see those in a different PR, yes. They are a bit more involved. For example relative selector querying, which is a CSS 4 thing. This would require patches in nwmatcher

PR #1102, opened 2015-04-25
DOMImplementation fixes

Joris-van-der-Wel:
It is possible to create a `Document` without a `Window` (null `defaultView`) using `DOMImplementation`: `newDocument = document.implementation.createHTMLDocument()`.

There are various places in jsdom that blow up on such a Document. I've fixed these.

There were also a few bad test cases that violate the specifications & browsers.


domenic:
Thanks! Merged with shortened commit messages.

PR #1105, opened 2015-04-27
[issue 1103] Resolve import stylesheet URL against baseUrl of current stylesheet

dbo:
Bugfix proposal which works for me.

Sorry, I don't have a test case attached.


IonicaBizau:
:+1:

domenic:
we can't accept fixes without test cases. When you have time to write one, please push that commit and we can reopen.

Issue #1049, opened 2015-03-01
Virtual console at startup


domenic:
Per the confusion in https://github.com/tmpvar/jsdom/pull/991#issuecomment-76561496, we should probably add some kind of option that allows setting up a virtual console at startup time. Not sure exactly what it would look like.

Alternately we could delay firing the console events for a tick?

jeffcarp:
Instead of receiving a `virtualConsole`, could you build one yourself and pass it in as a config option?

```javascript
var myConsole = new (require('events').EventEmitter)();

myConsole.on("log", function (message) {
  console.log(message);
});

jsdom.env({
  virtualConsole: myConsole
}};
```

On a different note, would it be useful to skip a layer of abstraction and let you just pipe output to a writable stream?

```javascript
jsdom.env({
  sendOutputTo: process.stdout
}};
```

domenic:
> Instead of receiving a virtualConsole, could you build one yourself and pass it in as a config option?

I think that was what I was envisioning in the OP, actually.

> On a different note, would it be useful to skip a layer of abstraction and let you pipe output to a writable stream?

Not really. The events are pretty important, especially given the upcoming #1039. And it's an easy enough adapter to write. I think it'd just be more code to support that kind of thing.

domenic:
> var myConsole = new (require('events').EventEmitter)();

Although this isn't great; we'd want `jsdom.createVirtualConsole()` probably.

PR #1108, opened 2015-04-28
Add createVirtualConsole

jeffcarp:
As outlined in #1049, this to fix the problem that it is currently impossible to capture console events between the time jsdom is initialized and `getVirtualConsole` is called. 

```js
var window = jsdom.jsdom().defaultView;
// Any console events fired here will be lost
var virtualConsole = jsdom.getVirtualconsole(window);
virtualConsole.on('log', console.log);
```

This PR allows you to create a virtual console and pass it in as a config option.

```js
var virtualConsole = jsdom.createVirtualConsole();
virtualConsole.on('log', console.log);

var window = jsdom.jsdom(null, {
  virtualConsole: virtualConsole
}).defaultView;
```

I also updated the README for this to be the default way of capturing console output, so that there is less confusion in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,9 +402,9 @@ var virtualConsole = jsdom.createVirtualConsole();
 
 virtualConsole.sendTo(console);
 
-var window = jsdom.jsdom(null, {
+var document = jsdom.jsdom(undefined, {
   virtualConsole: virtualConsole
-}).defaultView;
+});
 ```
 
 #### Get an event emitter for a window's console
@@ -418,9 +418,9 @@ virtualConsole.on("log", function (message) {
   console.log("console.log called ->", message);
 });
 
-var window = jsdom.jsdom(null, {
+var document = jsdom.jsdom(undefined, {
   virtualConsole: virtualConsole
-}).defaultView;
+});
 ```
 
 ## What Standards Does jsdom Support, Exactly?
--- a/README.md
+++ b/README.md
@@ -397,8 +397,13 @@ jsdom.env({
 
 ```js
 var jsdom = require("jsdom");
+
+var virtualConsole = jsdom.createVirtualConsole();
+
+virtualConsole.sendTo(console);
+
 var window = jsdom.jsdom(null, {
-  sendConsoleTo: console
+  virtualConsole: virtualConsole
 }).defaultView;
 ```
 
--- a/README.md
+++ b/README.md
@@ -397,22 +397,25 @@ jsdom.env({
 
 ```js
 var jsdom = require("jsdom");
-var window = jsdom.jsdom().defaultView;
-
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -397,22 +397,30 @@ jsdom.env({
...
 ```js
 var jsdom = require("jsdom");
-var window = jsdom.jsdom().defaultView;
 
-jsdom.getVirtualConsole(window).sendTo(console);
+var virtualConsole = jsdom.createVirtualConsole();
+
+virtualConsole.sendTo(console);
+
+var window = jsdom.jsdom(null, {
```
domenic: undefined not null

domenic: Also maybe just do `var document =` and omit `.defaultView`. If we ever do a major API overhaul it will probably become window-centric but for now jsdom's API is fairly document-centric so it makes sense to leave that in the examples.

lib/jsdom/browser/Window.js
```diff
@@ -68,7 +68,16 @@ function Window(options) {
   // HTMLFrameElement init function (see: level2/html.js).
   this._length = 0;
 
-  this._virtualConsole = new VirtualConsole();
+  if (options.virtualConsole) {
+    if (options.virtualConsole instanceof VirtualConsole) {
+      this._virtualConsole = options.virtualConsole;
+    } else {
+      throw new Error(
```
domenic: TypeError

test/jsdom/virtual-console.js
```diff
@@ -2,6 +2,7 @@
 
 var jsdom = require("../..");
 var EventEmitter = require("events").EventEmitter;
+var VirtualConsole = require("../../lib/jsdom/virtual-console");
```
domenic: Shouldn't test non-public APIs; use jsdom.createVirtualConsole() instead.

test/jsdom/virtual-console.js
```diff
@@ -21,10 +22,16 @@ var consoleMethods = [
   "warn"
 ];
 
-exports["jsdom.getVirtualConsole returns an instance of EventEmitter"] = function (t) {
+exports["VirtualConsole inherits from EventEmitter"] = function (t) {
```
domenic: I think it's OK to leave these tests as-is (as part of the don't-test-private-APIs thing).

domenic:
Looks great over all; thanks very much for taking this on. Some nits to clean up but I can take care of them later if you don't have time.

jeffcarp:
Cool! I can take care of them tonight if that's good.

jeffcarp:
Alright, changes have been made. Let me know if there's anything else.

domenic:
Thanks! Merged as dbf88666d1152576237ed1c741263f5516bb4005

Issue #822, opened 2014-07-07
HTML id and global variable


Mickael-van-der-Beek:
In normal browsers, if you have a DOM node containing an ```id```, the value of that ```id``` will be used as a key on the ```window``` object and the value will be a reference to the DOM node. e.g:

```<div id="testy">Hello World!</div>```

in this example, the ```window``` object will have a property called ```testy```.

Now, ```window['testy']``` will return the reference to the ```<div>``` we created earlier.

This is not the case in JSDOM.

domenic:
Fixed by #1116!

PR #1116, opened 2015-05-02
Improve named properties

Joris-van-der-Wel:
Fixes the following per html spec:

* Named properties no longer override existing properties. For example `<iframe name="addEventListener">`
* Named properties on the window also return an element or a list of elements based on their id (#822) and name. (currently only the `contentWindow` of iframes are returned)
* `getElementById` ignored  elements who had their id set *after* being inserted into the document

Also, I have removed a useless circular `require()` workaround. This workaround was causing issues. 
Node.js can figure it out if you set `module.exports` *before* you `require()`

Named properties should also exist for `<form>`, however I am not doing that in this PR because the tracking of form ownership needs a lot of work. This PR would get very big.

The tracking of named properties is very close to the spec, but not complete. The spec implements them using a default/fallback getter, which do not exist in v8 yet. The edge case where this difference matters has been documented using an inverted test case.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -9,6 +9,7 @@ var VirtualConsole = require("../virtual-console");
 var define = require("../utils").define;
 var inherits = require("../utils").inheritFrom;
 var EventTarget = require("../events/EventTarget");
+var namedPropertiesWindow = require("../living/named-properties-window");
 
 // NB: the require() must be after assigning `module.export` because this require() is circular
 module.exports = Window;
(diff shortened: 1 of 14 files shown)


domenic:
Amazing work as always \o/. The generic NamedPropertyTracker is quite helpful, as my new NamedNodeMap implementation also has to implement something similar.

Joris-van-der-Wel:
>  Amazing work as always \o/. The generic NamedPropertyTracker is quite helpful, as my new NamedNodeMap implementation also has to implement something similar.

:D 

Do note that indexed getters (window[0]) have different semantics, scripts are not allowed to set them. So a simple defineProperty will be appropriate for that.

Joris-van-der-Wel:
Updated

domenic:
Merged!

PR #1118, opened 2015-05-03
Overhaul attribute handling (but not Attr itself)

domenic:
See commit comments for details.

Would appreciate review from @Sebmaster and @Joris-van-der-Wel if you have the time :).

We need to figure out the fix for https://github.com/whatwg/dom/issues/31 before landing, I think.


Sebmaster:
The referenced bug in the spec is fixed now.

I looked over it a bit (very roughly) and didn't notice anything to nitpick, but I won't have a chance to test it myself.

Are there new w3c tests which cover attributes which we could add to our suite?

domenic:
> Are there new w3c tests which cover attributes which we could add to our suite?

There are, but they're tied to the tests for Attr being a non-Node, so I can't quite enable them just yet. But, enabling them (and ignoring those failures) is what found that spec bug.

PR #1124, opened 2015-05-07
implemented NonDocumentTypeChildNode

brandon-bethke-neudesic:
Imlemented NonDocumentTypeChildNode for core.Element according to specifications at: https://dom.spec.whatwg.org/#nondocumenttypechildnode

Please review and merge. Thank you.

### DIFF
--- a/lib/jsdom/living/non-document-type-child-node.js
+++ b/lib/jsdom/living/non-document-type-child-node.js
@@ -1,0 +1,45 @@
+"use strict";
+
+const defineGetter = require("../utils").defineGetter;
+
+module.exports = function (core) {
+  // https://dom.spec.whatwg.org/#nondocumenttypechildnode
+
+  const implementers = [core.Element, core.CharacterData];
+
+  implementers.forEach(function (Constructor) {
+    defineGetter(Constructor.prototype, "nextElementSibling", function () {
+      var element = this;
+      if (!element._parentNode) {
+        return null;
+      }
+      var sibling = null;
+      var index = element._parentNode._indexOf(element) + 1;
+      while (index < element._parentNode._childNodes.length) {
+        sibling = element._parentNode._childNodes[index];
+        if (sibling.nodeType === core.Node.ELEMENT_NODE) {
+          return sibling;
+        }
+        index++;
+      }
+      return null;
+    });
+
+    defineGetter(Constructor.prototype, "previousElementSibling", function () {
+      var element = this;
+      if (!element._parentNode) {
+        return null;
+      }
+      var sibling = null;
+      var index = element._parentNode._indexOf(element) - 1;
+      while (index !== -1) {
+        sibling = element._parentNode._childNodes[index];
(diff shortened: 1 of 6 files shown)


lib/jsdom/living/non-document-type-child-node.js
```diff
@@ -0,0 +1,36 @@
+"use strict";
+
+const defineGetter = require("../utils").defineGetter;
+
+module.exports = function (core) {
+  // https://dom.spec.whatwg.org/#nondocumenttypechildnode
+
+  const implementers = [core.Element];
```
domenic: Add `core.CharacterData`.

lib/jsdom/living/non-document-type-child-node.js
```diff
@@ -0,0 +1,36 @@
...
+  const ELEMENT_NODE = core.Node.ELEMENT_NODE;
+
+  implementers.forEach(function (Constructor) {
+    defineGetter(Constructor.prototype, "nextElementSibling", function () {
+      var element = this;
+      while ((element = element.nextSibling)) {
```
domenic: Please don't implement these in terms of other public APIs. If you look at the performance characteristics of nextSibling, you can see that repeated calls to it will be very bad. Instead, use a similar strategy to what `nextSibling`'s code does.

test/runner
```diff
@@ -46,6 +46,7 @@ var files = [
   "living-dom/query-selector-all.js",
   "living-dom/node-iterator.js",
   "living-dom/parent-node.js",
+  "living-dom/non-document-type-child-node.js",
```
domenic: Add to the browser test runner too please. Sorry that it's a pain.

domenic:
Thanks for working on this! I'm excited to have these supported.

test/living-dom/non-document-type-child-node.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+const load = require("../util").load(__dirname);
+
+exports["Element should implement NonDocumentTypeChildNode:nextElementSibling"] = function (t) {
```
domenic: Will need tests for CharacterData too (just testing `Text` and `Comment` is probably fine; no need to test `ProcessingInstruction`).

test/runner
brandon-bethke-neudesic: I don't understand what needs to be done here.

brandon-bethke-neudesic:
I have implemented all the changes, with the exception of the adding browser tests. I do not understand what needs to be done to add browser tests.

Sebmaster:
Add them in [this file](https://github.com/tmpvar/jsdom/blob/master/test/worker-runner.js). They need to be added to the `modules` array and the `fileFilter` default.

brandon-bethke-neudesic:
I added the browser tests according to my understanding. Please review changes and merge. Thank you.

domenic:
Thanks, merged as 691fca4b7790027a24d6c4cc106c4737a8b15b27!!

PR #1125, opened 2015-05-08
Load default html.css stylesheet. Fix #994.

### LINKED ISSUE #994: Showing element with jQuery does not update display style
I cannot 'show' an element with jQuery?

    it('should show elements', function() {
        document.body.innerHTML = '<style>.one {display: none;} </style><div class="one"><div class="inner">hello</div></div>';

        expect($('div.one').css('display')).to.equal('none');
        $('.one').show();
        expect($('div.one').css('display')).to.equal('block'); // fails, reports 'none'
    });

akhaku:
This probably needs some cleanup, so I'd appreciate some comments on where to move the relevant code - I've left some comments inline. Issue #994

### DIFF
--- a/lib/jsdom/browser/html.css
+++ b/lib/jsdom/browser/html.css
@@ -1,0 +1,1129 @@
+/*
+ * The default style sheet used to render HTML.
+ *
+ * Copyright (C) 2000 Lars Knoll (knoll@kde.org)
+ * Copyright (C) 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011 Apple Inc. All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


domenic:
Wow, this is really great! I didn't figure on us getting this fixed so fast :D. Comments inline...

domenic:
Thanks, merged as 89c1c77f9c553b8aebfbdaea6239f57fd3c22a9e!!

lib/jsdom/browser/html.css
```diff
@@ -0,0 +1,1129 @@
...
+
+tfoot {
+    display: table-footer-group;
+    vertical-align: middle;
+    border-color: inherit
+}
+
+/* for tables without table section elements (can happen with XHTML or dynamically created tables) */
+table > tr {
+    vertical-align: middle;
```
jeffcarp: Awesome PR - just curious, it seems like semicolon use is inconsistent in this file?

domenic: From the top of the file:

> This file is copied from https://chromium.googlesource.com/chromium/blink/+/96aa3a280ab7d67178c8f122a04949ce5f8579e0/Source/core/css/html.css

jeffcarp: Is there a reason it's inconsistent in that file?

Sebmaster: Looks consistent to me; last statement is always without semicolon.

Probably because it's not needed and that is blink's code style?

PR #1126, opened 2015-05-09
Add getVirtualConsole back to README

jeffcarp:
Realized that #1108 removed any documentation for `getVirtualConsole`. Since it's in the public API I felt like it should still be documented somewhere.


domenic:
Thanks! Merged as be5a90d5db037d8f65239ea57e12def5cb3fed24.

PR #1132, opened 2015-05-20
use xtend over custom impl.

dcousens:
Relatively simple pull request.

It just uses the `xtend` module over the custom implementation.

On the one hand,  it'd be nice to just use `Object.assign`,  but skipping that,  at least this way its less [untested] code that could (potentially, obviously, this is a very ideological example) potentially be erroneous.

Not really fussed either way,  just passing through :+1: , thanks for the module.

### DIFF
--- a/lib/jsdom.js
+++ b/lib/jsdom.js
@@ -5,6 +5,7 @@
 
 var fs = require("fs");
 var path = require("path");
+var xtend = require("xtend/mutable");
 var URL = require("url");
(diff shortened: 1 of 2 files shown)


domenic:
LGTM although we should add a TODO to replace with Object.assign once io.js gets V8... 4.4 I'd guess? Maybe 4.5. Will merge later.

domenic:
Merged as 8f5b1d232138e5d524762ee2f203da7b1ce75fd0; thanks!

PR #1134, opened 2015-05-21
Do not throw if a style attribute has syntax errors

Joris-van-der-Wel:
Do not throw if a style attribute has syntax errors. Same goes for `node.style.cssText`

As a bonus I updated lib/jsdom/level2/html.js for jscs and jshint

Thanks @chad3814 for fixing cssstyle


domenic:
Merged, thank you!

PR #1140, opened 2015-06-01
window.postMessage support

jeffcarp:
Here's a first draft of window.postMessage support.

```js
window.postMessage(message, targetOrigin, transfer)
```
Spec: https://html.spec.whatwg.org/#crossDocumentMessages

**Questions**:

- Is there a canonical place to get correct error message text? I used Chrome's for [lib/jsdom/living/postmessage.js](https://github.com/jeffcarp/jsdom/blob/postmessage/lib/jsdom/living/postmessage.js#L13).

**Deviations from the spec**:

- Does not implement `event.ports` (due to `MessagePort` not being in jsdom)
- Does not implement `event.source`
- Does not implement the third argument, `transfer`
- `message` is not cloned
- Is not asynchronous (to the best of my knowledge it all happens in the same event loop)
- Does not implement functionality where `targetOrigin` is `"/"`
- `targetOrigin` checking is implemented in this PR but is not fully integration tested

**Ideas**:

- Should VirtualConsole have an option to proxy messages from all windows? Currently the only way to get all console messages from all windows is to attach to them individually.
- Via discussion in #jsdom, should a warning be put in the README about the lack of cross-origin security enforcement?


lib/jsdom/living/postmessage.js
```diff
@@ -0,0 +1,25 @@
...
+module.exports = function (message, targetOrigin) {
+  let window = this;
+  let event;
+
+  if (!targetOrigin) {
+    throw new TypeError("'postMessage' requires a second argument 'targetOrigin'");
+  }
+
+  if (!isValidTargetOrigin(targetOrigin)) {
+    throw new SyntaxError("Failed to execute 'postMessage' on 'Window': " +
```
domenic: This is actually a DOMException type :-/. Super-confusing, I know.

lib/jsdom/living/postmessage.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
+const isValidTargetOrigin = require("../utils").isValidTargetOrigin;
+
+module.exports = function (message, targetOrigin) {
+  let window = this;
+  let event;
+
+  if (!targetOrigin) {
```
domenic: Should check `arguments.length < 2`, and then do `targetOrigin = String(targetOrigin)`. Passing `undefined` should be equivalent to passing `"undefined"`.

lib/jsdom/living/postmessage.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
+const isValidTargetOrigin = require("../utils").isValidTargetOrigin;
+
+module.exports = function (message, targetOrigin) {
+  let window = this;
+  let event;
```
domenic: Why not `const event` later.

lib/jsdom/living/postmessage.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
+const isValidTargetOrigin = require("../utils").isValidTargetOrigin;
+
+module.exports = function (message, targetOrigin) {
+  let window = this;
```
domenic: `const` not `let`

lib/jsdom/living/postmessage.js
```diff
@@ -0,0 +1,25 @@
...
+  if (!isValidTargetOrigin(targetOrigin)) {
+    throw new SyntaxError("Failed to execute 'postMessage' on 'Window': " +
+      "Invalid target origin '" + targetOrigin + "' in a call to 'postMessage'.");
+  }
+
+  if (targetOrigin !== "*" && targetOrigin !== window.origin) {
+    return;
+  }
+
+  event = window.document.createEvent("MessageEvent");
```
domenic: I think we have event constructors working... if so would be best to use them.

lib/jsdom/utils.js
```diff
@@ -277,3 +277,19 @@ exports.visitTree = function (root, callback) {
...
+  return (
+    parsedURL.protocol !== null &&
+    parsedURL.host !== null &&
+    parsedURL.path !== null
+  );
+}
+
+exports.isValidTargetOrigin = function (str) {
+  return (
+    str === "*" ||
```
domenic: Spec allows `"/"` also

lib/jsdom/utils.js
```diff
@@ -277,3 +277,19 @@ exports.visitTree = function (root, callback) {
   }
   visit(root);
 };
+
+function isValidAbsoluteURL(str) {
```
domenic: Put a `// TODO URL: be more rigorous` here so @Sebmaster can remember to base this on whatwg-url when that gets merged.

test/living-html/postmessage.js
```diff
@@ -0,0 +1,123 @@
+"use strict";
+
```
domenic: This file needs a lot more `const`s... hmm, gotta enable that eslint rule.

test/util.js
```diff
@@ -81,3 +81,15 @@ exports.todo = function (test, fn) {
     // add more as needed
   });
 };
+
+exports.injectIFrameWithScript = function (document, scriptStr) {
+  scriptStr = scriptStr || "";
+  let iframe = document.createElement('iframe');
```
domenic: Are we not linting this file? :( double quotes, const

test/util.js
```diff
@@ -81,3 +81,15 @@ exports.todo = function (test, fn) {
...
+  document.body.appendChild(iframe);
+
+  let scriptTag = iframe.contentWindow.document.createElement('script');
+  scriptTag.innerHTML = scriptStr;
```
domenic: textContent not innerHTML

lib/jsdom/living/postmessage.js
domenic: Actually it looks like you need to implement a new Event interface: https://html.spec.whatwg.org/multipage/comms.html#messageevent can be its own file in living.

lib/jsdom/living/postmessage.js
```diff
@@ -0,0 +1,25 @@
...
+  event.initEvent("message", false, false);
+  event.data = message;
```
domenic: Missing `origin`, `source`, and `ports`.

lib/jsdom/living/postmessage.js
```diff
@@ -0,0 +1,25 @@
...
+  window.dispatchEvent(event);
```
domenic: Should queue a task for this. I can't remember if we have something for task-queueing already, or whether just `setImmediate` is what you should go with.

lib/jsdom/living/postmessage.js
```diff
@@ -0,0 +1,25 @@
...
+  if (!targetOrigin) {
+    throw new TypeError("'postMessage' requires a second argument 'targetOrigin'");
+  }
+
+  if (!isValidTargetOrigin(targetOrigin)) {
+    throw new SyntaxError("Failed to execute 'postMessage' on 'Window': " +
+      "Invalid target origin '" + targetOrigin + "' in a call to 'postMessage'.");
+  }
+
+  if (targetOrigin !== "*" && targetOrigin !== window.origin) {
```
domenic: Needs `/` support. In general try to follow the spec's step 9 here more exactly.

lib/jsdom/living/postmessage.js
domenic: I think we should structured clone this. https://github.com/traviskaufman/cycloneJS was quite good last I checked.

test/living-html/postmessage.js
```diff
@@ -0,0 +1,123 @@
...
+    window.parent.iframeReceiver.contentWindow.postMessage("ack", "*");
+  `);
+
+  window.onload = function () {
+    test.ok(window.postMessageEvent.data === "ack");
+    test.done();
+  };
+};
+
+exports["postMessage respects absolute URL targetOrigins"] = function (test) {
```
domenic: s/respects/silently rejects/ I think?

test/living-html/postmessage.js
domenic: s/test/t/g in this file

domenic:
Overall a great start!! Very excited for this :D. The biggest missing piece is I think MessageEvent (some work) and structured cloning (hopefully just hooking up CycloneJS).

> Is there a canonical place to get correct error message text? I used Chrome's for lib/jsdom/living/postmessage.js.

Nope, implementations can make up their own errors---e.g. for localization purposes.

> - Does not implement event.source
> - Does not implement functionality where targetOrigin is "/"

Any reason for these? Shouldn't be too hard...

> Should VirtualConsole have an option to proxy messages from all windows? Currently the only way to get all console messages from all windows is to attach to them individually.

I think we should just automatically make this happen. When creating a sub-window, pass in your current virtual console.

lib/jsdom/living/postmessage.js
domenic: Hmm. Realizing that we'd have to add clone callbacks for all DOM classes gives me pause. We can leave that for the future ... eventually ...

test/living-html/postmessage.js
domenic: I think it should be possible to add tests that don't do script injection but instead reach directly into the iframe and call its postMessage/set its event handlers. No need to duplicate the existing ones all in that style, but one or two would be nice.

lib/jsdom/living/postmessage.js
jeffcarp: How about `arguments.length !== 2` with the error message `"'postMessage' requires 2 arguments: 'message' and 'targetOrigin'"`?

domenic: Extra arguments are allowed, so < 2 is better.

jeffcarp: Oops :grimacing: sounds good - I'll update the PR.

lib/jsdom/living/postmessage.js
jeffcarp: Is the correct invocation `throw new DomException("SyntaxError", "Failed to execute 'postMessage'...`?

domenic: Unfortunately not quite... I think it's `throw new DOMException(DOMException.SYNTAX_ERR, "Failed...")`.

You'll also need to require() the dom exception module, but thankfully that's nice and isolated these days.

jeffcarp: Cool, thanks! :smile:

lib/jsdom/living/postmessage.js
jeffcarp: Just to clarify, since I can't find a ton more information on the `/` option, using a `/` as `targetOrigin` means that it only proceeds if `sendingWindow.origin === targetWindow.origin`?

> If the targetOrigin argument is a single literal U+002F SOLIDUS character (/), and the Document of the Window object on which the method was invoked does not have the same origin as the responsible document specified by the entry settings object, then abort these steps silently.

test/living-html/postmessage.js
jeffcarp: The script injection is a remnant from when I was originally using separate HTML files for each test, similar to [test/window/frame.js](https://github.com/tmpvar/jsdom/blob/c59ad9a8f20508fc2b199195891ef62999ba7b8c/test/window/frame.js)

Now that the strings are completely static, as you pointed out, this approach doesn't make much sense anymore. Would it be alright if I removed the script injection and reached directly into the iframes for all tests?

lib/jsdom/living/postmessage.js
domenic: Yes, exactly.

test/living-html/postmessage.js
domenic: Sure! Maybe keep one script injection test just for coverage's sake.

jeffcarp:
Thanks for all the help so far! I addressed most of the issues but I need a bit of help with these two:

`MessageEvent`

- I wasn't sure exactly what this entails. I attempted to implement it in the files below, let me know if this is going in the right direction.
  - https://github.com/tmpvar/jsdom/pull/1140/files#diff-a6fe01c85d52438d462154c3e7eb7928R1
  - https://github.com/tmpvar/jsdom/pull/1140/files#diff-fa1e1623f9e3abbd91795adcfe328f87R76

Supporting `event.source`, `event.origin`, and `targetOrigin === "/"`

- I may be missing something obvious, but what is the best way to get the sending window in `lib/jsdom/living/postmessage.js`? The only way I can think of is to use some crazy `caller` reflection which seems less than kosher.
  - Relevant location: https://github.com/tmpvar/jsdom/pull/1140/files#diff-aaa4c6887643e04d4d3fdcd8d296c962R21

Also:

- Task queuing
  - I couldn't find any sort of event queuing code so I used `setImmediate`.

domenic:
MessageEvent is looking OK, but a few notes:

- Keep the inheritance in the same file
- Call the super-constructor so that you get _type, _bubbles, _cancelable set correctly, as well as all the rest of the state. This also allows you to omit the `typeof eventInit` test.
- Don't define getters on `this`. Use underscored properties on `this`, then define getters on the prototype.

To inherit from event, you'll probably have to move from the nice simple model where `module.exports` is the MessageEvent constructor, to the more convoluted model where you have a factory taking a `core` object, like in https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/element.js

> I may be missing something obvious, but what is the best way to get the sending window in lib/jsdom/living/postmessage.js

Hmm, yeah this may indeed be hard or impossible due to the fact that (unlike browsers) we don't have separate copies of each prototype for each window.

In fact it's going to be impossible in the general case, because what if io.js code calls postMessage?? What's the "calling window" then? Ugh.

Maybe just add a TODO linking to this comment for those.

jeffcarp:
Cool - cleaned up MessageEvent, turned it into a factory, and added TODOs in appropriate places.

> because what if io.js code calls postMessage?? What's the "calling window" then?

Since this is unspecified behavior, can jsdom just leave `event.source` empty? I don't think it would break existing code and it sounds like a reasonable limitation.

domenic:
Yeah probably. Would still have to figure out how to get the source window in the other cases though. Normally the browser figures that out because `destWindow.postMessage === sourceWindow.HTMLIFrameElement.prototype.postMessage` whereas `destWindow.postMessage !== destWindow.HTMLIFrameElement.prototype.postMessage`. But in jsdom they are the same.

domenic:
This is officially good enough that I'm going to adopt it and do any remaining tweaks myself. Thanks!! Working on it now.

domenic:
Merged as 14323d51e96f8a3aa5a8fe1de2a6c506473d6f1d with a few tweaks. Woohooooo!!!

jeffcarp:
Awesome!! Thanks for all the help!

robot dance

bchociej:
Hi all,

Just wanted to let you all know that the missing `event.origin` field breaks some tests of mine that use Stripe.js, which calls `e.origin.replace(...)` on the message event unconditionally. I'd be happy to dig in and see if I can help out with adding that functionality.

(Here is the exact error: ` Error: Cannot read property 'replace' of undefined (lib/stripe-debug.js:2051)`. That file is provided by Stripe at https://js.stripe.com/v2/stripe-debug.js as a debug version of their library.)

Meanwhile, if anyone else runs into this problem, the following workaround is sufficient for my integration tests (though you should not use it in real code!).

```javascript
window.addEventListener('message', function(e) {
	e.origin = window.location.origin;
});
```

domenic:
Hmm, maybe @Sebmaster can work on adding at least a dummy `event.origin` in #1258?

Sebmaster:
Yeah, it's already automatically generated, should work, just doesn't do any validation.

domenic:
It might be nice to add a line similar to @bchociej's, e.g. in the code that dispatches postMessage, it could set origin correctly. We can do that after #1258 lands... I'll open a new tracking bug.

dijs:
Is there an example usage of postMessage anywhere? I cannot get it to work properly.

My code:

```js
import { expect } from 'chai';
import sinon from 'sinon';
import jsdom from 'jsdom';

describe('post messaging', () => {
  it('should ...', (done) => {
    jsdom.env({
      url: "http://bar.com/",
        const frame = window.document.createElement('iframe');
        frame.setAttribute('src', 'http://foo.com/');
        window.document.body.appendChild(frame);
        console.log('Sending post message');
        window.addEventListener('message', e => {
          console.log('Window received message', e.data);
          done();
        });
        frame.contentWindow.postMessage({
          app: 'payment-prototype',
          type: 'ready',
        }, '*');
      } // done
    }); // env
  });
});
```

domenic:
It looks like you are posting the message to `frame.contentWindow` but adding your message handler to `window`.

dijs:
I am trying to post a message from the iframe to it's parent window.

domenic:
Then you should use `window.postMessage`, not `frame.contentWindow.postMessage`.

dijs:
Like this?

```js
jsdom.env({
      url: "http://bar.com/",
        const frame = window.document.createElement('iframe');
        frame.setAttribute('src', 'http://foo.com/');
        window.document.body.appendChild(frame);
        console.log('Sending post message');
        frame.contentWindow.addEventListener('message', e => {
          console.log('iframe received message', e.data);
          done();
        });
        window.postMessage({
          app: 'payment-prototype',
          type: 'ready',
        }, '*');
      } // done
    }); // env
```

domenic:
No. Now you are listening for the message on `frame.contentWindow`, but posting it to `window`. You need to put the message event handler on the same place you call `postMessage` on.

This is perhaps a better question for StackOverflow, as it is not jsdom specific. If you try code like that in a browser you will find it does not work either. See https://jsbin.com/jenututevu/edit?html,output

dijs:
You are right, I guess I do not fully understand this messaging. Sorry to waste time.

scarletsky:
Hi guys.
Just want to ask when will `event.origin` and `event.source` implement in postMessage ? Without these two property, my following test always failed. 😢 

```js
function onMessage(event) {
    event.source.postMessage('data', '*');
}
```

bchociej:
This is a really old pull request thread; it might be better to post a new issue or see if there's a jsdom chat/discussion someplace.

That said, if I recall correctly, it is essentially impossible to detect origins for emulated postMessage calls in the browser due to security restrictions.

scarletsky:
Thanks for your reply. So I think I should refactor my tests without
`event.source`.

Ben Chociej <notifications@github.com> 于2018年7月31日周二 上午2:29写道：

PR #1148, opened 2015-06-11
document.open() should return the Document on which the method was invoked

ryanseddon:
As mentioned in the last step of the [spec](https://html.spec.whatwg.org/multipage/webappapis.html#opening-the-input-stream:document-37) `document.open` should return the `Document` it was invoked on.

### DIFF
--- a/lib/jsdom/level2/html.js
+++ b/lib/jsdom/level2/html.js
@@ -297,6 +297,7 @@ inheritFrom(core.Document, core.HTMLDocument, {
     this._childNodes = [];
     this._documentElement = null;
     this._modified();
+    return this;
   },
   close : function() {
     this._queue.resume();
--- a/test/level2/html.js
+++ b/test/level2/html.js
@@ -19938,5 +19938,13 @@ exports.tests = {
 
     test.ok(div.toString() === '[object HTMLDivElement]', 'div.toString() should return "[object HTMLDivElement] just like a browser');
     test.done();
+  },
+
+  document_open_return_self: function(test) {
+    var doc = load("document");
+    var docOpen = doc.open();
+    doc.close();
+    test.ok(doc === docOpen, 'doc.open() should return the Document on which the method was invoked');
+    test.done();
   }
 };


ryanseddon:
So I have a scenario where I'm getting some weird behaviour with `this.body`.

I have the following code:

```js
var doc = iframe.contentWindow.document;

doc.open().hasLoaded = function() {
  // inject contents into iframe
  this.body.innerHTML = '...';
};
doc.write('<body onload="document.hasLoaded();">');
doc.close();
```

With this PR jsdom now correctly calls the onload method, but now `this.body` inside `hasLoaded` gets the parent body and not the iframe body like all actual browsers do. Any ideas on how to fix this?

Doing `this.getElementsByTagName('body')[0]` works correctly.

Sebmaster:
Haven't looked at it but my idea is this: the string given to even handlers is executed via `vm.runInContext`. However, we don't create seperate sandboxes for iframes so the globally defined document applies.

No idea how the reference to `this` fits in there though.

ryanseddon:
If I change [level2/html.js#L351-L358](https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/level2/html.js#L351-L358)

```js
get body() {
  var body = this.getElementsByTagName('body')[0];

  if (!body) {
    body = this.getElementsByTagName('frameset')[0];
  }
  return body;
}
```

It works fine.

Sebmaster:
We'd have to do that for every getter though. We'll need to get to the bottom of this.

ryanseddon:
Yep I have a feeling it's something here [level2/html.js#L166-L171](https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/level2/html.js#L166-L171)

ryanseddon:
Any idea on this?

domenic:
Looking at this problem it doesn't seem to be as weird as @Sebmaster is implying. (I might be missing something though.)

@ryanseddon I cannot reproduce within our test framework. Can you add a failing test? Here was my attempt; instead it just passes instead.

```js
  'iframe window, body, documentElement properties': function (t) {
    const document = jsdom.jsdom("<!doctype html>");
    const iframe = document.createElement("iframe");
    const window = document.defaultView;

    iframe.onload = function () {
      const iDocument = iframe.contentWindow.document;

      iDocument.hasLoaded = function () {
        t.notEqual(window, iframe.contentWindow, "window should not be equal");
        t.notEqual(window.document, iDocument, "document should not be equal");
        t.notEqual(window.document.documentElement, iDocument.documentElement,
          "documentElement should not be equal");
        t.notEqual(window.document.body, iDocument.body, "body should not be equal");

        t.equal(this, iDocument, "this should be correct");
        t.equal(this.defaultView, iframe.contentWindow, "window should be correct");
        t.equal(this.documentElement, iDocument.documentElement, "documentElement should be correct");
        t.equal(this.body, iDocument.body, "body should be correct");

        t.done();
      };

      iDocument.write("<body onload='document.hasLoaded();'></body>");
    };

    document.body.appendChild(iframe);
  }
```

(in test/window/frame.js)

domenic:
Anyway, merged as e136b5802ffeb222ed298355deda0ccb4776730c; maybe open another issue with the failing test for the other thing?

PR #1150, opened 2015-06-14
Benchmarking

Joris-van-der-Wel:
I would like for jsdom to care more about performance. This change adds a simple way to create benchmarks using http://benchmarkjs.com/. 

When the benchmarks are run in node, only the performance of jsdom is measured. This is useful to compare the effects of a proposed patch or to find out where a performance regression was introduced using `git bisect`.

If the benchmarks are run in chrome, a comparison will be made with the native DOM of chrome. This is useful to find good spots to improve jsdom. For example one of the benchmarks I included in this PR shows that `appendChild` is extremely slow compared to chrome if there are a lot of parents _(appending elements vertically appears to be something like `O(n^2)` instead of `O(n)`)_.

Currently the benchmarks scripts all use `var` because strict mode is not yet supported. But I will rebase this PR when support for it lands in the library I am using: https://github.com/bestiejs/benchmark.js/pull/97

Another point of interest is that if you specify setup, tearDown, cycle functions, they will actually all be part of the same function, so they are all in the same variable scope. You can see this in `benchmark/dom/tree-modification.js`. I am not sure if this is a desired style? Alternatives would include using globals or using `this` (must be non enumerable) ~~or a big closure around the benchmark.~~

A couple points on my implementation:

* The runner does not use `fs` at all, every benchmark file is included with a static `require()`. This makes it easier to browserify. This is different from jsdom's unit test runner
* You can specify what benchmarks to run using a command line flag: `node benchmark/runner -s jsdom/api,dom/construction/createElement`
* The browser bundle is generated using `node benchmark/runner --bundle`, you can then open the browser-runner.html in chrome
* You use the console (F12) in chrome to run benchmarks, it also reports to the console. There is no fancy html UI. This is much simpler and lets you more easily do fancy things like profiling: `console.profile(); run('dom/construction/createElement', 'jsdom'); console.profileEnd();`
* I included a few examples in this PR to demonstrate all the features (synchronous test, setup function, async test, etc), but this is of course not complete

Here is some sample output:
```
# dom/construction/createTextNode #
native x 1,848,191 ops/sec ±18.95% (26 runs sampled)
jsdom  x 6,000,818 ops/sec ±46.22% (64 runs sampled)
Fastest is "jsdom", which is 3.25x faster than the slowest "native"

# dom/tree-modification/appendChild: many parents #
native x 3,983,773 ops/sec ±1.45% (5 runs sampled)
jsdom  x 1,863 ops/sec ±97.50% (22 runs sampled)
Fastest is "native", which is 2138.46x faster than the slowest "jsdom "
```

Perhaps contributing.md should be updated to ask people to run a benchmark if they change something that might impact performance?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ v8.log
 node_modules
 test/worker-bundle.js
 npm-debug.log
+benchmark/browser-bundle.js
--- a/.jscsrc
+++ b/.jscsrc
@@ -27,6 +27,10 @@
 
   "excludeFiles": [
     "node_modules/**",
+
+    "benchmark/browser-bundle.js",
+    "test/worker-bundle.js",
+
     "lib/jsdom/browser/htmltodom.js",
     "lib/jsdom/browser/default-stylesheet.js",
     "lib/jsdom/level1/core.js",
--- a/.jshintignore
+++ b/.jshintignore
@@ -1,3 +1,6 @@
+benchmark/browser-bundle.js
+test/worker-bundle.js
+
 lib/jsdom/browser/htmltodom.js
 lib/jsdom/browser/default-stylesheet.js
 lib/jsdom/level1/core.js
--- a/benchmark/browser-runner.html
+++ b/benchmark/browser-runner.html
@@ -1,0 +1,18 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>JSDom benchmarks</title>
+    <script src="browser-bundle.js"></script>
+    <script>
+      window.run = require("jsdom-browser-runner");
+    </script>
+  </head>
+  <body>
+    <p>Execute <code>run('dom/construction/createElement')</code> to run a benchmark.
+      Or <code>run()</code> to run them all.
+      You can also specify to only run a benchmark for a specific implementation,
+      "jsdom" or "native": <code>run('dom/construction', 'jsdom')</code>.
+    </p>
+  </body>
+</html>
--- a/benchmark/browser-runner.js
+++ b/benchmark/browser-runner.js
@@ -1,0 +1,48 @@
+"use strict";
+
+const consoleReporter = require("./console-reporter");
+const pathToSuites = require("./path-to-suites");
+const benchmarks = require(".");
+
+module.exports = function (suites, documentImplementation) {
+  if (typeof suites === "string") {
+    suites = suites.trim().split(/,/);
+  }
+
+  let suitesToRun = pathToSuites(benchmarks, suites);
+  suitesToRun.forEach(consoleReporter);
+
+  suitesToRun = suitesToRun.map(function (suite) {
+    let runSuite = suite;
+    if (documentImplementation) {
+      runSuite = suite.filter(function (benchmark) {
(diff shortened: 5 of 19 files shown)


domenic:
Overall looking really good. I commented mostly on the benchmark files themselves, since that's what we'll be spending the most time maintaining in the future.

> The runner does not use fs at all, every benchmark file is included with a static require().

Could we experiment with generating the repetitive "manifest" files, like jsdom/benchmark/index.js, benchmark/jsdom/index.js, etc.? That is, write a script that generates them before bundling/running the benchmarks, and add the actual files themselves to .gitignore? (Also the script should output a comment saying "these files are generated" at the top.)

It would be cool to end up with a usable solution here that we can eventually care over to our unit tests.

Joris-van-der-Wel:
Made a lot of improvements today. 

A lot of style changes. Also the benchmark functions are no longer compiled, this took some time getting correct. 

I am not yet sure if I am happy with how the closures are set up now:
```javascript
exports["appendChild: many parents"] = (function () {
  const DEPTH = 500;
  let itData;
  let it;

  return suite({
    setup: function (document) {
      itData = new Array(this.count);

      for (let i = 0; i < this.count; ++i) {
        let nodes = new Array(DEPTH);

        for (let n = 0; n < DEPTH; ++n) {
          nodes[n] = document.createElement("div");
        }

        itData[i] = nodes;
      }

      it = 0;
    },
    fn: function () {
      let nodes = itData[it];

      for (let n = 1; n < DEPTH; ++n) {
        nodes[n - 1].appendChild(nodes[n]);
      }

      ++it;
    }
  });
}());
```

I have not yet thought about the manifest idea.

Joris-van-der-Wel:
I'm done with improvements unless someone has something to note.

I think that generating manifest files is not worth the effort. Adding a new benchmark file or test file is relatively rare. I prefer to be a little more verbose instead of doing a lot of clever things, it makes things more straightforward. 

It would also complicate things if your test file is using assets which should not be added to the manifest. For example `test/living-dom/attributes.js` uses the file `files/minimal.html`, but that html should obviously not be added to the manifest. A file such as `test/util.js` should not be added either. So you would need to rely on specific path matching, or on scanning javascript files for annotations.

domenic:
OK, all merged! We'll have to work on incorporating this into our workflow from now on. Your help prompting on PRs and such would be appreciated.

Also, maybe you could add a quick `npm run benchmark` script, and put that in Contributing.md? Maybe add the script in to Travis? (Although [I am not sure that Travis would be reliable to compare across runs](https://twitter.com/domenic/status/614388828816584704) ... maybe we'd do some kind of `git checkout HEAD~1` then benchmark the two?)

Joris-van-der-Wel:
Sure I'll do that in a bit.

I am also hesitant to use travis because it is virtualized. Other virtual machines could steal cpu in the middle of a benchmark. So even if you do `npm run benchmark; git checkout HEAD~1; npm install; npm run benchmark`, you could see a huge difference suddenly, even though all the change did was update README.

The solution would be to use a dedicated machine (which has a cpu without a turbo boost for bonus points), but I am not sure that is worth the effort

domenic:
Yeah I guess we could just ask contributors to run locally and report back. A dedicated machine would be pretty cool though :P.

PR #1158, opened 2015-07-04
Return instance from virtualConsole.sendTo

jeffcarp:
I noticed a possible shorthand when dealing with the virtual console and wanted to propose it here.

This PR makes `VirtualConsole.sendTo` return `this` so that you can write the shorthand:

```
jsdom.env("<p></p>", {
  virtualConsole: jsdom.createVirtualConsole().sendTo(console)
  done: function (err, window) {}
});
```

It also updates one place in the README to illustrate how it's done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,12 +398,8 @@ jsdom.env({
 ```js
 var jsdom = require("jsdom");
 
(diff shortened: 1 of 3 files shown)


domenic:
Nice! Published as 5.6.0.

Issue #1039, opened 2015-02-20
Replace node.raise()/node.errors/NOT_IMPLEMENTED with virtual console logging


domenic:
Supercedes #304.

Now that we have the virtual console infrastructure, we should be emitting these kinds of non-fatal errors (CSS parsing errors, not-implemented functionality errors, etc.) there. I am thinking the event name would be `"jsdomError"` or `"jsdomWarning"` maybe.

domenic:
Oh and script errors should trigger window.onerror.

Issue #587, opened 2013-03-18
window.onerror


mikermcneil:
Does jsdom support `window.onerror`?  I've set window.onerror, but haven't been able to get it to fire.  Perhaps I'm going about this the wrong way?

Thanks!

Sebmaster:
@domenic Was this fixed by #1159?

domenic:
Oh, yes it was!

PR #1159, opened 2015-07-05
Better errors

domenic:
This PR is meant to mainly address #1039 and #587. It implements `window.onerror` (including its funky parameter stuff) and `ErrorEvent`, then uses that to catch all script errors, instead of using our current system of a non-standard `raise` method that sends things to `node.errors` or `document.errors` or whatever.

It then proceeds to use the virtual console for all other error cases currently handled by raise. This is done via a new virtual console event, `jsdomError`. In sum, the virtual console now receives `jsdomError`s in the following cases:

- Errors loading scripts when using `jsdom.env`'s `scripts` array
- Script execution errors that are not handled by a window `onerror` event handler that returns `true` or calls `event.preventDefault()`
- Calls to methods, like `window.alert`, which jsdom does not implement, but installs anyway for web compatibility.

Finally, since script errors are now handled in this way, they are removed from the initialization flow. This means that the callbacks for `env` now take a single error object, instead of an array of errors; hooray!! This part is definitely backward-incompatible, although some of the rest is probably too.

Besides code review, I'd really appreciate comments on some of the details of this virtual console stuff. /ccing @jeffcarp for that purpose in particular, but any jsdom users are welcome. Here are some of my questions.

- Is one event, jsdomError, good enough? Or should we split this into one event per way that it's used? (See https://github.com/tmpvar/jsdom/commit/279b21d2fe362bc69bfcd98e11d3175b948945b3#diff-04c6e90faac2675aa89e2176d2eec7d8R451) If we don't split it into multiple events, should we perhaps add a .code property or something?
- What do you think of the use of Error objects with .detail properties?
- Should we use this mechanism for all resource loading errors, and not just jsdom.env script loading? I tend to think "yes."
- Should we forward these errors to nonVirtualConsole.error or similar, when doing sendTo?
- Will this new setup, especially for jsdom.env, "hide" errors that previously were easy to spot? Not everyone sets up a virtual console... Maybe we should store jsdomErrors at least in the virtual console for later inspection?

(Re: Travis: the build passes locally; presumably it's the browser build that's broken. I'll look into that soon.)


jeffcarp:
Cool! Looks like a great improvement! :sunglasses:

tl;dr: Better understanding the way people are using the virtual console (or not using it) might give the best answers to these questions.

Thoughts on specific points:

> Is one event, jsdomError, good enough? Or should we split this into one event per way that it's used? (See 279b21d#diff-04c6e90faac2675aa89e2176d2eec7d8R451) If we don't split it into multiple events, should we perhaps add a .code property or something?

I'd advocate for keeping it one event. The ability to add one handler and be confident you're catching all possible jsdom errors sounds like the most useful option. If you have different types of errors that you have to listen for and jsdom adds a new type of error, you'll miss those errors until you add listeners for them. 

Adding a `.code` property may or may not be useful based on whether `jsdomError` will be consumed more by machines or by people.

> What do you think of the use of Error objects with .detail properties?

I'm not sure I understand this completely. The `.detail` property is the original Error object?

> Should we use this mechanism for all resource loading errors, and not just jsdom.env script loading? I tend to think "yes."

I'm :+1: on this. It seems like this follows what browsers do.

> Should we forward these errors to nonVirtualConsole.error or similar, when doing sendTo?

This sounds slightly magical to me, but I can't think of a better alternative.

This is kind of a crazy idea, but is there a strong case against simply using `VirtualConsole.error` instead of `VirtualConsole.jsdomError` entirely? That would get rid of this issue.

> Will this new setup, especially for jsdom.env, "hide" errors that previously were easy to spot? Not everyone sets up a virtual console... Maybe we should store jsdomErrors at least in the virtual console for later inspection?

Yes, the README currently does a good job of explaining the new way error handling works, but I foresee there still being some confusion if some errors are moved to the virtual console.

After giving it some thought, the new way seems more similar to the way errors would happen in browsers. You might get an indication from the page that something wasn't right, but you'd need to pop open the console to see the actual error. Also I could be :+1: on storing errors for later consumption, but can't think of a ton of use cases right now, so possibly YAGNI.

domenic:
Almost there... got the virtual console errors for all resource loads working. The browser failures are a bit concerning though, so they need to be fixed before this is ready to be reviewed.

domenic:
OK, this is ready for review @Sebmaster and @Joris-van-der-Wel! I tried to make the commits small and atomic.

domenic:
@Joris-van-der-Wel said LGTM in IRC :)

Issue #1165, opened 2015-07-14
Efficient backing data structure


Joris-van-der-Wel:
I keep running into performance issues related to simple manipulations of the dom tree. There are a lot of `O(n)`  cases, even a few `O(n^2)`. 

I can probably get most of those down to `O(1)` by creating a linked tree data structure. So my current idea is to create a separate package for this purpose. It would provide many tree manipulation and querying methods dom implementations would need. 

Related: #1156

Joris-van-der-Wel:
First version: https://github.com/Joris-van-der-Wel/js-symbol-tree

Next up is the equivalent of Node._indexOf(), with caching. And various accessors used in the DOM standard, such as preceding and following.

domenic:
Looking great to me. Curious how much of a perf hit a WeakMap version would introduce, but that could be transparently changed at some point (e.g. if V8 gets faster with weak maps).

I opened a few cosmetic issues but overall happy to start using this whenever you've got a PR ready.

Joris-van-der-Wel:
Last I checked WeakMap was still pretty slow. Symbol will likely remain much faster because they could be optimized using v8's fast properties stuff (I am not sure if they are). Are you considering WeakMap because of `Object.getOwnPropertySymbols()`?

This will be very easy to compare after I integrate the lib. Although the name of the lib would be incorrect if it changes to WeakMap hah.

domenic:
> Are you considering WeakMap because of Object.getOwnPropertySymbols()?

That, and you can't add symbols to frozen objects. But in practice both gOPS and freezing will cause lots of havoc in jsdom today so it's more just an idle wondering than anything else.

PR #1169, opened 2015-07-20
Move the code dealing with trees to a different module

### LINKED ISSUE #1168: Some exceptions within unit tests are eaten
Some exceptions are eaten in unit tests without any warning on the terminal, the test simply times out.

I have noticed this in the callback for `env()`, but also in internal event handles such as the infamous `DOMNodeInsertedIntoDocument`.

One place where you could test this behaviour is here: https://github.com/tmpvar/jsdom/blob/96e763253cf24d2e1720efe9742783c2d9fb3e74/test/level2/style.js#L25. Try throwing something in that callback.

It looks like something is handling and ignoring uncaught exceptions. The exception is logged properly if i add the following:

```javascript
process.on('uncaughtException', function(err) {
  console.log('Caught exception: ' + err);
});
```

Joris-van-der-Wel:
See #1165
It also fixes #1168

Everything dealing with `_childNodes` and `_parent` has been converted to use https://github.com/jsdom/js-symbol-tree. This library uses `Symbol` to keep track of the entire tree. It has has a few efficient query methods. Its main benefit is that performance remains constant as you add more nodes.

So far I have patched the minimal what was needed to get jsdom working. I intend to move over some more utility functions and things like compareDocumentPosition. However I want to get some comments in first.

The tree modification benchmarks show a modest improvement for `removeChild()`, the biggest improvement is indeed in the tests where you call `removeChild()` on a node with a huge amount of siblings. 

`appendChild()` is a little slower probably because I am using for-of (the profiler is complaining about it). This hurts performance because of a limitation in v8. This has been recently fixed however: https://code.google.com/p/v8/issues/detail?id=3822. So I'm eagerly awaiting for that to land in io.js


domenic:
<3. Review done. Largely minor concerns. Some stuff can be pushed off to later, for sure.

Joris-van-der-Wel:
> Instead of doing three passes (iterate to create array; loop over array with map; loop over array with join) why not stick with a for-of over a children iterator?

The original code was using `join` so I assumed it was better for performance. However I just did a benchmark and it seems that simple string concatenations are faster in v8 than array + `join`. Some googling tells me that the `join` "optimization" is a relic of ancient IE.

Joris-van-der-Wel:
I've addressed all your points except the ones about `mapper` and `NodeList`. Addressing those two would be part of an effort to refactor all those collection classes. This is too big to include in this PR imo.

The symbol-tree API should be easier to understand now: https://github.com/jsdom/js-symbol-tree/blob/master/api.md
All optional arguments are passed an "options" object now. And the method names are more consistent with naming in DOM. For example `firstChild()` instead of `first()`, `hasChildren()` instead of `isEmpty()`, `appendChild()` instead of `insertLast()`.

`_indexOf()` has been removed, it was only used by the old `compareDocumentPosition`.

domenic:
LGTM! Maybe file a follow-up issue to refactor the collection classes.

Issue #1156, opened 2015-06-27
Optimize based on w3c/respec#453


domenic:
See @Joris-van-der-Wel's comments in https://github.com/w3c/respec/issues/453#issuecomment-115999039. Ideas:

- NamedPropertiesTracker is using getOwnPropertyDescriptor, which is known to be pretty slow. Maybe we can store these guys elsewhere, in some side table on the object, instead of as properties of the relevant getter function?
- compareDocumentPosition is implemented pretty naively per spec and maybe could be optimized. Although, why is the test case calling it so much? Maybe the test case is calling Node.prototype.contains, which delegates to compareDocumentPosition, but could probably be implemented in a faster way?

Joris-van-der-Wel:
The reason `NamedPropertiesTracker.prototype.maybeUntrack()` is so slow is actually because it needs to walk the document to see if a property can really be removed. Walking the document is also what makes compareDocumentPosition so slow.

So basically 90% of the time (in that test), jsdom is walking the document

domenic:
I don't see any document-walking in https://github.com/tmpvar/jsdom/blob/54c2ed4cb7eaa3cdf9798b052ed79641f26f3e9d/lib/jsdom/named-properties-tracker.js#L66 ...

domenic:
Oh, it's in `this.object[name]`, isn't it.

Joris-van-der-Wel:
Oh btw, if anyone wants to generate a profile: https://gist.github.com/Joris-van-der-Wel/4705395c018f12053968

NamedPropertiesTracker could be improved by tracking all id's and name's in a map for the entire document.

> I don't see any document-walking in https://github.com/tmpvar/jsdom/blob/54c2ed4cb7eaa3cdf9798b052ed79641f26f3e9d/lib/jsdom/named-properties-tracker.js#L66 ...

https://github.com/tmpvar/jsdom/blob/54c2ed4cb7eaa3cdf9798b052ed79641f26f3e9d/lib/jsdom/living/named-properties-window.js#L51

Joris-van-der-Wel:
A solution for both issues:

1. Whenever a `Node` is inserted or removed, store the node `index` (the number of preceding siblings)
2. Store an `id` to `Element` map in `Document` (this already exists)
3. Store a `name` to `Set<Element>` map in `Document`

Step 1 is used in both compareDocumentPosition and NamedPropertyTracker. Step 2 and 3 only in NamedPropertyTracker.

[...]

PR #1174, opened 2015-07-26
Optimize the named property tracker by not walking the entire tree

Joris-van-der-Wel:
This was one of the thing slowing down #1156 (w3c/respec#453).

The included benchmark has these results:

before: `jsdom  x 0.19 ops/sec ±68.89% (5 runs sampled)`
after: `jsdom  x 3,054 ops/sec ±7.63% (29 runs sampled)`

The named property tracker has been refactored by making the tracking and untracking more explicit. The tracker now keeps a list of potential values for a specific named property. The resolver function that is used for the `window` no longer walks the entire tree, but receives that list of potential values. It can then filter on that list and sort it in tree order using `compareDocumentPosition`. This works fast because `compareDocumentPosition` has been optimized in #1169.

### DIFF
--- a/benchmark/dom/index.js
+++ b/benchmark/dom/index.js
@@ -3,5 +3,6 @@
 module.exports = {
   construction: require("./construction"),
   "tree-modification": require("./tree-modification"),
-  "compare-document-position": require("./compare-document-position")
+  "compare-document-position": require("./compare-document-position"),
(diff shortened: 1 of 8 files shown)


domenic:
LGTM! No merge commits this time; --ff-only :)

PR #1176, opened 2015-07-29
Update browser vm shim to copy global built-ins to the sandbox

### LINKED ISSUE #1175: browser vm contexts do not have common globals
Haven't created a tiny repro yet, but this is the cause behind almost all of the current the failures in the better-errors branch. In that branch, the following test:

```js
exports["onerror catches exceptions thrown in inline event handlers"] = function (t) {
  const doc = jsdom(`<body onclick="throw new Error('oh no!')"></body>`, { url: "http://example.com" });

  doc.defaultView.addEventListener("error", function (event) {
    t.equal(event.message, "oh no!");
    t.equal(event.filename, "http://example.com");
    t.ok(event.lineno > 0, "lineno set");
    t.ok(event.colno > 0, "colno set");
    t.ok(event.error);
    t.done();
  });

  doc.body.click();
};
```

fails with "window.Error is not a function". If I add

```diff
diff --git a/lib/jsdom/vm-shim.js b/lib/jsdom/vm-shim.js
index 941d399..40ccb7d 100644
--- a/lib/jsdom/vm-shim.js
+++ b/lib/jsdom/vm-shim.js
@@ -13,6 +13,8 @@ exports.createContext = function (sandbox) {
     configurable: true,
     enumerable: false
   });
+
+  sandbox.Error = Error;
 };

 exports.isContext = function (sandbox) {
```

then almost all of the failures go away.

I am not sure what the correct solution is here. Do we set all the built-ins as inherited? Is there a way to get fresh copies of them (even in a worker, where there are no iframes to use)? @Sebmaster, thoughts?

domenic:
Fixes #1175 

Review please @Sebmaster? :)

### DIFF
--- a/lib/jsdom/vm-shim.js
+++ b/lib/jsdom/vm-shim.js
@@ -6,6 +6,15 @@ var acorn = require("acorn");
 var findGlobals = require("acorn-globals");
 var escodegen = require("escodegen");
 
+// From ES spec table of contents
+var builtInConsts = ["Infinity", "NaN", "undefined"];
+var otherBuiltIns = ["eval", "isFinite", "isNaN", "parseFloat", "parseInt", "decodeURI", "decodeURIComponent",
+  "encodeURI", "encodeURIComponent", "Array", "ArrayBuffer", "Boolean", "DataView", "Date", "Error", "EvalError",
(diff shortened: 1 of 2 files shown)


Sebmaster:
LGTM.

PR #1178, opened 2015-08-02
Update Location to use whatwg-url's URLUtils

domenic:
A jsdom/env test is failing currently due to https://github.com/jsdom/whatwg-url/issues/31

### DIFF
--- a/lib/jsdom/browser/history.js
+++ b/lib/jsdom/browser/history.js
@@ -1,5 +1,5 @@
 "use strict";
-var nodeURL = require("url"); // TODO Location: use new URL stuff
+var whatwgURL = require("whatwg-url-compat");
 var resolveHref = require("../utils").resolveHref;
 
 function StateEntry(data, title, url) {
@@ -11,10 +11,10 @@ function StateEntry(data, title, url) {
 module.exports = History;
 
 function History(window) {
-  this._states = [new StateEntry(null, "", window.location._url.href)];
+  this._states = [new StateEntry(null, "", window._document._URL)];
   this._index = 0;
   this._window = window;
-  this._location = window.location;
+  this._location = window._document._location;
 }
 
 History.prototype = {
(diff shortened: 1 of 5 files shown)


domenic:
whatwg-url-compat 0.6.1 now in use and fixes everything. Ready for review!

lib/jsdom/browser/location.js
```diff
@@ -1,103 +1,67 @@
...
+Location.prototype.assign = function () {
+  notImplemented("location.assign", this[document]._defaultView);
+};
 
-  toString: function () {
-    return this._url.href;
-  },
+Location.prototype.replace = function (url) {
+  // This is nowhere near spec compliant, but has worked so far.
+  whatwgURL.setTheInput(this, url);
```
Sebmaster: Wait wat?

lib/jsdom/browser/history.js
```diff
@@ -11,10 +11,10 @@ function StateEntry(data, title, url) {
 module.exports = History;
 
 function History(window) {
-  this._states = [new StateEntry(null, "", window.location._url.href)];
+  this._states = [new StateEntry(null, "", window._document._URL)];
   this._index = 0;
   this._window = window;
-  this._location = window.location;
+  this._location = window._document._location;
```
Sebmaster: Why go to document's location here?

lib/jsdom/browser/location.js
domenic: Not quite for replace. Replace generally does not cause a navigation. (It does, but with the replacement flag set.) The current code implements it as `this.href = val`.

lib/jsdom/browser/history.js
domenic: Because that's where it's really stored; window.location is a getter for this._document._location

Sebmaster:
LGTM then.

PR #1180, opened 2015-08-04
xhr shouldn't get cookies when requesting a file://

justinmchase:
There are two bugs here actually.

- if you request a local file then `XMLHttpRequest` has a null response object, calling `getResponseHeader` throws as a result.
- using `require("../utils").resolveHref` actually returns an invalid url on windows.

Regarding the first item, simply do not attempt to get the cookies if the url has a "file:" protocol.

Regarding the second item, using `URL.resolve` instead solves the problem. Specifically the problem is that a url such as:

```
file://c:/path/to/file.txt
```

is changed when using `require("../utils").resolveHref` into:
```
file:///c:/path/to/file.txt
```

Notice the triple `///` after file. It seems to be thinking that file paths should be preceded by a `/`, which isn't true on Windows.

I ran the tests on Windows:
```
OK: 7249 assertions (73055ms)
```

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -3,6 +3,7 @@
 var CSSStyleDeclaration = require("cssstyle").CSSStyleDeclaration;
 var XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest;
 var URL = require("url");
+var os = require("os");
 var notImplemented = require("./not-implemented");
 var History = require("./history");
 var VirtualConsole = require("../virtual-console");
@@ -19,6 +20,7 @@ module.exports = Window;
 var dom = require("../living");
 
 var cssSelectorSplitRE = /((?:[^,"']|"[^"]*"|'[^']*')+)/;
+var fileHrefWindowsRE = /file:\/\/\/([a-zA-Z]):/; // XMLHttpRequest requires non-triple slash for windows paths.
 
 var defaultStyleSheet = cssom.parse(require("./default-stylesheet"));
 
@@ -153,9 +155,12 @@ function Window(options) {
     var protocol = "";
     xhr._open = xhr.open;
     xhr.open = function (method, url, async, user, password) {
-      url = URL.resolve(window.document.URL, url);
-      lastUrl = url;
+      url = resolveHref(window.document.URL, url);
       protocol = URL.parse(url).protocol;
+      if (protocol === "file:" && os.platform() === "win32") {
+        url = url.replace(fileHrefWindowsRE, "file://$1:");
+      }
+      lastUrl = url;
       return xhr._open(method, url, async, user, password);
     };
     xhr._send = xhr.send;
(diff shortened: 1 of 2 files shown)


Sebmaster:
Regarding url: Triple slash is the correct format (as browsers do the same here and the spec is currently not-backwards-compatible) so this is probably a bug in XmlHttpRequest. Although we can fix it, but we have to special case windows paths here, e.g. match "file:///[A-Z]:/" and remove the first slash before forwarding it to XmlHttpRequest.

`url.resolve` is buggy and should not be used.

justinmchase:
@Sebmaster Ah ok, I'll defer to you on this. They are just using URL.parse(url).pathname which returns `"/c:/...."` which they are directly pasing into `fs.readFile`.

If you don't mind me special casing in the open function then I can do that.

Sebmaster:
Yeah, we should probably just special case open in that case then.

justinmchase:
Updated.
* Fixed lint issue.
* Don't use URL.resolve, instead use regexp to replace triple slash for windows file urls.

domenic:
This is still not quite right. Node URL module must not be used at all. (I think you want document.location.protocol.) Tests should use toFileUrl. I can fix up later though.

I think the cookie issue might be deeper. Cookie jar logic is probably wrong. Will investigate.

Sebmaster:
Don't switch on platform, we want to have the same behaviour cross-platform.

Just match directly on `file:///[A-Za-z]:` and replace, that should be fine.

domenic:
OK. I would like to fix this from first principles. With that in mind, I am confused with what the bugs are here.

- Is there a bug with using node-XMLHttpRequest with file URLs, that is not related to cookies? If so, how can we write a test for it?
- For the cookie bug, I think the cookie jar should be handling this. If we set a cookie with a file URL, then getting the cookies for the current URL later should not retrieve the cookie. That seems like a tough-cookie bug.

Sebmaster:
@domenic Yeah, node-xhr can't handle file schemas with triple slash - only double slash, so we have to strip the third one before passing it to xhr's open method.

domenic:
OK, planning to write a separate test and fix for that. Will PR when ready.

justinmchase:
The cookie bug is also in node-xhr, when getting a file the "request" is empty and attempting to get the headers causes an exception to be thrown.

I have a PR out to fix it in node-xhr:
https://github.com/driverdan/node-XMLHttpRequest/pull/109

But based on the number of pending PR's in that project I thought I would come here and submit another fix so that it would unblock me sooner.

But it seemed like getting cookies from a file get would probably not be desired anyway. I could be wrong.

domenic:
Oh, ugh.

Well, we can do a patch over that like you say; I'll send a jsdom PR for you to review that fixes that independently.

However, we've been looking for a long time for someone who's willing to fork node-XHR into jsdom/xhr or similar, and accept all those pull requests... Could that person be you!?

justinmchase:
This is dependent on this PR going in first:
https://github.com/tmpvar/jsdom/pull/1181

domenic:
As far as I can tell #1181 completely supercedes this PR. See https://github.com/tmpvar/jsdom/pull/1181/files#diff-a5bb7c6feb9a63afd8826ce81d3df489R174

justinmchase:
Interesting, you're right. I thought they were both issues but the cookies seem to be a red herring and the file path the only one. I'll close this, thanks for bearing with me!

PR #1181, opened 2015-08-05
Fix XHR for file URLs by working around node-XHR bug

domenic:
Closes #1180.

Please review @justinmchase. I was unable to write a test showing that getResponseHeader throws as long as the file URL was two-slashes only.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -151,9 +151,8 @@ function Window(options) {
     var lastUrl = "";
     xhr._open = xhr.open;
     xhr.open = function (method, url, async, user, password) {
-      url = resolveHref(window.document.URL, url);
-      lastUrl = url;
-      return xhr._open(method, url, async, user, password);
+      lastUrl = fixUrlForBuggyXhr(resolveHref(window.document.URL, url));
+      return xhr._open(method, lastUrl, async, user, password);
     };
     xhr._send = xhr.send;
     xhr.send = function (data) {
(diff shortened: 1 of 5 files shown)


lib/jsdom/browser/Window.js
```diff
@@ -412,3 +411,9 @@ function stopAllTimers(window) {
   });
   window.__timers = [];
 }
+
+function fixUrlForBuggyXhr(url) {
+  // node-XMLHttpRequest doesn't properly handle file URLs. It only accepts file://C:/..., not file:///C:/...
+  // See https://github.com/tmpvar/jsdom/pull/1180
+  return url.replace(/^file:\/\/\//, "file://");
```
justinmchase: Are you testing on darwin/linux? I'm thinking that actually 3 slashes would still work on darwin/linux and you need to specifically look for `/^file:\/\/\/([a-zA-Z]:)/`, since the url `file:///foo/bar.txt` would still be valid on darwin and linux as it would yield the path `/foo/bar.txt`. 

The problem case is when sending the url `/c:/bar.txt` on windows vs. `c:/bar.txt`, which the former node-xhr project fails on.

justinmchase: in other words:
```
return url.replace(/^file:\/\/\/([a-zA-Z]:)/, "file://$1")
```

justinmchase:
:+1: 

Thanks for being so responsive! This will unblock me and my project.

lib/jsdom/browser/Window.js
domenic: Right, Travis is indeed stalling, so that explains why this works locally.

domenic: Force-pushed revision with your suggestion.

justinmchase: I see that fixed the CI build too. I think this is :+1:

PR #1183, opened 2015-08-06
Issue node xhr file support

justinmchase:
This fixes the last two issues here for me.

`getAllResponseHeaders` needed to be patched to not throw for `file:` uri's, as there is a bug in node-xhr where it doesn't even attempt to set its internal response object.

Similarly the `response` getter needed to be patched to return the `responseText` for file urls since the node-xhr project doesn't properly set the `resonseType` like it should.

I added unit tests to cover these two cases.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -161,7 +161,7 @@ function Window(options) {
         // set the response for file urls.
         return null;
       }
-      return xhr.getAllResponseHeaders();
+      return xhr._getAllResponseHeaders();
     };
     xhr._send = xhr.send;
     xhr.send = function (data) {
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -154,6 +154,15 @@ function Window(options) {
       lastUrl = fixUrlForBuggyXhr(resolveHref(window.document.URL, url));
       return xhr._open(method, lastUrl, async, user, password);
     };
+    xhr._getAllResponseHeaders = xhr.getAllResponseHeaders;
+    xhr.getAllResponseHeaders = function () {
+      if (window.document.location.protocol === "file:") {
+        // Monkey patch this function for files, the node-xhr module doesn't properly
+        // set the response for file urls.
+        return "";
+      }
+      return xhr.getAllResponseHeaders();
+    };
     xhr._send = xhr.send;
     xhr.send = function (data) {
       var cookieJar = window.document._cookieJar;
@@ -193,6 +202,10 @@ function Window(options) {
           return this.responseText;
         } else if (this.responseType === "json") {
           return JSON.parse(this.responseText);
+        } else if (window.document.location.protocol === "file:") {
+          // node-xhr doesn't properly set the responseType for file requests.
+          // return the text anyway.
+          return this.responseText;
         } else {
           return null; // emulate failed requeest
         }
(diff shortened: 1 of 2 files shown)


justinmchase:
*ping*

It would be great if we could get this in, so I can proceed :)

lib/jsdom/browser/Window.js
```diff
@@ -193,6 +202,10 @@ function Window(options) {
           return this.responseText;
         } else if (this.responseType === "json") {
           return JSON.parse(this.responseText);
+        } else if (window.document.location.protocol === "file:") {
+          // node-xhr doesn't properly set the responseType for file requests.
+          // return the text anyway.
```
domenic: This doesn't seem like correct fallback behavior. responseType is set by authors. You don't get to decide that it always be text for files.

lib/jsdom/browser/Window.js
```diff
@@ -154,6 +154,15 @@ function Window(options) {
       lastUrl = fixUrlForBuggyXhr(resolveHref(window.document.URL, url));
       return xhr._open(method, lastUrl, async, user, password);
     };
+    xhr._getAllResponseHeaders = xhr.getAllResponseHeaders;
+    xhr.getAllResponseHeaders = function () {
+      if (window.document.location.protocol === "file:") {
```
domenic: You want to check lastUrl, I think.

lib/jsdom/browser/Window.js
```diff
@@ -154,6 +154,15 @@ function Window(options) {
...
+        // Monkey patch this function for files, the node-xhr module doesn't properly
+        // set the response for file urls.
+        return null;
+      }
+      return xhr.getAllResponseHeaders();
```
domenic: Infinite recursion. Probably need a test that this works in non-file cases to catch this kind of bug.

lib/jsdom/browser/Window.js
justinmchase: But if the type is set then it falls into one of the two conditionals above. In this case the `responseType` isn't being set and xhr isn't properly defaulting it for file urls.

lib/jsdom/browser/Window.js
justinmchase: Yup, good catch. Will do. I see it.

lib/jsdom/browser/Window.js
justinmchase: I'm not sure I understand what you mean here.

justinmchase:
I fixed the recursion issue and added a test to verify it.

lib/jsdom/browser/Window.js
domenic: I assume the breakage actually happens when lastUrl is file, not when the window location is file.

lib/jsdom/browser/Window.js
domenic: Got it. Needs a bit of clarification in the comments, but I can fix that while merging in.

test/misc/xhr-file-urls.js
```diff
@@ -32,3 +64,35 @@ exports["Getting a file URL should not throw for getResponseHeader"] = function
...
+      t.strictEqual(xhr.getAllResponseHeaders(), null);
+    });
+    t.done();
+  };
+
+  xhr.open("GET", toFileUrl(__filename), true);
+  xhr.send();
+};
+
+exports["Getting a non-file URL should not fail for getAllResponseHeaders"] = function (t) {
```
domenic: This test doesn't belong in xhr-file-urls.js anymore :)

domenic:
I can fix the nits from here. Thank you!

domenic:
OK, merged as c12b8384cf3ef324192c09edbac5cb593347e9ec and 6.0.1 going out shortly!

PR #1186, opened 2015-08-08
updateAllow (window): setTimeout / setInterval to take string param

AshCoolman:
Handle strings to be eval'd as function http://stackoverflow.com/questions/20625133/jsdom-throwing-error-for-some-urls


domenic:
Desparately needs tests. This won't work, since `global.eval(fn)` will just eval `fn` which is now a function, not a string. Also there's no point in binding if you don't use `this` inside the function.

PR #1187, opened 2015-08-09
Bold requirement of iojs instead of Node.js in readme

kevinold:


domenic:
I think it's clear enough as-is. Thanks though!

davidgilbertson:
Will you support node when v4 comes out?

domenic:
Yes, when io.js gets renamed to node.js, then io.js/node.js will continue working.

PR #1189, opened 2015-08-10
Proxy: include tunnel configutarion

danieljoppi:
https://github.com/request/request => tunnel - controls the behavior of HTTP CONNECT tunneling as follows:
 - undefined (default) - true if the destination is https or a previous request in the redirect chain used a tunneling proxy, false otherwise
 - true - always tunnel to the destination by making a CONNECT request to the proxy
 - false - request the destination as a GET request.


domenic:
I'm not really OK with the `|| true`. We should have our default be the same as whatever request's default is.

domenic:
Also, is it possible to test this behavior?

PR #1190, opened 2015-08-10
Add basic support for media elements.

cpojer:
This is a copy of the jest jsdom-compat file: https://github.com/facebook/jest/blob/235c859f982da61c501756bc9b615411c16275e0/src/lib/jsdom-compat.js

We can expand support later but I'd prefer to add basic support of these elements to jsdom and remove our jest jsdom-compat file.


domenic:
Merged as 703483ca87fbf2e00c78593e1048a94be8ca99e5 and 6.1.0 coming out shortly; thank you very much!!

PR #1196, opened 2015-08-15
Document config.virtualConsole options for jsdom.env().

ludovicofischer:
Document that the virtualConsole option is available when calling `jsdom.env()` (since https://github.com/tmpvar/jsdom/commit/e2d2481d00e53a4ad8d9a82161929287b60219ef)


domenic:
Thank you! Merged as dc7e5a38c64ba7378584724f73077885881de13a.

Issue #1198, opened 2015-08-15
DOMSubtreeModified events not fired


selam:
here is my html page;
```html
<!DOCTYPE html>
<html>
<head>
  <title></title>
  <link rel="stylesheet" href="https://duckduckgo.com/s1015.css" type="text/css">
  <script src="require.js"></script>
</head>
<body>
    <script src="test-js.js"></script>
    <script src="amd-fetch.js"></script>
</body>
</html>
```
here is my test-js.js
```javascript
body = document.getElementsByTagName("body")[0];
c = document.createElement("span");
c.innerHTML = "deneme inside";
body.appendChild(c);

window.document.addEventListener("DOMSubtreeModified", function() {
    console.log("from-test.js change event occurs")
});
```
here is my amd-fetch.js
```javascript
requirejs(["app"]);
```
app.js;
```javascript
require(["jquery"], function($) {
    //the jquery have been loaded.
    console.log("AMD Loaded application start working");
    $("<div>dom modification from rich application</div>").appendTo("body");

   // console.log(window.document.documentElement.outerHTML);

    require(["fake"], function(){
        $("body").append("<div>dom modification from rich application from fakejs callbacks in require</div>");
        console.log("from fake")
    });
});
```
Here is test.js 
```javascript
var jsdom   = require("jsdom");
var events = require('events');
var eventEmitter = new events.EventEmitter();

var virtualConsole = jsdom.createVirtualConsole();
virtualConsole.on("jsdomError", function (error) {
    console.error(error.message, error.detail);
});

jsdom.env("http://localhost/test-js.html", {
    virtualConsole: virtualConsole.sendTo(console),
    features: {
        FetchExternalResources : ["link", "script"],
      	ProcessExternalResources: ["link", "script"],
      	SkipExternalResources: false
    },
    resourceLoader: function(resource, callback) {
        console.log("we find external resource: %s", resource.url.pathname)
        if(/\.js$/.test(resource.url.pathname)) {
            console.log("external source is a script")
        }
        else if (/\.css$/.test(resource.url.pathname)) {
            console.log("external source is a stylesheet")
        }

        resource.defaultFetch(function(error, response, data){
            eventEmitter.emit("stillFetched")
            callback(null, response);
        });
    },
    onload: function(window) {
        console.log("onload");
        eventEmitter.addListener("stillFetched", function(){
            console.log("async loader working")
        });
        window.document.addEventListener("DOMSubtreeModified", function(){
            console.log("domchanged")
[...]

PR #1200, opened 2015-08-16
merge reqquest for #1198

selam:
i was able to fix this bug, i added all necessarily tests


domenic:
Thank you very very much for this contribution! I merged it as two commits: f56fef2d0399aae09b2dfcd3115b3ee7b4d73169 and e3ee5d0267dcad424e9e1b27b8891b129fc66975. You can check them out, and see how they differ from yours in minor ways, if you want to learn some best practices and style for contributing to jsdom (e.g. commit message format, use camelCase for variables instead of snake_case). I also did fix it to fire the event on the parent node instead of on the actual node that was changed.

I'm planning to do a release today, either 6.1.1 or 6.2.0 depending on how much I do, so by end of day today this should be published on npm.

PR #1119, opened 2015-05-03
Add web-platform-tests NodeIterator tests

domenic:
Required fixing a bug in w3ctest.js which would replace both <script> tags with our shim, instead of just the reporter. We did this in a more principled way by overriding the resourceLoader.

Fixes #1117.

PR #1201, opened 2015-08-16
Fix #1119: virtualConsole.sendTo should send jsdomErrors to console.error by default

domenic:
I also threw in a commit that classes up virtual-console.js.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -7,6 +7,7 @@
     "immed": true,
     "indent": 2,
     "latedef": "nofunc",
+    "loopfunc": true,
     "newcap": true,
     "noarg": true,
     "noempty": true,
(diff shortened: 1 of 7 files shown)


domenic:
Hmm. I am not able to reproduce the on-error.js Travis failure.

Worse, this PR has revealed an interesting bug that the w3c tests seem to trigger, where we try to fire events on window before it's created (i.e. before it has a `_document` property). These seem to originate from the resource-loading subsystem.

If anyone wants to help debug I'd appreciate it.

domenic:
OK, if I comment out window.close() in w3ctest.js, it goes away. So we should stop trying to fire events after the window closes. But, even if we do stop such things, it's unclear why we're *trying* to fire events after window closes.

domenic:
OK figured out the w3ctest.js. window.close() actually fires a load event, but the window is all destroyed-ey by that time. So it's expected, and the fix was easy.

Still not sure why the on-error failure isn't showing for me. Can someone with non-Windows test?

domenic:
Oh it's io.js 3 vs. 2, good times.

PR #1209, opened 2015-08-21
Add window.atob and window.btoa support, fixes #1184

### LINKED ISSUE #1184: `window.btoa` support?
When using the following, calling `window.btoa` returns `is not a function`

````
var document = jsdom.jsdom();
var window = document.defaultView
````

Is this currently supported?

### PR DESCRIPTION
**Pull request by jeffcarp**
In the browser, `btoa` throws an error if passed a string with UTF-8 characters. I have 2 questions around the implementation of that: 

1. Is the UTF-8 detection in the new `stringContainsUTF8Chars` method acceptable? (testing charCode > 255)

2. What type of error should be thrown when `btoa` is passed a non-ASCII string?

Right now Chrome throws this:

> Uncaught DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.

Firefox throws:

> InvalidCharacterError: String contains an invalid character

Other notes:
- I chose to implement the methods in jsdom since they're both essentially one liners, but if there's a strong preference for using something on npm, I can replace them with something like these:
  - https://github.com/node-browser-compat/atob
  - https://github.com/node-browser-compat/btoa
- I couldn't find anything similar to `stringContainsUTF8Chars` on npm, so if there really isn't anything like it out there, it might make a useful npm module.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -8,6 +8,7 @@ var VirtualConsole = require("../virtual-console");
 var define = require("../utils").define;
 var inherits = require("../utils").inheritFrom;
 var resolveHref = require("../utils").resolveHref;
+var stringContainsUTF8Chars = require("../utils").stringContainsUTF8Chars;
 var EventTarget = require("../events/EventTarget");
 var namedPropertiesWindow = require("../living/named-properties-window");
 var cssom = require("cssom");
@@ -145,6 +146,18 @@ function Window(options) {
   }
 
   this.postMessage = postMessage;
+
+  this.atob = function (base64String) {
+    return new Buffer(base64String, "base64").toString("binary");
+  };
+
+  this.btoa = function (plainString) {
+    if (stringContainsUTF8Chars(plainString)) {
+      // TODO: InvalidCharacterError? DOMError?
+      throw new Error("String contains an invalid character");
+    }
+    return new Buffer(plainString, "binary").toString("base64");
+  };
 
   this.XMLHttpRequest = function () {
     var xhr = new XMLHttpRequest();
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -8,7 +8,6 @@ var VirtualConsole = require("../virtual-console");
 var define = require("../utils").define;
 var inherits = require("../utils").inheritFrom;
 var resolveHref = require("../utils").resolveHref;
-var stringContainsUTF8Chars = require("../utils").stringContainsUTF8Chars;
 var EventTarget = require("../events/EventTarget");
 var namedPropertiesWindow = require("../living/named-properties-window");
 var cssom = require("cssom");
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -152,7 +152,8 @@ function Window(options) {
   this.atob = function (str) {
     const result = atob(str);
     if (result === null) {
-      throw new DOMException(DOMException.INVALID_CHARACTER_ERR, "The string to be encoded contains characters outside of the Latin1 range.");
+      throw new DOMException(DOMException.INVALID_CHARACTER_ERR,
+        "The string to be encoded contains characters outside of the Latin1 range.");
     }
     return result;
   };
@@ -160,7 +161,8 @@ function Window(options) {
   this.btoa = function (str) {
     const result = btoa(str);
     if (result === null) {
-      throw new DOMException(DOMException.INVALID_CHARACTER_ERR, "The string to be encoded contains characters outside of the Latin1 range.");
+      throw new DOMException(DOMException.INVALID_CHARACTER_ERR,
+        "The string to be encoded contains characters outside of the Latin1 range.");
     }
     return result;
   };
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by domenic** [comment]
> In the browser, btoa throws an error if passed a string with UTF-8 characters.

That's not really what it does. See [the spec](https://html.spec.whatwg.org/multipage/webappapis.html#dom-windowbase64-btoa). It throws if it contains any characters > U+00FF. Which is exactly what you're testing :).

> Is the UTF-8 detection in the new stringContainsUTF8Chars method acceptable? (testing charCode > 255)

Yes, but it seems to me it would be much better to do this while you're looping through each character, instead of doing it beforehand. But I guess if you're doing it by using Buffer then this is fine, if a bit inefficient.

> What type of error should be thrown when btoa is passed a non-ASCII string?

Per the spec, an InvalidCharacterError DOMException. I.e., `new DOMException(DOMException.INVALID_CHARACTER_ERR, "...")`, if I am remembering my stuff correctly.

> I chose to implement the methods in jsdom since they're both essentially one liners, but if there's a strong preference for using something on npm, I can replace them with something like these

Given that we have to throw our own DOMException types, this was probably the right decision.

> atob

This doesn't seem to follow the spec very closely, sadly. I would accept it for now with a big giant TODO and an open issue, just to help our users. But ideally we would implement it per spec.

The tests in https://github.com/w3c/web-platform-tests/blob/master/html/webappapis/atob/base64.html look pretty exhaustive. In fact, they have an implementation already in them. Now *that* could be a good npm package... something that does atob and btoa per spec, but returning `null` instead of throwing an error. Then jsdom can consume it and throw an error if the package returns `null`.

You should be able to make the tests run by adding `w3cTest("html/webappapis/atob/base64.html")` to the w3ctest/index.js file.

As always, thanks for your help!!

**Comment by jeffcarp** [comment]
Thanks for taking a look over this. My bad on the spec front, I'll look at the spec more closely in the future.

I'm up to use the implementations in `w3c/web-platform-tests` as a reference to implement `atob` and `btoa` as modules either in jsdom or standalone, however I'm a little unsure of how to do so without it looking mostly like plagiarism.

For future reference - including a TODO list for when the above question gets resolved:

- [ ] Change the `char > U+00FF` detection to happen during character iteration
- [ ] Make `btoa` throw an `InvalidCharacterError DOMException`
- [ ] Add `w3cTest("html/webappapis/atob/base64.html")` to `w3ctest/index.js`

**Comment by domenic** [comment]
> however I'm a little unsure of how to do so without it looking mostly like plagiarism.

Hah, I think it's fine: they're [dual-licensed under BSD](https://github.com/w3c/web-platform-tests/blob/master/LICENSE), and it's a time-honored tradition to create an npm package that pulls out code originally from some giant repo into its own nice little module :)

**Comment by jeffcarp** [comment]
Updated the PR to:

- Rely on the w3c implementation of atob/btoa, temporarily in `lib/base64.js`, which can be broken out and pushed to the jsdom GH org + npm
-

**Comment by jeffcarp** [comment]
Updated the PR to rely on the w3c implementation, temporarily residing in `lib/base64.js` so that it can be broken out and pushed to the jsdom GH org + npm.

TODO:

- [ ] Make `lib/base64.js` pass the linter

File: lib/jsdom/browser/Window.js [hunk 1]
```diff
@@ -146,6 +149,24 @@ function Window(options) {
...
+        "The string to be encoded contains characters outside of the Latin1 range.");
+    }
+    return result;
+  };
+
+  this.btoa = function (str) {
+    const result = btoa(str);
+    if (result === null) {
+      throw new DOMException(DOMException.INVALID_CHARACTER_ERR,
+        "The string to be encoded contains characters outside of the Latin1 range.");
```
domenic: I don't think this is quite the right error message, at least for btoa. Firefox shows "String contains an invalid character".

test/window/base64.js
```diff
@@ -0,0 +1,50 @@
+"use strict";
+
+var jsdom = require("../..").jsdom;
```
domenic: const!

lib/jsdom/browser/Window.js
domenic: Sorry, I meant for atob. E.g. `atob(":")` fails.

domenic:
Looking great!

I created a repo named "abab" (only relevant name I could find that wasn't taken on npm) and invited you to it. Let me know if you think of a better name. It's your project so feel free to run it however you want in terms of coding style, io.js version requirements, readme, linting, tests, whatever. The only real guideline I'd suggest is publishing a version 1.0.0 instead of sticking in the 0.x range. Let me know if you need anything set up (e.g. Travis, renames, change repository description, invite new collaborators).

I will basically land this PR after my queued-up PRs (#1209, #1217) get reviewed and land. It looks good enough I can tweak anything missing. But if you have time to do the repo move while I'm waiting for reviews, feel free, and I'll merge that version too.

jeffcarp:
Awesome thanks! I'm :+1: on `abab` for now.

It looks like I may not have push access to the repo yet, though.

domenic:
@jeffcarp oops, fixed that!

domenic:
Merged as a9e6c3a7295012b23800970183ac0d5d4bdb69ec!

jeffcarp:
Yay! :fireworks: :flags: Thanks for all the help! After I get the `abab` repo into shape, I'll submit a PR integrating that module and fixing [the exception message for `btoa`](https://github.com/tmpvar/jsdom/pull/1209#discussion_r38258398).

PR #1212, opened 2015-08-21
Add test for canvas instanceof HTMLElement, fixes #649

### LINKED ISSUE #649: Canvas element fails "instanceof HTMLElement"
At least the canvas element (I have not tested other) returns false on "element instanceof HTMLElement". It should return true.

jeffcarp:


domenic:
Lint fails :P
Hmm so #649 is largely about what happens when node-canvas gets installed. To test this properly we should probably have travis install node-canvas for us. (Which might require some prereqs.) It looks like @Sebmaster might be able to give you the right bash incantation, similar to his tricks to run test/ci.js in Travis... https://github.com/tmpvar/jsdom/blob/master/.travis.yml#L23

Sebmaster:
Should be able to just add a
```
  - npm i canvas
```
to the scripts section.

domenic:
Thanks for spurring us in to action on this. As of 28fd8b15e028b86dbaff4851023370388229c6a8 we now run a third configuration on travis, with node-canvas installed.

Unfortunately, adding this test (and a similar one) to CI shows that it in fact fails! https://travis-ci.org/tmpvar/jsdom/builds/77089927 So that at least confirms that the linked bug is not fixed :(

PR #1213, opened 2015-08-22
Add Element.classList

wacii:
A rough implementation of Element.classList.

Related w3c/web-platform-tests do not pass. Didn't know how to get some to work, so I just wrote my own based on the DOM standard that I think represent a reasonable attempt at the feature.

Didn't see a clean way to integrate my code, so I just stuck it in [level1/core.js](https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/level1/core.js#L674-L680).

### DIFF
--- a/lib/jsdom/living/class-list.js
+++ b/lib/jsdom/living/class-list.js
@@ -1,0 +1,7 @@
+"use strict";
+
+var createDOMTokenList = require("./dom-token-list").createDOMTokenList;
+
+exports.createClassList = function createClassList(el) {
+  return createDOMTokenList(el, "class");
+};
--- a/lib/jsdom/living/class-list.js
+++ b/lib/jsdom/living/class-list.js
@@ -1,6 +1,6 @@
 "use strict";
 
-var createDOMTokenList = require("./dom-token-list").createDOMTokenList;
+const createDOMTokenList = require("./dom-token-list").createDOMTokenList;
 
 exports.createClassList = function createClassList(el) {
   return createDOMTokenList(el, "class");
--- a/lib/jsdom/living/class-list.js
+++ b/lib/jsdom/living/class-list.js
@@ -1,7 +1,0 @@
-"use strict";
-
-const createDOMTokenList = require("./dom-token-list").createDOMTokenList;
-
-exports.createClassList = function createClassList(el) {
-  return createDOMTokenList(el, "class");
-};
--- a/lib/jsdom/living/dom-token-list.js
+++ b/lib/jsdom/living/dom-token-list.js
@@ -124,7 +124,7 @@ function update(list) {
 exports.DOMTokenList = DOMTokenList;
 
 // set dom token list without running update steps
-exports.resetDOMTokenList = function resetDOMTokenList(list, value) {
+exports.reset = function resetDOMTokenList(list, value) {
   splice.call(list, 0, list.length);
   value = (value || "").trim();
   if (value !== "") {
@@ -139,7 +139,7 @@ exports.resetDOMTokenList = function resetDOMTokenList(list, value) {
   }
 };
 
-exports.createDOMTokenList = function createDOMTokenList(element, attribute) {
+exports.create = function createDOMTokenList(element, attribute) {
   const list = Object.create(DOMTokenList.prototype);
 
   list[INTERNAL] = {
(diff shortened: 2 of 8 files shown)


wacii:
Defined classList on the prototype and used the _attrModified hook on Element to maintain a single instance of DOMTokenList rather than recreate it whenever requested.

DOMTokenList is not a subclass of Array, but that's the only way I know to add indexed properties.

wacii:
Still have to start using some of the newer js constructs, add/play around with benchmarks, and take another stab at the w3c tests...

Joris-van-der-Wel:
> DOMTokenList is not a subclass of Array, but that's the only way I know to add indexed properties.

You can just assign them: `this[3] = thirdItem; this.length = 4`. You can also use a lot of the array functions as follows:

```javascript
var arrayPush = Array.prototype.push;

function Foo() { }

var myObject = new Foo();
arrayPush.call(myObject, "stuff");
arrayPush.call(myObject, "more stuff");
console.log(myObject.length === 2);
console.log(myObject[0] === "stuff");
```

wacii:
Didn't know you could use functions pulled off array on non array-like objects. Thanks for that. Wasn't really in the mood to manage the properties manually, but I'll definitely scavenge array instead of inheriting from it.

lib/jsdom/living/class-list.js
```diff
@@ -0,0 +1,7 @@
+"use strict";
+
+var createDOMTokenList = require("./dom-token-list").createDOMTokenList;
+
+exports.createClassList = function createClassList(el) {
```
domenic: I don't think this file (or this function) is really necessary. You can just use createDOMTokenList at the appropriate place.

lib/jsdom/living/dom-token-list.js
```diff
@@ -0,0 +1,133 @@
...
+  var attribute = list[INTERNAL].attribute;
+  if (attribute !== undefined) {
+    list[INTERNAL].element.setAttribute(attribute, list.toString());
+  }
+}
+
+exports.DOMTokenList = DOMTokenList;
+
+// set dom token list without running update steps
+exports.resetDOMTokenList = function resetDOMTokenList(list, value) {
```
domenic: I think these exports should be renamed reset and create.

domenic:
Over all this is looking quite nice. Great structure. Thanks for doing the work. The major to-do left is to update to use ES2015 features: namely class syntax and `const`/`let` instead of `var`. (Prefer `const` wherever possible.)

lib/jsdom/living/dom-token-list.js
```diff
@@ -0,0 +1,133 @@
+"use strict";
```
domenic: You will also need to do something like this in `living/index.js`:

```js
core.DOMTokenList = require("./dom-token-list");
```

lib/jsdom/living/dom-token-list.js
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  if (/\s/.test(token)) {
+    throw new DOMException(DOMException.INVALID_CHARACTER_ERR);
+  }
+}
+
+function update(list) {
+  var attribute = list[INTERNAL].attribute;
+  if (attribute !== undefined) {
```
domenic: When would attribute be undefined?

lib/jsdom/living/dom-token-list.js
```diff
@@ -0,0 +1,133 @@
...
+  return true;
+};
+
+DOMTokenList.prototype.toString = function toString() {
+  return this.join(" ");
+};
+
+function validateToken(token) {
+  if (token === "") {
+    throw new DOMException(DOMException.SYNTAX_ERR);
```
domenic: These should probably have helpful messages

lib/jsdom/living/dom-token-list.js
wacii: I have no idea! But it's part of the specification. A DOMTokenList without an attribute is a DOMSettableTokenList. Which is a thing. And not related to classList, so I didn't look into it.

I can remove it if you'd like.

wacii:
Not going to play around with new benchmarks. Would take me awhile to accomplish anything and I don't think it's that important here.

Only one w3c test is failing; that requiring length to be readonly. Guess I'll do that tomorrow. Hopefully without rewriting a bunch of stuff. Or doing anything particularly heinous with js. No promises.

domenic:
> Only one w3c test is failing; that requiring length to be readonly.

Amazing!!

wacii:
I did stuff to make length readonly. So now all the tests pass. For the main classList file anywho.

domenic:
Thank you so much for this amazing work!! Merged as ad6c555587db2f314624c2fdaad07043a4076548 with a few minor tweaks. Also followed up with a33632cc03213c448b99543bdec4eda2d8eb6f9b.

Issue #1145, opened 2015-06-08
support node.js again


japrescott:
I was wondering since io.js and node.js will be merged together under the "open, neutral, Node.js Foundation", at what point will jsdom function on node.js again? What version of uv does node.js need to have, so that jsdom works? Or will it not be before the nodejs1.0 release regarding the history of joyent's speed?

domenic:
I already answered you in the previous thread. Node.js will be supported once a release is made that is based on the io.js code base. That is indeed the plan---now that the trademark is owned by a foundation instead of by Joyent, we can basically rename io.js to node.js, put together a new major release (my guess is Node.js 5.0), and be merged. At that time jsdom will support Node.js, since it is just a new name for the io.js platform jsdom already supports.

japrescott:
thank you for your detailed answer domenic!

ccarpita:
Is there a specific technical reason why node 0.12.x is not supported by jsdom, or is this more of a political issue?  I can understand why it might be the latter, but if it's the former I would really curious to understand the specific limitations and how they might affect other projects.

domenic:
There is indeed a technical reason. Please see the changelog for jsdom 4.0.0.

ccarpita:
I see you were able to drop the Contextify dependency, and I would say that qualifies as a solid technical reason!

Issue #1081, opened 2015-04-09
Node 0.12 support?


stevenvachon:
It has the same [`vm`](https://nodejs.org/api/vm.html) capabilities as io.js, does it not?

domenic:
No, it does not.

stevenvachon:
Care to elaborate? I have not yet tried out io.js

domenic:
The V8 version in Node 0.12 is too old to provide good support for vm.

In general we do not support legacy technology like Node.js.

stevenvachon:
Ok, thanks.

chapinkapa:
@domenic legacy? Really?

PR #1215, opened 2015-08-25
explanation why io.js instead of node.js in readme

joernhees:
makes it easier to understand why jsdom won't work with node.js and should reduce questions like #1187, #1145, #1081

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A JavaScript implementation of the WHATWG DOM and HTML standards, for use with [
 $ npm install jsdom
 ```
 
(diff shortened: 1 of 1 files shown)


domenic:
The second addition is not accurate. The original Node.js codebase will never be supported. io.js is being renamed to Node.js, and so we will continue supporting io.js in the future.

The first addition seems fine though and I'm happy to pull it in.

joernhees:
will change...

joernhees:
better?

joernhees:
aww f*** typos... will squash

joernhees:
squashed

domenic:
Looks good, will merge tonight in a batch of jsdom updates. Thanks.

PR #1217, opened 2015-08-26
Always run web platform tests from a server

domenic:
This fixes certain base URL tests I am soon hoping to add, which use hrefs like "/foo/bar.html" that don't work very well on the file system.

In theory we should be running their python server. But, it requires modifying your hosts file, which is pretty sad. We can just do this hack until we run in to the need to run tests that require that level of complexity.

### DIFF
--- a/test/w3c/w3ctest.js
+++ b/test/w3c/w3ctest.js
@@ -1,30 +1,30 @@
 "use strict";
 
-var fs = require("fs");
-var path = require("path");
-var request = require("request");
-var jsdom = require("../..");
-var toFileUrl = require("../../lib/jsdom/utils").toFileUrl;
+const http = require("http");
+const path = require("path");
+const request = require("request");
+const st = require("st");
+const jsdom = require("../..");
+const URL = require("../../lib/jsdom/utils").URL;
 
-function createJsdom(source, url, t) {
-  const reporterHref = toFileUrl(__dirname + "/tests/resources/testharnessreport.js");
+function createJsdom(urlPrefix, testPath, t) {
+  const reporterHref = urlPrefix + "resources/testharnessreport.js";
 
   jsdom.env({
-    html: source,
-    url: url,
+    url: urlPrefix + testPath,
     features: {
       FetchExternalResources: ["script", "frame", "iframe", "link"],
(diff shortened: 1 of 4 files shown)


test/w3c/w3ctest.js
```diff
@@ -50,40 +52,21 @@ function createJsdom(source, url, t) {
...
-
-        createJsdom(respBody, "http://w3c-test.org/" + url, t);
-      });
-    } else {
-      file = file.replace(/\/resources\//gi, toFileUrl(__dirname + "/tests/resources") + "/");
-      createJsdom(file, toFileUrl(path.resolve(__dirname, "tests", url)), t);
-    }
-  });
-}
+const server = http.createServer(ecstatic({ root: path.resolve(__dirname, "tests") })).listen(port, hostname);
```
domenic: I'm not terribly comfortable with this, but couldn't think of anything that was much better. Suggestions welcome :-/.

domenic:
Blargh, I guess the browser test runner was using the w3ctest.org fallback? Will have to rethink this somehow.

domenic:
Ah OK it was easy enough, just comment out the right line in worker-runner.js.

domenic:
@Sebmaster ready for another review. I added and tested fallback to online-hosted tests. (Which have moved URLs it seems, to https.)

test/w3c/w3ctest.js
```diff
@@ -50,40 +50,40 @@ function createJsdom(source, url, t) {
...
     };
   };
+
+  function fallbackToHostedVersion(req, res) {
+    // Problem getting it from disk. Let's try the online version!
+
+    const url = new URL(req.url, urlPrefix);
+
+    // TODO update when https://github.com/jsdom/whatwg-url/issues/32 is fixed
+    //      url.host = "http://w3c-test.org/";
```
Sebmaster: Already fixed?

Joris-van-der-Wel:
Instead of modifying the hosts file, we could just manually rewrite those domains the the proper ip from within jsdom (while keeping the Host header intact)

domenic:
Yeah, that's a good plan for next steps. I'm a little cautious because then we start requiring python, which is kind of annoying.

PR #1216, opened 2015-08-26
Update web platform tests and add a bunch of HTML element stubs

domenic:
Blocked on https://github.com/w3c/web-platform-tests/pull/2107

PR #1218, opened 2015-08-26
Implement correct base URL semantics

domenic:
Blocked on https://github.com/w3c/web-platform-tests/pull/2108, #1216, and #1217.


Joris-van-der-Wel:
This does not yet fix the performance issue right? (looking up the base element every time it is needed).

An easy fix would be to maintain an sorted array of attached `<base>` elements on the document. Something like:

```javascript
const treeOrderSorter = require("../utils").treeOrderSorter;

class HTMLBaseElement {
    _attach() {
        const baseElements = this._ownerDocument._baseElements;
        baseElements.push(this);
        baseElements.sort(treeOrderSorter)
        super._attach();
    }
    
    _detach() {
        const baseElements = this._ownerDocument._baseElements;
        const index = baseElements.indexOf(this);
        if (index >= 0) {
            baseElements.splice(index, 1);
        }
        super._detach();
    }
}
```
And any time you need the base, you simply loop over this array until you find one that has a href attribute.

domenic:
The performance issue comes because whenever a href attribute changes (e.g. during parsing) we update the URL state. The fix is to make that lazy.

This is pretty separate. It just changes base URL calculation, whenever that happens. When URL updates become lazy, base URL calculation will happen lazily (i.e. upon calling the .href getter), which will avoid the performance problem since it will no longer impact parsing.

Joris-van-der-Wel:
Ah, yes.

You would still walk the entire tree whenever you need to read out href of an anchor, although i am not sure what the impact would be in such a case.

domenic:
Yeah, I figure we'll wait until someone complains...

PR #1222, opened 2015-09-02
Prepare docs & scripts for the upcoming Node.js 4.0.0 release

mzgol:


mzgol:
`npm install` fails on io.js 3.3.0 but that's unrelated to my changes: https://travis-ci.org/tmpvar/jsdom/jobs/78442773

domenic:
Thanks, but this is in general inaccurate. We support both io.js and Node.js > 4. Whereas this removes all mention of io.js, including from examples.

mzgol:
Would you prefer to keep them as `io.js` or change to `io.js/Node.js`?

mzgol:
> We support both io.js and Node.js > 4. Whereas this removes all mention of io.js

That's incorrect, in places that mention support I left "io.js or Node.js 4 or newer". I only removed it in examples and texts that mentioned waiting for something from a "future io.js release" - in this case we know it will never arrive in io.js so I changed it to Node.js.

I can restore the use of io.js in examples or switch them to io.js/Node.js. Unless you prefer to re-do it by yourself. ;)

mzgol:
@domenic I've reverted the example and put `Node.js/io.js` in all places that apply to both. Let me know if you still don't like something in this PR (I've kept the copy of the old branch locally just in case).

domenic:
Thanks, merged as 65e0cef2447fdb2327c88d953cdffc21cacc6edd

Issue #1224, opened 2015-09-05
NodeList contains unspecified _length property which is enumerable


techkey:
The `NodeList` object as returned by e.g. `querySelectorAll` contains a unspecified property `_length` which is `enumerable`, example:

`var list = document.querySelectorAll('span');`

`Object.getOwnPropertyDescriptor(list, '_length')` returns:

`{configurable: true, enumerable: true, value: 448, writable: true, __proto__: Object}`

`list.propertyIsEnumerable('_length')` returns `true`, which messes up functions like `Object.keys(list)` because the `_length` property is in the returned array too.

`Object.keys(list)` returns an array like this:

```javascript
[
    '0',
    ...
   '447',
   '_length'
]
```

`list.propertyIsEnumerable('length')` returns `false`, which is good. 

https://dom.spec.whatwg.org/#interface-nodelist

domenic:
In general you're going to find lots of these properties all over jsdom, because we don't do private variables very well in general. But, we can certainly fix this particular instance, and any others you noticeas particularly problematic.

techkey:
@domenic A fix for this instance would be great as I (and maybe other programmers) use `NodeList` and `Object.keys` a lot. Then at least I can remove all the filters that I've build in my code that uses `jsdom` to filter out the `_length` property.

Almost always I use native ECMAScript code and report problematic code if it's not too much trouble for you.

It's on my agenda to look at the source of `jsdom` so that maybe I can be of any help.

domenic:
It's OK, I've already got a PR up :)

techkey:
Cool thanks, btw `jsdom` is doing a great job so far :)

domenic:
Fixed in master, new release coming shortly.

Issue #859, opened 2014-08-19
NodeList should not have a length setter


lawnsea:
Several `NodeList` methods use `Array.prototype.slice` and `Array.prototype.splice`. Unfortunately, both Firefox and Chrome expect to be able to set the `length` of arrays that are `slice`d or `splice`d. To work around this problem for #849, I added a `length` setter to `NodeList` that ignores the new value and uses `_length`.

This is not ok in the long run, however, as we want `NodeList` to look like the real thing, which exposes a read-only `length` property.

See https://github.com/TreehouseJS/jsdom/commit/15e3c4fa91d70ea803074aa21d61f36667db1b63 and tmpvar/jsdom#850 for details.

domenic:
Fixed in master, yay

PR #1225, opened 2015-09-08
Re-do NodeList and HTMLCollection

domenic:
- Removes all non-public properties and methods, in the process fixing #1224.
- Removes NodeList's no-op length setter, fixing #859. It was only used so that we could call Array.prototype.push on it; instead we manually copy over nodes.
- Removes their public constructors.
- Removes the inheritance relationship between them, which is incorrect.
- Removes HTMLOptionsCollection, which was just an alias for HTMLCollection. We'll use HTMLCollection instead, for now.

### DIFF
--- a/lib/jsdom/level1/core.js
+++ b/lib/jsdom/level1/core.js
@@ -22,6 +22,9 @@ var proxiedWindowEventHandlers = require("../living/helpers/proxied-window-event
 const domSymbolTree = require("../living/helpers/internal-constants").domSymbolTree;
 const NODE_TYPE = require("../living/node-type");
 const resetDOMTokenList = require("../living/dom-token-list").reset;
+const createLiveNodeList = require("../living/node-list").createLive;
+const updateNodeList = require("../living/node-list").update;
+const updateHTMLCollection = require("../living/html-collection").update;
 
 // utility functions
 var attachId = function(id,elm,doc) {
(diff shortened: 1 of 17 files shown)


lib/jsdom/level1/core.js
```diff
@@ -22,6 +22,10 @@ var proxiedWindowEventHandlers = require("../living/helpers/proxied-window-event
 const domSymbolTree = require("../living/helpers/internal-constants").domSymbolTree;
 const NODE_TYPE = require("../living/node-type");
 const resetDOMTokenList = require("../living/dom-token-list").reset;
+const createLiveNodeList = require("../living/node-list").createLive;
+const createStaticNodeList = require("../living/node-list").createLive;
```
Sebmaster: `createStatic`

lib/jsdom/living/parent-node.js
```diff
@@ -14,13 +16,14 @@ module.exports = function (core) {
...
-        this._childrenList = new core.HTMLCollection(this, function () {
+        this._childrenList = createHTMLCollection(this, function () {
           return domSymbolTree.childrenToArray(self, {filter: function (node) {
             return node.nodeType === NODE_TYPE.ELEMENT_NODE;
           }});
         });
       }
-      this._childrenList._update();
+
+      updateHTMLCollection(this._childrenList);
```
Sebmaster: `create` already does update. Put it into an `else` to avoid building the collection twice? Same deal with NodeCollections.

lib/jsdom/level1/core.js
Sebmaster: Just saw that this isn't even used in the

domenic: Oh nice catch... shouldn't jshint be catching me... meh.

lib/jsdom/living/parent-node.js
domenic: Good idea.

PR #1227, opened 2015-09-10
Automatically generate classes from IDL

Sebmaster:
Let's get some discussion going in here.
What do you think? Where're we going from here?

Missing points:
- [x] publish new webidl-conversions
- [x] publish new webidl2js
- [x] fix package.json versions for the packages


domenic:
I've added you to webidl-conversions and webidl-html-reflector. However, please add tests for your new additions before publishing new versions, and make sure to bump the major since the consumption is via require() instead of import now.

.npmignore
```diff
@@ -0,0 +1 @@
+!lib/jsdom/web-idl/generated/
```
domenic: We're using files in package.json, not npmignore.

lib/jsdom/level1/core.js
```diff
@@ -12,7 +12,7 @@ var HtmlToDom = require("../browser/htmltodom").HtmlToDom;
 var Location = require("../browser/location");
 var vm = require("vm");
 var CookieJar = require('tough-cookie').CookieJar;
-var EventTarget = require("../events/EventTarget");
+var EventTarget = require("../web-idl/generated/EventTarget");
```
domenic: I'd prefer ../living/generator. web-idl folder is for things defined in the actual Web IDL spec, like DOMException.

lib/jsdom/level1/core.js
```diff
@@ -245,7 +245,7 @@ core.DOMImplementation.prototype = {
 };
 
 core.Node = function Node(ownerDocument) {
-  EventTarget.call(this);
+  EventTarget.init.call(this);
```
domenic: Why is init a method instead of a function? (i.e. why `.call(this)` instead of just `(this)`?)

lib/jsdom/web-idl/implementation/EventTarget.js
```diff
@@ -0,0 +1,231 @@
+"use strict";
```
domenic: Do you think it might be easier to keep files next to their IDL, instead of in a sibling folder?

lib/jsdom/web-idl/implementation/EventTarget.js
```diff
@@ -0,0 +1,231 @@
...
+const DOMException = require("../DOMException");
+const reportException = require("../../living/helpers/runtime-script-errors");
+const domSymbolTree = require("../../living/helpers/internal-constants").domSymbolTree;
+
+class EventTarget {
+  addEventListener(type, callback, capture) {
+    if (callback === undefined || callback === null) {
```
domenic: All of this (lines 8 through 16) should be handled by webidl2js, no? It's OK if it isn't right now, but we should at least add comment delimiters.

lib/jsdom/level1/core.js
Sebmaster: So init doesn't have to do argument slicing (if called as part of the constructor) and can just be passed arguments via apply.

lib/jsdom/web-idl/implementation/EventTarget.js
```diff
@@ -0,0 +1,231 @@
...
+
+    return !event._canceled;
+  }
+}
+
+module.exports = {
+  init() {
+    this._events = Object.create(null);
+  },
+  interface: EventTarget
```
domenic: I think a better name is probably "implementation"? And the class should maybe be named EventTargetImpl?

scripts/webidl/pull.js
```diff
@@ -0,0 +1,49 @@
+"use strict";
```
domenic: Not convinced this needs to be in jsdom core.

scripts/webidl/convert.js
```diff
@@ -0,0 +1,44 @@
+"use strict";
+const fs = require("fs");
+const path = require("path");
+
+const recursive = require("recursive-readdir");
+const webidl2js = require("webidl2js");
+
+const optimist = require("optimist")
+  .usage("Pull all webidl segments froma a web page")
```
domenic: Seems like the wrong usage message?

scripts/webidl/convert.js
```diff
@@ -0,0 +1,44 @@
...
+  .string("input")
+  .alias("input", "i")
+  .describe("input", "the input file or directory")
+  .string("implDir")
```
domenic: Command line flags are traditionally kebab-case, not camelCase. So --impl-dir etc.

scripts/webidl/convert.js
```diff
@@ -0,0 +1,44 @@
...
+
+const inputStat = fs.statSync(optimist.argv.input);
+if (inputStat.isDirectory()) {
+  recursive(optimist.argv.input, readFiles);
+} else {
+  readFiles(null, [optimist.argv.input]);
+}
+
+function readFiles(err, files) {
+  let src = "";
```
domenic: Needs some better error handling; `throw err` seems fine potentially.

scripts/webidl/convert.js
```diff
@@ -0,0 +1,44 @@
...
+  files.forEach(function (f) {
+    src += fs.readFileSync(f, { encoding: "utf8" });
+  });
+
+  webidl2js.generate(src, optimist.argv.outputDir, optimist.argv.implDir, { suppressErrors: true });
```
domenic: I am a bit confused what the input to and output of webidl2js.generate is. It takes a string consisting of ... the concatenation of all IDL in the currently-being-recursed directory? But all calls use the same output and impl directory?

What errors are suppressed?

scripts/webidl/specs/events/EventTarget.idl
```diff
@@ -0,0 +1,6 @@
+[Exposed=(Window,Worker)]
```
domenic: Would prefer this to live in living or living/dom or something. Again probably nicer next to EventTarget-impl.js.

scripts/webidl/specs/events/EventTarget.idl
```diff
@@ -0,0 +1,6 @@
...
+interface EventTarget {
+  void addEventListener(DOMString type, EventListener? callback, optional boolean capture = false);
+  void removeEventListener(DOMString type, EventListener? callback, optional boolean capture = false);
+  boolean dispatchEvent(Event event);
+};
```
domenic: Newline at end of file!

domenic:
Review done. Excited to see progress on this, and happy that you found a good first integration attempt.

scripts/webidl/convert.js
Sebmaster: .generate automatically creates files for each interface. It therefore doesn't really care how much webidl you give it.

Suppressed errors are stuff webidl2js can't currently convert like `dictionary` and stuff (basically everything except interfaces).

lib/jsdom/level1/core.js
domenic: Hmm. I dunno, I think argument slicing will be pretty necessary anyway, given how few real constructors there are. Even those things with constructors have hidden parameters (e.g. ownerDocument on Text and friends).

Could you give an example where you think this will work out well?

scripts/webidl/convert.js
domenic: I see. Maybe to make that clearer you could turn the code into `getConcatenatedSource(input, function (err, source) { if (err) { throw err; } webidl2js.generate(source, ...); })`, or its promise equivalent.

lib/jsdom/level1/core.js
Sebmaster: The constructor of generated classes always does argument conversion and then redirects to generated.init, which does object creation (in cases where non is provided, i.e. instancing of objects without a constructor (HTMLElements, etc.)) and then forwards to impl.init which is the "proper" constructor.
I want to keep the init method as close to a real constructor as possible - which also has the added benefit of being able to just provide a `function Klass` definition as-is as both interface and init.

domenic: I still would appreciate an example. Given that the constructor of generated classes will usually be length-zero and take a secret.

lib/jsdom/web-idl/implementation/EventTarget.js
Sebmaster: `implementation` would then generate code like `Impl.implementation.myCall(args)` which isn't that elegant either :(

If we suffix everything with the `Impl`, we could use that name, although I don't really like the suffix either...

domenic: Well I'm more just concerned that `interface` is an invalid name here. This isn't the interface; it's the implementation class behind the interface.

lib/jsdom/level1/core.js
Sebmaster: So, this is mostly just about avoiding having to copy args and also allow a prototype-based declaration approach to allow a bit of elegance in the implementation: Let's imagine having a generated `Event` and we do our own extensions from there (like we currently have with EventTarget).

```js
function EventImpl(type, initArgs) {
  // my init code
}

module.exports = {
  init: EventImpl, // nice trick, impossible with class-based declaration syntax though
  interface: EventImpl
};
```

Event has a constructor, so we need to provide it type and initArgs. So our MouseEvent class would look (simplified) like this:
```js
function MouseEvent(type, initArgs) {
  Event.init.call(this, type, initArgs);
}
```
init then just does:
```js
function init() {
  Impl.init.apply(this, arguments);
}
```


If we do function instead of method:
```js
function MouseEvent(type, initArgs) {
    Event.init(this, type, initArgs);
}
```
`Event.init`
```js
function init(obj) {
  const args = []; // copying not required in first approach
  for (let i = 1; i < arguments.length; ++i) {
     args[i - 1] = arguments[i];
  }
  Impl.init.call(this, type, initArgs);
}
```

Third choice would be to forward the additional arg all the way through to the `Impl.init` method:
```js
function MouseEvent(type, initArgs) {
    Event.init(this, type, initArgs);
}
```
`Event.init`
```js
function init(obj) {
  Impl.init.apply(null, arguments);
}
```
`Impl.init`
```js
function init(obj) { // explicit declaration not required in previous two approaches
  const args = []; // copying not required in first approach
  for (let i = 1; i < arguments.length; ++i) {
     args[i - 1] = arguments[i];
  }
  EventImpl.apply(obj, args);
}
```

domenic: Why are you assuming Impl.init is a method? It should just be a function too.

Sebmaster: Yeah, but that'd mean either go with variant 3, or make using a function constructor directly as init method impossible.

domenic: I guess that is kind of the last example but it's only if you are deciding ahead of time that you want EventImpl to do the initialization. Don't do that.

domenic: Basically I am against trying to use things for purposes they were not meant for. The init stuff is clearly neither a constructor nor a method. We shouldn't use the language's built-in method and constructor functionality for that.

E.g. if I were to do the prototype approach for impls, then I would just declare prototype objects and not have them be attached to classes.

domenic: OK, that last comment really clarified things for me actually. I see two good approaches:

1. More explicitly using composition. In the constructor, do `this[impl] = new EventTargetImpl(options)`, then in the methods, do `this[impl].method.apply(this[impl], arguments)`. The impls are actually classes.

2. Or, just admit they're bags of methods not meant to be used as standalone classes, and use object literals. `Impl.interface.prototype.addEventListener.apply(this, args)` isn't actually any better than `implMethods.addEventListener.apply(this, args)`.

Sebmaster: 1 won't work since symbols don't transfer well over vm proxy right now, which would make it impossible to do in window.

I guess 2 is true, but since the impl classes are the thing you actually maintain, I think it should kind of resemble what it's used for instead of looking like random util classes.

domenic: I mean, window is a special case. I think we fixed symbols in new io.js, but there are probably issues there in general, yeah.

Still not entirely sure what the problem is though. All the symbols stay on the impl instance. Note that it's `this[impl].method.apply(this[impl], args)`, not `this[impl].method.apply(this, args)`.

domenic: Oh the problem is the `[impl]` symbol itself, duh.

Still, I'd be fine not being able to do Window until later...

Sebmaster: It's still `this[impl]`, you weren't able to retrieve that in io 2.x cross vm borders.

domenic: That was fixed in https://github.com/nodejs/node/issues/884, which first appeared in 3.0.0 :).

But, stuff like https://github.com/nodejs/node/issues/2734 is not encouraging.

Sebmaster: Welp, looks like we'll not be able to convert Window anyways...

scripts/webidl/specs/events/EventTarget.idl
Sebmaster: So we put .idl, interface.js and -impl.js into the same dir?

domenic: Where interface.js is generated? I think that seems reasonable...

.npmignore
```diff
@@ -0,0 +1,3 @@
+/*
+!lib/
+lib/**/*.idl
```
domenic: Newline at end of file. Get an editorconfig plugin for your editor?

lib/jsdom/living/generator/events/EventTarget.idl
```diff
@@ -0,0 +1,6 @@
+[Exposed=(Window,Worker)]
```
domenic: What does "generator" mean in the file path?

scripts/webidl/convert.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+function generateClasses(src, outputDir, implDir, utilPath) {
+  webidl2js.generate(src, outputDir, implDir, { suppressErrors: true, implSuffix: "-impl", utilPath: utilPath });
+}
+
+function doConversion(inputPath) {
+  let isDir;
+
+  Q.nfcall(fs.stat, inputPath)
```
domenic: return this promise, then the last line becomes `doConversion(...).done()`, to ensure errors get propagated to the console.

lib/jsdom/living/generator/events/EventTarget.idl
Sebmaster: Heh, I just went by https://github.com/tmpvar/jsdom/pull/1227#discussion_r39228742 on this one.

domenic: Wow, sorry, I have no idea what I was thinking when I said that. Probably either living/generated/events or living/events? My bad.

domenic:
OK, so, major things in my mind that need to be resolved:

- What do generated and subclass constructors look like? I noticed that Node has to do `EventTarget.init.call(this)`. That's not OK to me, since `init` is not a method, so `.call` usage here is abuse.
  - Ideally we'd do something like `super(secret)`, maybe? That would require converting Node and other subclasses to real classes, but that's probably fine. (We'd just convert the constructor, to start.) The idea being that the generated EventTarget constructor takes care of calling the impl's init functionality, somehow.
  - Alternately, I'd be fine with `EventTarget.init(this)`. Not quite as nice, but probably easier to implement.
- What are impls, really? I see two approaches, as outlined. I think we need to adopt one of them.
  - Actual classes, for which instances get created. This would involve each `EventTarget` creating an `EventTargetImpl` and delegating using `this[impl].method(...args)`. The constructor for the impl performs the initialization of the necessary stuff.
  - Bags of methods. We stop creating impl classes and just create impl object literals. Then the delegating is done with `impl.method.call(this, ...args)`. Basically they're mixins, although not the kind you transparently mix in via transferring to another object. In this case there is no constructor, but there is an `init(obj)` function that comes as a package deal with the methods. So the generated constructor does `impl.init(this)`.

What do you think among these alternatives?

Sebmaster:
Okay so, I'd actually like to go with the class approach, but since we don't have private Symbols (yet) and I don't want to publicise any non-spec properties I'd probably just go with the method-bag/mixin approach.

Regarding super:
If the descendant is also generated, it can just super(...arguments) (although that's currently broken since spread args are not in iojs2 and in v4 only with flags. I'll check in how far reworking init in combination with the others makes sense.

domenic:
Well, the method-bag/mixin approach will probably introduce a (non-private) property or symbol or two of its own, for the private internals of the class. E.g. `_events` or `[events]`. So I don't see much of a difference in that regard.

Sebmaster:
But that's for the implementor to decide :stuck_out_tongue:

domenic:
I guess that's true. But on the flip-side if we encapsulate it entirely in one public symbol in the binding layer, in the future we could do an upgrade to zero public symbols in one small code change.

domenic:
But anyway, happy to go with whichever you decide.

PR #1229, opened 2015-09-11
Add jsdom.nodeLocation

### LINKED ISSUE #1194: Line numbers/offsets for nodes
I have need of knowing what line number in the source document a node comes from. This is so I can display useful errors to authors when they have errors in their markup. You could imagine a symbol on the node that gives row/col info.

Since row/col comes with some cost to performance, you could instead have a symbol for offset and let consumers map to row/col themselves, although the total overhead of consumers re-scanning the document for linebreaks is larger than just doing that work at parse-time.

domenic:
Fixes #1194, pending further parse5 updates to make it even better: https://github.com/inikulin/parse5/issues/67

/cc @bterlson. Should meet your needs for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ jsdom.env({
 });
 ```
 
-### Canvas
+## Canvas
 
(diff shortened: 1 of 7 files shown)


Sebmaster:
LGTM.

bterlson:
:+1:

Issue #1208, opened 2015-08-21
Allow overwrite of window.top


cpojer:
We have a few tests in jest that attempt to simulate a parent-child relationship between a window and the window of an iframe. With recent versions of jsdom it seems like it isn't possible to overwrite `window.top` any longer. Are you guys interested in supporting this use case? :)

domenic:
You should be able to do so using `Object.defineProperty` without any issue...

domenic:
(Alternately, jsdom's iframe support is pretty solid, so you might be able to use that directly)

domenic:
Let me know if anything isn't working with defineProperty! Otherwise, closing for now.

cpojer:
Sorry for taking so long to get back to you, it seems like defineProperty doesn't work in this case, which is why I opened this issue in the first place (although, I did forget whether I tried this before). Unfortunately for now I can't use the iframe support in jsdom – I need my code to run in an iframe-like environment and I'm not ready yet to enable env switching in jest.

I tried this:

```js
      if (!isTop) {
        console.log('top is set before?', !!window.top);
      }
      Object.defineProperty(window, 'top', {
        writable: false,
        value: null,
      });
      if (!isTop) {
        console.log('top is set?', !!window.top);
      }
```

And `window.top` is always set and equals `window` no matter what I use as a `value`. Using `get` doesn't work either.

domenic:
Investigating. This appears to be an io.js bug with the global/global proxy. `Object.getOwnPropertyDescriptor(window, "top")` is returning a data descriptor, not an accessor descriptor, which is not good news... :-S

domenic:
Commenting out the line

```js
        this._ownerDocument._defaultView._globalProxy = vm.runInContext("this", this._ownerDocument._global);
```

in level1/core.js "fixes" it, but I have no idea why...

cpojer:
woah, no idea what is going on here.

domenic:
Yeah just kind of giving live updates as I debug... unfortunately this is not as easy as I'd hoped. I'm hoping to find a workaround at the very least though.

domenic:
OK, I think I see the issue.

Any "contextified" object, whose getters/setters are handled via the V8 named property callbacks, can only have data properties, it seems. That is, when you ask it what properties it has, it will always tell you it has a data property. Even if we've put a getter on the object.

[...]

PR #1230, opened 2015-09-11
Add jsdom.reconfigureWindow to solve #1208

domenic:
@cpojer, does this look like it will meet your needs?


cpojer:
Yes, absolutely! Thanks so much for spending time on this.

PR #1233, opened 2015-09-12
Re-do getElementsByClassName

### LINKED ISSUE #1232: getElementsByClassName() doesn't work with multiple class names
For convenient reference, [here's the relevant spec](
https://dom.spec.whatwg.org/#dom-document-getelementsbyclassname).

This example is admittedly little contrived (I'm working with a different document/DOM in my real code), but it sufficiently illustrates the issue.

### Test Case

```Javascript
import { env } from "jsdom";

new Promise(resolve => {
  env('https://github.com', (err, window) => resolve(window.document));
})
.then(document => {
  console.log(document.getElementsByClassName('header header-logged-out'));
});

// expect: (one element)
// result: {}
```

&nbsp;
However, making the same query from the devtools in an ***Incognito window*** for the URI https://github.com, it returns a non-zero number of elements:

screenshot

I'm seeing the same behavior in my test suite -- I can run it in the browser and it works as expected, but with the very same document-parsing code and JSDOM, I get an empty results set given a `getElementsByClassName()` query with more than one class name:

`document.getElementsByClassName('classA classB')`

Also, for what it's worth, if I change the above query to use either one of...
- `document.getElementsByClassName('classA')`
- `document.getElementsByClassName('classB')`

... I get a list of results.

:smile:

domenic:
Fixes #1232 plus a bunch of other edge cases around whitespace and such. Also returns a HTMLCollection now instead of a NodeList, and memoizes the result, as is explicitly allowed in the spec.

The change from _clearMemoizedQueries() to _modified() in the base _attrChanged was necessary for the newly-introduced web platform tests test to pass, and likely fixes other subtle bugs people may have encountered.


Joris-van-der-Wel:
looks good

PR #1234, opened 2015-09-14
ChildNode implementation

zaucy:
I attempted to create the [ChildNode Interface](https://dom.spec.whatwg.org/#interface-childnode). I just used the parent's functions `insertBefore()`, `appendChild()`, and `removeChild()` to implement them. I'm sure it could be more efficient if using the internals of jsdom, but this should provide the correct behaviour.

I couldn't get `npm test` to work on my machine for some reason, but I did put some tests for ChildNode. These tests are untested. I did some basic tests of my own on some other projects I'm working on and didn't find an issue with the implementation yet.


domenic:
You can use the Travis build to make sure your tests and lint are passing. For example https://travis-ci.org/tmpvar/jsdom/jobs/80188055 shows a lot of lint errors.

test/living-dom/child-node.js
```diff
@@ -0,0 +1,51 @@
+"use strict";
+
+var jsdom = require("../..");
+var types = require("../lib/jsdom/node-type.js");
```
domenic: Do not use jsdom internals in tests. you can use `doc.TEXT_NODE` or similar.

test/runner
```diff
@@ -50,6 +50,7 @@ var files = [
   "living-dom/node-iterator.js",
   "living-dom/parent-node.js",
   "living-dom/non-document-type-child-node.js",
+  "living-dom/child-node.js",
```
domenic: Add to the browser runner too. (It requires two lines worth of additions)

lib/jsdom/living/child-node.js
```diff
@@ -0,0 +1,91 @@
...
+        for(var i=0; nodes.length > i; i++) {
+          var node = nodes[i];
+          this.parentNode.insertBefore(node, this);
+        }
+      }
+    },
+
+    // Inserts nodes just after node, while replacing strings in nodes with equivalent Text nodes.
+    // https://dom.spec.whatwg.org/#dom-childnode-after
+    after: function() {
```
domenic: Use `after() {` instead of `after: function() {`.

lib/jsdom/living/child-node.js
```diff
@@ -0,0 +1,91 @@
...
+      if(this.parentNode !== null) {
+
+        var nodes = ensureNodeArguments(this._ownerDocument, arguments);
```
domenic: One thing the linter does not catch is that you should always use `const` where possible, and `let` where not.

lib/jsdom/living/child-node.js
```diff
@@ -0,0 +1,91 @@
...
+        var nextSibling = this.nextSibling;
```
domenic: See https://dom.spec.whatwg.org/#dom-childnode-replacewith. It's the first following sibling *not in _nodes_*.

domenic:
In general I think you need to follow the spec a bit more closely. E.g. looking at https://dom.spec.whatwg.org/#dom-childnode-replacewith there are a lot of subtle things done to take care of cases where e.g. one of the arguments is the node itself or the parent node.

domenic:
Hey, sorry I didn't see that you added more commits the day after I commented.

I've rebased and fixed a few style issues in https://github.com/tmpvar/jsdom/tree/childnode-take2. However, I also enabled the relevant web platform tests, which show how this implementation is still not very spec compliant. (Do `npm test -- -s web-platform-tests` to see those specifically.) If you want to try basing some work off that, and getting the tests to pass, then we can accept this in to jsdom.

PR #1235, opened 2015-09-14
Iterable nodelist

darobin:


domenic:
Published as 6.5.0, yay!

Issue #1207, opened 2015-08-19
Replace prototype chain of our objects in window


Sebmaster:
Currently it's quite easy to break out of jsdom and get access to modules like `fs` and similiar, which allow raw access to the systems. The most obvious way is walking the prototype chain up to function and evaling to get process and from there require. However, replacing other methods, e.g. on object will also lead to jsdom executing those unintentionally.

Fix this by overwriting all our prototypes in dom by dummy chains derived from a newly created vm context. This can be done at require time (which doesn't lead to quite correct behaviour, but it doesn't regress, so it's fine for now).

Joris-van-der-Wel:
Content Security Policy (the `CSP` header) in chrome lets you disable any use of `eval` (`setTimeout`, `Function`, etc). Perhaps io.js could expose this functionality for vm?

Sebmaster:
Whatever's in VM is not a big deal, we can overwrite that stuff ourselves (we do, in fact, with set timeout and set interval at least), but we currently kill the sandbox by introducing outside objects into it with their prototype chain.

Sebmaster:
Just as a note:
We can actually introduce this without penalizing our require time, nor our runtime. We can actually just add a method like `jsdom.secureSandbox()` which can be called explicitly to harden jsdom against untrusted sites.

Joris-van-der-Wel:
I am not sure if `env()` could ever be secured properly without running jsdom itself in a sandbox. I could bypass that prototype chain workaround by simply abusing the internals of jsdom, this would turn into an exercise of securing every part of jsdom' api.

Sebmaster:
Running the whole of jsdom in the vm sandbox was also one of our considerations, however that comes with a reeeeally bad window init perf, which is not something we can do (I think, untested). We'd also need to implement some kind of message-passing for loading resources.

Making internals inaccessible is an ongoing effort - every time we rewrite stuff we try to migrate to Symbols (and hopefully, we can make them private some day).

Joris-van-der-Wel:
Such changes would be good, however I am not confident that it will be secure.

Sebmaster:
It'll never be (also we can't pay for detecting security leaks like Chrome :stuck_out_tongue:), but it's just about reducing our attack surface - some methods of executing code are just too obvious.

PR #1244, opened 2015-09-24
add proper networking

nicolashenry:
Hello,

I took some time to enhance jsdom networking.
I added a new XMLHttpRequest implementation based on "w3c-xmlhttprequest" but using "request" for requests and "deasync" for the synchronous mode.
I added some request management options (see pool, agentOptions and custom loader in the readme file).

I added some tests from "web-platform-tests" XMLHttpRequest tests but using the python server. This implies to add the "web-platform-tests" hosts in order to be able to start all tests.

Everything seems ok with Travis : https://travis-ci.org/nicolashenry/jsdom/builds/81976743


domenic:
Wow!! This looks amazing.

I am going to review and leave lots of nits, but please don't let them discourage you. You've implemented a lot of stuff here, which is so great.

README.md
```diff
@@ -134,6 +134,8 @@ jsdom.env(config);
 - `config.done`, `config.onload`, `config.created`: see below.
 - `config.concurrentNodeIterators`: the maximum amount of `NodeIterator`s that you can use at the same time. The default is `10`; setting this to a high value will hurt performance.
 - `config.virtualConsole`: a virtual console instance that can capture the window’s console output; see the "Capturing Console Output" examples.
+- `config.pool`: an object describing which agents to use for the requests, defaults to {maxSockets: 6}, see [request module](https://github.com/request/request#requestoptions-callback) for more details.
```
domenic: Both here and on the next line:

s/requests, defaults/requests; defaults/

Put spaces after { and before }, and make the default options `code`.

README.md
```diff
@@ -304,6 +306,9 @@ jsdom lets you intercept subresource requests using `config.resourceLoader`. `co
   - `error`: either `null`, if nothing goes wrong, or an `Error` object.
   - `body`: a string representing the body of the resource.
 
+You can return an object containing an `abort(emitEvent)` function which will be called if the window is closed or stopped before the request ends.
```
domenic: s/function/method

What is the difference between window closed and window stopped?

What is an abort error?

Why is it called emitEvent?

`emitEvent`, `false`, and `true` should all be code.

lib/jsdom.js
```diff
@@ -72,6 +72,17 @@ exports.jsdom = function (html, options) {
       `"xml", "auto", or undefined`);
   }
 
+  options.userAgent = options.userAgent || "Node.js (" + process.platform + "; U; rv:" + process.version + ")";
```
domenic: This needs readme documentation

lib/jsdom.js
```diff
@@ -353,6 +375,17 @@ function getConfigFromArguments(args) {
     config.url = toFileUrl(config.file);
   }
 
+  config.pool = typeof config.pool !== "undefined" ? config.pool : {
```
domenic: I know this code is messy. But is there any way to avoid duplicating the defaulting code?

lib/jsdom.js
```diff
@@ -353,6 +375,17 @@ function getConfigFromArguments(args) {
...
+    maxSockets: 6
+  };
+
+  config.agentOptions = typeof config.agentOptions !== "undefined" ? config.agentOptions : {
```
domenic: Use `config.agentOptions !== undefined`, not `typeof`.

lib/jsdom/browser/Window.js
```diff
@@ -1,13 +1,13 @@
 "use strict";
 
 var CSSStyleDeclaration = require("cssstyle").CSSStyleDeclaration;
-var XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest;
+var file = require("./file");
```
domenic: Can you follow the style used in `living`, putting these files there, instead of putting these as siblings in `jsdom/browser`? Example:

https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/url.js plus https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/index.js#L30

lib/jsdom/browser/Window.js
```diff
@@ -168,70 +171,40 @@ function Window(options) {
...
-    xhr.send = function (data) {
-      var cookieJar = window.document._cookieJar;
-      var cookieStr = cookieJar.getCookieStringSync(lastUrl, {http: true});
-
-      if (cookieStr) {
-        xhr.setDisableHeaderCheck(true);
-        xhr.setRequestHeader("cookie", cookieStr);
-        xhr.setDisableHeaderCheck(false);
-      }
+  this.File = file.File;
```
domenic: As per the above comment, the main export of `file.js` should be a function like

```js
module.exports = core => {
  core.File = File;
  core.Blob = Blob;
};
```

Then no code needs to appear here in window.js at all.

lib/jsdom/browser/Window.js
```diff
@@ -168,70 +171,40 @@ function Window(options) {
...
+  this.Blob = file.Blob;
 
-      function setReceivedCookies() {
-        if (xhr.readyState === xhr.HEADERS_RECEIVED) {
-          var receivedCookies = xhr.getResponseHeader("set-cookie");
+  this.FileReader = file.createFileReader({
+    window: window
```
domenic: Style nit, here and elsewhere: do `file.createFileReader({ window })`, using shorthand object literal syntax and all on one line.

lib/jsdom/browser/Window.js
```diff
@@ -168,70 +171,40 @@ function Window(options) {
...
-          return JSON.parse(this.responseText);
-        } else {
-          return null; // emulate failed request
-        }
-      },
-      enumerable: true,
-      configurable: true
-    });
-    return xhr;
+  this.ArrayBuffer = ArrayBuffer;
```
domenic: None of these are needed. They come automatically when we contextify. Please don't add them.

lib/jsdom/browser/Window.js
```diff
@@ -168,70 +171,40 @@ function Window(options) {
...
+  this.Uint8Array = Uint8Array;
+  this.Uint8ClampedArray = Uint8ClampedArray;
+  this.Int16Array = Int16Array;
+  this.Uint16Array = Uint16Array;
+  this.Int32Array = Int32Array;
+  this.Uint32Array = Uint32Array;
+  this.Float32Array = Float32Array;
+  this.Float64Array = Float64Array;
+
+  this.stop = function () {
```
domenic: Wow, I did not know about this function. Nice.

lib/jsdom/browser/file/blob.js
```diff
@@ -0,0 +1,78 @@
+"use strict";
+
+function Blob() {
```
domenic: For new code, please use ES2015 class syntax and style. (E.g. `const` everywhere possible, `let` when not, never `var`.)

If you're feeling brave, you can try using webidl2js, like we are doing for EventTarget. But that is still very early so it might not be ready yet.

lib/jsdom/browser/file/blob.js
```diff
@@ -0,0 +1,78 @@
...
+
+  const buffers = [];
+
+  if (parts) {
+    var l = +parts.length;
+    for (var i = 0; i < l; i++) {
+      var part = parts[i];
+      var buffer;
+      if (part instanceof ArrayBuffer) {
+        buffer = new Buffer(new Uint8Array(part));
```
domenic: Why use Buffer instead of just sticking with Uint8Array exclusively? Buffers are kind of gross.

lib/jsdom/browser/file/blob.js
```diff
@@ -0,0 +1,78 @@
...
+        part instanceof Float64Array) {
+        buffer = new Buffer(new Uint8Array(part.buffer, part.byteOffset, part.byteLength));
+      } else {
+        buffer = new Buffer(typeof part === "string" ? part : String(part));
+      }
+      buffers.push(buffer);
+    }
+  }
+
+  this._buffer = Buffer.concat(buffers);
```
domenic: Another thing that new-style code does is use symbols for private state. This is one example: https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/hash-change-event.js

lib/jsdom/browser/file/filereader.js
```diff
@@ -0,0 +1,202 @@
+"use strict";
+
+var DOMException = require("../../web-idl/DOMException");
+var EventTarget = require("../../living/generated/events/EventTarget");
+var core = require("../../level1/core");
+require("../../level2/core");
```
domenic: This kind of requiring is not good (it easily leads to circular dependencies, due to jsdom's messy legacy codebase nature). Instead, use the style in living/, where the module.exports function takes `core` as an argument.

lib/jsdom/browser/file/filereader.js
```diff
@@ -0,0 +1,202 @@
...
+    self.dispatchEvent(new ProgressEvent("loadstart"));
+
+    process.nextTick(function () {
+
+      var data = file._buffer;
+      if (!data) {
+        data = new Buffer("");
+      }
+      self.dispatchEvent(new ProgressEvent("progress", {
+        lengthComputable: (!isNaN(file.size)) ? true : false,
```
domenic: You can just do `!isNaN(file.size)`, no need for the ternary

lib/jsdom/browser/file/filereader.js
```diff
@@ -0,0 +1,202 @@
...
+    if (file._closed) {
+      self.error = new DOMException(DOMException.INVALID_STATE_ERR);
+      self.dispatchEvent(new ProgressEvent("error"));
+    }
+
+    self.readyState = self.LOADING;
+    self.dispatchEvent(new ProgressEvent("loadstart"));
+
+    process.nextTick(function () {
+
```
domenic: Nit: excess blank line, here and below

lib/jsdom/browser/file/filereader.js
```diff
@@ -0,0 +1,202 @@
...
+        process.nextTick(function () {
+          if (self.readyState !== self.LOADING) {
+            self.dispatchEvent(new ProgressEvent("loadend"));
+          }
+        });
+      });
+    });
+  }
+
+  var constants = {
```
domenic: We have an easier way to do this. Search for usages of `addConstants`.

lib/jsdom/browser/resource-loader.js
```diff
@@ -36,14 +36,32 @@ function createResourceLoadHandler(element, resourceUrl, document, loadCallback)
   };
 }
 
-function readFile(url, callback) {
-  var filePath = url
-    .replace(/^file:\/\//, "")
-    .replace(/^\/([a-z]):\//i, "$1:/")
-    .replace(/%20/g, " ");
+exports.readFile = function (filePath, callback) {
+  var readableStream = fs.createReadStream(filePath);
```
domenic: Style nit: while updating new code, use `const` and `let`.

lib/jsdom/browser/resource-loader.js
```diff
@@ -36,14 +36,32 @@ function createResourceLoadHandler(element, resourceUrl, document, loadCallback)
...
+  readableStream.on("data", function (chunk) {
+    data += chunk;
+  });
+
+  readableStream.on("end", function () {
+    callback(null, data);
+  });
+
+  return {
+    abort: function (emitEvent) {
```
domenic: use shorthand: `abort(emitEvent)`.

lib/jsdom/browser/resource-loader.js
```diff
@@ -55,12 +73,42 @@ function wrapCookieJarForRequest(cookieJar) {
   return jarWrapper;
 }
 
-function fetch(urlObj, cookieJar, referrer, callback) {
-  if (urlObj.hostname) {
-    exports.download(urlObj, null, cookieJar, referrer, callback);
+function fetch(element, urlObj, cookieJar, referrer, requestManager, callback) {
+  var req;
+  var _callback = requestManager ? function () {
```
domenic: No underscores on variables. Call it wrappedCallback perhaps.

lib/jsdom/browser/resource-loader.js
```diff
@@ -55,12 +73,42 @@ function wrapCookieJarForRequest(cookieJar) {
...
+    requestManager.remove(req);
+    callback.apply(this, arguments);
+  } : callback;
+  if (urlObj.protocol === "data:") {
```
domenic: WOW!!! data URL support? You are amazing!

lib/jsdom/browser/xhr/formdata.js
```diff
@@ -0,0 +1,42 @@
+"use strict";
+
+var stream = require("stream");
+
+var Blob = require("../file/blob");
+
```
domenic: Since this is an incomplete implementation of FormData, add a TODO note at the top to finish it out.

lib/jsdom/browser/xhr/utils.js
```diff
@@ -0,0 +1,235 @@
...
+  while (n--) {
+    var key = keys[n];
+    if (key.toLowerCase() === header.toLowerCase()) {
+      return requestHeaders[key];
+    }
+  }
+  return null;
+}
+
+utils.getRequestHeader = getRequestHeader;
```
domenic: Can you just add these directory to exports instead of creating a utils object?

lib/jsdom/browser/xhr/utils.js
```diff
@@ -0,0 +1,235 @@
...
+
+var base64Regexp = /^(?:[A-Z0-9+\/]{4})*(?:[A-Z0-9+\/]{2}==|[A-Z0-9+\/]{3}=|[A-Z0-9+\/]{4})$/i;
+
+utils.createClient = function createClient(settings, body, callback, callbackError) {
+  var xhr = this;
+  var window = settings.options.window;
+  var baseUrl = window.document.URL === "about:blank" ? window.parent.document.URL : window.document.URL;
+  var uri = url.resolve(baseUrl.toString(), settings.uri);
+  var urlObj = url.parse(uri);
+  var response;
```
domenic: Don't declare variables before they are used

lib/jsdom/browser/xhr/utils.js
```diff
@@ -0,0 +1,235 @@
+"use strict";
+
+var request = require("request");
+var url = require("url");
```
domenic: The built-in URL module is not allowed in jsdom. Instead use the URL constructor from util.js

lib/jsdom/browser/xhr/utils.js
```diff
@@ -0,0 +1,235 @@
...
+}
+
+utils.getRequestHeader = getRequestHeader;
+
+var base64Regexp = /^(?:[A-Z0-9+\/]{4})*(?:[A-Z0-9+\/]{2}==|[A-Z0-9+\/]{3}=|[A-Z0-9+\/]{4})$/i;
+
+utils.createClient = function createClient(settings, body, callback, callbackError) {
+  var xhr = this;
+  var window = settings.options.window;
+  var baseUrl = window.document.URL === "about:blank" ? window.parent.document.URL : window.document.URL;
```
domenic: Use lib/jsdom/living/helpers/document-base-url.js

lib/jsdom/browser/xhr/utils.js
```diff
@@ -0,0 +1,235 @@
...
+  var rmReq;
+
+  var requestManager = window.document._requestManager;
+
+  if (urlObj.protocol === "file:") {
+    response = new EventEmitter();
+    response.statusCode = 200;
+    response.rawHeaders = [];
+    response.headers = {};
+    var filePath = urlObj.pathname
```
domenic: This should not be necessary when using the real URL parser

lib/jsdom/browser/xhr/utils.js
```diff
@@ -0,0 +1,235 @@
...
+    }
+  }
+  return null;
+}
+
+utils.getRequestHeader = getRequestHeader;
+
+var base64Regexp = /^(?:[A-Z0-9+\/]{4})*(?:[A-Z0-9+\/]{2}==|[A-Z0-9+\/]{3}=|[A-Z0-9+\/]{4})$/i;
+
+utils.createClient = function createClient(settings, body, callback, callbackError) {
```
domenic: Can you document what a "client" is, and how it behaves? It's a bit unclear why this is necessary.

lib/jsdom/browser/xhr/xmlhttprequest.js
```diff
@@ -0,0 +1,891 @@
+"use strict";
+
+var HTTP_STATUS_CODES = require("http").STATUS_CODES;
+var deasync;
+try {
+  deasync = require("deasync");
```
domenic: Augh :(. We cannot use deasync, because it is a native module, and those cause a lot of pain.

Could you instead investigate the techniques in https://github.com/ForbesLindesay/sync-request ?

lib/jsdom/browser/xhr/xmlhttprequest.js
```diff
@@ -0,0 +1,891 @@
...
+var fieldValueRegexp = /^(?:(?![\x00-\x1F\x7F])[\x00-\xFF])*$/;
+
+function createXMLHttpRequest(options) {
+  function XMLHttpRequest() {
+    if (!(this instanceof XMLHttpRequest)) {
+      throw new TypeError("DOM object constructor cannot be called as a function.");
+    }
+    XMLHttpRequestEventTarget.call(this);
+    Object.defineProperties(this, {
+      upload: {
```
domenic: This needs to be a getter like the other properties

lib/jsdom/browser/xhr/xmlhttprequest.js
```diff
@@ -0,0 +1,891 @@
...
+].join("|") + ")$", "i");
+
+var allowedRequestMethods = ["OPTIONS", "GET", "HEAD", "POST", "PUT", "DELETE"];
+var forbiddenRequestMethods = ["TRACK", "TRACE", "CONNECT"];
+
+var tokenRegexp = /^(?:(?![\x00-\x1F\x7F \t\(\)<>@,;:\\"\/\[\]?={}])[\x00-\x7F])+$/;
+var fieldValueRegexp = /^(?:(?![\x00-\x1F\x7F])[\x00-\xFF])*$/;
+
+function createXMLHttpRequest(options) {
+  function XMLHttpRequest() {
```
domenic: As in other files, use class syntax and symbols for this one.

lib/jsdom/browser/xhr/xmlhttprequest.js
```diff
@@ -0,0 +1,891 @@
...
+      enumerable: true,
+      get: function () {
+        return options.window.document;
+      }
+    }
+  });
+
+  return {
+    XMLHttpRequest: XMLHttpRequest,
+    XMLHttpRequestUpload: XMLHttpRequestUpload
```
domenic: XMLHttpRequestEventTarget also needs to be exposed on Window

domenic: Oh, but maybe you take care of that below.

lib/jsdom/level2/events.js
```diff
@@ -157,6 +157,27 @@ inheritFrom(core.Event, core.MutationEvent, {
     REMOVAL      : 3
 });
 
+core.ProgressEvent = function(eventType, eventInit) {
```
domenic: For this, can you follow the pattern used for https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/hash-change-event.js, and place it in its own file, and so on? (You can also use class syntax if you like, or even webidl2js, but at least following that pattern would be helpful, and would ensure that things like getters vs. instance properties are correct.)

lib/jsdom/level2/html.js
```diff
@@ -164,11 +166,46 @@ ResourceQueue.prototype = {
   }
 };
 
+function RequestManager(options) {
```
domenic: Again, use class syntax for new code :)

lib/jsdom/level2/html.js
```diff
@@ -164,11 +166,46 @@ ResourceQueue.prototype = {
...
+  this.pool = options.pool;
```
domenic: Why does this have pool and agentOptions? It's not actually used by this class...

lib/jsdom/level2/html.js
```diff
@@ -452,6 +490,9 @@ define('HTMLFormElement', {
 define('HTMLLinkElement', {
   tagName: 'LINK',
   proto: {
+    get _accept() {
```
domenic: Instead of using a underscore-prefixed getter, use a new internal symbol. Add it to https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/helpers/internal-constants.js and then do e.g. `[accept]: "text/css,*/*,q=0.1"` after requiring it.

lib/jsdom/level2/html.js
```diff
@@ -1886,14 +1927,15 @@ function loadFrame (frame) {
 
   var src = frame.src.trim() === '' ? 'about:blank' : frame.src;
   var parentDoc = frame._ownerDocument;
-
+
```
domenic: Nit: unnecessary whitespace insertion

lib/jsdom/level2/html.js
```diff
@@ -1886,14 +1927,15 @@ function loadFrame (frame) {
 
   var src = frame.src.trim() === '' ? 'about:blank' : frame.src;
   var parentDoc = frame._ownerDocument;
-
+  
   // If the URL can't be resolved or the src attribute is missing / blank,
   // then url should be set to the string "about:blank".
   // (http://www.whatwg.org/specs/web-apps/current-work/#the-iframe-element)
   var url = resourceLoader.resolveResourceUrl(parentDoc, src);
+  var urlObj = url.match(/(.+?):(.*)/);
```
domenic: Use utils's URL constructor and check the resulting protocol, instead of using regexes.

lib/jsdom/living/generated/events/EventTarget-impl.js
```diff
@@ -190,12 +190,8 @@ function getListeners(target, type, capturing) {
 
       // Will be falsy for windows that have closed
       if (document) {
-        var implementation = document.implementation;
```
domenic: Why make this change? It seems bad to break the API in this way...

test/misc/cancel-requests.js
```diff
@@ -0,0 +1,107 @@
+"use strict";
+const env = require("../..").env;
+const http = require("http");
+
+var routes = {
```
domenic: throughout the file, s/var/const where possible, let where not

test/misc/cancel-requests.js
```diff
@@ -0,0 +1,107 @@
...
+  "/html": "<!DOCTYPE html><html>" +
+    "<head><script>window.test = true;</script><script src=\"/js\"></script></head>" +
+    "<body></body>" +
+    "</html>",
+  "/js": "window.testJs = true;",
+  "/xhr": "test"
+};
+
+exports["aborting env request should stop window creation"] = function (t) {
+
```
domenic: no extra line to start functions

test/misc/cancel-requests.js
```diff
@@ -0,0 +1,107 @@
...
+      res.end(routes[req.url]);
+    }, 200);
+  })
+  .listen(33334, function () {
+    env({
+      url: "http://127.0.0.1:" + 33334 + "/html",
+      created: function (err, window) {
+        process.nextTick(function () {
+
+          var requestManager = window.document._requestManager;
```
domenic: Tests should not test internal APIs. You need to find a different way of testing this.

test/misc/keep-alive-connections.js
```diff
@@ -0,0 +1,88 @@
...
+
+var routes = {
+  "/html": "<!DOCTYPE html><html><head><script src=\"/js\"></script></head><body></body></html>",
+  "/js": "var xhr = new window.XMLHttpRequest();" +
+          "xhr.open(\"GET\", \"/xhr\", true);" +
+          "xhr.send();",
+  "/xhr": "test"
+};
+
+exports["only one connection should be openned on sequenced calls"] = function (t) {
```
domenic: s/openned/opened

test/misc/keep-alive-connections.js
```diff
@@ -0,0 +1,88 @@
+"use strict";
+const env = require("../..").env;
+const http = require("http");
+
+var routes = {
+  "/html": "<!DOCTYPE html><html><head><script src=\"/js\"></script></head><body></body></html>",
+  "/js": "var xhr = new window.XMLHttpRequest();" +
```
domenic: Use template strings instead of concatenation

test/misc/keep-alive-connections.js
```diff
@@ -0,0 +1,88 @@
...
+    res.writeHead(200, { "Content-Length": routes[req.url].length });
+    res.end(routes[req.url]);
+  })
+  .on("connection", function () {
+    t.ok(true);
+    t.done();
+  })
+  .listen(33336, function () {
+    env({
+      url: "http://127.0.0.1:" + 33336 + "/html",
```
domenic: Just include it in the string, no concatenation

test/misc/keep-alive-connections.js
```diff
@@ -0,0 +1,88 @@
...
+      pool: false,
+      created: function () {},
+      features: {
+        FetchExternalResources: ["script"],
+        ProcessExternalResources: ["script"]
+      }
+    });
+  });
+};
+
```
domenic: These are some impressive and thorough tests of the new HTTP functionality. Nicely done.

test/w3c/config.jsdom.json
```diff
@@ -0,0 +1,26 @@
+{"host": "web-platform.test",
```
domenic: Can you format this using `JSON.stringify(JSON.parse(contents), undefined, 2)`?

test/w3c/index.js
```diff
@@ -59,3 +59,138 @@ w3cTest("html/infrastructure/urls/terminology-0/document-base-url.html");
 w3cTest("html/syntax/serializing-html-fragments/outerHTML.html");
 w3cTest("html/webappapis/atob/base64.html");
 
+
```
domenic: this is AMAZING. You are the best person ever. Thank you so much for getting all these passing!!!!

test/w3c/w3ctest.js
```diff
@@ -1,17 +1,15 @@
 "use strict";
 
-const http = require("http");
-const path = require("path");
-const request = require("request");
-const st = require("st");
 const jsdom = require("../..");
-const URL = require("../../lib/jsdom/utils").URL;
+
+var globalPool = {maxSockets: 6};
```
domenic: const everywhere.

Spaces after { and before }.

test/w3c/w3ctest.js
```diff
@@ -56,31 +55,76 @@ function createJsdom(urlPrefix, testPath, t) {
...
+
 module.exports = function (exports) {
-  const staticFileServer = st({ path: path.resolve(__dirname, "tests"), url: "/", passthrough: true });
-  const server = http.createServer(function (req, res) {
-    staticFileServer(req, res, function () {
-      fallbackToHostedVersion(req, res);
+  var server = new EventEmitter();
+  server.started = false;
+
+  dns.lookup("web-platform.test", function (err) {
```
domenic: Could we use a custom jsdom resource loader instead of requiring people to modify their hosts file?

domenic:
FYI I have added ESLint rules that enforce a lot of the style issues I was talking about. Hopefully that will make it easier. But, they do not yet enforce using classes instead of Object.defineProperty and functions.

nicolashenry:
I am currently trying to change everything about your remarks, it will take me some time.

README.md
nicolashenry: I have changed my mind about that. I added an example for more clarity.

lib/jsdom.js
nicolashenry: I made a function to this pupose.

lib/jsdom/browser/Window.js
nicolashenry: Done but I still had to do some weird things like this because the class need to have the current window/document :
```
this.XMLHttpRequest = this.createXMLHttpRequest(this);
```

lib/jsdom/browser/Window.js
nicolashenry: Ok but this would be break Blob and FileReader because instanceof and new instances use the classes from the global context.
Maybe the Window dependencies should be bootstrapped from the window context execution instead of normal require dependencies.

lib/jsdom/browser/file/blob.js
nicolashenry: Because it's simpler to use than Uint8Array.

lib/jsdom/browser/xhr/xmlhttprequest.js
nicolashenry: I replaced it by a spawnSync call but this is really slower than deasync and the http agents cannot be shared between two differents nodejs processes.

lib/jsdom/level2/html.js
nicolashenry: This is used by XMLHttpRequest and resource-loader, I could put them in the document directly with symbols but putting them inside the request manager seems more logical to me.

lib/jsdom/browser/Window.js
domenic: Yeah, this is totally understandable, for classes for which the spec gives working constructors. We can't avoid creating a new constructor per-window/document for those. Seems like a good pattern.

lib/jsdom/browser/Window.js
domenic: Hmm, that is surprising. But, you can use ArrayBuffer.isView instead of instanceof to avoid the issue I think.

If it remains a problem then it's OK to add these to get the tests to pass but add a TODO saying something like "// TODO: necessary for Blob and FileReader due to different-globals weirdness; investigate how to avoid this."

lib/jsdom/level2/html.js
domenic: Hmm. I think it is weird to store them as member variables of the class if they don't actually relate to the class. So symbols on the document (or window) seems better.

lib/jsdom/browser/xhr/xmlhttprequest.js
domenic: Thank you! I think it's worth it. Sync XHR is an evil feature anyway and it's OK if it's slow.

lib/jsdom/browser/file/blob.js
domenic: OK. But Buffer is not part of the web, so for example it should not be supported as an argument to the Blob constructor. And if you don't use too many special Buffer features then I'd prefer Uint8Array, but I'll leave that decision up to you.

domenic:
Did a quick skim while at work and things are looking great; thanks so much for taking the time to update. I'll do a more in-depth review after work, and in the meantime I'm happy to answer any questions. You can also bug us in the #jsdom IRC channel on freenode.

lib/jsdom/living/generated/events/EventTarget-impl.js
nicolashenry: I removed it because it broke some existing tests because XMLHttpRequest and FileReader does not really have "inline listener". I put it back but by trying to test if the target is a Node or not.

test/w3c/w3ctest.js
nicolashenry: No because w3c/web-platform-tests server needs them and does not just serve static files, there is specific paths for testing cross domains, headers, uploads, redirects, etc.

lib/jsdom/browser/file/blob.js
nicolashenry: I used the argument for the slice method and I forgot to change it. I think I could use Uint8Array but I don't have so much time in my life :), you still can change it later anyway.

Joris-van-der-Wel:
Just a head ups, last time I checked "leaking" `arguments` causes a lot of performance issues in v8 (node/chrome). For example `function() { someFunc(arguments) }`. 

This is why you often see something like this in jsdom:
```javascript
function doSomething() {
  const args = new Array(arguments.length);
  for (let i = 0; i < arguments.length; ++i) args[i] = arguments[i];
  somethingElse(args.slice(1));
}
```

I have even seen some projects (bluebird) that use a precompiler for this purpose.

Note that, `func.apply(something, arguments)` has been special cased in v8, and does not cause performance issues. I could imagine that `Array.from` is also special cased (but it probably is not). I'll see if I can test this.

lib/jsdom/browser/Window.js
nicolashenry: Note that I have not added some tests because of "instanceof Function" which does not work with classes/functions created outside the Window context. (Blob instanceof Function, Window instanceof Function, etc)

Sebmaster:
.apply is the only special-cased function for `arguments`. Can't even alias it and access it via the alias....

nicolashenry:
`arguments` is still useful to get optional parameters :
```javascript
"use strict";

class Blob1 { constructor(arg1, arg2) {  } }
class Blob2 { constructor() { const arg1 = arguments[0]; const arg2 = arguments[1]; } }

console.log(Blob1.length) // => 2
console.log(Blob2.length) // => 0
```
For example, see http://w3c-test.org/FileAPI/blob/Blob-constructor.html :
```javascript
  assert_equals(Blob.length, 0, "Blob.length should be 0.");
```

Sebmaster:
Yeah of course, it's just important to not access/provide it anywhere without either an index, or `.length`.

lib/jsdom/browser/Window.js
domenic: Hmm OK. Could you add those tests but commented out, with a brief comment explaining the problem? It sounds like I should spend some time investigating how to fix this. Maybe #1207 might help.

lib/jsdom/browser/resource-loader.js
```diff
@@ -40,14 +41,30 @@ function createResourceLoadHandler(element, resourceUrl, document, loadCallback)
   };
 }
 
-function readFile(url, callback) {
-  const filePath = url
-    .replace(/^file:\/\//, "")
-    .replace(/^\/([a-z]):\//i, "$1:/")
-    .replace(/%20/g, " ");
+exports.readFile = function (filePath, callback) {
+  const readableStream = fs.createReadStream(filePath);
```
domenic: Nit: can remove the next line in favor of `fs.createReadStream(filePath, { encoding: "utf8" })`

lib/jsdom/browser/resource-loader.js
```diff
@@ -59,12 +76,34 @@ function wrapCookieJarForRequest(cookieJar) {
...
 }
 
-function fetch(urlObj, cookieJar, referrer, callback) {
-  if (urlObj.hostname) {
-    exports.download(urlObj, null, cookieJar, referrer, callback);
+function fetch(element, urlObj, cookieJar, referrer, pool, agentOptions, userAgent, callback) {
+  let req = null;
+  if (urlObj.protocol === "data:") {
+    process.nextTick(() => {
+      const pathname = urlObj.href.match(/^data:(.+?)(?:;(base64))?,(.*)$/);
```
domenic: pathname is not the right name for this variable. Not sure what the best name is though.

lib/jsdom/level2/events.js
```diff
@@ -32,7 +40,7 @@ core.Event = function(eventType, eventInit) {
     this._target = null;
     this._currentTarget = null;
     this._eventPhase = 0;
-    this._timeStamp = null;
+    this._timeStamp = (new Date()).getTime();
```
domenic: Use `Date.now()` instead of this pattern.

lib/jsdom/level2/html.js
```diff
@@ -958,6 +995,14 @@ define('HTMLInputElement', {
         this._value = String(val);
       }
     },
+    get files() {
+      if(this.type === "file") {
```
domenic: Space after `if`.

lib/jsdom/living/blob.js
```diff
@@ -0,0 +1,76 @@
...
+class Blob {
+  constructor() {
+    if (!(this instanceof Blob)) {
+      throw new TypeError("DOM object constructor cannot be called as a function.");
+    }
+    const parts = arguments[0];
+    const properties = arguments[1];
+    if (arguments.length > 0) {
+      if (!parts || typeof parts !== "object" || parts instanceof Date || parts instanceof RegExp) {
+        throw new TypeError("blobParts should be nor a non-objects nor Dates nor RegExps nor null");
```
domenic: Nit: a better error message would be "Blob parts must be objects that are not Dates or RegExps".

lib/jsdom/living/file-reader.js
```diff
@@ -0,0 +1,159 @@
...
+        this.readyState = FileReader.EMPTY;
+        this.result = null;
+        this.onloadstart = null;
+        this.onprogress = null;
+        this.onload = null;
+        this.onabort = null;
+        this.onerror = null;
+        this.onloadend = null;
+      }
+      get EMPTY() {
```
domenic: Constants should not be getters; use addConstants to add them (even for classes).

lib/jsdom/living/file-reader.js
```diff
@@ -0,0 +1,159 @@
...
+function FileReaderEventTarget() {
+  if (!(this instanceof FileReaderEventTarget)) {
+    throw new TypeError("DOM object constructor cannot be called as a function.");
+  }
+  EventTarget.setup(this);
+}
+
+FileReaderEventTarget.prototype = Object.create(EventTarget.interface.prototype);
+
+module.exports = function (core) {
```
domenic: What do you think of a structure where instead of adding `createFileReader` to core and then deleting it from window, you return it? So the usage code would be

```js
const createFileReader = require("../living/file-reader");

...

this.FileReader = createFileReader(this);
```

(and living/index.js would not use file-reader.js)

What you have now is OK too thought if you think that's better.

lib/jsdom/living/form-data.js
```diff
@@ -0,0 +1,79 @@
+"use strict";
+
+const formDataSymbols = require("./form-data-symbols");
+
+module.exports = function (core) {
+  const Blob = core.Blob;
+  const File = core.File;
+
+  class FormData {
```
domenic: Wow, you implemented all of it!!

I think it would be a lot more efficient if you used a Map instead of an array? (I might be missing something.)

You've done enough work already that if you just want to leave a "TODO" comment about Map vs. array that's fine too :)

lib/jsdom/living/progress-event.js
```diff
@@ -0,0 +1,26 @@
+"use strict";
+
+module.exports = function (core) {
+  class ProgressEvent extends core.Event {
+    constructor(eventType, eventInit) {
+      super(eventType);
+      if (eventInit) {
+        this.lengthComputable = eventInit.lengthComputable || 0;
```
domenic: Do `Number(eventInit.lengthComputable) || 0` here and below to get correct type coercion.

lib/jsdom/living/xmlhttprequest.js
```diff
@@ -0,0 +1,722 @@
+"use strict";
+
+const HTTP_STATUS_CODES = require("http").STATUS_CODES;
+const spawnSync = require("child_process").spawnSync;
+
+const jsdom = require("../../jsdom");
```
domenic: Is there any way to avoid this circular require?

lib/jsdom/living/xmlhttprequest.js
```diff
@@ -0,0 +1,722 @@
...
+          responseCache: null,
+          responseTextCache: null,
+          responseXMLCache: null,
+          readyState: XMLHttpRequest.UNSENT,
+          status: 0,
+          statusText: ""
+        };
+        this.onreadystatechange = null;
+      }
+      get UNSENT() {
```
domenic: Use addConstants for these

lib/jsdom/virtual-console.js
```diff
@@ -17,9 +17,10 @@ module.exports = class VirtualConsole extends EventEmitter {
 
     for (const method of Object.keys(anyConsole)) {
       if (typeof anyConsole[method] === "function") {
-        this.on(method, () => {
```
domenic: Why change this?

test/web-platform-tests/run-to-upstream-web-platform-test.js
```diff
@@ -0,0 +1,86 @@
+"use strict";
```
domenic: Could we use the python runner for to-upstream tests too?

domenic:
OK, all reviewed! Only a few minor nits and questions. I will pull this in after your next revision. Thank you so much, this is amazing work!!!

lib/jsdom/living/form-data.js
nicolashenry: That's was my first thought but the specification says that FormData is "an ordered list of entries" and the algorithms described match the list use. (the form data should be sent in this exact order too, even it's not currently the case in my XMLHttpRequest implementation).
https://xhr.spec.whatwg.org/#interface-formdata

lib/jsdom/living/xmlhttprequest.js
nicolashenry: I don't know how to create an XML document object differently, if you have a better idea... :)

lib/jsdom/virtual-console.js
nicolashenry: Because arguments[0] === anyConsole and arguments[1] === options when used in arrow function. This means that with arrow function, the console will always output itself and options, unless I miss something.

test/web-platform-tests/run-to-upstream-web-platform-test.js
nicolashenry: I think that the right way to do would be to fork the web-platform-tests repository, include the tests in the fork and change the web-platform-tests submodule url to this new repository.

lib/jsdom/living/form-data.js
domenic: Sure, but Maps in JS are ordered too.

lib/jsdom/living/xmlhttprequest.js
domenic: Would the `window.Document` constructor work? Or `window.document.implementation.createDocument`?

lib/jsdom/virtual-console.js
domenic: Oh, I see! Good catch.

test/web-platform-tests/run-to-upstream-web-platform-test.js
domenic: No, that sounds like too much work. Browsers also have non-upstreamed tests and don't do that. I guess I can investigate later.

lib/jsdom/living/form-data.js
nicolashenry: No, because Maps can only contain one value per key.
For example:
```javascript
var formData = new FormData();
formData.append("test1", "value1");
formData.append("test2", "value2");
formData.append("test1", "value3");
var request = new XMLHttpRequest();
request.open("POST", "/test");
request.send(formData);
```
Should send the payload below :
```
------WebKitFormBoundaryPTFAfBO65hSkn3gb
Content-Disposition: form-data; name="test1"

value1
------WebKitFormBoundaryPTFAfBO65hSkn3gb
Content-Disposition: form-data; name="test2"

value2
------WebKitFormBoundaryPTFAfBO65hSkn3gb
Content-Disposition: form-data; name="test1"

value3
------WebKitFormBoundaryPTFAfBO65hSkn3gb--
```
If I put the values in Map inside lists, the values will be grouped by key and that's not how FormData should work.

lib/jsdom/living/xmlhttprequest.js
nicolashenry: Done with `window.Document`.

domenic:
I finally managed to merge this! Thank you so much. Here are the commits:

- 2c8933905e9b81436843fe8d429918876a277e54 for your original work
- 3eae7e0650853133be918173bc2f5ec5b5cdca2e for some additional work on the test runner
- a3f608a3abc825af89ecac1b4abec216c3aaf34a to disable some tests which failed when I ran them on my Windows machine. They all seem to have to do with the error event. If you have a chance to look in to these, I would be very grateful. But I understand not everyone has Windows and so I can certainly do it myself.

Thank you so much for this truly amazing contribution to jsdom. I hope we can see more of you in the future!

PR #1247, opened 2015-09-27
[WIP] Re-do Attr using webidl2js

domenic:
Missing pieces:
- [TreatNullAs=EmptyString] is not implemented in webidl2js (but is implemented in webidl-conversions)
- Generated getters are missing semicolons
- Speculative assumption about how impl constructors will change.
- Lots of TODO Attr to find and clean up
- attributes.js needs lots of updates, to stop looking for no-longer-there underscored properties, and instead look at impls. Kind of not sure what the best way to handle that is. Lots of discussion below.
- Not sure about mismatch between public name and spec name, e.g. namespaceURI vs. namespace or ownerElement vs. element. Also underscored vs. non-underscored.

@Sebmaster, thoughts especially on the last couple issues? In more, rambly, detail:

Code outside of Attr-impl.js, like that in attributes.js or in the definition of Element or similar, wants to access things that the spec spells as "the attribute's element" or "the attribute's namespace" or "the attribute's value". Given my current setup, these translate to:

```js
attributeImpl(wrapper).ownerElement
attributeImpl(wrapper).namespaceURI
attributeImpl(wrapper)._value // (!)
```

whereas you might expect something like

```js
attributeImpl(wrapper).element
attributeImpl(wrapper).namespace
attributeImpl(wrapper).value
```

or even

```js
attributeImpl(wrapper)._element
attributeImpl(wrapper)._namespace
attributeImpl(wrapper)._value
```

I think I could make `_element`/`_namespace`/`_value` work, at the cost of creating useless forwarding getters in the impl class: `get ownerElement() { return this._element; }`.

Making `element`/`namespace`/`value` work seems impossible since `impl.value` must be a getter that performs the special getter logic.

Or, I could just live with the inconsistency of `ownerElement`/`namespaceURI`/`_value`, and say that those parts of the code are tightly coupled to knowing the implementation detail of `AttrImpl`. Seems bad...

We could even do something more elaborate, like exporting some utilities, so that you do something like

```js
const attr = require("./generated/attributes/Attr");

attr.setSlots(wrapper, { element, namespace, value });
attr.getSlots(wrapper); // returns { element, namespace, value } object
```

These could either be custom-built for Attr, or they could be part of a general pattern generated by webidl2js... maybe AttrImpl exports some map, like `slots: { element: "ownerElement", ... }`??

----

Seems like the best idea is either `_element`/`_namespace`/`_value` or the slots idea. The slots idea is nice since code rarely if ever has to worry about actual impl objects. But it adds a lot of complexity. 

Hmm, maybe we could combine them!!

New idea. The impl class uses `_element`/`_namespace`/`_value`, and has some forwarding getters. But, we add these two new exports to the generated `Attr.js`:

```js
exports.getSlot = (wrapper, name) => idlUtils.implForWrapper(wrapper, "Attr")["_" + name];
exports.setSlot = (wrapper, name, value) => idlUtils.implForWrapper(wrapper, "Attr")["_" + name] = value;
```

Then consumers use them like

```js
const attrGet = require("./generated/attributes/Attr").getSlot;
const attrSet = require("./generated/attributes/Attr").setSlot;

// later

    if (attrGet(attr, "namespace") === namespace && attrGet(attr, "localName") === localName) {
      return attr;
    }

// elsewhere

    const oldValue = attrGet(attr, "value");
    attrSet(attr, "value", value);
```

Hmm, maybe this is overkill, just to avoid thinking about wrappers vs. impls... Maybe just the `_element`/`_namespace`/`_value` naming convention is enough to decouple.


Sebmaster:
Okay, so: I don't like the slots idea at all, it seems to unnecessarily complicate things, however we might not need it unless I have misunderstood something:

If we take [namespace](https://dom.spec.whatwg.org/#dom-attr-namespaceuri) as an example, we have the `namespaceURI` wrapper getter, however in the spec it references a `namespace` name. This is especially sweet because you can easily differentiate between (algorithm-) local and (element-) state info like this in all specs I've read up until now.

So we implement a namespaceURI getter in impl which forwards to a `_namespace` property imo. That's how I've handled it up until now at least and I think that's pretty clear - it keeps a pretty strong spec equivalency (if we keep and enforce it as general convention).

So _namespace, _element and _value seemt to be the way to go with "useless" getters. I think this just keeps this the most clear.

domenic:
Sounds good. I guess I was just hoping to be able to use the nice trick you told me about where no forwarding getters were necessary. But it does have clarity benefits anyway.

Sebmaster:
Yeah, that only works if the spec doesn't use special getter behaviour and directly uses the same terminology.

domenic:
Giving this another try. However i've run into circular dependency issues which I am not sure how to solve.

- Attr-impl.js depends on attributes.js for changeAttribute.
- attributes.js depends on generated/Attr.js for attrGenerated.create()
- generated/Attr.js depends on Attr-impl.js to instantiate implementations

We could accept this circular dependency, however, webidl2js is barfing on it. At least, I think it is; https://github.com/jsdom/webidl2js/issues/11 makes it hard to tell. But when I comment out the Attr-impl.js -> attributes.js dependency it starts actually using the implementation; otherwise it uses the empty-object implementation.

In more general terms, the problem is:

- Attr-impl needs to use utility methods
- utility methods need to create new Attrs
- to create new Attrs you need the Attr-impl

I think I could work around this by partitioning the utility methods into ones that need to create new Attrs, and ones that don't. But that's not pretty. Not sure what's the best idea here...

Sebmaster:
Yeah, I'm aware that's kinda an issue, I thought

> I think I could work around this by partitioning the utility methods into ones that need to create new Attrs, and ones that don't. But that's not pretty.

was sufficient until now. Maybe it's not and we'll have to look for something else though...

domenic:
Meh, I'm giving it a try and it's not the worst thing ever. Probably OK for now. Just wondering if you had a better idea.

lib/jsdom/living/attributes.js
```diff
@@ -1,6 +1,17 @@
 "use strict";
 const DOMException = require("../web-idl/DOMException");
 const defineGetter = require("../utils").defineGetter;
+const idlUtils = require("./generated/utils");
+const attrGenerated = require("./generated/Attr");
+const changeAttributeImpl = require("./attributes/Attr-impl").changeAttributeImpl;
+
+function attrImpl(attrWrapper) {
+  return idlUtils.implForWrapper(attrWrapper, "Attr");
```
Sebmaster: 2nd arg in `implForWrapper` is gone in webidl2js v3.

lib/jsdom/living/attributes.js
```diff
@@ -1,6 +1,17 @@
...
+}
+
+function isAttr(obj) {
+  return Boolean(obj && obj.constructor && obj.constructor.name === "Attr"); // TODO branding
```
Sebmaster: `implForWrapper(obj) instanceof AttrImpl`?

domenic: Oh nice. Can you have Attr.js export a brand check for me in all generated classes?

domenic:
@Sebmaster this is ready for review :)

Sebmaster:
Ugh this is big, got through Attr and compared with spec, but I gotta do this in multiple steps...

Sebmaster:
Am through (I think), nothing else to add.

Issue #1157, opened 2015-07-02
Removed style elements do not get removed from document.styleSheets


stuartlangridge:
Demonstration:

    var jsdom = require("jsdom");
    var pageHTML = '<!doctype html><html><head>' +
        '<style>p { color: green; }</style><style>div { color: red; }</style>' +
        '</head><body></body></html>';
    var removeScript = "var style2 = document.getElementsByTagName('style')[1]; " +
        "style2.parentNode.removeChild(style2);";
    jsdom.env({
        html: pageHTML,
        src: removeScript,
        done: function(err, window) {
            console.log("Number of style elements in the page:", 
                window.document.getElementsByTagName("style").length);
            console.log("Length of document.styleSheets:",
                window.document.styleSheets.length);
        }
    });

This should print 1, 1 and instead prints 1, 2. 

This is similar to https://github.com/tmpvar/jsdom/issues/690 but that's about leaking memory; this is more concerned with the data being _wrong_ and so iterating over `document.styleSheets` gives incorrect answers.

Joris-van-der-Wel:
Looks like someone left a todo note for precisely this issue: [lib/jsdom/level2/style.js:80](https://github.com/tmpvar/jsdom/blob/6e52bcdc4805323ea2407887d5a35d484ee6d1f7/lib/jsdom/level2/style.js#L80)

PR #1251, opened 2015-10-02
1157 remove style elem fix

AVGP:
Handling the `DOMNodeRemovedFromDocument` for `HTMLStyleElement`s.

This is resolving #1157

### DIFF
--- a/lib/jsdom/level2/style.js
+++ b/lib/jsdom/level2/style.js
@@ -149,8 +149,6 @@ core.HTMLLinkElement._init = function() {
       fetchStylesheet.call(this, this.href, this.sheet);
     }
   });
-  this.addEventListener('DOMNodeRemovedFromDocument', function() {
-  });
 };
 
 assert.equal(undefined, core.HTMLStyleElement._init);
--- a/lib/jsdom/level2/style.js
+++ b/lib/jsdom/level2/style.js
@@ -174,14 +174,14 @@ core.HTMLStyleElement._init = function() {
     evaluateStylesheet.call(this, content, this.sheet, this._ownerDocument.URL);
   });
 
-  this.addEventListener('DOMNodeRemovedFromDocument', function() {
+  this.ownerDocument.addEventListener('_descendantRemoved', function(e) {
     var index = this.ownerDocument.styleSheets.indexOf(this.sheet);
     if(index > -1) {
-      this.ownerDocument.styleSheets = this.ownerDocument.styleSheets.splice(index, 1);
+      this.ownerDocument.styleSheets.splice(index, 1);
     }
   });
-};
 
+};
 
 // https://html.spec.whatwg.org/multipage/semantics.html#htmllinkelement:
 // HTMLLinkElement implements LinkStyle
--- a/lib/jsdom/level2/style.js
+++ b/lib/jsdom/level2/style.js
@@ -174,13 +174,6 @@ core.HTMLStyleElement._init = function() {
     evaluateStylesheet.call(this, content, this.sheet, this._ownerDocument.URL);
   });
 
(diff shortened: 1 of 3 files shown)


lib/jsdom/level2/style.js
```diff
@@ -173,6 +173,13 @@ core.HTMLStyleElement._init = function() {
 
     evaluateStylesheet.call(this, content, this.sheet, this._ownerDocument.URL);
   });
+
+  this.addEventListener('DOMNodeRemovedFromDocument', function() {
```
domenic: We're trying not to use mutation events going forward, since they are no longer spec compliant, and should be removed from jsdom. Instead of creating a new event listener for every stylesheet, maybe you could add a _descendantRemoved handler to Document?

lib/jsdom/level2/style.js
```diff
@@ -173,6 +173,13 @@ core.HTMLStyleElement._init = function() {
...
+    var index = this.ownerDocument.styleSheets.indexOf(this.sheet);
+    if(index > -1) {
+      this.ownerDocument.styleSheets = this.ownerDocument.styleSheets.splice(index, 1);
```
domenic: You should just do `this.ownerDocument.styleSheets.splice(index, 1)`, without the reassignment.

test/level2/style.js
```diff
@@ -19,6 +19,23 @@ exports.tests = {
...
+    var pageHTML = '<!doctype html><html><head>' +
+        '<style>p { color: green; }</style><style>div { color: red; }</style>' +
+        '</head><body></body></html>';
+    var removeScript = "var style2 = document.getElementsByTagName('style')[1]; " +
+        "style2.parentNode.removeChild(style2);";
+    jsdom.env({
+      html: pageHTML,
+      src: removeScript,
+      done: function(err, window) {
+        test.equal(1, window.document.getElementsByTagName("style").length);
```
domenic: Add a `test.ifError(err);` line.

lib/jsdom/level2/style.js
AVGP: Ah alright, I just took the handlers from the LinkElement. Good to know! So I'll also remove the (anyway empty) handlers from there, too... it's misleading to other newcomers.

lib/jsdom/level2/style.js
AVGP: Good catch, thanks!

lib/jsdom/level2/style.js
domenic: Awesome, thanks! Yeah, jsdom has a lot of legacy code :(

AVGP: Hm, honestly? I don't even know where to do that. There's nothing in jsdom that does this, it's all using the mutation events so far :(

domenic: So, if you edit level1/core.js's definition of the Document prototype, you can add a _descendantRemoved method, which receives (parent, child) as arguments. Just be sure to call the super-class `Node.prototype._descendantRemoved.apply(this, arguments)`.

AVGP:
@domenic I'm trying your suggestion but I'm getting something wrong. Any hints?

AVGP:
@domenic nevermind, I got it. Silly me :/ But could you check if I'm doing it right?

domenic:
This looks great to me. Thanks for sticking with us. I'll merge this as soon as I get a free day to do a new jsdom release... soon (TM)...

AVGP:
Cool, thanks!

I would have continued earlier but life happened ;)

Cheers!
Am 12.10.2015 8:00 nachm. schrieb "Domenic Denicola" <
notifications@github.com>:

domenic:
Merged as 5dc416a3aaec1440520ad78945e434c2ceca104b, and expect a 7.0.0 with this fix soon!

PR #1255, opened 2015-10-03
Implement loose cooking parsing

Sebmaster:
This enables correct execution of code like `document.cookie = 'foo'`.

Blocked on either https://github.com/request/request/pull/1811 or https://github.com/SalesforceEng/tough-cookie/pull/56.

### DIFF
--- a/lib/jsdom.js
+++ b/lib/jsdom.js
@@ -36,7 +36,7 @@ exports.getVirtualConsole = function (window) {
 };
 
 exports.createCookieJar = function () {
-  return new CookieJar();
+  return new CookieJar(null, { looseMode: true });
 };
 
 exports.nodeLocation = function (node) {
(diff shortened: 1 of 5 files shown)


domenic:
LGTM except for the commit message (s/cooking/cookie), so once one of the prereqs gets merged feel free to merge.

Sebmaster:
@domenic Added loose mode to the constructor too now. The problem right now is (without request merging the PR), if a user of jsdom provides a cookie jar without loose mode, we'll handle malformatted cookies in `document.cookie` differently, than if we get the cookie from a resource request, which kind of sucks. We could set `jar.enableLooseMode`, but I'm not sure if it's considered public API.

Sebmaster:
Modified to only set loose mode on jars we create ourselves. If a user passes us a jar without loose mode, we'll behave strict and refuse malformed cookies.

domenic:
LGTM!

Issue #1065, opened 2015-03-17
template.content is undefined


dchambers:
If you create a `<template/>` element than it should have a `content` property containing a `DocumentFragment` instance that contains the templates content. Based on #941 and #944 it sounds like this is supported by jsdom, yet if I write code like this:

```js
var jsdom = require("jsdom").jsdom;
var doc = jsdom("<html><body></body></html>");
var templateElem = doc.createElement('template');
templateElem.id = 'my-id';
templateElem.innerHTML = '<span></span>';
doc.body.appendChild(templateElem);
doc.importNode(doc.getElementById('my-id').content, true);
```

then I just get the following error on the final line of code:

```
/home/dominicc/git-dev/template-spike/node_modules/jsdom/lib/jsdom/level2/core.js:227
      newNode = importedNode.cloneNode(deep, function(a, b) {
                            ^
TypeError: Cannot read property 'cloneNode' of undefined
    at HTMLDocument.core.Document.importNode (/home/dominicc/git-dev/template-spike/node_modules/jsdom/lib/jsdom/level2/core.js:227:29)
    at Object.<anonymous> (/home/dominicc/git-dev/template-spike/src/Templates.js:9:5)
```

I'm using jsdom 3.1.2. Have I done something wrong?

domenic:
Oh, hmm, I guess we might not actually support that yet, like #941 implies. We have parsing support but never actually implemented HTMLTemplateElement. I think.

dchambers:
Fair enough. To get my tests running on IE (which also doesn't support the `content` attribute) I ended up using this [template polyfill](http://jsfiddle.net/brianblakely/h3EmY/). I imagine you'd need something similar within jsdom?

domenic:
Yeah, I guess it should work, but honestly it should be a pretty quick patch to get a working HTMLTemplateElement, one of us can probably knock it out soon.

dchambers:
Excellent news :+1:

rickychien:
+1

rickychien:
@domenic Unfortunately, I ran into same problem here. How can I polyfill with HTMLTemplateElement or is there someone working on this issue?

domenic:
@rickychien take a look at CONTRIBUTING.md to learn how to contribute features like HTMLTemplateElement to jsdom.

domenic:
Fixed by 86ef3543de49deb21d75e07179baf5a45a56c7c3!

dchambers:
[...]

PR #1256, opened 2015-10-07
Support HTMLTemplateElement.content for #1065

rickychien:
See also #1065 

@domenic here is my quick patch for supporting template.content, I'm not sure what's the best approach to hack that things. Please take a look!

### DIFF
--- a/lib/jsdom/level2/html.js
+++ b/lib/jsdom/level2/html.js
@@ -2217,7 +2217,21 @@ define('HTMLSourceElement', {
   ]
 });
 
-define('HTMLTemplateElement', { tagName: 'TEMPLATE' });
+define('HTMLTemplateElement', {
+  tagName: 'TEMPLATE',
+  proto: {
+    get content() {
+      if (!this._content) {
+        var content = this._ownerDocument.createDocumentFragment();
+        for (const child of domSymbolTree.childrenIterator(this)) {
+          content.appendChild(child);
+        }
+        this._content = content;
+      }
+      return this._content;
+    }
+  }
+});
 
 define('HTMLTrackElement', {
   tagName: 'TRACK',
--- a/test/level2/html.js
+++ b/test/level2/html.js
@@ -16670,6 +16670,13 @@ exports.tests = {
     test.done();
   },
 
+  /**
+   *
+   HTMLTemplateElement.content should contain elements which are nested inside
+   <template> tag
+   * @author Ricky Chien
+   * @see https://html.spec.whatwg.org/multipage/scripting.html#the-template-element
+   */
   HTMLTemplateElement01: function(test) {
     var doc = jsdom.jsdom('<template><p>TEST</p></template>');
     var template = doc.getElementsByTagName('template').item(0);
@@ -16681,6 +16688,12 @@ exports.tests = {
     test.done();
   },
 
+  /**
+   *
+   HTMLTemplateElement.content is a DOCUMENT_FRAGMENT_NODE (value = 11)
+   * @author Ricky Chien
+   * @see https://html.spec.whatwg.org/multipage/scripting.html#the-template-element
+   */
   HTMLTemplateElement02: function(test) {
     var doc = jsdom.jsdom();
     var template = doc.createElement('template');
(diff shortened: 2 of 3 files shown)


Sebmaster:
The [spec](https://html.spec.whatwg.org/multipage/scripting.html#template-contents) definitely requires a lot more details with template elements, especially in regards to cloning so modifying the result of content doesn't modify the template element itself.

domenic:
Thanks for working on this!

My biggest concern here is that I'm pretty sure template.content and template.firstChild should actually be different. The former should be a document fragment, whereas the latter should be the first node nested under the template tag in the source.

It'd be OK to have an incomplete implementation, but we should at least get that right.

For the tests, could you either find appropriate web platform tests, or create new ones in https://github.com/tmpvar/jsdom/tree/master/test/web-platform-tests/to-upstream ?

rickychien:
Hey @domenic,

Unfortunately, there are lots of tests cannot passed web platform tests, so I leaved some passable tests and updated the patch.

rickychien:
I was thinking the spec [1] is a little bit unclear to me and I spent lots of time doing trial and errors on real browser to figure out real behavior behind that. (Or maybe I missed something from spec?)

[1] https://html.spec.whatwg.org/multipage/scripting.html#template-contents

lib/jsdom/level2/html.js
```diff
@@ -2217,7 +2217,29 @@ define('HTMLSourceElement', {
...
+              content.appendChild(child);
+            }
+          }
+        };
+        createContent(content, this);
+        this._content = content;
+      }
+      return this._content;
+    },
+    set content(val) {}
```
domenic: No empty setter.

test/level2/html.js
```diff
@@ -16670,6 +16670,33 @@ exports.tests = {
     test.done();
   },
 
+  HTMLTemplateElement01: function(test) {
+    var doc = load('template');
```
domenic: Instead of using `load()`, can you just use `jsdom.jsdom()`?

domenic:
This is probably ready to pull when I finally get some free time for a jsdom release. But, I'm curious:

- What did you find unclear about the spec?
- Were you not able to get any of the other tests in https://github.com/w3c/web-platform-tests/tree/master/html/semantics/scripting-1/the-template-element passing?

rickychien:
There are only two additions-to-the-steps-to-clone-a-node tests are able to pass after I checked all tests of the-template-element. Some tests got stuck without any error messages so I guess it probably ran into problem during parsing? 

We need to do more works to support correct behavior for template like as only append template's childNodes into template.content in HTML parsing stage.
(I found that it will create a documentFragment object as its first child after HTML parsing)

After that, we can do more improvement :)

rickychien:
About unclear part, for instance:

1. Let doc be the template element's node document's appropriate template contents owner document.
what's doc mean?
2. Create a DocumentFragment object whose node document is doc.
Could you give me an code example how to do this? It may help me understand it quickly.

thank!

rickychien:
Descriptions between [1] and [2] are a little bit different. Keyword node document appears in [1] but  ownerDocument appears in [2]. Which one is better?

[1] https://html.spec.whatwg.org/multipage/scripting.html#template-contents
[2] http://www.w3.org/TR/html5/scripting-1.html#the-template-element

domenic:
Please don't use the outdated W3C fork when implementing jsdom features. Use http://html.spec.whatwg.org/multipage/ instead.

rickychien:
ok I got it, but description of appropriate template contents owner document is same as outdated one...lol

I've updated my patch to move ```jscs:disable maximumLineLength``` in index.js.

domenic:
OK. To create a new Document without a browsing context, you can do `new core.HTMLDocument({ parsingMode: "html" })`.

But yeah, this is still good to pull in when I have the time to do a release, just clarifying :)

domenic:
Merged as 86ef3543de49deb21d75e07179baf5a45a56c7c3; look for a 7.0.0 soon with this in it! Thanks so much for contributing; this is a great feature to have.

rickychien:
nice!!

PR #1257, opened 2015-10-07
Make webidl conversions work with multiple categories

Sebmaster:
Since all generated js files are so interdependent on each other, splitting them into folders according to their category screws up all relative paths between each other.

Instead, we'll put all generated js files into a single folder and split .idl and -impl files according to their category, which works fine.

**Blocked on webidl2js 2.0.0.**


scripts/webidl/convert.js
```diff
@@ -46,4 +46,4 @@ function onlyIDL(filePath) {
   return path.extname(filePath) === ".idl";
 }
 
-doConversion("lib/jsdom/living/generated/events").done();
+doConversion("lib/jsdom/living/events").done();
```
domenic: Should this maybe just become jsdom/living?

domenic:
LGTM assuming the build passes and webidl2js 2.0.0 is released and stuff.

scripts/webidl/convert.js
Sebmaster: But then we'd have all files in a single location, that'll get confusing really quickly I'd imagine.

Or do you mean the convert script should group according to folders by itself?

domenic: I thought it searched recursively?

Sebmaster: It does, but in concats into a single "source file" which is then used for generation and still needs to be able to find the right implementation directory, which is then not possible anymore since its regarded as a single entity.

domenic: Ah OK, I see. Well eventually I want to avoid us having to update this list explicitly but for now this is fine.

Sebmaster:
Passing :tada:

Sebmaster:
Let's just merge this together with #1258.

PR #1242, opened 2015-10-08
`FetchExternalResources: ["script"]` causes jsdom to pass different window instances to its callbacks

Sebmaster:
### Versions

jsdom: 6.5.1
node: 4.1.0

### What steps will reproduce the problem?

1. Save this code to a file:

    ```
var jsdom = require("jsdom");
var assert = require("chai").assert;

var instances = {};
jsdom.env({
    html: "",
    features: {
        FetchExternalResources: ["script"],
    },
    created: function (error, w) {
        assert.isNull(error);
        assert.isTrue(!!w);
        instances["created"] = w;
    },
    onload: function (w) {
        assert.isTrue(!!w);
        instances["onload"] = w;
    },
    done: function (error, w) {
        assert.isNull(error);
        assert.isTrue(!!w);
        instances["done"] = w;
        var keys = Object.keys(instances);
        for (var i = 0, key1; (key1 = keys[i]); ++i) {
            for (var j = i + 1, key2; (key2 = keys[j]); ++j) {
                console.log(key1 + " === " + key2 + ": " +
                            (instances[key1] === instances[key2]));
            }
        }
    }
});
```

2. Install jsdom and Chai.

3. Run the code.

### What is the expected output?

```
created === onload: true
created === done: true
onload === done: true
```

### What do you see instead?

```
created === onload: false
created === done: false
onload === done: true
```

### Observations

1. The expected output is obtained if this is removed from the configuration:

    ```
    features: {
        FetchExternalResources: ["script"],
    },
```

2. `html:""` is not a factor. I can reproduce the same problem with complex pages.

3. This issue is a problem if the window instance passed to `created` is saved for future use, as this appears to be a `Window` instance which is not fully formed. (For instance, `Array` is not defined on it.) The solution is to save the window instance passed to `onload` or `done` instead.

### DIFF
--- a/lib/jsdom.js
+++ b/lib/jsdom.js
@@ -96,11 +96,11 @@ exports.jsdom = function (html, options) {
     virtualConsole: options.virtualConsole
   });
 
-  if (options.created) {
-    options.created(null, window);
-  }
-
   documentFeatures.applyDocumentFeatures(window.document, options.features);
+
+  if (options.created) {
+    options.created(null, window.document.defaultView);
+  }
 
   if (options.parsingMode === "html") {
     if (html === undefined || html === "") {
--- a/test/jsdom/env.js
+++ b/test/jsdom/env.js
@@ -560,3 +560,60 @@ exports["done should be called only once, after all src scripts have executed"] 
     }
   });
 };
+
+exports["window instances should be initialized when provided to callbacks"] = t => {
+  t.expect(5);
+
+  env({
+    html: "<div></div>",
+    features: {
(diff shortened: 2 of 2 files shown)


Sebmaster:
Fixed, attached PR.

domenic:
Test plz

Sebmaster:
Added.

domenic:
LGTM, thanks!

domenic:
Merged as b667ed3c89545f60035d3e13b71eda2d9cf1bfdd; expect a 7.0.0 release with this later today.

PR #1258, opened 2015-10-10
Refactor events to use webidl2js

Sebmaster:
The implementation itself is also very closely aligned with the spec now.

Blocked on #1257.


Sebmaster:
Not done with re-implementing all the event subclasses yet - UI, Mouse and Mutation are still missing.

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -0,0 +1,237 @@
...
+        (event.eventPhase === Event.CAPTURING_PHASE && !listener.capture) ||
+        (event.eventPhase === Event.BUBBLING_PHASE && listener.capture)) {
+      continue;
+    }
+
+    try {
+      listener.callback.call(eventImpl.currentTarget, event);
+    } catch (e) {
+      let window = null;
+      if (target._document) {
```
Sebmaster: This is a bit weird - I had to rewrite it and check this a bit more explicitly, since there might be a Window which has its document removed, which logged a whole lot of errors to jsdomErrors.

I'm not sure why those errors weren't there before though.

domenic: I've seen a variety of problems with this area of the code. The XHR PR also made some fixes there. So yeah, understandable. Maybe checking for _document isn't the best way to identify a window though.

Sebmaster: We could just do instanceof checks? Or .constructor == Window?

Branding would be best (which is btw not solved yet in webidl2js 3.0) but we don't have that yet in Window. Could just patch it in though...

domenic: I figure just pick a property that's unique-ish to window. I guess instanceof/.constructor is also fine but would require a reference to the Window class which might be hard to get at this point in the file.

Sebmaster: Since it was _document before too, I'd just leave as-is until someone complains or we get window to webidl2js.

Sebmaster:
Implemented all events which are necessary for [createEvent](https://dom.spec.whatwg.org/#dom-document-createevent) and fixed the remaining test failures.

Gonna work on the new webidl2js major next so we can get this out the door soon.

lib/jsdom/living/events/CustomEvent-impl.js
```diff
@@ -0,0 +1,26 @@
+"use strict";
+
+const EventImpl = require("./Event-impl").implementation;
+const CustomEventInit = require("../generated/CustomEventInit");
+
+class CustomEventImpl extends EventImpl {
+  constructor(args, privateData) {
+    const type = args[0]; // TODO: Replace with destructuring
+    const eventInitDict = args[1] || CustomEventInit.convert(undefined);
```
domenic: We really need to fix this before this is merge-ready. It should not be necessary to define this constructor.

Sebmaster: Yeah I know, it's still on my todo.

domenic: \o/

Sebmaster: Fixed

lib/jsdom/living/events/CustomEvent-impl.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
+
+const EventImpl = require("./Event-impl").implementation;
+
+class CustomEventImpl extends EventImpl {
+  constructor(args, privateData) {
```
domenic: Can't you just omit this entire constructor now?

lib/jsdom/living/events/Event.idl
```diff
@@ -0,0 +1,35 @@
...
+
+  void initEvent(DOMString type, boolean bubbles, boolean cancelable);
+};
+
+dictionary EventInit {
+  boolean bubbles = false;
+  boolean cancelable = false;
+};
+
+dictionary CustomEventInit : EventInit {
```
domenic: Duplicated here and in CustomEvent.idl?

lib/jsdom/living/events/TouchEvent.idl
```diff
@@ -0,0 +1,9 @@
+interface TouchEvent : UIEvent {
+      readonly    attribute TouchList touches;
+      readonly    attribute TouchList targetTouches;
+      readonly    attribute TouchList changedTouches;
+      readonly    attribute boolean   altKey;
+      readonly    attribute boolean   metaKey;
+      readonly    attribute boolean   ctrlKey;
+      readonly    attribute boolean   shiftKey;
+  };
```
domenic: Missing end of file newline :P

Sebmaster:
Done :+1:

lib/jsdom/living/events/TouchEvent-impl.js
```diff
@@ -0,0 +1,11 @@
+"use strict";
+
+const UIEventImpl = require("./UIEvent-impl").implementation;
+
+class TouchEventImpl extends UIEventImpl {
```
domenic: A bit sad that all of this could be reduced to `exports.implementation = class extends UIEventImpl {}` but I guess I am OK with the verboseness for the sake of uniformity...

domenic:
This LGTM after the webidl2js release happens!

Sebmaster:
Done :tada:

domenic:
Wooohoooooooo!

PR #1265, opened 2015-10-19
Reimplement HTMLHyperlinkElementUtils, Location, and History

domenic:
Mostly HTMLHyperlinkElementUtils for now, just so you can see how it works. Depends on https://github.com/jsdom/whatwg-url/pull/36 and still lots to do before it's mergable, including Location and getting rid of a few other uses of the old API (e.g. in History).

This should solve the performance issues since we no longer react to every `href` attribute change.


rgbkrk:
Would these also make `URL.createObjectURL()` available in JSDom?

domenic:
No

domenic:
@Sebmaster this is ready for review, although it depends on https://github.com/jsdom/whatwg-url/pull/36

lib/jsdom/browser/Window.js
```diff
@@ -384,6 +392,7 @@ function Window(options) {
 
   ///// INITIALIZATION
 
+  // TODO: delay the load event until everything loads.
```
Sebmaster: Does this still apply?

lib/jsdom/living/index.js
```diff
@@ -27,6 +28,8 @@ exports.MutationEvent = require("./generated/MutationEvent").interface;
 exports.ProgressEvent = require("./generated/ProgressEvent").interface;
 exports.EventTarget = require("./generated/EventTarget").interface;
 
+exports.Location = require("./generated/Location").interface;
```
Sebmaster: Can we add `window.History` too?

lib/jsdom/browser/Window.js
domenic: Yeah I think I misread this code; can remove.

lib/jsdom/living/index.js
domenic: Good call

domenic:
@Sebmaster rebased with review comments applied.

Sebmaster:
Still missing the spec comparison, will get that done tonight too.

lib/jsdom/living/window/Location-impl.js
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  _locationObjectSetterNavigate(url) {
+    // Extra steps here to determine replacement flag are not implemented, and are not applicable to our rudimentary
+    // "navigation" implementation.
+
+    return this._locationObjectNavigate(url);
+  }
+
+  _locationObjectNavigate(url) {
```
Sebmaster: Doesn't accept the options object with replacement?

Sebmaster:
Okay, went over the anchor and location code (except getters/setters, I assume those were mainly just copying anyways). Had a look over history and didn't notice anything, but didn't compare with spec as I assume that wouldn't make a lot of sense anyways with lots of stuff missing.

Sebmaster:
I might be missing something, but it looks to me like the url isn't being reparsed after href has been set (if the URL is non-special). I may have to try to repro to get why it doesn't work.

Sebmaster:
Here we go:
```js
"use strict";

const jsdom = require(".");

const document = jsdom.jsdom("<a href='mycustomprotocol:abc'>Hello</a>");

const anchor = document.getElementsByTagName("a")[0];

console.log(anchor.protocol);
anchor.href = 'otherprot:test.js';

console.log(anchor.protocol);
```

Doesn't do what you'd expect. Works with the URL constructor however.

lib/jsdom/living/window/Location-impl.js
domenic: It does, but see the comment. We don't support replacement.

domenic: I will add a commented out argument and keep the call site. What do you think?

domenic:
Nice catch; will add a test. I think it is covered by the clause

> When elements implementing the HTMLHyperlinkElementUtils mixin are created, and whenever those elements have their href content attribute set, changed, or removed, the user agent must set the url.

Which we don't implement, since the following note says 

> This is only observable for blob: URLs as parsing them involves the structured clone algorithm.

I guess that note is not completely true. We still don't want to implement it though due to the perf consequences during parsing.

I will just add an explicit call in the setter.

lib/jsdom/living/window/Location-impl.js
Sebmaster: :+1:

domenic:
OK, fixed in a slightly different way since it is indeed a spec bug, in a separate commit on top of the others.

PR #1270, opened 2015-10-22
issue with babel compiled source and jsdom 7.0.1

Sebmaster:
I'm trying to upgrade jsdom to 7.0.1 and I bisected a big issue to https://github.com/tmpvar/jsdom/commit/2c8933905e9b81436843fe8d429918876a277e54 although I'm not exactly sure *why* things are going wrong given that this change was so large.

Here is a repro for open source:
```
git clone https://github.com/facebook/relay.git
cd scripts/babel-relay-plugin
npm install
# cd into `node_modules/jest-cli`, edit the package.json to jsdom 7.0.1 and run npm update
npm test
```

test output: https://gist.github.com/cpojer/b15a99a7939cfe79c2f4

with jsdom 6.5 we didn't have any of these issues. At FB I have about 70 failing tests because of this, unfortunately.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -177,9 +177,6 @@ function Window(options) {
 
   // TODO: necessary for Blob and FileReader due to different-globals weirdness; investigate how to avoid this.
   this.ArrayBuffer = ArrayBuffer;
-  this.Date = Date;
-  this.RegExp = RegExp;
-
   this.Int8Array = Int8Array;
   this.Uint8Array = Uint8Array;
   this.Uint8ClampedArray = Uint8ClampedArray;
--- a/test/misc/cancel-requests.js
+++ b/test/misc/cancel-requests.js
@@ -201,5 +201,3 @@ exports["stopping window should close xhr but abort event should be triggered"] 
     });
   });
 };
-
-


Sebmaster:
Just as additional info: The real problem is, that RegExps can be created by either constructing them, or just by using their literal. When overwriting the constructor, the literal will create a different object than the constructor which causes a lot of havoc.

Attached a PR to fix this for now.

domenic:
This fails the build; fix that and it LGTM.

> We can't remove all *Array and ArrayBuffer shimming, since createContext doesn't insert those.

This isn't actually true. I think the reason they were added is that sometimes jsdom code creates outside-sandbox arrays/arraybuffers/etc., and then code inside the sandbox does `instanceof ArrayBuffer` and it fails. This was causing some web-platform-tests failures.

The correct longer-term fix, I believe, is to try to always create those instances inside the sandbox.

Sebmaster:
> The correct longer-term fix, I believe, is to try to always create those instances inside the sandbox.

Right, I tried that first but did a thing wrong:

```js
vm = require("vm");
ctx = {};
vm.createContext(ctx);
ctx.ArrayBuffer // undefined
vm.runInContext("this", ctx).ArrayBuffer // ArrayBuffer
```
I only checked `ctx.ArrayBuffer` and was confused when that was undefined. We'll probably have to major the strict distinction change. :confused:

domenic:
Merged as 3f1a82baacb6b56920a34b8380bc40a4230509a1 and 2b4a5fc7b4c61d353c7853e5ff95142a4a6ad81c. 7.0.2 coming out shortly!

cpojer:
thank you!

PR #1252, opened 2015-10-22
Travis browser builds are broken, only for PRs from forks I think

Sebmaster:
This used to work :-/. Might need to disable them.


Sebmaster:
Here we go, attached a commit which makes it work as expected for both PR + push scenarios.

PR #1275, opened 2015-10-27
Create composer.json

paladox:
Adding composer.json so it can be added to packagist.

This repo is required by some repos using packagist.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,3 @@
+{
+	"name": "node-jsdom"
+}


domenic:
Not interested. Use npm.

paladox:
It won't. I searched for the package on packagist but won't come up with it. I try adding the repo to it and it says Carnot find composer.json.

Joris-van-der-Wel:
Why not use something like this: https://github.com/francoispluchino/composer-asset-plugin/blob/master/Resources/doc/index.md

paladox:
Didn't know of that one. How would I just do something like 
{
    "extra": {
        "asset-repositories": [
            {
                "type": "bower-vcs",
                "url": "https://github.com/vendor/exemple-asset-name.git"
            }
        ]
    }
}
for this. 

Since it would be harder for others to depend on this repo since you would have to tell them to do it like this.

PR #1286, opened 2015-11-08
Use the "request" module instead of "browser-request"

### LINKED ISSUE #1285: [Browser] JSDom is unable to pull external dependencies when run inside a browser
I followed the method used [here](https://github.com/tmpvar/jsdom/issues/1018#issuecomment-73269131) to create a browserified jsdom script.

It seems that jsdom is not able to load the ``request`` library properly. This is the error I see in the Chrome browser console:
screenshot from 2015-11-07 18 32 26

This the HTML page I loaded:
```
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title></title>
    <script src='jsdom_browserify.js'></script>
    <script>
        window.onload = function () {
            var jsdom = require('jsdom_capsule');
            var virtualConsole = jsdom.createVirtualConsole().sendTo(console);
            jsdom.env({
                html: '<!DOCTYPE html><html><head></head><body></body></html>',
                scripts: [
                    'https://cdnjs.cloudflare.com/ajax/libs/jquery/1.11.2/jquery.min.js',
                    'https://cdnjs.cloudflare.com/ajax/libs/lodash.js/3.6.0/lodash.min.js'
                ],
                    console.log('I\'m done');
                },
                features: {
                    FetchExternalResources: ['script', 'link'],
                    ProcessExternalResources: ['script']
                },
                virtualConsole: virtualConsole
            });
        }
    </script>
</head>
<body>

</body>
</html>
```
I guess I could solve the issue by using a custom ``resourceLoader``, but I wonder why this is happening in the first place. The ["browser" directive in the ``package.json``](https://github.com/tmpvar/jsdom/blob/master/package.json#L55-L59) should ideally take care of this, because `request` is being replaced by `browser-request`.

Joris-van-der-Wel:
Currently, fetching files within the browser is broken because `browser-request` does not support `request.jar()`, also we are passing arguments to `request()` in such a way that it is not compatible with `browser-request`. The `request` library seems to support browsers just fine nowadays

This fixes #1285 @czardoz

I am going to try to update this PR with a simple test case to prevent regressions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
     "abab": "^1.0.0",
     "acorn": "^2.4.0",
     "acorn-globals": "^1.0.4",
-    "browser-request": ">= 0.3.1 < 0.4.0",
     "cssom": ">= 0.3.0 < 0.4.0",
     "cssstyle": ">= 0.2.29 < 0.3.0",
     "escodegen": "^1.6.1",
@@ -54,8 +53,7 @@
   },
   "browser": {
     "canvas": false,
-    "vm": "./lib/jsdom/vm-shim.js",
-    "request": "browser-request"
+    "vm": "./lib/jsdom/vm-shim.js"
   },
   "scripts": {
     "prepublish": "npm run convert-idl",
--- a/test/jsdom/env-browser.js
+++ b/test/jsdom/env-browser.js
@@ -1,0 +1,19 @@
+/* global location */
+
+"use strict";
+
+// This suite tests env() within a WebWorker
+
+const env = require("../..").env;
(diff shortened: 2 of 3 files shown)


domenic:
Awesome catch, thank you. Will pull in once we get a test case. The inside-worker-smoke-tests.js file seems appropriate.

Joris-van-der-Wel:
Added a test case

domenic:
Merged as 56323fc29e79fb567253d96a77886e41a7a37e03; thanks!!

PR #1288, opened 2015-11-10
Adds logic for retrieval of items by name and includes a test for it

julianhille:
document.form['someName'] was not possible.
I added this functionality to the nodeList class and keep the length to the original.
This reflects the browser behavior.

I was not sure where to put the test so i just added it to level2.

Hopefully your going fine with that as we have some legacy js which uses this alot.
Also for document.images['name'].

### DIFF
--- a/lib/jsdom/living/node-list.js
+++ b/lib/jsdom/living/node-list.js
@@ -64,6 +64,11 @@ function resetNodeListTo(nodeList, nodes) {
 
   for (let i = 0; i < nodes.length; ++i) {
     nodeList[i] = nodes[i];
+    if (typeof nodes[i].getAttribute !== "undefined") {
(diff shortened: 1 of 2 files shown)


domenic:
The problem with this approach is that it does not account for what happens when names/ids are removed or changed. To do this properly you need to use the named property tracker. See for example https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/named-properties-window.js and how each of the exported methods is used throughout the codebase.

julianhille:
As far as i understood the code: it is that the function updateNodeList  is called every time the list of items are accessed. And so the update would have been written into the object.
And because the snapshot is not changed but rather the returned list, it also has no updating issue.
But i get the point about the part with named-properties, which i just didn't see in the first place.

Issue #1295, opened 2015-11-21
Incorrect implementation of template element


clecap:
In my opinion the <template> element in HTML5 is treated incorrectly by this parser.

According to the html5rocks tutorial as well as the Mozilla implementation, the content of a template element should be placed into its .content and not into its .childNodes.

This deviation from what I understand is the accepted standards currently breaks code which is running in Firefox and Chrome when using jsdom parser.

domenic:
can you provide a small minimal code sample where jsdom behaves differently from browsers?

clecap:
Thanx domenic for your quick and very helpful reply.

TL;DR: Case solved. Issue closed. :-)

Longer: The minimal offending code sample is:

```javascript
   var resu = jsdom.env ("<html><head></head><body><template>Test</template></body></html>",
       function(e,d){
         console.warn (d.document.body.firstChild.content);
   });
```
and it produced "undefined" instead of a DocumentFragment.

However, when re-checking this I noticed that higher up in the code I was not loading jsdom but rather the old node-jsdom. Calling the variable jsdom was not my smartest choice of variable names, as I ended up complaining here instead of at node-jsdom. When I corrected this and updated my package definition the error vanished. 

Thanx for helping me debug a completely different part of my code, I appreciate this, and sorry for taking up your time on this. I think this should now un-break the rest of my code :-) if not I'll be back for more.

clecap:
Sorry, domenic, that I am back with this issue. Now I have a test case and a suggestion wrt the reason behind what I suspect to be a bug in jsdom

**In the browser**, the content inside of a template tag x can be found in the dom at x.content, where it has the form of a document fragment node with the appropriate number of nodes, i.e. in x.content.childNodes (in the example it is 3 childNodes). 

**In jsondom**, the content inside of a template tag x can be found in the dom at x.content, where it first is packaged in a document fragment node with ONE child node; this child node consists of another document fragment node, which then contains the appropriate number of nodes, i.e. in x.content.childNodes.item(0).childNodes. There is an additional, in my opinion unwanted, document fragment wrapper around the content of the template.

Here comes the example.

**In the browser** (firefox) this snippet produces

```
<html><head></head><body><template id="7">a<b>b</b>c</template>
<script>
function domPrintDebugMini (dom) {
  switch (dom.nodeType) {
    case 1:     
      if (dom.tagName == "TEMPLATE") {
[...]

PR #1297, opened 2015-11-22
Update package.json

clecap:
Also want to include scripts for npm processing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,5 +66,6 @@
     "benchmark-browser": "node ./benchmark/runner --bundle",
     "convert-idl": "node ./scripts/webidl/convert"
   },
+  "files": ["scripts/webidl"],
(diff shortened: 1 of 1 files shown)


Sebmaster:
What do you need them for? They shouldn't be included in the npm distribution.

domenic:
Yes, we're not going to do this, and certainly not with a commit message like this one.

clecap:
In my understanding I need the scripts to generate the contents of the /generated/ directory, as I am getting the code directly from git and not from the npm registry. I do this as I was not able to find a documentation for the build process for this situation. I pull from git instead of npm registry due to unfixed issue #1295.

This PR, of course, should not have been directed to github and certainly not to the master branch, but to my local mirror. I apologize for the confusion caused by a click in the wrong window.

PR #1303, opened 2015-11-28
Completely refactor htmltodom into a proper class

Sebmaster:
I took on the beast :tada: 

It's not quite there yet, for example I want to replace htmlparser2 as a default for xml with something else, but I thought it was a good start to get some eyes on it.

The no-parser support has been completely dropped and there are no more weird method switching hacks in there.


domenic:
Nice! Remove this from .eslintignore? Should be good to go as it looks like mostly moving code around (in a good way).

Sebmaster:
Okay, should be good to go.

This uses a very ugly hack in node-sax for now, which buffers script text, since we don't implement some of the intricacies of parser-appended script tags which I hope to replace soon, but it should work for most use-cases as-is.

Sebmaster:
Added all those tests.

Sebmaster:
Fixed up the commit.

PR #1304, opened 2015-11-29
Add extra request options

calebmer:
I want to make `POST` requests using `jsdom.env`, so this pr adds the ability to define extra [request options](https://www.npmjs.com/package/request#request-options-callback) in `config.request`.

I can add tests if you want, I just need to know where to put those tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ jsdom.env(config);
 - `config.virtualConsole`: a virtual console instance that can capture the window’s console output; see the "Capturing Console Output" examples.
 - `config.pool`: an object describing which agents to use for the requests; defaults to `{ maxSockets: 6 }`, see [request module](https://github.com/request/request#requestoptions-callback) for more details.
 - `config.agentOptions`: the agent options; defaults to `{ keepAlive: true, keepAliveMsecs: 115000 }`, see [http api](https://nodejs.org/api/http.html) for more details.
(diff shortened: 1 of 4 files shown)


Joris-van-der-Wel:
A good place to place a test case for this would be `test/jsdom/env.js`. Here is an example from that file:

```javascript
exports["explicit config.url, valid"] = t => {
  const html = "<html><head><title>Example URL</title></head><body>Example!</body></html>";
  const responseText = "<!DOCTYPE html>" + html;

  const server = http.createServer((req, res) => {
    res.writeHead(200, { "Content-Length": responseText.length });
    res.end(responseText);
    server.close();
  }).listen(8976);

  env({
    url: "http://localhost:8976/",
    done(err, window) {
      t.ifError(err);
      t.equal(serializeDocument(window.document), responseText);
      t.equal(window.location.href, "http://localhost:8976/");
      t.equal(window.location.origin, "http://localhost:8976");
      t.done();
    }
  });
};
```
This creates a temporary node http server. You could add tests such as `t.equal(req.method, 'POST")` (see https://nodejs.org/api/http.html#http_http_incomingmessage)

calebmer:
Ok, added a test and updated the README.

calebmer:
Whoops. My bad. Fixed.

domenic:
I don't really understand this PR. If you want all the flexibility of request, you should just use request:

```js
request({ ... lots of options ... }, function (error, response, body) {
  jsdom.jsdom(body, jsdomOptions);
});
```

This also has the advantage of making it clear that you're using request, instead of having to read the jsdom documentation which refers you to the request documentation.

PR #1307, opened 2015-11-30
Fix typo in symbols file name

jfremy:
The correct file containing the symbols for filellist.js is filelist-symbols, not file-symbols.

### DIFF
--- a/lib/jsdom/living/filelist.js
+++ b/lib/jsdom/living/filelist.js
@@ -1,6 +1,6 @@
 "use strict";
 
-const filelistSymbols = require("./file-symbols");
+const filelistSymbols = require("./filelist-symbols");
 
 module.exports = function (core) {
   class FileList {


domenic:
Thanks. This shows a pretty bad gap in our test coverage, so I'll wait to merge this until we can get a test showing the failure before the change.

jfremy:
I was thinking about ways to test this and the only thing I can come up with is a direct access to the array through its correct symbol name. If I use the methods in FileList, then since there is only one symbol for it, they will work fine (the FileList will have an array named 'undefined' that can be accessed and will behave correctly).

It might be generalized to other classes relying on this pattern:
- create object
- enumerate symbols from symbols list for the object
- for each symbol, check the object has a property with this key and verify the type if needed
Do you see any better way?

domenic:
Hmm, I see what you mean; this is not that easy to test after all. I guess we can just merge it as-is. Thanks for explaining.

PR #1308, opened 2015-11-30
INPUT tag of type 'file' should return an array for files

jfremy:
The INPUT tag of type 'file' returns a FileList object for the files property at the moment. This is not correct, it should return an array of Files.
Change the method so that it returns the Array backing the FileList instead of returning the FileList itself.

### DIFF
--- a/lib/jsdom/level2/html.js
+++ b/lib/jsdom/level2/html.js
@@ -875,6 +875,7 @@ define('HTMLOptionElement', {
 });
 
 const filesSymbol = Symbol("files");
+const fileListSymbol = require("../living/filelist-symbols")
 
 define('HTMLInputElement', {
   tagName: 'INPUT',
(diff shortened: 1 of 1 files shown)


domenic:
I'm sorry, this is not correct. See https://html.spec.whatwg.org/multipage/forms.html#htmlinputelement.

```js
  readonly attribute FileList? files;
```

jfremy:
Sorry, you're right. At the moment, that's the case.
But if you look at https://w3c.github.io/FileAPI/#filelist-section, it seems to be moving toward an Array.
Also, the example given by the w3c shows an array like access that does not work today with the current implementation.

Mozilla also already makes such an assumption (https://developer.mozilla.org/en/docs/Using_files_from_web_applications)

My fix is clearly incorrect, but there will probably be some changes needed in there in the future.

domenic:
We should make fileList[i] work, I agree. That will require hackery as we have done in the past. I will open a new issue for that.

PR #1312, opened 2015-12-04
Use `window.setTimeout` instead of `setTimeout`

cpojer:
(I made this through the GitHub UI, so.. didn't test? Is travis going to take care of this?)

### DIFF
--- a/lib/jsdom/browser/location.js
+++ b/lib/jsdom/browser/location.js
@@ -43,7 +43,7 @@ whatwgURL.mixinURLUtils(
         newURL: this.href
       });
 
-      setTimeout(() => {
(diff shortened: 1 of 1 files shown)


domenic:
Merged as 6472fcb6fbbe071e5c661673bdb19f8ea27cab04; new release soon!

cpojer:
Thanks @domenic. Someone should probably take a look whether there are any timer function calls that aren't using `window.`. :)

domenic:
Yeah, maybe. In the end there should be a different timeout-customizing mechanism here; after all in browsers overriding window.setTimeout doesn't change the timing of postMessage. But when we get around to that, it'll be a different story.

PR #1313, opened 2015-12-06
relative path example of jquery

ambodi:
The example is setting the relative path without __dirname which produces ENOENT error.


domenic:
It depends on how you run it. If you run it inside the examples directory, it works.

Regardless, this is a reasonable change. But please use `path.resolve(__dirname, "jquery.js")` instead of string concatenation.

ambodi:
@domenic nice input, thanks.

PR #1316, opened 2015-12-09
Use new parse5 streaming interface

### LINKED ISSUE #1315: Update parse5 to 2.0.0 to get line/col numbers in LocationInfo
As discussed in #1194 it would be nice to have line and column numbers for the location info. This has been implemented in the new version of parse5, 2.0.0 (see inikulin/parse5#67). 

It would be great if we could upgrade the included version of parse5 from 1.5.1 to 2.0.0. However there are some breaking changes in that update which means that it requires more work than just changing the version number.

Sebmaster:
Fixes #1315.

ToDo:
- [ ] Add stylesheet removal test
- [ ] Add test for link `styleSheets` insertion timing (before / after load)
- [ ] Test that stylesheets are already added/removed (in `styleSheets`) when dom mutation events are emitted
- [ ] Fix script insertion
- [ ] Add test for incomplete tag `document.write`s


Sebmaster:
This doesn't use the new JS capabilities parse5 got, but gives us the new location info that was added.

lib/jsdom/browser/htmltodom.js
```diff
@@ -237,7 +267,26 @@ function setChild(core, parent, node) {
     return null;
   }
 
-  newNode[locationInfo] = node.__location;
+  if (node.__location) {
+    newNode[locationInfo] = {
```
Sebmaster: Not too sure about this. This is mainly needed for backwards compat because start/end got renamed to (start|end)Offset. We could just alias them, or major ourselves...

domenic:
I'd like to be a bit more ambitious than this. We should do a major version bump and:

- Stop using parse5's htmlparser2 adapter, thus solving problems mentioned in https://github.com/inikulin/parse5/issues/78
- Remove support for all parsers except sax and parse5. Figure out the best way to change the `parser` option so that `if` statements are minimal. If people want to use other parsers they must ape the parse5 or sax interface. E.g. maybe we require HTML parsers to have parse, parseFragment, serialize, serializeFragment, and ParserStream?? We can remove the `parser` option until we switch parse5's new interface, if that makes it easier.
- Clean up dead code where we are patching up old parsers, e.g. parseDoctype.

Sebmaster:
Sounds good to me, will do that.

Sebmaster:
Finished up our document adapter, however there're a few issues there, one of them being that we don't handle scripts correctly now, not sure if there are others as well. I think parse5 creates a script, appends it to the dom and inserts text after, but we don't notice the appended text. The HTML spec describes how to handle parser inserted scripts correctly - I hope that's do-able with the jsdom-parse5 interface(s).

sykaeh:
Any updates on this?

Sebmaster:
@sykaeh It's the next thing I'll tackle, the node rewrite got in the way, but I'll have to re-implement scripts for this to work and that (or a different part afterwards) might reveal some more issues I haven't thought of yet.

lib/jsdom/browser/htmltodom.js
```diff
@@ -47,61 +45,27 @@ class HtmlToDom {
...
-      const instance = new this.parser.Parser(htmlparser2Adapter, { locationInfo: true });
-      dom = instance.parse(html);
-    }
-
-    const parsed = dom.children;
-    for (let i = 0; i < parsed.length; i++) {
-      setChild(this.core, element, parsed[i]);
+      const instance = new this.parser.ParserStream({
+        treeAdapter: adapter(element),
+        locationInfo: true
```
inikulin: I would recommend to not enable `locationInfo` by default. It has significant impact on performance and since it's not commonly used, I suggest to make it optional and disabled by default in jsdom.

domenic:
Wishlist: allow "invoking the fragment parsing algorithm with _text_ as markup, and _context_ as the context element" from other parts of jsdom. See #1355. Maybe requires parse5 support, haven't checked.

Sebmaster:
Our resource loader is currently killing my progress - it's way too tightly integrated with our delay-load-event logic, but is also the only place which contains our logic on how to correctly resolve urls to either files or urls and how to handle a custom fetch callback :(

Will need to work around it or clean it up...

MeirionHughes:
Any progress?

TimothyGu:
An update: parts of the cleanup is now being developed in https://github.com/tmpvar/jsdom/pull/2098. Resource loader is still there however, for now.

domenic:
Got some new work in this area coming up. Closing this older attempt.

PR #1318, opened 2015-12-11
Regression with serialisation of <math>

Sebmaster:
Serialising `math` elements generates a namespace declaration with `:xmlns` instead of `xmlns`.

Example code:

```js
var jsdom = require('jsdom');
var html = '<!DOCTYPE html><math xmlns="http://www.w3.org/1998/Math/MathML"></math>';
jsdom.env(html, (err, window) => {
  if (err) return console.error('ERROR: ' + err);
  console.log(jsdom.serializeDocument(window.document));
});
```

Output:

```html
<!DOCTYPE html><html><head></head><body><math :xmlns="http://www.w3.org/1998/Math/MathML"></math></body></html>
```

### DIFF
--- a/lib/jsdom/browser/htmltodom.js
+++ b/lib/jsdom/browser/htmltodom.js
@@ -242,7 +242,7 @@ function setChild(core, parent, node) {
   if (node.attribs) {
     for (let localName in node.attribs) {
       const value = node.attribs[localName];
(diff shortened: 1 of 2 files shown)


Sebmaster:
Attached fix here.

domenic:
Released as 7.2.1!

darobin:
You kids rock!

PR #1325, opened 2015-12-22
Make base64 contents optional in data urls

sebmck:
I wasn't sure where to put a test so I created a new `misc/utils` file. Let me know if there's a more appropriate place.

### DIFF
--- a/test/jsdom/resource-loading.js
+++ b/test/jsdom/resource-loading.js
@@ -71,3 +71,8 @@ exports["<frame> loading errors show up as jsdomErrors in the virtual console"] 
   jsdom.jsdom(`<frameset><frame src="http://localhost:12345/foo.html"></frameset>`,
     { virtualConsole, features: { FetchExternalResources: ["frame"] } });
 };
+
+exports["empty base64 data urls should be blank"] = t => {
+  jsdom.jsdom(`<link rel="stylesheet" href="data:text/css;base64,">`);
+  t.done();
+};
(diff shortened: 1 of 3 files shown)


domenic:
https://github.com/tmpvar/jsdom/blob/master/test/jsdom/utils.js is probably the right place.

Would you be up for adding a more integration-ey test in the context you encountered this in (e.g. `<link rel="stylesheet" href="data:...">`)? Probably in https://github.com/tmpvar/jsdom/blob/master/test/jsdom/resource-loading.js

sebmck:
Ah, didn't see that. Done.

test/jsdom/resource-loading.js
```diff
@@ -71,3 +71,10 @@ exports["<frame> loading errors show up as jsdomErrors in the virtual console"]
   jsdom.jsdom(`<frameset><frame src="http://localhost:12345/foo.html"></frameset>`,
     { virtualConsole, features: { FetchExternalResources: ["frame"] } });
 };
+
+exports["empty base64 data urls should be blank"] = t => {
+  jsdom.jsdom(`<link rel="stylesheet" href="data:text/css;base64,">`, error => {
```
domenic: This isn't a callback-taking function...

sebmck:
Thought I was using `jsdom.env` instead of `jsdom.jsdom`. I've removed the callback. `t.done()` doesn't appear to be required but I added it anyway. Sorry for the PR churn.

domenic:
Thanks so much! Bugfix release coming out shortly...

PR #1333, opened 2016-01-02
Transform all elements to a node infrastructure

Sebmaster:
@domenic It is finally done :tada: 

HTMLAppletElement causes a failing test because of the unimplemented `HTMLAppletElement`. The test also [fails in Chrome](http://w3c-test.org/dom/nodes/Node-cloneNode.html). Do we PR a removal from the wpt, disable the test or implement the element for now?


domenic:
If implementing it is easy, let's do that for now. If it's nontrivial, let's disable the test for now with a comment. In both cases, let's open an issue at least on WPT asking to remove it.

domenic:
I rewrote most of the commit messages to reference that they were changing tests, and updated the last commit message to explain a bit more. The only thing we need to do before merging this is publish the new version of webidl2js.

However, would you be up for moving the HTML elements into a subfolder (or sibling folder?) of nodes? I think it would be nice to be able to look for HTML element implementations or for node subtypes without having them mixed together.

Sebmaster:
> However, would you be up for moving the HTML elements into a subfolder (or sibling folder?) of nodes?

Not possible with webidl2js currently due to it's API / conversion algorithm steps :(

Fixed the rest of the issues.

inikulin:
:tada:

PR #1334, opened 2016-01-04
Post-1333 cleanup

domenic:
This includes a few commits that clean up in the wake of #1333.

We shouldn't merge this until master has passing tests; there are some characterData problems @Sebmaster is going to look in to first.


Sebmaster:
Fixed characterdata by fixing unsigned conversion in webidl-conversions. However now we have an issue with width/height in canvas - our tests try to set width/height to -1 which is wrapped to MAX_INT by webidl2js. In browsers this results in the canvas being reset to the default dimensions, but I can't find where in the spec that would happen.

Chrome totally just has width/height declared as long, so they might specifically handle negative values, Gecko has it as unsigned long in their binding layer, no idea how they manage to detect MAX_INT specifically and reset to default (maybe they convert it back in their actual implementation?).

domenic:
Regarding canvas, the issue is that reflecting an `unsigned long` sets to the default value when outside the range [0, 2147483647]. MAX_INT is outside that range, so it should get set the default value. So the specs do reflect implementations and we need to update our reflection support. And in this case I guess we can't use reflect anyway, since we don't support reflecting with default values yet.

domenic:
Pushed new commits that are basically complete. I am getting one failing web platform test having to do with ProgressEvent, which I will have to investigate tomorrow; I presumably did something wrong but I can't imagine what since I didn't really touch that code.

domenic:
Hmm, WPT doesn't fail on Travis; maybe I have an outdated local copy or something.

Sebmaster:
> Hmm, WPT doesn't fail on Travis; maybe I have an outdated local copy or something.

Might be an update either in webidl-conversions or webidl2js, I fiddled around a bit with them to make that work. I actually think webidl2js, with the inherited unforgeable isTrusted prop.

domenic:
OK in that case this is ready for review. I saw your commit with CSSOM View; looks good, thanks.

Sebmaster:
Changes LGTM. Have you checked if we're missing anything from the previously (incorrectly) reflecting attributes now?

domenic:
Hmm what do you mean exactly? I mostly didn't comment out any reflect stuff. Do you mean the define() calls that we used to have? I guess I should check that yeah...

Sebmaster:
> Do you mean the define() calls that we used to have?

Yeah, we defined all the reflecting attrs in those and I may have forgotten to pull all of them over (I only really noticed if I missed them because of tests).

domenic:
Awesome, found a few missing attributes by checking that. Thanks for the idea.

Issue #1241, opened 2015-09-20
Setting element border style loses some of the border values


dpvc:
If you set `el.style.border = "1px solid back"`, then `el.style.border` will be set to `black`, but the `1px` and `solid` are lost.  It appears that only the last word in the string is used, so `el.style.border = "1px solid"` sets it to `solid`.

Here is a test case:

```
var jsdom = require("jsdom").jsdom;
var doc = jsdom("");
var el = doc.createElement("div");
el.style.border = "1px solid black";
console.log(el.style.border);          // prints "black"
console.log(el.style.cssText);         // prints "border: black"
```

This is with jsdom 6.5.1.

pkra:
Is there a chance on an update (or suggestion) on this? FWIW, still seeing this on jsdom 7.1.0.

domenic:
The best thing to do would be to submit a PR to https://github.com/chad3814/CSSStyleDeclaration fixing the issue.

pkra:
Thanks for the pointer!

dpvc:
I spent quite some time looking into this, and it turns out that the issue is not with CSSStyleDeclaration itself, but in how jsdom uses it.  The problem seems to be the on-change handler in [level2/html.js](https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/level2/html.js#L341-L343).  If you remove the callback, then the example code above produces the correct results.

It appears that these lines are designed to get the updated style definition added to the element as the `style` attribute, and indeed if you remove the callback from `html.js` and add  `console.log(div.outerHTML)` to the end of the example above, the result is `<div></div>` with no `style` attribute.

Although the situation is pretty complicated, it seems that what is happening is that during the `that.setAttribute('style', newCssText)` call, the style attribute is being modified for each of the individual style attributes from the `border` (i.e., the style is set for `border-width:1px`, `border-style:solid`, and `border-color:black` separately).  This causes the previous settings to be lost, and so only the last one is being retained.

My suggested fix is to replace those three lines with

```
      if (!that._settingCssText) {
        that._settingCssText = true;
        that.setAttribute('style', newCssText);
        that._settingCssText = false;
      }
```

so that recursive calls to `setAttribute()` aren't made.  This seems to resolve the issue for me.  Of course, I don't know the codebase for jsdom very intimately, so and not sure if that will cause any other problems.  But the change doesn't seem to have caused any problems during my testing of the change in MathJax-node, which can do a lot of style setting.

[...]

PR #1337, opened 2016-01-04
Allow style changes using shorthands to set all the sub-styles properly

dpvc:
This pull request resolves issue #1241.  To test, use

```
var jsdom = require("jsdom").jsdom;
var doc = jsdom("");
var el = doc.createElement("div");
el.style.border = "1px solid black";
console.log(el.style.border);          // should be "1px solid black"
console.log(el.style.cssText);         // should be  "border: 1px solid black"
console.log(el.outerHTML);             // should be '<div style="border: 1px solid black"></div>'
el.style = "border: 1px solid black";
console.log(el.style.border);          // should be "1px solid black"
console.log(el.style.cssText);         // should be  "border: 1px solid black"
console.log(el.outerHTML);             // should be '<div style="border: 1px solid black"></div>'
```

Without the patch the border would be reported as just `black` rather than `1px solid black`.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLElement-impl.js
@@ -14,9 +14,9 @@ class HTMLElementImpl extends ElementImpl {
 
     this._style = new this._core.CSSStyleDeclaration(newCssText => {
       if (!this._settingCssText) {
-	this._settingCssText = true;
+        this._settingCssText = true;
         this.setAttribute("style", newCssText);
-	this._settingCssText = false;
+        this._settingCssText = false;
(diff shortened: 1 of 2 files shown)


domenic:
Thank you! Would you be up for adding a test to https://github.com/tmpvar/jsdom/blob/master/test/level2/style.js ?

dpvc:
OK, I've added a test, as requested.

Note that the test suite fails on some canvas tests, but these are not related to my changes (the master branch fails on these same tests).

domenic:
Merged as 124456566b2f93acf75a06ab363849f21e4e1a7a; thank you!!

PR #1338, opened 2016-01-05
Properly catch out-of-bounds args in <canvas>#width/height

Sebmaster:
This should finally fix master after #1333.


domenic:
Annoying that Travis still counts it as a half-red build, but, yay!

Issue #1291, opened 2015-11-16
`setTimeout` and `setInterval` does not support strings as callbacks


diegoperini:
Browsers tend to evaluate `setTimeout('myFunction()',0)` type of usage. Current javascript client implementation in jsdom causes a crash. I forked the project to silence such errors but I believe an implementation that mimics browsers would be more appropriate.

Here is the [fork](https://github.com/diegoperini/jsdom-se)

overflowz:
Hi, I need this feature asap, I could not find the implementation where should I change it. I'm just interested, when you will support this feature or can you point me where it is implemented?

Regards.

domenic:
This will be implemented when someone, perhaps you, submits a pull request (that matches the spec). The place to do so would be https://github.com/tmpvar/jsdom/blob/d7b2f7b91e7aa64c072db7a8c3bbd16f9d1d2095/lib/jsdom/browser/Window.js#L131-L139

PR #1342, opened 2016-01-07
Fix setTimeout for string attributes

Sebmaster:
Only read up on the spec a tiny bit (until I dive into the Window rework), but this should be a pretty nice fix for #1291, going mostly by rudimentary behaviour here.

Spec says
> Create a script using script source as the script source, the URL where script source can be found, and settings object as the environment settings object.

I'm not sure if the url there means that stack traces should spit out that url as the source, but added that with a nice trick. We'll never get the stacktrace as a whole correct I think, sadly.

@domenic You think this would suffice?

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -18,6 +18,7 @@ const internalConstants = require("../living/helpers/internal-constants");
 const createFileReader = require("../living/file-reader");
 const createXMLHttpRequest = require("../living/xmlhttprequest");
 const Document = require("../living/generated/Document");
+const reportException = require("../living/helpers/runtime-script-errors");
 
 // NB: the require() must be after assigning `module.exports` because this require() is circular
 // TODO: this above note might not even be true anymore... figure out the cycle and document it, or clean up.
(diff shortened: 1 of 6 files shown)


lib/jsdom/browser/Window.js
```diff
@@ -403,6 +404,21 @@ function matchesDontThrow(el, selector) {
 }
 
 function startTimer(window, startFn, stopFn, callback, ms) {
+  if (typeof callback !== "function") {
+    const code = String(callback);
+    callback = window._globalProxy.eval.bind(window,
+      `(function(){` + code + `}())\n//# sourceURL=${window.location.href}`);
```
domenic: Does this sourceURL actually work in node? Or is this purely for browser users?

It seems better to use vm since we can specify the filename there.

lib/jsdom/browser/Window.js
```diff
@@ -403,6 +404,21 @@ function matchesDontThrow(el, selector) {
...
+  }
+
+  const oldCallback = callback;
+  callback = () => {
+    try {
+      oldCallback();
```
domenic: Need to pass arguments, and a this value of the window

lib/jsdom/browser/Window.js
Sebmaster: Actually works, it's a v8 thing. vm is ~5x slower than eval, see https://github.com/facebook/jest/pull/635#issuecomment-162708673.

lib/jsdom/browser/Window.js
Sebmaster: Arguments to a setTimeout callback?

domenic: setTimeout((a, b) => console.log(a + b), 1, 2)

lib/jsdom/browser/Window.js
domenic: Why do we wrap it in an IIFE?

domenic:
It would be nice to write some to-upstream tests for argument-passing behavior and for exception-reporting behavior

lib/jsdom/browser/Window.js
Sebmaster: I wanted to make sure `arguments.callee.toString` doesn't include the sourceURL, but executed code can't access arguments anyways, so removed.

domenic:
LGTM just missing tests

Sebmaster:
Added tests and rebased on #1350 so we don't lose that one.

domenic:
Updated the tests a bit and merged

PR #1320, opened 2016-01-07
removeEventListener doesn't work when useCapture is specified

Sebmaster:
```javascript
var jsdom = require("jsdom");

jsdom.env(
  "<p id='test'></p>",
  function (err, window) {
	var ticks = 0,
		listener = function() { ticks++ },
		elm = window.document.getElementById('test');
	
	// Attach 
    elm.addEventListener('test', listener, false);
    
    // Dispatch
	event = new window.CustomEvent('test');
    elm.dispatchEvent(event);
    
    console.log(ticks); // 1
    
    // Detach
    elm.removeEventListener('test', listener);
    
    // Dispatch
	event = new window.CustomEvent('test');
    elm.dispatchEvent(event);
    
    console.log(ticks); // 2 (should be 1)
  }
);
```

Bad behavior with those values:

```javascript
elm.addEventListener('test', listener, false);
elm.removeEventListener('test', listener);

elm.addEventListener('test', listener);
elm.removeEventListener('test', listener, false);

elm.addEventListener('test', listener, undefined);
elm.removeEventListener('test', listener, false);

elm.addEventListener('test', listener, false);
elm.removeEventListener('test', listener, undefined);
```

Correct behavior with those values:

```javascript
elm.addEventListener('test', listener);
elm.removeEventListener('test', listener);

elm.addEventListener('test', listener, true);
elm.removeEventListener('test', listener, true);

elm.addEventListener('test', listener, false);
elm.removeEventListener('test', listener, false);

elm.addEventListener('test', listener, null);
elm.removeEventListener('test', listener, null);

elm.addEventListener('test', listener, undefined);
elm.removeEventListener('test', listener, undefined);
```

jsdom 7.2.1
node 4.1.2

### DIFF
--- a/lib/jsdom/living/events/Event-impl.js
+++ b/lib/jsdom/living/events/Event-impl.js
@@ -7,13 +7,7 @@ class EventImpl {
     const type = args[0]; // TODO: Replace with destructuring
     const eventInitDict = args[1] || EventInit.convert(undefined);
 
-    if (privateData.noInitialize) {
-      this.type = "";
-      this._initializedFlag = false;
-    } else {
-      this.type = type;
-      this._initializedFlag = true;
-    }
+    this.type = type;
 
     const wrapper = privateData.wrapper;
     for (const key in eventInitDict) {
@@ -26,6 +20,7 @@ class EventImpl {
     this.currentTarget = null;
     this.eventPhase = 0;
 
+    this._initializedFlag = true;
     this._stopPropagationFlag = false;
     this._stopImmediatePropagationFlag = false;
     this._canceledFlag = false;
--- a/lib/jsdom/living/events/EventTarget-impl.js
+++ b/lib/jsdom/living/events/EventTarget-impl.js
@@ -26,6 +26,10 @@ class EventTargetImpl {
       return;
     }
 
+    if (capture === undefined) {
+      capture = false;
+    }
+
     if (!this._eventListeners[type]) {
       this._eventListeners[type] = [];
     }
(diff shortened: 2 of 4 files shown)


Sebmaster:
Attached PR

domenic:
Why isn't this taken care of by the IDL conversion? `undefined` to `optional boolean` should give `false` automatically.

Sebmaster:
Webidl spec says

> It is strongly suggested not to use default value of true for boolean-typed arguments, as this can be confusing for authors who might otherwise expect the default conversion of undefined to be used (i.e., false).

which sounds more like a `should not` instead of a `must not`.

domenic:
That doesn't have anything to do with the Web IDL conversions though; it's just non-normative text guiding spec authors to not do `optional boolean capture = true`.

Sebmaster:
Couldn't a spec describe special default behaviour in the actual description though? I'm not sure `= false` is implicit in this case.

domenic:
A spec could define special case behavior. But this one doesn't. It's pretty clear:

- optional means "don't throw when undefined is passed"
- boolean means "convert argument using Boolean()"

domenic:
Oh, I see what you mean, hmm.

domenic:
Yeah, OK, makes sense.

What will happen if I pass `1`? Will webidl2js correctly convert that to `true`?

Sebmaster:
Yeah, only `undefined` is special cased.

domenic:
So uh this also needs tests

Sebmaster:
Done :tada:

PR #1343, opened 2016-01-08
Fix browser build after node refactor

### LINKED ISSUE #1339: Now the browser test runner is broken on Travis
https://travis-ci.org/tmpvar/jsdom/jobs/100284756

The pain never ceases. Maybe we actually broke the browser tests? I don't have Java installed currently; will have to investigate tomorrow.

domenic:
#1333 introduced some dynamic requires, which are not good for the browser build. Fixes #1339.


Sebmaster:
:+1:

Issue #1219, opened 2015-08-28
Support for `insertAdjacentHTML`


danawoodman:
Is there plans to support [`element.insertAdjacent`](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML) in jsdom?

```js
document.body.insertAdjacentHTML('afterbegin', '<p>Hello World</p>')
```

Results in:

```
TypeError: document.body.insertAdjacentHTML is not a function
```

Joris-van-der-Wel:
Yes, I was planning to add this eventually

mbrevda:
+1. Fresh jquery convert here: is there another way that I can use in the meantime?

domenic:
You can just use `document.body.innerHTML = '<p>Hello World</p>' + document.body.innerHTML`. It will be slower as it has to re-parse everything, but it will work. You can also use DOM APIs directly.

mbrevda:
Oh, cool. Thanks! Seems `insertAdjacentHTML` is much more efficient, but I guess this will do in the mean time

Joris-van-der-Wel:
Here is how you can avoid a reparse. This would be similar to how `insertAdjacentHTML` could be implemented:

```javascript
function htmlToFragment(ownerDocument, markup) {
    const container = ownerDocument.createElement('div');
    container.innerHTML = markup;
    const fragment = ownerDocument.createDocumentFragment();
    while (container.firstChild) {
        fragment.appendChild(container.firstChild);
    }
    return fragment;
}
```
(might contain typo's)

Then you can do something like: `node.appendChild(htmlToFragment(document, '<p>hello world</p>'))`

domenic:
Yeah, that will generally work, although it doesn't work for e.g. inserting `<tr>`s into tables.

patrickfatrick:
+1. Trying out jsdom for the first time today and came across this limitation.

tombh:
You can write your own with something like `window.HTMLElement.prototype.insertAdjacentHTML = function(){}`

inikulin:
@domenic It will work even for `<tr>` if you'll use `<template>` as `container`.

PR #1355, opened 2016-01-14
Implement Element.insertAdjacentHTML()

kasperisager:
Ref. #1219. Grabbed and slightly tweaked the implementation from https://github.com/assaf/zombie/blob/005b4726c9c8bd02a234cb26cb756d90358bf0a4/src/dom/jsdom_patches.js#L60-L90

A test case has been added, but both this as well as the implementation might need some more work. It's a start though!


kasperisager:
Also, the WHATWG spec doesn't yet contain a definition of insertAdjacentHTML but I've added a reference to https://dom.spec.whatwg.org/#dom-element-insertadjacenthtml as I wasn't sure what else to put there (https://w3c.github.io/DOM-Parsing/#widl-Element-insertAdjacentHTML-void-DOMString-position-DOMString-text maybe?).

domenic:
Ah excellent! Could you enable https://github.com/w3c/web-platform-tests/blob/master/domparsing/insert-adjacent.html as well (and make sure it passes)? And probably move your test into to-upstream/domparsing.

kasperisager:
Ah, I totally missed that one! Should make my test case obsolete then.

lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -430,6 +430,43 @@ class ElementImpl extends NodeImpl {
 
     return attr;
   }
+
+  insertAdjacentHTML(position, text) {
+    const parentNode = this.parentNode;
+    const container = this.ownerDocument.createElementNS("http://www.w3.org/1999/xhtml", "_");
```
domenic: You should use `this.ownerDocument.createElement("template")` as the container. Also you should add some tests showing how using `"_"` fails for elements like `thead`, `tr`, `td`.

domenic:
The spec is actually at https://w3c.github.io/DOM-Parsing/#widl-Element-insertAdjacentHTML-void-DOMString-position-DOMString-text

lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -430,6 +430,43 @@ class ElementImpl extends NodeImpl {
 
     return attr;
   }
+
+  insertAdjacentHTML(position, text) {
```
domenic: Looks like "If context is null or a document, throw a DOMException with name "NoModificationAllowedError"." would be good to add a test for this too.

lib/jsdom/living/nodes/Element-impl.js
domenic: Hmm reading the spec it seems it does something a bit more complicated, e.g. defaulting to body in certain cases. I think it's OK to just use template as it will be most correct, but be careful and check browsers. If there are some edge cases that neither "template" nor "body" cover, then it's OK, we can just leave a TODO.

domenic:
FYI the existing web platform test is failing; not sure exactly why...

kasperisager:
Read through the spec more thoroughly and re-did what I had copied from zombie. The tests pass on my end now :+1:

kasperisager:
I should note that not all conditions of step 2 are checked.

lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -430,6 +430,58 @@ class ElementImpl extends NodeImpl {
...
+    position = position.toLowerCase();
+
+    let context;
```
domenic: So as-is context is unused. I think the spec is not going to be very useful for a guidance on this since it includes "invoking the fragment parsing algorithm with text as markup, and context as the context element" which I don't think we have the ability to do (yet). I made a note of that in https://github.com/tmpvar/jsdom/pull/1316#issuecomment-172030345

Instead I think you should remove context, and just check this.parentNode in line 442. Lines 542 and 460-462 can go away.

lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -430,6 +430,58 @@ class ElementImpl extends NodeImpl {
...
+
+    switch (position) {
+      case "beforebegin":
+      case "afterend":
+        context = this.parentNode;
+
+        if (context === null || context.nodeType === NODE_TYPE.DOCUMENT_NODE) {
+          throw new DOMException(DOMException.NO_MODIFICATION_ALLOWED_ERR, "Modifications are not allowed for this" +
```
domenic: I don't believe the existing test suite covers these exceptions, so adding to-upstream tests for them would be lovely.

domenic:
Thank you very much; lovely work! Merged as 33b93417f93c41e67ba01e95faedfb1d542fdff7.

kasperisager:
Cheers!

RReverser:
Looks like this should be trivial to extend to support `insertAdjacentText`?

Zirro:
@RReverser We would certainly appreciate a pull request doing so, assuming it passes the relevant tests.

PR #1356, opened 2016-01-15
added elem.getClientRects

mciparelli:
didn't add tests as I didn't find any for `getBoundingClientRect`, but I could if you want me to

### DIFF
--- a/lib/jsdom/living/nodes/HTMLElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLElement-impl.js
@@ -46,6 +46,14 @@ class HTMLElementImpl extends ElementImpl {
       width: 0
     };
   }
+  getClientRects() {
+    return [{
+      bottom: 0,
+      left: 0,
+      right: 0,
+      top: 0
+    }];
+  }
   focus() {
     this._ownerDocument.activeElement = this;
   }


domenic:
Yeah, for stubs I guess there's no need. Looks good, I will merge tonight or tomorrow.

domenic:
Fixed in a bit more spec-compliant way in 892c63d271168005e19adeb53a84778103fc406c. The IDL needed updating, both for getBoundingClientRect and getClientRects.

danieler1981:
Why **getClientRects** always return [] and **getBoundingClientRect** always return {0,0,0,0,0}?
Any workaround?

Sebmaster:
> Any workaround?

No. You can overwrite the methods if there's a specific value you're expecting in tests, but jsdom doesn't do layout, so it can't calculate these accurately. You'll have to go with a headless browser if you need that.

danieler1981:
Thank You; i'm switching this part of code on Google Chrome Protocol.

I saw more in details the readme of JSDOM where is pretty clear this portion of code is out of scope for JSDOM.

> **Beyond just features that we haven't gotten to yet, there are two major features that are currently outside the scope of jsdom**

PR #1358, opened 2016-01-19
Implement and use adoptNode

dmethvin:
Still a WIP at this point. It's passing all tests including the basic adoptNode tests as added but there are no tests yet that use the auto adoption steps. I see some in level3/core.js but that file as a whole can't be run. What's the best way to add tests? Should I write fresh ones for to-upstream, or instead try to rescue the ones in core.js somehow?

### DIFF
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -657,9 +657,8 @@ class DocumentImpl extends NodeImpl {
     if (node.nodeType === NODE_TYPE.DOCUMENT_NODE) {
       throw new DOMException(DOMException.NOT_SUPPORTED_ERR, "Cannot adopt a document node");
     }
-    if (node.nodeType === NODE_TYPE.DOCUMENT_FRAGMENT_NODE) {
-      throw new DOMException(DOMException.HIERARCHY_REQUEST_ERR, "Cannot adopt a shadow root node");
-    }
+    // TODO: Determine correct way to detect a shadow root
+    // See also https://github.com/w3c/webcomponents/issues/182
 
     if (node.parentNode) {
       node.parentNode.removeChild(node);
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -650,6 +650,28 @@ class DocumentImpl extends NodeImpl {
     return clone(this._core, node, this, deep);
   }
 
+  adoptNode(node) {
+    if (!("_ownerDocument" in node)) {
+      throw new TypeError("First argument to adoptNode must be a Node");
+    }
+    if (node.nodeType === NODE_TYPE.DOCUMENT_NODE) {
+      throw new DOMException(DOMException.NOT_SUPPORTED_ERR, "Cannot adopt a document node");
+    }
+    if (node.nodeType === NODE_TYPE.DOCUMENT_FRAGMENT_NODE) {
+      throw new DOMException(DOMException.HIERARCHY_REQUEST_ERR, "Cannot adopt a shadow root node");
+    }
+
+    if (node.parentNode) {
+      node.parentNode.removeChild(node);
+    }
+    node._ownerDocument = this;
+    for (const descendant of domSymbolTree.treeIterator(node)) {
+      descendant._ownerDocument = this;
+    }
+
+    return node;
+  }
+
   get cookie() {
     return this._cookieJar.getCookieStringSync(this._URL, { http: false });
   }
--- a/lib/jsdom/living/nodes/Node-impl.js
+++ b/lib/jsdom/living/nodes/Node-impl.js
@@ -213,6 +213,11 @@ class NodeImpl extends EventTargetImpl {
         oldParentImpl.removeChild(newChildImpl);
       }
 
+      // adopt the node if it's not in this document
+      if (this._ownerDocument !== newChildImpl._ownerDocument) {
+        this._ownerDocument.adoptNode(newChildImpl);
+      }
+
       if (refChildImpl === null) {
         domSymbolTree.appendChild(this, newChildImpl);
       } else {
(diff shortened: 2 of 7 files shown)


lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -650,6 +650,28 @@ class DocumentImpl extends NodeImpl {
...
 
+  adoptNode(node) {
+    if (!("_ownerDocument" in node)) {
+      throw new TypeError("First argument to adoptNode must be a Node");
+    }
+    if (node.nodeType === NODE_TYPE.DOCUMENT_NODE) {
+      throw new DOMException(DOMException.NOT_SUPPORTED_ERR, "Cannot adopt a document node");
+    }
+    if (node.nodeType === NODE_TYPE.DOCUMENT_FRAGMENT_NODE) {
+      throw new DOMException(DOMException.HIERARCHY_REQUEST_ERR, "Cannot adopt a shadow root node");
```
dmethvin: I'm not sure how to detect a shadow root node. Advice?

Sebmaster:
to-upstream is definitely preferred for tests. We'll get them into the wpt tests eventually.

lib/jsdom/living/nodes/Document-impl.js
domenic: We don't support shadow roots yet so let's just leave this out.

domenic:
This looks quite nice to me.

I think you can leave adopting steps out for now. They only apply to `template` and the implementation there isn't complete enough for them to matter. Just add a // TODO line and we'll be fine.

So yeah, just add a TODO line and remove the shadow root check (add a TODO for that too I suppose) and we should be good to go.

dmethvin:
Turns out I definitely need to remove the shadow root check because it's legit to adopt a document fragment. However, in order to get this to work with jQuery 2.2.0 I need the adopting steps to work because we essentially do this:
```js
var doc = document.implementation.createHTMLDocument("");
var div = doc.createElement("div");
div.innerHTML = "<p>test</p>";
document.body.appendChild(div.firstChild);
```
I've pushed an update to do the adoption, but it's failing 9 level1 tests that expect a WRONG_DOCUMENT_ERR. How do you usually deal with those kinds of situations?

domenic:
> However, in order to get this to work with jQuery 2.2.0 I need the adopting steps to work because we essentially do this:

Hmm I still don't quite understand why that would be important. Adopting steps only exist for template elements, which I don't see any of in your example.

It would still be good to add a "TODO: implement adopting steps" at the bottom of adoptNode, but I don't see why it's necessary to implement them for jQuery 2.2.0.

> it's failing 9 level1 tests that expect a WRONG_DOCUMENT_ERR. How do you usually deal with those kinds of situations?

Delete the tests!! level1 tests are only good until they are not good anymore.

lib/jsdom/living/nodes/Node-impl.js
```diff
@@ -181,19 +181,19 @@ class NodeImpl extends EventTargetImpl {
...
-    // search for parents matching the newChild
-    for (const ancestor of domSymbolTree.ancestorsIterator(this)) {
-      if (ancestor === newChildImpl) {
-        throw new DOMException(DOMException.HIERARCHY_REQUEST_ERR);
+    if (this._ownerDocument !== newChildImpl._ownerDocument) {
+      // adopt the node when it's not in this document
+      this._ownerDocument.adoptNode(newChildImpl);
+    } else {
+      // search for parents matching the newChild
+      for (const ancestor of domSymbolTree.ancestorsIterator(this)) {
```
domenic: Why is this in the else clause?

dmethvin: If the node was in another document and we are adopting it, it cannot be an ancestor of `newChildImpl`. I suppose it could be left as it was, but for that case the search will always be fruitless.

dmethvin: > Hmm I still don't quite understand why that would be important. Adopting steps only exist for template elements, which I don't see any of in your example.

The also appear in https://dom.spec.whatwg.org/#mutation-algorithms as part of the pre-insertion validity checks?

domenic: > If the node was in another document and we are adopting it, it cannot be an ancestor of newChildImpl. I suppose it could be left as it was, but for that case the search will always be fruitless.


Oh, good point! Maybe add a note so that we remember this in the future.

> The also appear in https://dom.spec.whatwg.org/#mutation-algorithms as part of the pre-insertion validity checks?

Not quite understanding. Adopting steps only appear in https://dom.spec.whatwg.org/#concept-node-adopt step 4 and are only defined for template elements (around https://html.spec.whatwg.org/#appropriate-template-contents-owner-document)

dmethvin: In https://dom.spec.whatwg.org/#mutation-algorithms which seems to be the prelude to `appendChild` and such, it says " To pre-insert a node into a parent before a child, run these steps: ... 4. Adopt node into parent’s node document."  which links to https://dom.spec.whatwg.org/#concept-node-adopt.

domenic: Ah, I guess we are talking about different things. You think "adopting steps" means "the step in the algorithm which adopts the node". I am referring to https://dom.spec.whatwg.org/#concept-node-adopt-ext which is referred to as "For each descendant in node’s inclusive descendants, in tree order, run the adopting steps with descendant and oldDocument."

dmethvin: So many magic phrases. Okay, to rephrase, I need  https://dom.spec.whatwg.org/#mutation-algorithms to work, in particular the part in "4. Adopt node into parent’s node document.". 

I think once I delete out the level1 failing tests this should be done.

domenic: Got it. Yes, agreed!

domenic:
Merged as d00f0af3a005175d55de79838bac3a47f0841d9f! We should be doing a 8.0.0 release within a few days.

PR #1363, opened 2016-01-21
Upgrade cssstyle and add test showing it fixes #1353

### LINKED ISSUE #1353: Padding and margin don't set the -top, -right, -bottom, and -left properties
When `style.padding` or `style.margin` are set, the corresponding `style.paddingTop`, `style.paddingRight`, etc. properties are not set.  In particular, setting `style.padding = ""` does not clear existing `style.paddingTop` or the other properties.

This is actually a CSSSyleDeclaration bug, and I have made a [pull request](https://github.com/chad3814/CSSStyleDeclaration/pull/35) for it, so it should be fixed upstream.  I just wanted you to be aware of it.

Actually, there are a number of similar bugs in CSSStyleDeclaration (e.g., setting `style.font` is quite broken).  I'm not going to be able to take time to fix them all.  Also, the library seems pretty inefficient.  I suspect that fixing some of the inefficiencies would help improve jsdom's performance.

domenic:
Test does not yet pass (so we can't merge this PR yet) due to https://github.com/chad3814/CSSStyleDeclaration/issues/37

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "acorn-globals": "^1.0.4",
     "array-equal": "^1.0.0",
     "cssom": ">= 0.3.0 < 0.4.0",
-    "cssstyle": ">= 0.2.29 < 0.3.0",
+    "cssstyle": ">= 0.2.34 < 0.3.0",
     "escodegen": "^1.6.1",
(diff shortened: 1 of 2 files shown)


chad3814:
this passes with cssstyle >= 0.2.33

Issue #1036, opened 2015-02-20
Clean up tests


Sebmaster:
Our test framework has become pretty unwieldy; nodeunit isn't maintained anymore, there're all kind of unused stuff in our tests, as well as tests which are not run regularly (if at all).

We should begin a clean up of the tests, some suggestions include:
- [ ] Migration to mocha or web-platform-tests as appropriate. Remaining tests:
  - [x] level1/core.js
  - [ ] level2/core.js
  - [ ] level2/html.js
  - [ ] level2/style.js
  - [ ] level2/events.js
  - [ ] level3/textContent.js
  - [ ] level3/xpath.js
  - [x] living-dom/attributes.js
  - [ ] living-dom/class-list.js
  - [ ] living-dom/compare-document-position.js
  - [ ] living-dom/dom-implementation.js
  - [ ] living-dom/event-target.js
  - [ ] living-dom/node-clone-node.js
  - [ ] living-dom/node-contains.js
  - [ ] living-dom/node-list.js
  - [ ] living-dom/node-owner-document.js
  - [ ] living-dom/node-parent-element.js
  - [ ] living-dom/query-selector.js
  - [ ] living-dom/query-selector-all.js
  - [ ] living-dom/node-iterator.js
  - [ ] living-dom/parent-node.js
  - [ ] living-dom/non-document-type-child-node.js
  - [x] living-html/cookie.js
  - [ ] living-html/current-script.js
  - [x] living-html/focus.js
  - [ ] living-html/htmlanchorelement.js
  - [x] living-html/htmlbuttonelement.js
  - [ ] living-html/htmlelement.js
  - [ ] living-html/htmlcanvaselement.js
  - [ ] living-html/htmlimageelement.js
  - [ ] living-html/htmlinputelement.js
  - [x] living-html/htmloptionelement.js
  - [x] living-html/htmltextareaelement.js
  - [x] living-html/htmlformelement.js
  - [ ] living-html/inline-event-handlers.js
  - [ ] living-html/location.js
  - [ ] living-html/message-event.js
  - [x] living-html/navigator.js
  - [ ] living-html/on-error.js
  - [ ] living-html/post-message.js
  - [x] misc/url.js
  - [ ] misc/xhr-file-urls.js
  - [ ] misc/xhr-requires-server.js
  - [x] misc/cancel-requests.js
  - [x] misc/keep-alive-connections.js
  - [ ] window/history.js
  - [ ] window/script.js
  - [x] window/console.js
[...]

PR #1375, opened 2016-02-04
Use Mocha & Karma for test cases

Joris-van-der-Wel:
See #1036

This PR contains all the infrastructure needed:
* Mocha is used for the test cases and as the runner.
* [mocha-sugar-free](https://github.com/Joris-van-der-Wel/mocha-sugar-free) is used so that we can use the style of coding we are used to. For example `it()`, `describe()`, etc are globals in mocha, mocha-sugar-free lets you define them using `require()`. It also adds support for arrow functions, which mocha does not like. One extra feature I added which is convenient in jsdom is that you can exclude a test case from being run in the browser by passing options to mocha's `it()` function. Instead of having to blacklist/whitelist test files (what we did in the past)
* Karma is used instead of selenium. This speeds up the tests on your own machine a lot, and it also lets you easily test mobile browsers (all you have to do is navigate to your lan ip address).
* Browser tests are now being run both with and without a Worker. Previously tests were only run within a Worker. I created a karma plugin for this: [karma-mocha-webworker](https://www.npmjs.com/package/karma-mocha-webworker)
* [karma-browserify](https://www.npmjs.com/package/karma-browserify) is used instead of having to generate a bundle ourselves. This plugin supports fast incremental rebuilds (which then trigger a fresh test run). Getting this to work actually took me the most time of everything because I had to fix a few issues in browserify and karma-browserify

So far I have converted one test case to mocha (test/jsdom/env.js). I made a small tool for myself which takes care of most of the conversion. Until all test cases have been converted, the mocha tests can live alongside the nodeunit tests (and the npm test commands and the travis build will run both).

Here is a successful travis build: https://travis-ci.org/Joris-van-der-Wel/jsdom/builds/107108618


Sebmaster:
Is the options object with `async: true` a mocha-sugar-free thing? I think mocha supports Promises natively, why not use those / make mocha-sugar-free also accept promises?

Will check the actual infrastructure commit later today.

Joris-van-der-Wel:
> Is the options object with async: true a mocha-sugar-free thing? I think mocha supports Promises natively, why not use those / make mocha-sugar-free also accept promises?

Yes that is a mocha-sugar-free thing. Normally mocha decides that your test is async by checking `func.length > 0`. I would rather make this explicit, therefor `{async:true}`.

Promises are also [supported](https://github.com/Joris-van-der-Wel/mocha-sugar-free/blob/4e968012bd1f87f50106979ba08a8a65bad1f3ab/test/bdd.js#L31), simply return them (everything that works in mocha should work in mocha-sugar-free).

I have not used promises in env.js simply because we did not use them before. I could add them if that is a style we would like to follow, however it is a bit of a work because jsdom does not support them in its API:

```
it("explicit config.html, with overridden config.url", { async: true }, testCase => {
  env({
    html: "Hello",
    url: "http://example.com/",
    done(err, window) {
      t.ifError(err);
      t.equal(serializeDocument(window.document), "<html><head></head><body>Hello</body></html>");
      t.equal(window.location.href, "http://example.com/");
      t.equal(window.location.origin, "http://example.com");
      t.equal(window.location.search, "");
      t.equal(window.location.hash, "");
      testCase.done();
    }
  });
});
```

```
it("explicit config.html, with overridden config.url", () => {
  return new Promise( (resolve, reject) => {
    env({
      html: "Hello",
      url: "http://example.com/",
      done(err, window) {
        err ? reject(err) : resolve(window);
      }
    });
  }).then(window => {
    t.equal(serializeDocument(window.document), "<html><head></head><body>Hello</body></html>");
    t.equal(window.location.href, "http://example.com/");
    t.equal(window.location.origin, "http://example.com");
    t.equal(window.location.search, "");
    t.equal(window.location.hash, "");
  });
});
```

Joris-van-der-Wel:
> Why Chai? These days I tend to prefer the built-in assert. But I'm happy to be convinced if you think there's a killer feature we should be using.

It has more assertion functions like `assert.isFunction(fn)`

The plugins that chai has could also be useful:

Promises:
```javascript
return assert.eventually.equal(Promise.resolve(2 + 2), 4, "This had better be true, eventually");
```

DOM:
```javascript
expect(document.querySelector('main article')).to.have.attr('foo').match(/bar/)
```

And if we come up with useful assertions for jsdom, we could easily release them as a chai plugin, instead of using a hypothetical `test/util.js"

I do not feel strongly towards chai though, whatever works is fine with me.

Joris-van-der-Wel:
> Given that we haven't changed all our test names to follow the describe/it pattern, maybe use describe/specify instead?

Mocha uses the BDD interface by default (describe & it), it also offers the TDD interface (suite & test). Perhaps TDD would be more appropriate? We can also alias the functions to whatever we would like to use.

domenic:
Regarding assertion library: those are fair arguments in favor of Chai; I'm happy to go with them. On the other hand, https://www.npmjs.com/package/power-assert looks pretty awesome....

Mocha allows `specify` as an alias of `it`, so I was just suggesting describe/specify.

domenic:
Hmmm, nevermind about power-assert; I didn't realize it requires transforms of the test code. Chai it is!

Joris-van-der-Wel:
Rebased & new build https://travis-ci.org/Joris-van-der-Wel/jsdom/builds/107894787

domenic:
This looks good to me! It looks like the Travis failures might just be timeouts? The failures are on WPT which should not be an issue. I restarted the build to see.

Do you want to update CONTRIBUTING.md, or shall I? I can do it when I merge (probably the weekend).

Amazing stuff, really appreciated.

Joris-van-der-Wel:
Thanks!,

Yea the travis build is suffering from timeouts a lot. Either saucelabs has no browsers available, or WPT timeouts somehow. Nice thing is that karma is able to retry starting the browser, it tries 3 times by default so that should help a bit with the suacelabs issue.

I can take care of CONTRIBUTION.md 

btw, I just got test fixtures (html files, etc) to work in browers, will commit tommorow:

```javascript
specify("should set namespace-related properties in HTML-SVG documents created by jsdom.env()", () => {
    return readTestFixture("jsdom/files/ns-svg.html").then(content => {
      const doc = jsdom.jsdom(content);
      testDocumentWithSVG(doc);
    });
  });
```
We could go even further if we use karma's `proxy` option (karma's http server will proxy a path to a different server).

domenic:
> btw, I just got test fixtures (html files, etc) to work in browers, will commit tommorow:

Oooh yesss that is super-great news. I was looking at all those tests that are skipped just because they use fixtures and being sad.

domenic:
I'm going to work on the tweaks here and merge this now :)

domenic:
This is merged!!!

One thing I noticed while updating Contributing.md is that it would be really nice if we could move the web platform tests to mocha. Then we can kind of lump all the nodeunit stuff as "these are old tests, run them only to check that you haven't broken everything".

domenic:
Oh, and figuring out how to get the web platform tests running in the browser would be... amazing.

PR #1377, opened 2016-02-06
Added referrer field to iframe

mfrye:
I'm not sure this is the best way to implement this. Does this look okay?


domenic:
This is pretty good! Just needs a test.

mfrye:
What is the best dir to put that? I see that the code is kind of in flux.

domenic:
The spec step where this is done is https://html.spec.whatwg.org/#creating-a-new-browsing-context so I think a new directory test/web-platform-tests/to-upstream/browsers/windows/nested-browsing-contexts is the place, based on looking at https://github.com/w3c/web-platform-tests/tree/master/html/browsers/windows/nested-browsing-contexts

test/web-platform-tests/to-upstream/browsers/windows/nested-browsing-contexts/frame.js
```diff
@@ -0,0 +1,29 @@
+var path = require("path");
```
domenic: Oh, I didn't see that you had added a test.

You need to look at the nearby files to see how these tests are written. They are HTML pages, not JS files, and they aren't jsdom specific; they can be run in any browser.

PR #1384, opened 2016-02-16
Disable tests that now fail on Node 5.6.0

### LINKED ISSUE #1380: Web platform tests for XHR have started failing
Master was fine, I re-ran the build, now it is not fine: https://travis-ci.org/tmpvar/jsdom/jobs/107099285

I wonder if one of our dependencies changed.

domenic:
PRing this to see if Travis likes it:

This is due to a regression in Node's HTTP parser that makes it less capable than browsers: https://github.com/nodejs/http-parser/issues/281. Closes #1380. We probably won't be able to fix this until Node either fixes the regression or makes HTTP header parsing customizable.


Sebmaster:
:+1:

PR #1387, opened 2016-02-17
Fix #1382: this inside event listener object handleEvent was wrong

### LINKED ISSUE #1382: Event handler objects are not `this` in handleEvent.
The snippet `addEventListener('click', { handleEvent() { console.log(this) } }, false)` will log the object passed in all major browsers, but logs the window object in jsdom.

https://dom.spec.whatwg.org/#concept-event-listener-invoke doesn't seem to necessarily require the current browser behavior, but it would nice to be consistent.

thetalecrafter:
Leave `callback` a reference to the object, and only check for `handleEvent` method at invoke time.

### DIFF
--- a/lib/jsdom/living/events/EventTarget-impl.js
+++ b/lib/jsdom/living/events/EventTarget-impl.js
@@ -16,9 +16,7 @@ class EventTargetImpl {
     // webidl2js currently can't handle neither optional arguments nor callback interfaces
     if (callback === undefined || callback === null) {
       callback = null;
-    } else if (typeof callback === "object") {
-      callback = callback.handleEvent;
-    } else if (typeof callback !== "function") {
+    } else if (typeof callback !== "object" && typeof callback !== "function") {
       throw new TypeError("Only undefined, null, an object, or a function are allowed for the callback parameter");
     }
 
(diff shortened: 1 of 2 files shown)


domenic:
> and only check for handleEvent method at invoke time.

That doesn't sound quite right. Can you add the following two tests, and make sure they pass? (They're in a jsbin so I can be sure they pass in browsers. Which, well, only Chrome supports this level of ES6; so embarassing.)

http://jsbin.com/nuyavoroqe/1/edit?html,output

thetalecrafter:
The tests pass for me, would you like me to add them to `EventTarget-this-of-listener.html` or to a new test file?

domenic:
Same file works!

Would you mind trying another test? Change obj.handleEvent to a different function after addEventListener but before dispatchEvent, and see which one browsers call and which one your implementation calls.

thetalecrafter:
I tested that in Safari while working on this, and it calls whatever the current handleEvent property is. Changing it after addEventListener works. I'll make tests for that too.

domenic:
Great stuff, thank you very much! Will release 8.0.4 with this shortly.

PR #1390, opened 2016-02-19
Input selection revision

### LINKED ISSUE #1351: Support full set of selectable input types
JSDOM does not support the full list of selectable input types (according to WHATWG) and throws an exception if an unsupported element is selected. Please update the `_allowSelection` condition to include the full list of supported input types. 

List of selectable input types according to WHATWG:

* Text (type=text)
* Search (type=search)
* URL (type=url)
* Telephone (type=tel)
* E-mail (type=password)
* Password (type=password)
* Date (type=date)
* Month (type=month)
* Week (type=week)
* Time (type=time)
* Local Date and Time (type=datetime-local)
* Number (type=number)
* Color (type=color)
* File Upload (type=file)

/source https://html.spec.whatwg.org/multipage/forms.html#the-input-element

Besides the condition update I suggest to fail select calls on unsupported input types silently as propose by WHATWG:

> In the case of input elements, if the control has no text field, then the method must do nothing.

/source https://html.spec.whatwg.org/multipage/forms.html#dom-textarea/input-select

Many thanks in advance! :bow:

yaycmyk:
There's now a distinction between inputs that can be selected outright, versus specific textual inputs that allow for variable-length selection via the `selectionStart`, etc attributes and methods.

Fixes #1351; updated the wpt submodule SHA to include the new tests and removed the ones I added to wpt/to-upstream, as they are no longer valid.


yaycmyk:
@domenic When you get a chance - I ran my updated wp tests against the code and it's happy.

yaycmyk:
@domenic The tests are failing due to something off with jQuery. Do you have any suggestions?

domenic:
It looks like something in this commit broke something jQuery is depending on, probably during the startup phase so that it's not even available. I'd suggest hooking up the virtual console to those tests---ideally all tests should do this automatically somehow, but that's another battle. Hopefully it will output a helpful script error.

yaycmyk:
I fixed it locally but haven't pushed yet. Will ping you again later tonight!

yaycmyk:
Let's actually hold off on merging this until the wpt test changes are merged.

yaycmyk:
@domenic All settled now

domenic:
Awesome!! Will merge tonight-ish.

domenic:
Merged as 3cc731fa0c87823ba3f9e541d208db36cba2f88a; thank you very much!

PR #1399, opened 2016-02-28
Add click proxy behavior for label elements

yaycmyk:
Clicking on a label element should trigger activation on an associated element via the `for` attribute linking, or simple nesting (if the associated element meets the [labeling criteria](https://html.spec.whatwg.org/multipage/forms.html#the-label-element:the-label-element-10)).

I would also like to note that I tried very hard to get the `NodeIterator` API to work, but it doesn't seem to be hooked up, unfortunately.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLLabelElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLLabelElement-impl.js
@@ -1,9 +1,69 @@
 "use strict";
 
 const HTMLElementImpl = require("./HTMLElement-impl").implementation;
+const MouseEvent = require("../generated/MouseEvent");
 const closest = require("../helpers/traversal").closest;
+const domSymbolTree = require("../helpers/internal-constants").domSymbolTree;
+
+function isLabelable(node) {
+  /* labelable logic defined at: https://html.spec.whatwg.org/multipage/forms.html#category-label */
+  if (node.nodeType !== 1) { return false; }
+
+  switch (node.tagName) {
+  case "BUTTON":
+  case "KEYGEN":
+  case "METER":
+  case "OUTPUT":
+  case "PROGRESS":
+  case "SELECT":
+  case "TEXTAREA":
+    return true;
+
+  case "INPUT":
+    return node.type !== "hidden" ? true : false;
+  }
+
+  return false;
+}
+
+function sendClickToAssociatedNode(node) {
+  node.dispatchEvent(
+    MouseEvent.createImpl(["click", {
+      bubbles: true,
+      cancelable: true,
+      view: node.ownerDocument ? node.ownerDocument.defaultView : null,
+      screenX: 0,
+      screenY: 0,
+      clientX: 0,
+      clientY: 0,
+      button: 0,
+      detail: 1,
+      relatedTarget: null
+    }])
+  );
+}
 
 class HTMLLabelElementImpl extends HTMLElementImpl {
+  constructor(args, privateData) {
+    super(args, privateData);
+
+    this._eventDefaults.click = () => {
+      if (this.hasAttribute("for")) {
+        const node = this.ownerDocument.getElementById(this.getAttribute("for"));
+        if (node && isLabelable(node)) {
+          sendClickToAssociatedNode(node);
+        }
+      } else {
+        for (const descendant of domSymbolTree.treeIterator(this)) {
+          if (isLabelable(descendant)) {
+            sendClickToAssociatedNode(descendant);
+            break;
+          }
+        }
+      }
+    };
+  }
+
   get form() {
     return closest(this, "form");
   }
--- a/test/web-platform-tests/to-upstream/html/semantics/forms/the-label-element/proxy-click-to-associated-element.html
+++ b/test/web-platform-tests/to-upstream/html/semantics/forms/the-label-element/proxy-click-to-associated-element.html
@@ -1,0 +1,50 @@
+<!DOCTYPE HTML>
+<title>label element click proxying via "for" attribute or nested labelable element</title>
+<link rel="author" title="yaycmyk" href="mailto:evan@yaycmyk.com">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/forms.html#the-label-element:the-label-element-10">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<div id="log"></div>
+<form id="test">
+    <input id="foo" type="checkbox" />
+    <label id="foo-label" for="foo">foo</label>
+
+    <label id="bar-label">
+        <input id="bar" type="checkbox" /> bar
+        <input id="baz" type="checkbox" /> baz
+    </label>
+</form>
+<script>
+  "use strict";
+
+  async_test(t => {
+    const label = document.getElementById('foo-label');
+    const input = document.getElementById('foo');
+
+    input.addEventListener("click", () => {
+      t.step(() => {
+        assert_true(true);
+        t.done();
+      });
+    });
+
+    label.click();
+
+  }, "label with for attribute should proxy click events to the associated element");
+
+  async_test(t => {
+    const label = document.getElementById('bar-label');
+    const input = document.getElementById('bar');
+
+    input.addEventListener("click", () => {
+      t.step(() => {
+        assert_true(true);
+        t.done();
+      });
+    });
+
+    label.click();
+
+  }, "label without for attribute should proxy click events to the first labelable child");
+
+</script>


lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -4,9 +4,62 @@ const HTMLElementImpl = require("./HTMLElement-impl").implementation;
...
   }
+
+  _handleClick(event) {
+    if (this.hasAttribute('for')) {
+      const node = this.ownerDocument.getElementById(this.getAttribute('for'));
+      if (node && this._isLabelable(node)) {
+        this._dispatchClickEvent(node);
+      }
+    } else if (this.children.length) {
+      const kids = Array.prototype.slice.call(this.querySelectorAll('*'));
```
domenic: Can you use domSymbolTree instead? E.g.

```js
} else {
  for (const descendant of domSymbolTree.treeIterator(this)) {
    if (this._isLabelable(descendant)) {
      this._dispatchClickEvent(descendant);
      break;
    }
  }
}
```

This will avoid query-selector, plus creating an array and then mutating it a bunch.

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -4,9 +4,62 @@ const HTMLElementImpl = require("./HTMLElement-impl").implementation;
...
+          this._dispatchClickEvent(kids[0]);
+          break;
+        }
+
+        kids.shift();
+      }
+    }
+  }
+
+  _isLabelable(node) {
```
domenic: This should be a function, outside the class body, not a method

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -4,9 +4,62 @@ const HTMLElementImpl = require("./HTMLElement-impl").implementation;
...
+      return true;
+
+    case 'INPUT':
+      return node.type !== 'hidden' ? true : false;
+    }
+
+    return false;
+  }
+
+  _dispatchClickEvent(node) {
```
domenic: `require` and use the `MouseEvent` constructor. Then this can be a function outside the class body.

domenic: Also maybe name it "fire a click event" per https://html.spec.whatwg.org/#fire-a-click-event, and make sure that https://html.spec.whatwg.org/#fire-a-synthetic-mouse-event matches the details for the mouse event.

test/web-platform-tests/to-upstream/html/semantics/forms/the-label-element/proxy-click-to-associated-element.html
```diff
@@ -0,0 +1,50 @@
...
+    const input = document.getElementById('foo');
+
+    input.addEventListener("click", () => {
+      t.step(() => {
+        assert_true(true);
+        t.done();
+      });
+    });
+
+    label.click();
```
domenic: Have you tested that this test actually passes in browsers? It's not obvious to me that programmatic `click()` works the same as user-initiated click.

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -4,9 +4,62 @@ const HTMLElementImpl = require("./HTMLElement-impl").implementation;
 const closest = require("../helpers/traversal").closest;
 
 class HTMLLabelElementImpl extends HTMLElementImpl {
+  constructor(args, privateData) {
+    super(args, privateData);
+
+    this._eventDefaults.click = this._handleClick.bind(this);
```
domenic: Use an arrow function in the constructor instead of creating a method and then binding it.

Joris-van-der-Wel:
> I would also like to note that I tried very hard to get the NodeIterator API to work, but it doesn't seem to be hooked up, unfortunately.

Here are some examples: https://github.com/tmpvar/jsdom/blob/master/test/living-dom/node-iterator.js

yaycmyk:
It's not actually in the ownerDocument. The functionality needs to be ported and added to the DocumentImpl. I know the code is in that module, but it's functionally inaccessible.

Joris-van-der-Wel:
createNodeIterator works for me, can you post a small snippit that demonstrates how it should work in other browsers? 

Or are you talking about using createNodeIterator internally? In that case it might be better to use: https://github.com/jsdom/js-symbol-tree/blob/master/api.md#module_symbol-tree--SymbolTree+treeIterator. createNodeIterator is a bit heavy weight.

yaycmyk:
Yeah, I was trying to do this.ownerDocument.createNodeInterator(). It works in Chrome. 

Sent via mobile. Please excuse brevity / typos.

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
yaycmyk: I had no idea that existed - sure!

test/web-platform-tests/to-upstream/html/semantics/forms/the-label-element/proxy-click-to-associated-element.html
yaycmyk: @domenic Yes I did in Chrome via JSBIN.

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -1,12 +1,74 @@
...
+  case 'INPUT':
+    return node.type !== 'hidden' ? true : false;
+  }
+
+  return false;
+}
+
+function sendClickToAssociatedNode(node) {
+  node.dispatchEvent(
+    MouseEvent.createImpl(['click', {
```
domenic: This should be `new MouseEvent` where `MouseEvent` is `require("../generated/MouseEvent").interface`, I think. Right, @Sebmaster?

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -1,12 +1,74 @@
...
+      relatedTarget: null
+    }])
+  );
+}
 
 class HTMLLabelElementImpl extends HTMLElementImpl {
+  constructor(args, privateData) {
+    super(args, privateData);
+
+    this._eventDefaults.click = () => this._handleClick();
```
domenic: Just inline the logic here. No need for the class method.

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -1,12 +1,74 @@
...
     return closest(this, "form");
   }
+
+  _handleClick() {
+    if (this.hasAttribute('for')) {
+      const node = this.ownerDocument.getElementById(this.getAttribute('for'));
+      if (node && isLabelable(node)) {
+        sendClickToAssociatedNode(node);
+      }
+    } else if (this.children.length) {
```
domenic: No need for this `if` at all. Just let the loop loop zero times.

domenic: No need for this `if` at all. Just let the loop loop zero times.

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
Sebmaster: No, fine as-is. You only pass impls in internal calls.

domenic:
Needs to not break the build.

test/web-platform-tests/to-upstream/html/semantics/forms/the-label-element/proxy-click-to-associated-element.html
```diff
@@ -0,0 +1,50 @@
...
+
+  }, "label with for attribute should proxy click events to the associated element");
+
+  async_test(t => {
+    const label = document.getElementById('bar-label');
+    const input = document.getElementById('bar');
+
+    input.addEventListener("click", () => {
```
domenic: I have been informed that the best way to do this is `input.addEventListener("click", t.step_func_done())`. step_func_done is not documented, which is kind of annoying, but I'll try to get that fixed.

yaycmyk:
@domenic Should be all good now, thanks! Had some linting errors. (and took care of the review comments)

domenic:
Awesome! LGTM, will merge tonight-ish or on next jsdom release day.

domenic:
Hmm the new test you added is now failing on Travis. Did my `step_func_done` suggestion break it?

yaycmyk:
Yes unfortunately. I'm going to revert it.

yaycmyk:
@domenic Happy and green.

Issue #1398, opened 2016-02-28
ajax calls which return response with charset=utf-16 not parsed correctly


alvarorahul:
Digging a bit into this, it turns out that in xmlhttprequest.js, responseBuffer.toString() which seems to be causing the issue

If I try passing in responseBuffer.toString("ucs2") things start working

Sebmaster:
We don't do any encoding stuff at all yet. Byte streams are always assumed to be utf8. See also https://github.com/tmpvar/jsdom/issues/1364#issuecomment-173748986.

alvarorahul:
I took a stab at addressing this in the following pull request #1400 
The fix is limited to the response buffers returned in XmlHttpRequest, If you think this is ok, I can add tests for it.

PR #1400, opened 2016-02-29
ajax calls which return response with charset=utf-16 not parsed correctly

alvarorahul:


alvarorahul:
#1398

domenic:
Needs tests and needs to not break the build

domenic:
Should probably be general over any encoding, not just utf-16.

lib/jsdom/living/xmlhttprequest.js
```diff
@@ -590,6 +594,30 @@ module.exports = function createXMLHttpRequest(window) {
...
+      });
+    }
+
+    return result;
+  }
+
+  function bufferToString(buffer, contentType) {
+    const parsed = parseContentType(contentType);
+    let encoding;
+    if (parsed.params["charset"] === "utf-16") {
```
domenic: Needs to compare lowercased

lib/jsdom/living/xmlhttprequest.js
```diff
@@ -590,6 +594,30 @@ module.exports = function createXMLHttpRequest(window) {
     DONE: 4
   });
 
+  function parseContentType(contentType) {
+    const result = { type: "", params: { } };
+
+    if (contentType) {
+      const parts = contentType.split(";");
+      result.type = parts[0];
+      parts.slice(1).forEach(part => {
```
domenic: Just use a `for (let i = 1; ...)` loop; no need to create a new array just to use `.forEach` on it.

lib/jsdom/living/xmlhttprequest.js
```diff
@@ -590,6 +594,30 @@ module.exports = function createXMLHttpRequest(window) {
...
+        const kvp = part.split("=");
+        result.params[kvp[0].trim()] = kvp[1].trim();
```
domenic: Need to remove quotes from potentially-quoted values (e.g. `Content-Type: text/html; charset="utf-8"`)

domenic: Needs to handle cases where people to invalid things like `Content-Type: text/html; charset; foo; baz=bar=qux`.

lib/jsdom/living/xmlhttprequest.js
```diff
@@ -590,6 +594,30 @@ module.exports = function createXMLHttpRequest(window) {
     DONE: 4
   });
 
+  function parseContentType(contentType) {
```
domenic: Actually, this would all be a lot easier if maybe we just used https://www.npmjs.com/package/contenttype :))

alvarorahul: agreed... will switch to use the contenttype lib

domenic: https://www.npmjs.com/package/con-type looks pretty good too; maybe it is better at handling invalid cases.

lib/jsdom/living/xmlhttprequest.js
alvarorahul: switched to use https://www.npmjs.com/package/contenttype

lib/jsdom/living/xmlhttprequest.js
alvarorahul: switching to https://www.npmjs.com/package/contenttype

lib/jsdom/living/xmlhttprequest.js
alvarorahul: switching to https://www.npmjs.com/package/contenttype

lib/jsdom/living/xmlhttprequest.js
alvarorahul: how about this one https://www.npmjs.com/package/content-type?

Sebmaster:
> Should probably be general over any encoding, not just utf-16.

Is there a more recent specification than the [IANA Charset spec](http://www.iana.org/assignments/character-sets/character-sets.xhtml) on which charsets exist / how to parse them?

I don't think the current switch case with hex/binary/etc. is a good thing, it probably produces a lot of invalid results.

Sebmaster:
Found it, we should strive to support these encodings and aliases: https://encoding.spec.whatwg.org/#names-and-labels. I guess we'll need to add iconv sooner or later, but we should at least support all the aliases for the encodings we can do with node's internal encodings.

PR #1406, opened 2016-03-01
Don't interpret script tags when type is not in spec ECMA262

### LINKED ISSUE #1402: application/json throws exception in virtualConsole
Let's say you've got this page:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title></title>
  </head>
  <body>
    <script type="application/json" id="json">
      { "test": 42 }
    </script>
    <script type="text/javascript">
      console.log('hi');
    </script>
  </body>
</html>
```

and fetch it with:

```js

const virtualConsole = jsdom.createVirtualConsole();
virtualConsole.sendTo(console);

jsdom.env({
  url: 'http://localhost:8000/',
  features: {
    FetchExternalResources: ['script'],
    ProcessExternalResources: ['script'],
  },
  virtualConsole: virtualConsole,
  done: (err, window) => {
    if (err) throw err;
    const json = JSON.parse(window.document.querySelector('#json').innerHTML);
    console.log('json:', json);
  }
})
```

you'll get the following output:

```
Error: Uncaught [SyntaxError: Unexpected token :]
    at reportException (/Users/Jasper/Desktop/node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:56:24)
    at processJavaScript (/Users/Jasper/Desktop/node_modules/jsdom/lib/jsdom/living/nodes/HTMLScriptElement-impl.js:73:7)
    at HTMLScriptElementImpl._eval (/Users/Jasper/Desktop/node_modules/jsdom/lib/jsdom/living/nodes/HTMLScriptElement-impl.js:27:7)
    at /Users/Jasper/Desktop/node_modules/jsdom/lib/jsdom/browser/resource-loader.js:25:22
    at Object.check (/Users/Jasper/Desktop/node_modules/jsdom/lib/jsdom/living/nodes/Document-impl.js:83:11)
    at /Users/Jasper/Desktop/node_modules/jsdom/lib/jsdom/living/nodes/Document-impl.js:101:12
    at HTMLScriptElement.<anonymous> (/Users/Jasper/Desktop/node_modules/jsdom/lib/jsdom/living/nodes/HTMLScriptElement-impl.js:18:74)
    at invokeEventListeners (/Users/Jasper/Desktop/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:209:27)
    at HTMLScriptElementImpl._dispatch (/Users/Jasper/Desktop/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:129:9)
    at HTMLScriptElementImpl.dispatchEvent (/Users/Jasper/Desktop/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:91:17) [SyntaxError: Unexpected token :]
hi
json: { test: 42 }
```

I can't figure out exactly what causes this, but:

- it doesn't fail when the `features`-config is removed
- it doesn't fail when the `script`-element is replaced by a `div`
- It still fails when the code in `done` is removed (so it seems to be before this point)
- the rest of the page seems to function fine (as 'hi' is still logged)

Jpunt:
Fixes #1402.


domenic:
Thanks! The code looks great, but could you follow Contributing.md and submit a web platform test instead of a jsdom-specific test?

Jpunt:
Damn, shouldn't have skipped the web-platforms-part that fast.. Will take a look at it!

Jpunt:
I'm afraid I need some help with these.. I've found the appropriate web-platform-tests:

[script-language-type.html](https://github.com/w3c/web-platform-tests/blob/f685409553687752d593a435e56f87f4094c0fd9/html/semantics/scripting-1/the-script-element/script-language-type.html)
[script-languages-01.html](https://github.com/w3c/web-platform-tests/blob/f685409553687752d593a435e56f87f4094c0fd9/html/semantics/scripting-1/the-script-element/script-languages-01.html)
[script-languages-02.html](https://github.com/w3c/web-platform-tests/blob/f685409553687752d593a435e56f87f4094c0fd9/html/semantics/scripting-1/the-script-element/script-languages-02.html)

The first two pass, but the last one fails (even without my changes to the code):

```
AssertionError: Failed in "Script should run with type=""":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="application/ecmascript"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="application/javascript"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="application/x-ecmascript"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="application/x-javascript"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="APPLICATION/ECMASCRIPT"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="APPLICATION/JAVASCRIPT"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="APPLICATION/X-ECMASCRIPT"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="APPLICATION/X-JAVASCRIPT"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="application/ecmascript "":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="application/javascript "":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="application/x-ecmascript "":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="application/x-javascript "":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type=" application/ecmascript"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type=" application/javascript"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type=" application/x-ecmascript"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type=" application/x-javascript"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="application/ecmascript\t"":
assert_equals: expected true but got false
AssertionError: Failed in "Script should run with type="application/javascript\t"":
assert_equals: expected true but got false

... continues with all types defined in the test, in lowercase and uppercase
```

Any idea what we should do here?

Jpunt:
Also, the test validates the use of both the `language` and `type` attribute, while I believe `language` is deprecated. What's jsdom's approach to this? Should we check for language as well?

domenic:
Haven't had time to look into the test failures, but yes, we should definitely support language! Deprecated for authors doesn't mean browsers can stop supporting it :)

Jpunt:
Alright, in that case it's nice we've got these kind of tests to comply too :+1: 

Unfortunately, I'm having troubles with making those tests pass, which may be because I don't have a lot of experience on how browsers (or jsdom) work internally.. 

It seems to me that "real" browsers wait for a newly injected script-tag to be interpreted, before continuing running other code, while jsdom does not. To illustrate what I mean, I've made two examples:

https://gist.github.com/Jpunt/09dfa4d37f8d201d12d9
https://gist.github.com/Jpunt/01134873c76126f36159

The first one will give you an exception that `ran` doesn't exist (yet), while the second one will succeed. However, the first one will work when you run it in chrome for example.

This is the same thing as what happens in the web-platform-tests. A bunch of script-tags are injected and immediately asserted on their result, while `jsdom` did inject the element, but not yet interpreted it. 

How should we go about fixing this? I've got a hunch that changing the order of parsing/interpreting these elements isn't trivial, if possible at all?

domenic:
Ah, yeah, if it's due to something like that, we're in real trouble :(. I'll try to take a look myself later to confirm. But what I would do in that case is to create versions of those tests that don't trigger the core jsdom bug, but just test the type/language stuff, inside to-upstream. They'll need a HTML comment like `<!-- this duplicates ..., so probably don't upstream it; see https://github.com/tmpvar/jsdom/pull/1406 -->`.

Sebmaster:
> How should we go about fixing this? I've got a hunch that changing the order of parsing/interpreting these elements isn't trivial, if possible at all?

I'm rewriting the script tag for #1316 which should resolve that, but that's gonna take a while.

Jpunt:
Considering that the rewrite will take a while, I've made a custom "not so upstream" test and changed the implementation to comply to that in the meantime.

domenic:
(If you could rebase and squash this it'd be a bit easier to review)

This isn't quite correct. The logic isn't "is type or language in a set containing both types and languages". It's instead, per https://html.spec.whatwg.org/#prepare-a-script:

- Determine type string as follows:
  - If there's a type attribute with empty string value, return "text/javascript"
  - If there's no type attribute but there's a language attribute with empty string value, return "text/javascript"
  - If there's no type or language attribute, return "text/javascript"
  - Otherwise, if there's a type attribute, return it, with whitespace trimmed
  - Otherwise, if there's a langauge attribute, return "text/" + language attribute value

Then, you check type string vs. the list at https://html.spec.whatwg.org/#javascript-mime-type

PR #1410, opened 2016-03-03
add initial cors support && add strictSSL option && various fixes

nicolashenry:
Hello,

I took some time to add cors support.
I also added some missing tests and added an option called "strictSSL" for self-signed ssl certificates.

I uncommented :
- the "windows broken" tests because I tried them on Windows 7 without any problem (even if I had to disable my corporate proxy and to disable the "autocrlf" git config).
- the tests linked to https://github.com/nodejs/http-parser/issues/281 because the tests will pass on the next stable node release (2.7.x is ok).

I can comment these tests again if you want.


lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -215,7 +231,9 @@ class DocumentImpl extends NodeImpl {
       this._contentType = this._parsingMode === "xml" ? "application/xml" : "text/html";
     }
 
-    this._URL = whatwgURL.parseURL(privateData.options.url === undefined ? "about:blank" : privateData.options.url);
+    const urlOption = privateData.options.url === undefined ? "about:blank" : privateData.options.url;
+    this._URL = whatwgURL.parseURL(urlOption);
+    this[internalConstants.origin] = new whatwgURL.URL(urlOption).origin;
```
domenic: Since we now have this -impl.js vs. generated wrapper split, there's no need to use symbols for these. Normal `_origin` properties will work fine and won't be exposed to the user.

lib/jsdom/living/nodes/HTMLScriptElement-impl.js
```diff
@@ -11,13 +11,15 @@ const domSymbolTree = internalConstants.domSymbolTree;
...
-    this.addEventListener("DOMNodeInsertedIntoDocument", () => {
-      if (this.src) {
-        resourceLoader.load(this, this.src, this._eval);
-      } else {
-        resourceLoader.enqueue(this, this._ownerDocument.URL, this._eval)(null, this.text);
-      }
-    });
+  }
+
+  _attach() {
```
domenic: Nice! Does this fix something in particular, or is it just general cleanup? Always happy to remove internal mutation event uses.

lib/jsdom/living/xhr-sync-worker.js
```diff
@@ -4,6 +4,10 @@ const util = require("util");
 
 const jsdom = require("../../jsdom");
 const xhrSymbols = require("./xmlhttprequest-symbols");
+const tough = require("tough-cookie");
+
+const doc = jsdom.jsdom();
```
domenic: Why move these here?

test/web-platform-tests/index.js
```diff
@@ -228,48 +236,84 @@ const runWebPlatformTest = require("./run-web-platform-test")(exports, path.reso
...
   "XMLHttpRequest/open-url-base-inserted-after-open.htm",
+  // "XMLHttpRequest/open-url-bogus.htm", // todo
+  // "XMLHttpRequest/open-url-encoding.htm", // todo
   "XMLHttpRequest/open-url-fragment.htm",
   "XMLHttpRequest/open-url-javascript-window-2.htm",
   "XMLHttpRequest/open-url-javascript-window.htm",
+  "XMLHttpRequest/open-url-multi-window.htm",
+  "XMLHttpRequest/open-url-multi-window-2.htm",
+  "XMLHttpRequest/open-url-multi-window-3.htm",
+  // "XMLHttpRequest/open-url-multi-window-4.htm", // todo
```
domenic: In general if you could leave more informative comments on why a test fails in jsdom that would be nice. I see you do that in a few places. If it's too much trouble to debug I understand.

domenic:
This is really amazing, thank you so much (again)!I I only had minor nits in reviewing the code. It's a bit of a shame this is all in the XHR implementation, and eventually we'll need to move it into something following https://fetch.spec.whatwg.org/#fetching with XHR built on top, but that can be left for another day.

Regarding the tests, I will try to re-run on Windows using Node 5.7.1 and see what happens. Right now Travis is stuck on 5.6.0 so I don't think we should enable the failing tests there until we can get our build to pass (presumably by waiting for Travis to upgrade itself). If the tests still fail on my Windows then maybe we need to tweak .gitattributes somehow, either here or in web-platform-tests?

domenic:
OK, I ran the tests twice on Node 5.7.1 with your patch on Windows 10 and here is the only failure I am getting (both times):

```
AssertionError: Timout in "XmlHttpRequest: The send() method: Fire a progress event named error when Network error happens (synchronous flag is unset)":
Test timed out

null
    at Object.ok (C:\Users\Domenic\Dropbox\GitHub\tmpvar\jsdom\node_modules\nodeunit\lib\types.js:83:39)
    at C:\Users\Domenic\Dropbox\GitHub\tmpvar\jsdom\test\web-platform-tests\run-web-platform-test.js:40:15
    at http://web-platform.test:9000/resources/testharness.js:1867:21
    at forEach (http://web-platform.test:9000/resources/testharness.js:2521:26)
    at Tests.notify_result (http://web-platform.test:9000/resources/testharness.js:1864:9)
    at http://web-platform.test:9000/resources/testharness.js:1885:30
    at Array.forEach (native)
    at Tests.complete (http://web-platform.test:9000/resources/testharness.js:1881:20)
    at Tests.timeout (http://web-platform.test:9000/resources/testharness.js:1801:14)
    at http://web-platform.test:9000/resources/testharness.js:1792:55
    at callback (C:\Users\Domenic\Dropbox\GitHub\tmpvar\jsdom\lib\jsdom\browser\Window.js:434:19)
    at Timer.listOnTimeout (timers.js:92:15)

AssertionError: test harness should not timeout: XMLHttpRequest/send-network-error-async-events.sub.htm
    at Object.ok (C:\Users\Domenic\Dropbox\GitHub\tmpvar\jsdom\node_modules\nodeunit\lib\types.js:83:39)
    at C:\Users\Domenic\Dropbox\GitHub\tmpvar\jsdom\test\web-platform-tests\run-web-platform-test.js:49:13
    at http://web-platform.test:9000/resources/testharness.js:1903:22
    at forEach (http://web-platform.test:9000/resources/testharness.js:2521:26)
    at Tests.notify_complete (http://web-platform.test:9000/resources/testharness.js:1900:9)
    at Tests.complete (http://web-platform.test:9000/resources/testharness.js:1891:14)
    at Tests.timeout (http://web-platform.test:9000/resources/testharness.js:1801:14)
    at http://web-platform.test:9000/resources/testharness.js:1792:55
    at callback (C:\Users\Domenic\Dropbox\GitHub\tmpvar\jsdom\lib\jsdom\browser\Window.js:434:19)
    at Timer.listOnTimeout (timers.js:92:15)
```

If you have any ideas we can try to debug this otherwise we can comment it out with "doesn't work on Domenic's computer" :)

lib/jsdom/living/nodes/Document-impl.js
nicolashenry: Ok, I did not know that. I will change this.

lib/jsdom/living/nodes/HTMLScriptElement-impl.js
nicolashenry: This fixes open-url-multi-window* tests.

lib/jsdom/living/xhr-sync-worker.js
nicolashenry: To define these sooner but it does not change anything, I can put these back to the "end" callback if you prefer.

test/web-platform-tests/index.js
nicolashenry: I will try.

nicolashenry:
Maybe this is an issue with your network, you can try to put this into your hosts file :
```
0.0.0.0     nonexistent-origin.web-platform.test
```

domenic:
I think adding that line to my hosts file fixed the problem, so after addressing review comments this should be good to merge :)

nicolashenry:
I tried to use "con-type" module but it can't be used in every context. I didn't find any module which :
- is rfc7231 compliant
- is permissive on parameters
- can change parameter values without any other changes (for order, separators, duplicated/invalid values, ...)

For now, I created a parsing method in ```utils.js```, it can be used for #1400 :
```javascript
const contentType = parseContentType("text/html;charset=utf-16");
if (contentType) {
  console.log(contentType.get("charset")); // "utf-16"
}
```

domenic:
Merged as f5274e32d136d21ccc75c96974a21e89b5e654ac; thank you so much!!

lehni:
Is there a way to turn CORS off through the features object? I am using XMLHttpRequest it to load files locally in paper.js, and these examples stopped working since I switched to the newly released v8.3.0. I believe it's due to CORS now being supported.

lehni:
I found a way to circumvent this for now, but it seems very hackish. Wondering what the right way should be:

```js
var document = jsdom.jsdom('<html><body></body></html>', { url: 'file://', });
```

This gives the document the origin `file://`, which will match the origin in the XMLHttpRequest...

domenic:
Hmm. Browsers don't let you bypass CORS, so I am not sure we should have that in jsdom. On the other hand, the fact that we let you set an arbitrary URL does mean any security checks here are kind of moot. I'll have to think about this more...

crabmusket:
@domenic I'm about to run up against this CORS issue as well. I know I can sidestep it by setting the document URL, which is what I plan to do. I just wanted to say that having a way to disable CORS would be very helpful. jsdom may be emulating a browser, but it isn't a browser.

PR #1414, opened 2016-03-09
Implement frameElement for Window.

mickylad:
`frameElement` is currently `undefined ` for all `Window`s. This will set it to `null` for the top-level window, and as the `HTMLFrameElement` for attached frames.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -54,6 +54,7 @@ function Window(options) {
   // Set up the window as if it's a top level window.
   // If it's not, then references will be corrected by frame/iframe code.
   this._parent = this._top = this._globalProxy;
+  this._frameElement = null;
 
   // List options explicitly to be clear which are passed through
   this._document = Document.create([], {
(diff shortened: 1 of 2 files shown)


mickylad:
I suspect you'll want tests for this, there appear to be some relevant ones in [`web-platform-tests`](https://github.com/w3c/web-platform-tests/blob/master/html/browsers/windows/nested-browsing-contexts/frameElement.sub.html) but I'm struggling to work with them.

domenic:
It looks like to pass those tests you'll need to do an origin check. https://html.spec.whatwg.org/#dom-frameelement step 5, basically check the two documents' `._URL.origin` should do the trick probably.

domenic:
I took a look at those tests and actually I don't think we can pass them given jsdom's current set up. They rely on some crazy stuff where the object model actually looks totally different cross-origin. So I will pull this in without tests.

domenic:
Merged as 17996d39f955ca2e6ab72a0a6c192ed251815517; thank you!

PR #1415, opened 2016-03-10
Correct the logic as to uppercase if NOT xhtml AND html.

mjgp2:
Note that currently this breaks many, many tests.

From MDN:

In XML (and XML-based languages such as XHTML), tagName preserves case.
On HTML elements in DOM trees flagged as HTML documents, tagName returns
the element name in the uppercase form. The value of tagName is the same
as that of nodeName.

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -239,7 +239,7 @@ class ElementImpl extends NodeImpl {
   }
   get tagName() {
     let qualifiedName = this._prefix !== null ? this._prefix + ":" + this._localName : this._localName;
-    if (this.namespaceURI === "http://www.w3.org/1999/xhtml" && this._ownerDocument._parsingMode === "html") {
+    if (this.namespaceURI !== "http://www.w3.org/1999/xhtml" && this._ownerDocument._parsingMode === "html") {
       qualifiedName = qualifiedName.toUpperCase();
     }
     return qualifiedName;


mjgp2:
Hey @tmpvar , I came across this issue because using htmlparser2 means `this.namespaceURI` is always `null`, and hence `document.body` which depends upon the case of `tagName` being upper case breaks.

domenic:
This is not correct. MDN is not a spec, and I think you are reading it wrong anyway. Regardless, the spec is at https://dom.spec.whatwg.org/#dom-element-tagname and step 2 does what the current code does (which is the opposite of what your change does). That is why you are breaking all the tests.

mjgp2:
OK @domenic thanks for pointing me to that. I guess the confusion comes that the HTML namespace is `http://www.w3.org/1999/xhtml` which is somewhat confusing :)

The bug here then would seem to be that when using the htmlparser2 parser the namespace is not being set, i.e. it is `null`. This then causes the `tagName` to be lowercase, which breaks all the methods that depend on the value being upper case such as `document.body`.

domenic:
Well, in general we only support parse5, because of issues exactly like this; we are working on removing htmlparser2 support entirely for the next release, so in the meantime I guess it's deprecated.

mjgp2:
OK, thanks for the info. Is there a ticket for that? The performance of parser5 seems to be pretty poor, I guess if pluggable parsers are being removed I would be keen to understand if there's another extension mechanism.

domenic:
#1316. In general we find parsing performance to be the least important part of jsdom performance, and the correctness improvements are drastic (just look at the massive number of issues closed by introducing parse5: https://github.com/tmpvar/jsdom/issues?q=is%3Aissue+label%3Aparsing+is%3Aclosed). But parse5 always accepts pull requests.

The alternative in the future will be for other parsers to create shim interfaces that behave exactly like parse5's, so they can pretend to be parse5. That's a ways off though.

inikulin:
@mjgp2 

> The performance of parser5 seems to be pretty poor,

Nah, it's just fine (should be compared *somehow* with `htmlparser2-dom`):
```
gumbo-parser   : 34.9208 ms/file ± 21.4238
html-parser    : 24.8224 ms/file ± 15.8703
html5          : 419.597 ms/file ± 264.265
htmlparser     : 60.0722 ms/file ± 384.844
htmlparser2-dom: 12.0749 ms/file ± 6.49474
htmlparser2    : 7.49130 ms/file ± 5.74368
hubbub         : 30.4980 ms/file ± 16.4682
libxmljs       : 14.1338 ms/file ± 18.6541
parse5         : 22.0439 ms/file ± 15.3743
sax            : 49.6513 ms/file ± 26.6032
```

mjgp2:
@inikulin thanks for this, I think I was conflating html5 vs parse5 :)

PR #1417, opened 2016-03-11
Fix undefined location access in Window.startTimer

vacero:
I have encountered some instances where if an error occurs on a timer callback after the window has been closed (from the `done` callback of `jsdom.env`), accessing the `href` property of `window.location` throws an error because location is undefined.

The thing is that stopAllTimers is called, but this callback is still being called. You can test this with `http://www.github.com`

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -421,11 +421,13 @@ function matchesDontThrow(el, selector) {
 }
 
 function startTimer(window, startFn, stopFn, timerId, callback, ms, args) {
-  const sourceURL = window.location.href;
+  if (!window.document) {
+    return;
+  }
 
   if (typeof callback !== "function") {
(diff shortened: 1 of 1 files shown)


Sebmaster:
> The thing is that stopAllTimers is called, but this callback is still being called. You can test this with http://www.github.com

This sounds like the actual bug here. If we can repro this, we should fix that too.

domenic:
Yes. In any case, nothing will be accepted without tests.

vacero:
You are right, I just found out that the problem is the timer being set after the document is already deleted (probably from a callback of an ajax call). So the fix was not really solving anything.

I think the real solution would be checking for `if (!window.document)` at the beginning of `startTimer` to avoid setting new timers.

I'll try to add tests for this later, but I'm not used with the testing structure so it'll probably take me some time.

PR #1421, opened 2016-03-17
Run web-platform-tests using mocha

Joris-van-der-Wel:
fs-readdir-recursive is no longer used for the to-upstream tests (instead a list is used), this is so that we can run them more easily in the browser later on.

I've not yet added browser support. I have a good idea about how I want to get about this, however it too much work for now. So I will revisit that after I've completed converting all of the old tests to mocha.


test/index.js
```diff
@@ -14,3 +14,6 @@ require("./jsdom/serialization.js");
 require("./jsdom/utils.js");
 require("./jsdom/virtual-console.js");
 require("./jsdom/xml.js");
+
+// keep this one last so that you can cancel the test run early
+require("./web-platform-tests/index.js");
```
domenic: Maybe we shouldn't even mix these up together; what do you think?

test/web-platform-tests/run-web-platform-test.js
```diff
@@ -1,73 +1,95 @@
...
-      }
-    },
-    created(err, window) {
-      if (err) {
-        t.ifError(err, "window should be created without error");
-        t.done();
-        return;
-      }
+  let created;
+  const loaded = nodeResolverPromise(loadedResolver => {
```
domenic: Can't wait until we overhaul the API and make promises supported by default...

test/web-platform-tests/run-web-platform-test.js
```diff
@@ -1,73 +1,95 @@
...
       };
-    },
-
-    loaded(err) {
-      t.ifError(err);
-    }
-  });
+    });
+  })
+  .then(loaded); // make sure that the loaded callback does not error
```
domenic: I think this is supposed to be `.then(() => loaded)`

test/web-platform-tests/run-web-platform-test.js
```diff
@@ -1,73 +1,95 @@
...
 }
 
 const childProcess = require("child_process");
```
domenic: While you're here, move these up to the top?

test/web-platform-tests/run-to-upstream-web-platform-test.js
```diff
@@ -4,64 +4,88 @@ const http = require("http");
 const path = require("path");
```
domenic: Sigh, so much duplicated code here between the two files. No need to fix it in this PR though.

test/web-platform-tests/run-to-upstream-web-platform-test.js
```diff
@@ -4,64 +4,88 @@ const http = require("http");
...
       };
-    },
-
-    loaded(err) {
-      t.ifError(err);
-    }
-  });
+    });
+  })
+  .then(loaded); // make sure that the loaded callback does not error
```
domenic: `.then(() => loaded)`

test/web-platform-tests/run-to-upstream-web-platform-test.js
```diff
@@ -4,64 +4,88 @@ const http = require("http");
...
 const fs = require("fs");
 const st = require("st");
+const specify = require("mocha-sugar-free").specify;
+
+const nodeResolverPromise = require("../util").nodeResolverPromise;
+const inBrowserContext = require("../util").inBrowserContext;
 const jsdom = require("../..");
 
-function createJsdom(urlPrefix, testPath, t) {
+const globalPool = { maxSockets: 6 };
```
domenic: What's the motivation for this? Maybe comment it?

test/web-platform-tests/to-upstream.js
```diff
@@ -1,8 +1,37 @@
 "use strict";
 const path = require("path");
-const readdirRecursive = require("fs-readdir-recursive");
```
domenic: Can we remove this devDependency from package.json now?

Kind of a shame to lose the ease of use here, but maybe in the future I'll move toward pretest manifest generation.

test/web-platform-tests/run-web-platform-test.js
Joris-van-der-Wel: yup

test/web-platform-tests/run-web-platform-test.js
Joris-van-der-Wel: yea, that would clean up a lot of test cases.

test/web-platform-tests/run-to-upstream-web-platform-test.js
Joris-van-der-Wel: Not sure, it is there because I copied over the code from `run-web-platform-test.js` for consistencies sake. I will make sure this part is reused between `run-to-upstream-web-platform-test.js` and `run-web-platform-test.js`

test/web-platform-tests/run-web-platform-test.js
Joris-van-der-Wel: Also this callback does not even exist anymore in jsdom, the proper one is "onload". There is also a small issue in the readme that i'll fix.

test/web-platform-tests/to-upstream.js
Joris-van-der-Wel: It is still used by `scripts/webidl/convert.js`

I had to remove it here so that we can get a proper "skipped" notice in karma. I agree generating the manifest is probably easier than trying to do this dynamically while running the tests.

Joris-van-der-Wel:
updated

domenic:
<3 thank you!!

PR #1427, opened 2016-03-27
Support element.remove()

Sebmaster:
http://devdocs.io/dom/childnode/remove


Sebmaster:
@domenic Attached a PR for it.

PR #1429, opened 2016-03-28
Visibility api

### LINKED ISSUE #1405: thoughts about window.navigator.noUI and page visibility api
Hi,
this is a follow-up of https://github.com/tmpvar/jsdom/issues/642.

The [Page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) looks like a way for client code to know if it's running in a "half browser" or not:
`document.visibilityState == "prerender"`

I like that because:
* it does not define a non standard property
* for some use cases (isomorphic server rendering) the name is natural
* the check is harmless, browsers not supporting the api return undefined
* it might not even be up to jsdom to set that value - userland can set it to prerender
in its server rendering code if needed.

kapouer:
Closes #1405.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -341,7 +341,7 @@ function Window(options) {
       get appVersion() {
         return process.version;
       },
-      noUI: true,
+      noUI: true, // TODO MAJOR https://github.com/tmpvar/jsdom/issues/1405
       get cookieEnabled() {
         return true;
       }
--- a/lib/jsdom/living/nodes/Document.idl
+++ b/lib/jsdom/living/nodes/Document.idl
@@ -115,3 +115,12 @@ partial interface Document {
 //  readonly attribute FrozenArray<DOMString> styleSheetSets;
 //  void enableStyleSheetsForSet(DOMString? name);
 };
+
+// http://w3c.github.io/page-visibility/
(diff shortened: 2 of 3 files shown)


lib/jsdom/browser/Window.js
```diff
@@ -341,7 +341,11 @@ function Window(options) {
       get appVersion() {
         return process.version;
       },
-      noUI: true,
+      get noUI() {
```
domenic: As discussed, let's just comment this but not spew console warnings

lib/jsdom/living/nodes/Document.idl
```diff
@@ -115,3 +115,10 @@ partial interface Document {
 //  readonly attribute FrozenArray<DOMString> styleSheetSets;
 //  void enableStyleSheetsForSet(DOMString? name);
 };
+
+enum VisibilityState { "hidden", "visible", "prerender", "unloaded" };
```
domenic: Maybe add a spec link?

kapouer:
Updated commits.
Note that i found the correct "living spec", this explains the changes in the IDL.

kapouer:
Forgot to say "please" to lint :)

domenic:
Thanks! Merged as 0e528323309b60b59e00f25baffdd326ff666b08 and 38484b4ba761b6919f76941d867e0ef0ff7c02bd.

PR #1432, opened 2016-03-31
add proxy config doc

gsouf:
Proxy was not documented


domenic:
Thanks; merged in slightly different form as 5cc248661aa327cdbc888216372e787f5e69a944.

PR #1439, opened 2016-04-06
Clarify that document.origin is now set/evaluated based on config.url.

eloquence:
This bit me in a package upgrade, hence making a quick note in case others encounter the same issue :)


domenic:
Thanks! Fixed with slightly different wording in 96908118af7b26b1b4b1f964b96db6ce0302800d.

PR #1440, opened 2016-04-06
add encoding handling

nicolashenry:
Hello,

I tried to implement character encoding using the iconv-lite module.
The encoding can come from :
- the content-type header
- the bom
- the charset attribute (for script/link tags)
- the charset from meta tag (for document/iframe)

The "@charset" CSS property is still ignored for the time being.


lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -227,6 +227,8 @@ class DocumentImpl extends NodeImpl {
     }
 
     this._contentType = privateData.options.contentType;
+    this._encoding = privateData.options.encoding;
```
domenic: I would like to normalize things before we reach the Document level. Per https://dom.spec.whatwg.org/#concept-document-encoding each document has an encoding, and not a characterSet. Can we make that true, and get rid of documentImpl._characterSet?

domenic: This may require implementing a copy of https://encoding.spec.whatwg.org/#names-and-labels or something similar, that maps all the possible encoding labels (which I think is what you call characterSet) to iconv-lite-compatible strings.

lib/jsdom/browser/resource-loader.js
```diff
@@ -170,7 +200,7 @@ exports.download = function (url, options, cookieJar, referrer, callback) {
   };
 };
 
-exports.load = function (element, url, callback, encoding) {
+exports.load = function (element, url, callback, encoding, detectMetaCharset) {
```
domenic: Can you use an options object instead of a boolean parameter? See http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html.

If you'd like to reorganize this while you're here, I think the best signatures would be

- download(url, options, callback) with options containing all the things it contains now plus cookieJar, referrer, and detectMetacharset
- load(element, url, options, callback) with options containing encoding and detectMetaCharset

lib/jsdom/utils.js
```diff
@@ -336,6 +348,246 @@ exports.parseContentType = function parseContentType(contentType) {
   return null;
 };
 
+exports.decodeString = function decodeString(buffer, contentType, defaultEncoding, detectMetaCharset) {
```
domenic: Wow, it is amazing you implemented all this! Thank you!!

Can you move this and its helpers to a separate file, e.g. lib/jsdom/living/helpers/encoding.js? And use an options object for detectMetaCharset?

test/jsdom/encoding.js
```diff
@@ -0,0 +1,464 @@
...
+    });
+  });
+
+  specify("UTF-16LE", { async: true }, t => {
+    jsdom.env({
+      url: testHost + "/utf16le",
+      done(err, window) {
+        assert.ifError(err);
+
+        assert.equal(window.document.characterSet.toUpperCase(), "UTF-16LE", "document.characterSet");
```
domenic: Don't test with toUpperCase(). Browsers are supposed to return the canonical name, which I believe is uppercase. We should match them, and have strict tests to prove it.

test/jsdom/misc.js
```diff
@@ -115,7 +115,7 @@ describe("jsdom/miscellaneous", () => {
       process.nextTick(() => {
         req.emit("response", res);
         process.nextTick(() => {
-          res.emit("data", "window.attachedHere = 123");
+          res.emit("data", new Buffer("window.attachedHere = 123"));
```
domenic: Why this change?

test/jsdom/encoding.js
```diff
@@ -0,0 +1,464 @@
...
+  let testHost;
+
+  before(() => {
+    return new Promise((done, rej) => {
+      portfinder.getPort((err, port) => {
+        if (err) {
+          return rej(err);
+        }
+
+        server = http.createServer((req, res) => {
```
domenic: This is a big ask of you, but could you possibly turn these into to-upstream web-platform-tests, instead of Node specific tests?

For the tests which don't have any custom headers, it shouldn't be too hard: just manually create the right html files.

For those with custom headers, I think it would not be easy because we don't use the Python runner to run those yet, so we can leave those in Node-specific form for now.

README.md
```diff
@@ -138,6 +138,7 @@ jsdom.env(config);
 - `config.pool`: an object describing which agents to use for the requests; defaults to `{ maxSockets: 6 }`, see [request module](https://github.com/request/request#requestoptions-callback) for more details.
 - `config.agentOptions`: the agent options; defaults to `{ keepAlive: true, keepAliveMsecs: 115000 }`, see [http api](https://nodejs.org/api/http.html) for more details.
 - `config.strictSSL`: if `true`, requires SSL certificates be valid; defaults to `true`, see [request module](https://github.com/request/request#requestoptions-callback) for more details.
+- `config.encoding`: the default encoding; defaults to `UTF-8`
```
domenic: We should specify that this is an _encoding label_ from https://encoding.spec.whatwg.org/#names-and-labels, and make sure we normalize all labels to their corresponding encoding using some sort of lookup table.

We should also list somewhere (maybe in the readme, or maybe in a wiki page if it's too long) which encodings we do and don't support from that list.

lib/jsdom/living/nodes/Document-impl.js
nicolashenry: I used "encoding" as the default encoding used for this document and others sub-documents and "characterSet" for the effective encoding. The names are maybe wrong, I can rename "encoding" as "defaultEncoding" and "characterSet" as "encoding".

domenic: Hmm, so defaultEncoding is the user-supplied encoding?

Maybe we shouldn't allow user-supplied encodings at all. Browsers generally do not, but instead follow the spec algorithms for sniffing.

nicolashenry: The default encoding should depend on language (https://html.spec.whatwg.org/#determining-the-character-encoding:concept-encoding-confidence-8) but I didn't want to add it now. I can remove the parameter but I will have to change some tests.

domenic: Ah, I see. Yes, I think it would be good to remove it for now.

lib/jsdom/browser/resource-loader.js
nicolashenry: I will try if I have some time.

lib/jsdom/utils.js
nicolashenry: Ok, I will do that.

test/jsdom/encoding.js
nicolashenry: I will try to normalize them thanks to this file : https://encoding.spec.whatwg.org/encodings.json

test/jsdom/misc.js
nicolashenry: Because the response is not expected to be already decoded as a string here. iconv-lite can only decode buffers so I change it to a buffer.

domenic: Oh, I wasn't looking at the larger context for this code. It makes sense. Thanks.

test/jsdom/encoding.js
nicolashenry: All encoding tests I added are expecting an encoding in header or not encoding at all. I can try but I don't know if the "st" module will produce a charset in content-type or not in the future (see https://github.com/isaacs/st/pull/53).

lib/jsdom/living/nodes/Document-impl.js
nicolashenry: Mmh, in order to follow the specification, the suggested default encoding is UTF-8 but only "In controlled environments or in environments where the encoding of documents can be prescribed". Browsers don't usually use "UTF-8" as default but "windows-1252" (for an english locale). I think jsdom should follow the browsers behaviour, so should I change the default encoding to "windows-1252" ?

nicolashenry: Or maybe should it be "UTF-8" if the string content is provided by the user and "windows-1252" if the content is decoded from jsdom.

domenic: Very interesting. I like the idea of passing a string => UTF-8, otherwise windows-1252.

I will have to think carefully about this when we redesign the public jsdom API to see if there is a better solution. E.g. if we start accepting binary data instead of just strings, or similar.

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -302,13 +304,13 @@ class DocumentImpl extends NodeImpl {
     return (this._parsingMode === "xml" || this.doctype) ? "CSS1Compat" : "BackCompat";
   }
   get charset() {
-    return "UTF-8";
+    return this._encoding || this._defaultEncoding;
```
domenic: I think we should not have a _defaultEncoding option, but instead set _encoding to be either the passed encoding or the passed defaultEncoding.

test/jsdom/encoding.js
domenic: Right, I was thinking about the ones with no encoding in the header.

If st changes to add encoding to the header we can override it. I left a note in https://github.com/isaacs/st/pull/53#issuecomment-210682912

domenic:
Thank you so much! Merged as 8e8f333e2b9bffc45f5f53a156f64d8b60bfe061.

PR #1444, opened 2016-04-09
[Fix #1442] Explicitly enumerate direct node attributes

### LINKED ISSUE #1442: Error Serializing Inherited Attributes
Extending the `Object.prototype` causes `Element.outerHTML()` to break.

```js
var jsdom = require('jsdom');

Object.prototype.test_name = 'test_value';

var dom = jsdom.jsdom('<div id="0"></div>');

var div = dom.getElementById('0');

console.log(div.test_name);
console.log(div.outerHTML);
```

```html
test_value
<div id="0" test_value:test_name="test_value"></div>
```

This tells me that the issue is with the jsdom since its [`documentAdapter`](https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/browser/documentAdapter.js) is being passed to the serializer.

I would expect jsdom to skip attributes inherited from Object.

Oddly if `Object.prototype` is extended after the DOM is initialized it works as expected.

```js
var jsdom = require('jsdom');

var dom = jsdom.jsdom('<div id="0"></div>');

Object.prototype.test_name = 'test_value';

var div = dom.getElementById('0');

console.log(div.test_name);
console.log(div.outerHTML);
```

```html
test_value
<div id="0"></div>
```

I suspect the `HTMLElement` inheritance tree affects the attributes when creating the div, but requires explicit calls through the DOM APIs to change, so the fix will likely be during the attributes initialization.

deckar01:
The issue in #1442 is caused by directly accessing properties and using the `in` operator on an object that is meant to be a dictionary or map. These operations traverse the object's prototype chain which makes them unsuitable for enumerating the members of a collection.

 1. The attribute name shows up as `value:name`, because `node["x-attribsPrefix"]["name"]` will always equal `"value"` when `Object.prototype.name = "value";` has been set. The fix is a more explicit check of the object's _direct properties_ using [`Object.prototype.hasOwnProperty()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty).
 2. The attribute is added to the node's attribute map, because `for (let localName in node.attribs) {` will always iterate over `"name"` even when `node.attribs` is an "empty" object. The fix is a more explicit enumeration of the object's _direct properties_ using [`Object.keys()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys).

 > Note: This pitfall is a byproduct of JavaScript mixing the concepts of objects and dictionaries. Python separates these concepts into the `Object` class and the `dict` type, and provides dynamic access to object properties using the `{get,set,has,del}attr()` built-in functions.

### DIFF
--- a/lib/jsdom/browser/htmltodom.js
+++ b/lib/jsdom/browser/htmltodom.js
@@ -239,7 +239,7 @@ function setChild(core, parentImpl, node) {
   newNode[locationInfo] = node.__location;
 
   if (node.attribs) {
-    for (let localName in node.attribs) {
+    Object.keys(node.attribs).forEach(localName => {
       const value = node.attribs[localName];
       let prefix =
         node["x-attribsPrefix"] &&
@@ -255,7 +255,7 @@ function setChild(core, parentImpl, node) {
         prefix = null;
       }
       attributes.setAttributeValue(newNode, localName, value, prefix, namespace);
-    }
+    });
   }
 
   if (node.children) {
--- a/lib/jsdom/browser/htmltodom.js
+++ b/lib/jsdom/browser/htmltodom.js
@@ -241,8 +241,14 @@ function setChild(core, parentImpl, node) {
   if (node.attribs) {
     for (let localName in node.attribs) {
       const value = node.attribs[localName];
-      let prefix = node["x-attribsPrefix"] && node["x-attribsPrefix"][localName] || null;
-      const namespace = node["x-attribsNamespace"] && node["x-attribsNamespace"][localName] || null;
+      let prefix =
+        node["x-attribsPrefix"] &&
+        node["x-attribsPrefix"].hasOwnProperty(localName) &&
+        node["x-attribsPrefix"][localName] || null;
+      const namespace =
+        node["x-attribsNamespace"] &&
+        node["x-attribsNamespace"].hasOwnProperty(localName) &&
+        node["x-attribsNamespace"][localName] || null;
       if (prefix === "xmlns" && localName === "") {
          // intended weirdness in node-sax, see https://github.com/isaacs/sax-js/issues/165
         localName = prefix;
--- a/lib/jsdom/browser/htmltodom.js
+++ b/lib/jsdom/browser/htmltodom.js
@@ -243,11 +243,11 @@ function setChild(core, parentImpl, node) {
       const value = node.attribs[localName];
       let prefix =
         node["x-attribsPrefix"] &&
-        node["x-attribsPrefix"].hasOwnProperty(localName) &&
+        Object.prototype.hasOwnProperty.call(node["x-attribsPrefix"], localName) &&
         node["x-attribsPrefix"][localName] || null;
       const namespace =
         node["x-attribsNamespace"] &&
-        node["x-attribsNamespace"].hasOwnProperty(localName) &&
+        Object.prototype.hasOwnProperty.call(node["x-attribsNamespace"], localName) &&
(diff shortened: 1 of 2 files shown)


domenic:
Thanks, I'll merge this in next release day. But the better fix would be for parse5 to use `Object.create(null)` when creating objects-as-maps. @inikulin?

lib/jsdom/browser/htmltodom.js
```diff
@@ -239,17 +239,23 @@ function setChild(core, parentImpl, node) {
...
 
   if (node.attribs) {
-    for (let localName in node.attribs) {
+    Object.keys(node.attribs).forEach(localName => {
       const value = node.attribs[localName];
-      let prefix = node["x-attribsPrefix"] && node["x-attribsPrefix"][localName] || null;
-      const namespace = node["x-attribsNamespace"] && node["x-attribsNamespace"][localName] || null;
+      let prefix =
+        node["x-attribsPrefix"] &&
+        node["x-attribsPrefix"].hasOwnProperty(localName) &&
```
domenic: This will fail if someone creates an attribute with the prefix `"hasOwnProperty"`.

deckar01: What about `Object.prototype.hasOwnProperty.call(node["x-attribsPrefix"], localName)`? Should I add an additional test for this?

domenic: Yeah, that would be great

deckar01:
I pushed the test and the fix separately, because I wanted the travis build history to show it initially failed. The [build for `857204d`](https://travis-ci.org/deckar01/jsdom/builds/122064347) failed on my account as expected, but Travis seems to have hit a race condition when running the PR builds.

test/jsdom/parsing.js
```diff
@@ -100,6 +100,33 @@ describe("jsdom/parsing", () => {
...
+    delete Object.prototype.attribute;
+    const els = document.getElementsByTagName("element");
+
+    assert.equal(els.length, 1);
+    assert.equal(els[0].getAttribute("attribute"), undefined);
+    assert.equal(els[0].attributes.length, 0);
+    assert.equal(els[0].outerHTML, "<element></element>");
+  });
+
+  specify.only("prefix on attribute named 'hasOwnProperty' (GH-1444)", () => {
```
domenic: No .only

deckar01: Oops. Lint did not catch that.

deckar01:
It looks like the build failed due to a timeout in an [unrelated test](https://travis-ci.org/tmpvar/jsdom/jobs/122094348#L1879). Looking at the build history indicates that this test was fluctuating between 40 and 60 seconds.

[A build for the same commit](https://travis-ci.org/deckar01/jsdom/jobs/122094344#L1858) ran on my branch and passed. It took [41 seconds](https://travis-ci.org/deckar01/jsdom/jobs/122094344#L1858) for `cors/origin.htm` to run which indicates that this commit did not introduce a performance regression.

Restarting the latest build...

domenic:
Thanks! Merged as 96908118af7b26b1b4b1f964b96db6ce0302800d.

inikulin:
> But the better fix would be for parse5 to use Object.create(null) when creating objects-as-maps

Agreed. Can you create an issue in parse5 repo, please?

deckar01:
@inikulin https://github.com/inikulin/parse5/issues/119 opened.

Thanks for working on this issue.

PR #1456, opened 2016-04-21
Fixed iFrame sandbox attribute so setter does not utilize this.sandbox.value

palmerj3:


domenic:
This is incorrect. The IDL given there is the IDL from the spec.

What is the actual problem you are trying to solve?

palmerj3:
I noticed my unit tests would fail with code like so

```javascript
var sandboxParams = ['allow-top-navigation', 'allow-script'];
var iframe = document.createElement('iframe');
iframe.sandbox = sandboxParams.join(' ');
```

This would fail when the iframe setter would try to access this.sandbox.value.

domenic:
It looks like jsdom does not support the sandbox attribute at all right now. The correct fix is commenting out that IDL line.

palmerj3:
@domenic alright I'll close and re-submit. Thanks.

PR #1457, opened 2016-04-21
Commented out sandbox attribute for iFrame element

palmerj3:
Currently code such as this in your tests will fail when JSDom iFrame sandbox setter attempts to access this.sandbox.value.

```javascript
var sandboxParams = ['allow-top-navigation', 'allow-script'];
var iframe = document.createElement('iframe');
iframe.sandbox = sandboxParams.join(' ');
```


palmerj3:
This PR is in reference to my previous PR which I closed. https://github.com/tmpvar/jsdom/pull/1456

domenic:
Thanks, merged as ecf37b955552f6a679d5bc03f2d84dc77eaf490b!

Issue #1460, opened 2016-04-23
A Document created using createHTMLDocument() throws if "onscroll" is set


Joris-van-der-Wel:
Here's an interesting exception (8.4.0) :

```javascript
const jsdom = require('jsdom');

const document = jsdom.jsdom().implementation.createHTMLDocument('');
document.body.setAttribute('onscroll', 'foo'); // throws
```

```
P:\node.js\jsdomz\lib\jsdom\living\nodes\Element-impl.js:181
          self[name] = function (event) {
                     ^

TypeError: Cannot set property 'onscroll' of undefined
    at HTMLBodyElementImpl._attrModified (P:\node.js\jsdomz\lib\jsdom\living\nodes\Element-impl.js:181:22)
    at HTMLBodyElementImpl._attrModified (P:\node.js\jsdomz\lib\jsdom\living\nodes\HTMLElement-impl.js:92:25)
    at Object.exports.hasAttributeByNameNS.exports.appendAttribute (P:\node.js\jsdomz\lib\jsdom\living\attributes.js:204:11)
    at Object.exports.hasAttributeByNameNS.exports.setAttributeValue (P:\node.js\jsdomz\lib\jsdom\living\attributes.js:379:13)
    at P:\node.js\jsdomz\lib\jsdom\browser\htmltodom.js:257:18
    at Array.forEach (native)
    at setChild (P:\node.js\jsdomz\lib\jsdom\browser\htmltodom.js:242:31)
    at HtmlToDom._parseWithparse5v1 (P:\node.js\jsdomz\lib\jsdom\browser\htmltodom.js:90:7)
    at HtmlToDom.appendHtmlToElement (P:\node.js\jsdomz\lib\jsdom\browser\htmltodom.js:39:48)
    at HTMLBodyElementImpl.outerHTML (P:\node.js\jsdomz\lib\jsdom\living\nodes\Element-impl.js:297:25)

```

https://github.com/tmpvar/jsdom/blob/f6871893538db625de034ffc70e363ccf0d74bdb/lib/jsdom/living/nodes/Element-impl.js#L181

PR #1468, opened 2016-04-29
Fix setting proxied inline event handlers without a browsing context

Joris-van-der-Wel:
See issue #1460 

There are more issues with our handling of proxied inline event handlers. However there might be some gaps in the HTML LS spec, I'll have to do some more experimentation to be sure. At the very least, I've discovered some disagreement between the major browsers, so that might be worth submitting a WPT for.

So for now just a fix for the exception.


domenic:
Merged as 58de9c5c1ddb244d077d11a361fe256325317795, thanks!

PR #1469, opened 2016-04-30
Implement Navigator to-spec

### LINKED ISSUE #1467: navigator.javaEnabled is not implemented
When processing a segment of javascript, i get the following error;

> TypeError: navigator.javaEnabled is not a function

It would appear that this function has not been implemented.

domenic:
Adds a variety of missing properties to Navigator:

- appCodeName, product, productSub, vendor, and vendorSub from NavigatorID
- language and languages from NavigatorLanguage
- onLine from NavigatorOnLine
- javaEnabled() from NavigatorPlugins (fixes #1467)

We now emulate Chrome a bit more closely: we follow the spec with "navigator compatibility mode" of Chrome, which locks down appName, platform, and version to be more browser-like instead of Node.js-like. The default userAgent similarly gains a suffix "AppleWebKit/537.36 (KHTML, like Gecko)".

---

Blocked on https://github.com/jsdom/webidl2js/issues/22 which breaks `navigator.platform` and thus prevents the web platform tests from passing.

@Sebmaster might also want to give it a look for advice on webidl2js usage; given all the partial interfaces floating around I wasn't sure this was the best setup.


Sebmaster:
Looking pretty good to me. :tada:

PR #1470, opened 2016-05-01
Convert browser & jsonp & level1 tests to mocha

Joris-van-der-Wel:
I have corrected `jscs` for "level1". However `eslint` for "level1" is too much work, there are about 2000 violations that would have to be fixed manually (I did run `eslint --fix` though)


domenic:
Great stuff, thanks! Of course some of these should become to-upstream web platform tests, but baby steps.

domenic:
All merged :)

Ms2ger:
I occasionally submit tests from those test suites into wpt; https://bitbucket.org/ms2ger/old-dom-tests tracks my progress.

Joris-van-der-Wel:
That's great, we plan to do the same with our tests so that we can submit them. For example https://github.com/tmpvar/jsdom/blob/master/test/living-dom/node-iterator.js contains a lot of tests which are not covered by WPT.

PR #1471, opened 2016-05-05
Dgg

bbbryanb:


bbbryanb:
Yeah you cunts are trying to set me up but have I got some major surprises
and evidence. You ain't seen nothing

bbbryanb:
I've been all over your surveillance since the start and you are wrong lol. I read about the element a Month ago but I finally decided to pull it down and give samples and copies of 9 usb sticks of ur shit from the moment I tried to help. Now I just read a communique between u guys that confirmed your intention to set me up for a prison term. I've been very passive until now but I'm stepping up to yo
image
u the EVIL ppl you are. Come in my room or even wrong spot of the yard and expect to be stick with needles dipped in excrement and HIV.  That's just one tactic in your assault on our privacy. I hope those involved lose children to cancer and the most horrific things life has to offer as you have made ours hell. I got so many moves you are gonna flip. Come at me bitchs

Issue #1475, opened 2016-05-09
How to access DOM elements appended by ajax fetch?


coodoo:
I have a page with a "Load More" button, once clicked it will fetch next 10 items and append it to the page. 

Just wondering how do I access these 10 elements? I tried waiting for 10 secs then query the document again but nothing showed up, just want be sure this is a use case supported by jsdom? Thanks.

Sebmaster:
Yes, that should be possible.

coodoo:
That was my thought too but didn't seem to work, mind pointing me at a working example?

Btw, Do I have to explicitly allow external scripts running?

Sebmaster:
You do if you use .env. See [here](https://github.com/tmpvar/jsdom#external-resources).

coodoo:
Unfortunately it still doesn't work, here's a simple test case, mind having a look?

```javascript
var jsdom = require( 'jsdom' );

let url = 'http://www.nytimes.com/section/us?action=click&contentCollection=Science&module=SiteIndex&region=Footer&pgtype=article'

let config = {

	onload: function(window){
		console.log( 'load: ', window )
	},

	done: function( err, window ){
		let $elem = window.$('.initial-set article.story .story-body .headline')

		// 20 items
		console.log( 'original size: ', $elem.length, ' >last item:', $elem.last().text() )

		// click load more button
		let $btn = window.$('#latest-panel button.load-more-button')
		console.log( 'btn=', $btn.text() )
		$btn.trigger('click')

		// wait for 3sec, should have 40 items, but still 20
		setTimeout( function(){
			let $a = window.$('.initial-set article.story .story-body .headline')
			console.log( 'new size 1: ', $a.length, ' >last item: ', $elem.last().text() )
		}.bind(this), 3000)

		// wait for 6 sec, should have 40 items, but still 20
		setTimeout( function(){
			let $a = window.$('.initial-set article.story .story-body .headline')
			console.log( 'new size 2: ', $a.length, ' >last item: ', $elem.last().text() )
		}.bind(this), 6000)

	},

	features: {
		FetchExternalResources: ["script"],
		ProcessExternalResources: ["script"],
		SkipExternalResources: false
	  },
}

jsdom.env(
  url,
  ['http://code.jquery.com/jquery.js'],
  config,
);
```

vsemozhetbyt:
@coodoo 
[...]

PR #1478, opened 2016-05-10
Dispatch focus and blur event when the methods are called on HTMLElement

perrin4869:
Also, make sure the right version of webidl2js is installed.

This makes it so that:

```js
const input = document.createElement('input');
input.addEventListener('focus', () => console.log('Dispatching focus'));
input.focus();
```

Outputs `Dispatching focus`.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLInputElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLInputElement-impl.js
@@ -336,6 +336,14 @@ class HTMLInputElementImpl extends HTMLElementImpl {
     }
   }
 
+  focus() {
+    super.focus();
+
+    if (!this.getAttribute("disabled")) {
+      this.dispatchEvent(Event.createImpl(["focus"]));
+    }
+  }
+
   set maxLength(value) {
     if (value < 0) {
       throw new DOMException(DOMException.INDEX_SIZE_ERR);
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "st": "^0.5.5",
     "watchify": "^3.7.0",
     "wd": "0.3.12",
-    "webidl2js": "^4.3.0"
+    "webidl2js": "^4.3.3"
   },
   "browser": {
     "canvas": false,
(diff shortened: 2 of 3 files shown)


domenic:
This isn't quite correct. The spec is much more complex than just always firing the event. For example, `<div>`s should not get focus events fired, or `<button disabled>`s.

See https://html.spec.whatwg.org/#focusing-steps.

perrin4869:
I see! Guess it really was too good to be true ^^;;
not sure I can implement that...

vsemozhetbyt:
By the way, does jsdom support `document.hasFocus()`? I've [come upon](https://github.com/tmpvar/jsdom/issues/1475#issuecomment-217978339) an error in #1475.

dmethvin:
What does the spec say about focus on an element that isn't in a document? The section on [nested browsing contexts}(https://html.spec.whatwg.org/#nested-browsing-context) seems to assume it is already in one as far as I can tell. FWIW, focus on a detached element is not consistent in real browsers. https://jsfiddle.net/9ck8m8h5/

perrin4869:
Not entirely sure this one is correct, but it's as far as I can go now...

domenic:
> By the way, does jsdom support document.hasFocus()? I've come upon an error in #1475.

No, although if I am reading the spec right, I think we can implement it as returning true for documents that are not inside iframes.

> What does the spec say about focus on an element that isn't in a document?

To me it looks like this comes down to the definition of "focusable area," which is left a bit vague on purpose, but perhaps we should tighten it. That is, there's nothing that demands that the element be in the document. It only says "if the element is a focusable area". Arguably, it's logical that out-of-document elements are not focusable. I'll open a HTML issue.

> Not entirely sure this one is correct, but it's as far as I can go now...

This is reasonable for now, thanks. Will merge this weekend.

domenic:
Oh, it turns out the spec is actually fairly clear that elements that are not being rendered are not focusable areas. See the table at https://html.spec.whatwg.org/multipage/interaction.html#focusable-area.

perrin4869:
Looks awesome, thanks!

PR #1479, opened 2016-05-10
Remove document event listeners before setting readyState to complete

WojciechNowak:
Hi.

Test scenario: 
1) load page with \<script\>document.addEventListener("DOMContentLoaded", function() {})\</script\>
2) close window.

Problem: On window close, document implementation fires DOMContentLoaded event, which in turn executes function added within addEventListener.

Proposed solution: clear document eventListeners before setting it to readyState.

Tests added to test/window/script.js. If pull request is accepted and there is a requirement to write tests in Mocha please tell me in which file it should be rewritten.

Best regards,
Wojciech Nowak

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -252,7 +252,7 @@ function Window(options) {
       if (this._document.close) {
         // It's especially important to clear out the listeners here because document.close() causes a "load" event to
         // fire.
-        this._document._listeners = Object.create(null);
+        idlUtils.implForWrapper(this._document)._eventListeners = Object.create(null);
         this._document.close();
       }
       const doc = idlUtils.implForWrapper(this._document);
--- a/test/window/script.js
+++ b/test/window/script.js
@@ -105,6 +105,19 @@ exports.tests = {
     test.strictEqual(window.DONE, 1);
     test.done();
   },
+  
+  script_should_ignore_DOMContentLoaded_after_window_close: function(test){
+    var window = jsdom.jsdom('<html><head>\
+      <script>\
+        var a = 0;\
+        function handle(){ a++; };\
+        document.addEventListener("DOMContentLoaded", handle, false);\
+      </script>\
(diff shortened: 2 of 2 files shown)


domenic:
Hmm. `window.close()` already handles this use case. But `document.close()` should not blow away event listeners ([see spec](https://html.spec.whatwg.org/multipage/webappapis.html#dom-document-close)).

This seems wrong to me.

WojciechNowak:
Event listeners added to a window are in fact cleared and the problem doesn't exist. However we can see a comment in close function in document-impl.js:
// Set the readyState to 'complete' once all resources are loaded.
// As a side-effect the document's load-event will be dispatched. 
...
ev.initEvent("DOMContentLoaded", false, false);
this.dispatchEvent(ev);

Is emmiting this event defined by spec or is it some kind of jsdom hack? Maybe we should consider removing only DOMContentLoaded from eventListeners?

domenic:
document.close() should definitely never remove any event listeners. If you are trying to close the window and stop event listeners, use window.close(), not document.close().

It sounds like it might be a separate issue where document.close() should not be what fires the DOMContentLoaded event. Instead, the parser probably should.

WojciechNowak:
I call window.close which in turn calls document.close which causes DOMContentLoaded to be fired. If proposed solution to this problem is incorrect what should be done next? Shall I create an issue to track this bug?

domenic:
Yes. window.close() needs to clear the event handlers; if it is not doing so then something is wrong. But document.close() should not clear event handlers.

WojciechNowak:
I checked window.close() function and I think I found the bug. I have provided a new solution. What do you think about it?

domenic:
Oh, wow, excellent find, thank you! Yes, I'll include this in the next release (probably this weekend).

WojciechNowak:
Great, I'm glad I could help.

domenic:
Merged as 868560b63ab6544ab5b10913c640fcca0e814a60; thanks!

PR #1481, opened 2016-05-11
use proxy setting for all requests

### LINKED ISSUE #1464: Is it possible run all requests via proxy server ?
It's very interesting fact, only first request goes via proxy server but other ones through  local ip host where jsdom launched.

I have used this url http://myip.ru/ for testing connection to site via proxy.

For  testing this issue, needs to load that url and wait nearly 5 second, because there's runs ajax request for detecting your ip.

Thanks for yours understanding.

nicolashenry:
This should fixes #1464 ~~but sadly the dependency I added for testing it does'nt work well with browserify (this will probably have to wait nodejitsu/node-http-proxy#975 to be merged).~~

### DIFF
--- a/lib/jsdom.js
+++ b/lib/jsdom.js
@@ -108,6 +108,7 @@ exports.jsdom = function (html, options) {
     pool: options.pool,
     agentOptions: options.agentOptions,
     strictSSL: options.strictSSL,
+    proxy: options.proxy,
     userAgent: options.userAgent
   });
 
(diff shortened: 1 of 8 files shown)


nicolashenry:
I removed the http-proxy dependency in favor of a simple proxy implementation, so everything should be ok with browserify now.

domenic:
Thank you! Merged as bcecb46ec75dfd51daf6155eb0b5ef8715e42b22.

Issue #696, opened 2013-10-29
WHATWG HTML requires appending event attributes to event queue


terinjokes:
In the current WHATWG HTML spec, handlers added via the "traditional" event attributes on a DOM element should be appended to the same queue as events added via the DOM spec.
#### 7.1.6.1 Event handlers

> […]
> 
> When an event handler H of an element or object T implementing the EventTarget interface is first set to a non-null value, the user agent must append an event listener to the list of event listeners associated with T with type set to the event handler event type corresponding to H, capture set to false, and listener set to the event handler processing algorithm defined below.
> 
> […]
> 
> This example demonstrates the order in which event listeners are invoked. If the button in this example is clicked by the user, the page will show four alerts, with the text "ONE", "TWO", "THREE", and "FOUR" respectively.
> 
> ```
> <button id="test">Start Demo</button>
> <script>
>  var button = document.getElementById('test');
>  button.addEventListener('click', function () { alert('ONE') }, false);
>  button.setAttribute('onclick', "alert('NOT CALLED')"); // event handler listener is registered here
>  button.addEventListener('click', function () { alert('THREE') }, false);
>  button.onclick = function () { alert('TWO'); };
>  button.addEventListener('click', function () { alert('FOUR') }, false);
> </script>
> ```
> 
> […]

Sebmaster:
Fixed in 5f3cbd6.

PR #1063, opened 2016-05-14
Expose on[event] attributes on prototypes

Sebmaster:
[Twitter Conversation](https://twitter.com/Rich_Harris/status/576486021543649280) which made me check this.

Looks like we should be exposing on\* attributes on our prototypes (Window and HTMLElement and Document): https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements,-document-objects,-and-window-objects https://html.spec.whatwg.org/multipage/webappapis.html#idl-definitions

Currently we just check if an attribute named on[event] is defined on the element and execute it when triggering the event. We might just be able to fix order of the event trigger too while we're at it (#696).


domenic:
@Sebmaster what's the latest on this? I have cherry-picked the "enable tests" commits but all the work here seems great and it would be great to rebase/finish it :)

Sebmaster:
The problem here was that window is not split into wrapper + impl class which makes mixins really difficult, so I first have to convert window to finish this.

lib/jsdom/living/nodes/GlobalEventHandlers-impl.js
```diff
@@ -1,6 +1,202 @@
...
+
+  _eventChanged(event) {
+    const propName = "on" + event;
+    if (!events.has(event)) {
+      return;
+    }
+
+    const val = this.getAttribute(propName);
+
+    if (inheritedAttributes.has(event) && (this.tagName === "BODY" || this.tagName === "FRAMESET")) { // still not great at detecting this
```
domenic: I guess we could at least pull this out into a shared function.

lib/jsdom/living/nodes/GlobalEventHandlers-impl.js
```diff
@@ -1,6 +1,202 @@
...
+          }
+          argNames.push(wrapperBody);
+          fn = Constructor(...argNames)(...args);
+        }
+        this._eventHandlers[event] = fn;
+      }
+      return this._eventHandlers[event];
+    },
+    set(val) {
+      if (!(typeof val === "function")) {
```
domenic: As per discussion, since we need to mix in the impl directly for Window, I guess we need this. In which case it needs a comment explaining this.

lib/jsdom/living/nodes/GlobalEventHandlers-impl.js
```diff
@@ -1,6 +1,202 @@
...
+        // TODO: location
+        if (element !== null) {
+          formOwner = element.form || null;
+        }
+
+        try {
+          // eslint-disable-next-line no-new-func
+          Function(body); // properly error out on syntax errors
+        } catch (e) {
+          // TODO: Report the error
```
domenic: Shouldn't be that hard with the utils we have; I think it's in the deleted code.

lib/jsdom/browser/Window.js
```diff
@@ -20,6 +20,8 @@ const Navigator = require("../living/generated/Navigator");
 const reportException = require("../living/helpers/runtime-script-errors");
 const { contextifyWindow } = require("./documentfeatures.js");
 
+const GlobalEventHandlersImpl = require("../living/nodes/GlobalEventHandlers-impl").implementation;
```
Sebmaster: Can't use the interface in non-wrapped classes. I think I may have an idea on how to incrementally move window over, but let's punt on that either to the end of the PR or a new one.

lib/jsdom/living/nodes/GlobalEventHandlers-impl.js
```diff
@@ -1,6 +1,202 @@
...
+  "select", "show",
+  "sort", "stalled",
+  "submit", "suspend",
+  "timeupdate", "toggle",
+  "volumechange", "waiting"
+]);
+
+const inheritedAttributes = new Set(["blur", "error", "focus", "load", "resize", "scroll"]);
+
+function appendHandler(el, eventName) {
```
Sebmaster: Elements are always impls in an impl class (except in the few instances we leak).

lib/jsdom/living/nodes/GlobalEventHandlers-impl.js
```diff
@@ -1,6 +1,202 @@
...
+  "resize", "scroll",
+  "seeked", "seeking",
+  "select", "show",
+  "sort", "stalled",
+  "submit", "suspend",
+  "timeupdate", "toggle",
+  "volumechange", "waiting"
+]);
+
+const inheritedAttributes = new Set(["blur", "error", "focus", "load", "resize", "scroll"]);
```
Sebmaster: [They are](https://html.spec.whatwg.org/#event-handler-content-attributes), "kind of".

lib/jsdom/living/nodes/GlobalEventHandlers-impl.js
```diff
@@ -1,6 +1,202 @@
...
+
+    if (eventName === "mouseover" || (eventName === "error" && ErrorEvent.isImpl(event))) {
+      if (returnValue) {
+        event._canceledFlag = true;
+      }
+    // TODO: before unload
+    } else if (!returnValue) {
+      event._canceledFlag = true;
+    }
+  }, false);
```
Sebmaster: Yep

lib/jsdom/living/nodes/GlobalEventHandlers-impl.js
```diff
@@ -1,6 +1,202 @@
...
+      if (!(typeof val === "function")) {
+        return;
+      }
+      if (inheritedAttributes.has(event) && (this.tagName === "BODY" || this.tagName === "FRAMESET")) { // still not great at detecting this
+        this.ownerDocument.defaultView["on" + event] = val;
+        return;
+      }
+
+      if (!this._eventHandlers[event]) {
+        appendHandler(this, event);
```
Sebmaster: Done

Sebmaster:
There's got to be a better way to to the body proxy thing. I think it should be possible to do write this in the actual body interface instead of (shittily) detecting this at runtime.

lib/jsdom/living/helpers/create-event-accessor.js
```diff
@@ -0,0 +1,135 @@
...
+          Function(body); // properly error out on syntax errors
+          // Note: this won't execute body; that would require `Function(body)()`.
+        } catch (e) {
+          // TODO: Report the error
+          return null;
+        }
+
+        let fn;
+        const createFunction = vm.isContext(document._global) ? document.defaultView._globalProxy.Function : Function;
+        if (event === "error" && element === null) {
```
domenic: @Sebmaster, this isn't quite right; error events also get the `with() {}` wrapping, so putting that all in the `else` doesn't work, I think.

snuggs:
What's the `wrapperBody` used for @domenic ? Well done on this by the way to you and @Sebmaster.

Coinidentally just finished a similar iteration for mapping `on*` events from a `class` to an existing `HTMLElement`

https://github.com/devpunks/snuggsi/blob/master/elements/global-event-handlers.es#L53-L110

domenic:
wrapperBody is used to implement a crazy part of the spec where inside an event handler everything on document, the target element, and possibly the target element's form owner is in scope. We do this via a bunch of with statements. Basically it's all about implementing step 10 of https://html.spec.whatwg.org/#getting-the-current-value-of-the-event-handler .

Issue #1388, opened 2016-02-17
Stubbing/patching `window.location.href` in JSDom 8?


dylanpyle:
We've got some code in use in tests that worked in jsdom 7 but appears not to work in jsdom 8.

Our JSDom setup:

```
  const doc = jsdom.jsdom('<html><body></body></html>', {
    referrer: 'https://example.com/baz',
    url: 'https://example.com/foo'
  });

  global.document = doc;
  global.window = doc.defaultView;
```

And the test itself:
```
      window.location.href = 'https://newurl.com/foo';
      // some assertions that depend on window.location
```

Adding some logging immediately after setting `window.location.href` shows that the value wasn't actually updated or changed, and other means overriding (e.g. stubbing with `sinon`) don't seem to work either since it's implemented via getters/setters rather than an actual property. What's the best way to override this value?

It looks like setting is supposed to do *something* based on https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/window/Location-impl.js#L46 but I can't figure out what the final effect of that is supposed to be.

domenic:
So, the issue is that you can't set location in browsers, without navigating. And jsdom is trying to behave like a browser. (Although it doesn't do the navigation part yet.)

The way we generally work around this is adding methods like `jsdom.reconfigureLocation(window, { href: "..." })` which allow you to modify the jsdom objects from the outside, using "superpower" APIs that are not part of the normal DOM. Would that suffice for your use case?

domenic:
Alternately, it might be a better solution to add something like `jsdom.changeURL(window, ...)` and not mess with `Location` directly. Then you still get all the usual behaviors from `window.location`, but everything in the DOM (including e.g. `document.URL`) now uses a new page URL.

dylanpyle:
Ah yeah, makes sense. An API for either of those sounds great — didn't realize there was already a `reconfigureWindow` but adding it there seems reasonable? I'm basically looking for something with the semantics of "override the stuff I set up in `config` initially, without recreating the whole DOM".

sterpe:
I am also running into with test written for `jest@0.8`, which depends on `jsdom@7` upgrading to `jest@latest` which bumps to `jsdom 8`.  Testing some authentication/oauth code that sets a new window location (e.g., to an oauth server).

domenic:
Can you explain more how that works, and how you would test that code in browsers?

sterpe:
@domenic, something like this

```
jest.autoMockOff()
[...]

PR #1487, opened 2016-05-17
Added ability to update window.location via _URL

mummybot:
changeURL fix for #1388 

This works, however you may be unhappy with reproducing the setup from document-impl to window.js > update method. Happy to rework this, but I couldn't see an easy way to DRY the set logic in the document-impl constructor so it could be reused. Glad for any pointers.

The history API is broken. Based on your comment:

> And jsdom is trying to behave like a browser. (Although it doesn't do the navigation part yet.)

I assume this is okay until jsdom accepts updating of the window.location object directly. Given we are calling it via changeURL this will hopefully be clear to users.

I called the window method update as this in theory could be used by other update functions. However it checks for and only updates _URL, and is called by the specific changeURL method.

### DIFF
--- a/lib/jsdom.js
+++ b/lib/jsdom.js
@@ -20,6 +20,8 @@ const locationInfo = require("./jsdom/living/helpers/internal-constants").locati
 const idlUtils = require("./jsdom/living/generated/utils");
 const blobSymbols = require("./jsdom/living/blob-symbols");
 
+const whatwgURL = require("whatwg-url");
+
 require("./jsdom/living"); // Enable living standard features
 
 /* eslint-disable no-restricted-modules */
@@ -49,6 +51,19 @@ exports.nodeLocation = function (node) {
 exports.reconfigureWindow = function (window, newProps) {
   if ("top" in newProps) {
     window._top = newProps.top;
+  }
+};
+
+exports.changeURL = function (window, urlString) {
+  const pattern = /^((https?|file):\/\/)([\da-z\.-]+)(\.?([a-z\.]{2,6})([\/\w \.-]*)*)?(:[0-9]*)?\/?$/;
+
+  if (typeof urlString === "string" && pattern.test(urlString)) {
+    const doc = idlUtils.implForWrapper(window._document);
+
+    doc._URL = whatwgURL.parseURL(urlString);
+    doc._origin = whatwgURL.serializeURLToUnicodeOrigin(doc._URL);
+  } else {
+    throw new Error(urlString + " is not a valid URL. URL is still " + window._document.URL + ".");
   }
 };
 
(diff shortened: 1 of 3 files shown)


lib/jsdom/browser/Window.js
```diff
@@ -149,6 +151,13 @@ function Window(options) {
   this.dispatchEvent = this.dispatchEvent.bind(this);
 
   ///// METHODS
+  this.update = function (privateData) {
```
domenic: You can't add a new nonstandard method to `window`. Instead, move the contents of this to lib/jsdom.js, which is allowed to access `window._document`.

lib/jsdom/browser/Window.js
```diff
@@ -149,6 +151,13 @@ function Window(options) {
...
+    if (privateData.url) {
+      const doc = idlUtils.implForWrapper(this._document);
+      doc._URL = whatwgURL.parseURL(privateData.url);
+      doc._origin = new whatwgURL.URL(privateData.url).origin;
```
domenic: Use `whatwgURL.serializeURLToUnicodeOrigin(privateData.url)

lib/jsdom.js
```diff
@@ -52,6 +52,12 @@ exports.reconfigureWindow = function (window, newProps) {
   }
 };
 
+exports.changeURL = function (window, newProps) {
```
domenic: The signature should be `(window, urlString)`

test/misc/update.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
```
domenic: Let's not write any more old-style tests. Instead, add a new test (using Mocha) to https://github.com/tmpvar/jsdom/blob/master/test/jsdom/misc.js

test/misc/update.js
```diff
@@ -0,0 +1,13 @@
...
+const jsdom = require("../..");
+
+exports["Update option url"] = t => {
+  const window = jsdom.jsdom("", { url: "http://example.com/" }).defaultView;
+
+  t.strictEqual(window.location.href, "http://example.com/");
+
+  jsdom.changeURL(window, { url: "http://localhost/" });
+  t.strictEqual(window.location.href, "http://localhost/");
+
```
domenic: Also test `window.document.URL` just to be extra safe.

test/misc/update.js
```diff
@@ -0,0 +1,13 @@
...
+  t.done();
+};
```
domenic: Add a test that it throws a TypeError if you pass it an invalid URL (like `"asdf"`).

domenic:
Awesome, thanks! A great start, just some feedback on how to do things a bit better. Much appreciated.

domenic:
Thanks, looking good. Still some minor tweaks I'd make (e.g. not changing Window.js, and getting rid of the useless updateURL function) but I can make those while merging and doing a release, which should be at the latest this weekend.

domenic:
Thank you; merged as 83c410049655701c842d7e8cc3647cb0608d475d!

PR #1492, opened 2016-05-19
fix #1491

### LINKED ISSUE #1491: Content-Length is blocked in CORS requests
While checking into sindresorhus/fast-cli#2, I noticed that we're blocking client-JS from reading the Content-Length header, because it fails [this](https://github.com/tmpvar/jsdom/blob/72b53d9682dd60aa98047439c4f810584027de6b/lib/jsdom/living/xmlhttprequest.js#L805) check.

It seems to me the [spec](https://fetch.spec.whatwg.org/#cors-safelisted-response-header-name) says that basically all headers except Set-Cookie / Set-Cookie2 are allowed (due to `Any value in list that is not a forbidden response-header name.`)? That kinda seems too permissive to me... But Content-Length definitely sounds like something that should be exposed, so there's a bug here somewhere...

/cc @nicolashenry

nicolashenry:
It still need a test, I will do it when I have some time.

### DIFF
--- a/lib/jsdom/living/xmlhttprequest.js
+++ b/lib/jsdom/living/xmlhttprequest.js
@@ -46,6 +46,10 @@ const forbiddenRequestHeaders = new Set([
   "upgrade",
   "via"
 ]);
+const forbiddenResponseHeaders = new Set([
+  "set-cookie",
+  "set-cookie2"
+]);
 const uniqueResponseHeaders = new Set([
   "content-type",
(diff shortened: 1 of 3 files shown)


nicolashenry:
Tests added.

domenic:
Thanks, merged as 8eebf8bce2f431c3b0ed6e5dbceab9ce275ec8f1!

PR #1502, opened 2016-05-24
Changing StyleSheetList.prototype.item to match specification

Ginden:
Per [spec](https://drafts.csswg.org/cssom/#dom-stylesheetlist-item)
> The item(index) method must return the indexth CSS style sheet in the collection. If there is no indexth object in the collection, **then the method must return null.**

### DIFF
--- a/lib/jsdom/level2/style.js
+++ b/lib/jsdom/level2/style.js
@@ -62,7 +62,7 @@ module.exports = core => {
   StyleSheetList.prototype.__proto__ = Array.prototype;
 
   StyleSheetList.prototype.item = function item(i) {
-    return this[i];
+    return Object.prototype.hasOwnProperty.call(this, i) ? this[i] : null;
   };
 
   core.StyleSheetList = StyleSheetList;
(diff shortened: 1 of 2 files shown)


domenic:
Test please?

Ginden:
I doubt if tests in `test/level2/` are called anywhere, so I don't know where should I put this test.

`test/level2/style.js` is referenced by `test/worker-runner.js`. `test/worker-runner.js` is referenced by `test/worker.js`, but I couldn't find any file referencing `test/worker.js`.

Putting `throw new Error()` on top of file doesn't change output of `npm test`.

Joris-van-der-Wel:
The file you are looking for is `test/runner`. test/level2 still has to be converted to mocha, you can use `npm run test-old` to run it. Normal `npm test` should also work, but you will have to wait a while because test/level2 will be on of the last tests that will be executed.

Ginden:
Test added. BTW, tests seem to fail because of timeouts.

domenic:
Merged as e91dec8195273d8690421b1e4763808874f4b1eb; thanks!

Issue #1438, opened 2016-04-05
Errors in XHR event handlers get swallowed


domenic:
This code https://github.com/tmpvar/jsdom/blob/3ad126cc9783efd498cfb32a7e64e4888caf9687/lib/jsdom/living/events/EventTarget-impl.js#L212-L221 has always been troublesome. One issue noticed in IRC today is that since XHR objects don't have an ownerDocument, it can't find the virtual console to report to.

I'd also like to fix the document-less window business by associating all document-less windows with their creator window, maybe.

PR #1516, opened 2016-06-07
Swallowing errors in invokeEventListeners

ehpc:
Hi!
I've had some problem using jsdom with simple _jQuery.ajax.then_. There was an error inside _then_ callback and it was swallowed. So everything was fine, no errors, but I had never received any data. So I started debugging and found that piece of code. I think it it misleading to just swallow errors without any warning. Maybe it has some purpose in the big picture, but for my use case it was destructive.

### DIFF
--- a/lib/jsdom/living/events/EventTarget-impl.js
+++ b/lib/jsdom/living/events/EventTarget-impl.js
@@ -220,6 +220,9 @@ function invokeEventListeners(listeners, target, eventImpl) {
       if (window) {
         reportException(window, e);
       }
+      else {
+        throw e;
+      }
       // Errors in window-less documents just get swallowed... can you think of anything better?
     }
   }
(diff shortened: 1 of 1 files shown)


domenic:
Can you give a small (10-20 line) test case where this error manifests, so that we can use it as a regression test in the future?

I don't think the fix you propose here is the best way to do this, but once we have a small test case, we can use it to work on better solutions, like forwarding it to the jsdom virtual console.

ehpc:
var jsdom = require('jsdom').jsdom,
	targetUrl = 'https://code.jquery.com/jquery-1.11.3.js';
	document = jsdom('<script src="' + targetUrl + '"></script>'),
	window = document.defaultView;

document.body.onload = function () {
	window.jQuery.ajax({
		url: targetUrl,
		type: 'GET',
		dataType: 'text',
		data: {},
		async: false
	}).then(function (data) {
		// The next line should throw something
		undefinedVariable.property = true;
		console.log('>>>DATA:', data.substring(0, 80));
	}, function (jqXHR, textStatus, errorThrown) {
		console.error('>>>ERROR:', textStatus, errorThrown);
	});
	console.log('body.onload end');
};

console.log('script end');

domenic:
Ah, thank you. This is #1438 then. We should solve that in a different way.

ehpc:
Ok, thanks.

PR #1521, opened 2016-06-14
Make button[type=submit]'s default preventable

marcandre:
If a click on a submit button has its default prevented, the `submit` event should not be fired. This PR fixes that and adds a test.

*Note*: `input[type=submit]` already uses `_activationStep` and thus only works fine, only `button` is affected by this bug.


marcandre:
I doubled check this behavior on Chrome, and here's an [excerpt of the spec](https://www.w3.org/TR/html5/forms.html#the-button-element):

> When a button element is not disabled, its **activation behavior** element is to run the steps defined in the following list for the current state of the element's type attribute [...]

domenic:
Merged as ad235e133a10d4a3925bc1870e2ea8fa739e46f5; thank you!!

PR #1522, opened 2016-06-14
Make label's default action preventable

marcandre:
Similar issue to #1521, but with `<label>`


marcandre:
Here's an excerpt [of the WhatWG  spec](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) that might help:

"The label element's exact default presentation and behaviour, in particular what its **activation behaviour** might be, if anything, should match the platform's label behaviour. [...] For example, on platforms where clicking a checkbox label checks the checkbox, clicking the label in the following snippet could trigger the user agent to run synthetic click activation steps on the input element, as if the element itself had been triggered by the user. On other platforms, the behavior might be just to focus the control,"

domenic:
Merged as 88407c540066ecba9fb2b0567e30f96b45fdcd39, thank you!!

PR #1523, opened 2016-06-14
Allow for button and label's children

marcandre:
`<button>` and `<label>` can have children, and clicking on those should also submit a form / proxy clicks. Currently, nothing happens...

This PR attempts to address this issue.
First, I simplify the interface to the event behavior hooks. I found the mix of `event.target` and `this` to be confusing (when they were always the same). This makes the fix easier, otherwise an alternative solution would have been to use `currentTarget` instead of `target` everywhere and set that property before calling them. I see no point to that, these methods are internal only and `currentTarget` would always be `this` anyways.

The fix then consists of finding the first element with event behavior hooks, if any and call them on that element.

This builds on #1521 and #1522

### DIFF
--- a/lib/jsdom/living/nodes/HTMLButtonElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLButtonElement-impl.js
@@ -5,11 +5,10 @@ const HTMLElementImpl = require("./HTMLElement-impl").implementation;
 const closest = require("../helpers/traversal").closest;
 
 class HTMLButtonElementImpl extends HTMLElementImpl {
-  _activationBehavior(event) {
-    const target = event.target;
-    const form = target.form;
+  _activationBehavior() {
+    const form = this.form;
     if (form) {
-      if (target.type === "submit") {
+      if (this.type === "submit") {
         form._dispatchSubmitEvent();
       }
     }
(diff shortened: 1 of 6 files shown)


marcandre:
The test that fails (in 2 of the 3 environments) seem to simply be flaky (and completely unrelated to these changes).

Anything I should change, on this PR or the other 2?

domenic:
I haven't had time to review this yet, sorry. One thing that I'll need to check is that the spec actually prescribes activation behavior instead of pre-click activation steps.

marcandre:
Right. I double checked in Chrome.

Here's an excerpt [of the spec](https://www.w3.org/TR/html5/forms.html#the-label-element) that might help:

"The label element's exact default presentation and behavior, in particular what its activation behavior might be, if anything, should match the platform's label behavior. [...] For example, on platforms where clicking a checkbox label checks the checkbox, clicking the label in the following snippet could trigger the user agent to run synthetic click activation steps on the input element, as if the element itself had been triggered by the user. On other platforms, the behavior might be just to focus the control,"

domenic:
Hmm, that spec is not one that anyone implements. Can you check https://html.spec.whatwg.org/multipage/ instead? Also we'll need to check all the non-label classes you changed too.

I imagine all of these outcomes are probably correct BTW; I'm just saying that we'll need to ensure it's done the right way.

marcandre:
Sure, I understand perfectly.
I updated the spec references in both PRs; both w3 and whatwg use the same terms, in particular refer only to *activation behavior*.

domenic:
Awesome, thank you! I'll try to make time to review within the next day or two!!

marcandre:
While the other two should be fine, I need to refactor this PR a bit. I'll reopen when done

PR #1529, opened 2016-06-21
Button label children

marcandre:
`<button>` and `<label>` can have children, and clicking on those should also submit a form / proxy clicks. Currently, nothing happens...

This PR attempts to address this issue.
First, I simplify the interface to the event behavior hooks. I found the mix of `event.target` and `this` to be confusing (when they were always the same). This makes the fix easier, otherwise an alternative solution would have been to use `currentTarget` instead of `target` everywhere and set that property before calling them. I see no point to that, these methods are internal only and `currentTarget` would always be `this` anyways.

The fix then consists of finding the first element with event behavior hooks, if any and call them on that element.

This builds on #1521 and #1522 

It's probably pretty clear that clicking on children should fire the activation behavior, but here's an excerpt [from the spec](https://html.spec.whatwg.org/multipage/interaction.html#activation):
> Let *e* be the nearest activatable element of target (defined below), if any [...], run pre-click activation steps on it [...] if the click event is not canceled, run post-click activation steps [...]

I called the utility function `nearestActivatableElement`, as it aims to follow [the spec](https://html.spec.whatwg.org/multipage/interaction.html#nearest-activatable-element)

### DIFF
--- a/lib/jsdom/living/nodes/HTMLButtonElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLButtonElement-impl.js
@@ -5,11 +5,10 @@ const HTMLElementImpl = require("./HTMLElement-impl").implementation;
 const closest = require("../helpers/traversal").closest;
 
 class HTMLButtonElementImpl extends HTMLElementImpl {
-  _activationBehavior(event) {
-    const target = event.target;
-    const form = target.form;
+  _activationBehavior() {
+    const form = this.form;
     if (form) {
-      if (target.type === "submit") {
+      if (this.type === "submit") {
         form._dispatchSubmitEvent();
       }
     }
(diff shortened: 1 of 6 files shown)


domenic:
So, this doesn't seem to be correct to me, at least per spec. *Users clicking* on children should find the nearest activatable element. However, *calling `.click()`* should not. This is the difference between "run synthetic click activation steps" and "run authentic click activation steps".

And yet... somehow, http://jsbin.com/fakufixiqo/edit?html,console,output shows browsers passing the test you added. Hmm. But I believe that if it really follows the logic you added, it would cause "child got click event" to not happen.

At this point I am suspecting a spec bug. Do you have any thoughts?

In any case, I'll pull in the first of the two new commits as 03a212f5cf2ea653ce5bb96c7526d05d9f529650.

marcandre:
I re-read the spec more carefully and you're right, the point 5 of the first section ("If this click event is not canceled, run post-click activation steps on the element.") should probably be modified to talk about the first targetable element instead.
Should I create a PR at whatwg for this?

domenic:
A PR or an issue makes sense to me. I'm a little worried that just changing it to include the first targetable element would mean that the child would not get the click event.

Some previous inconclusive discussion: https://github.com/whatwg/html/issues/737

PR #1534, opened 2016-06-24
Move "named access on window" tests to to-upstream

### LINKED ISSUE #1532: Name access on the window object for iframes
This is a follow-up from a comment in another issue: https://github.com/tmpvar/jsdom/issues/1030#issuecomment-227855373

The contention is whether:

```html
<iframe id="foo"></iframe>
<script>
  console.log(window.foo);
</script>
```

Should `window.foo` be the iframe element or the iframe's contentWindow.

I filed bugs against [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1281604) and [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=622437#c1).  Firefox closed it as invalid saying that because iframe doesn't have a `name` it doesn't set the browsing context and therefore the WindowProxy is not returned. If so then jsdom has a bug.  I'd like to fix it.  Will wait to see what Chromium says first.

@domenic suggested moving the named access tests to the WPT format and getting them upstreamed, I've started working on that in a branch by putting them in the `to-upstream` folder.

matthewp:
This commit moves the tests for [named access on the window
object](https://html.spec.whatwg.org/#named-access-on-the-window-object)
to the to-upstream folder so that they can be upstreamed to WPT.

This is part of (but does not close #1532).


domenic:
This is really great work!! LGTM, and much appreciated. I will merge sometime this weekend when I get a jsdom day.

PR #1535, opened 2016-06-27
Some work on DOMParser; needs help

domenic:
@Sebmaster I was hoping you could take a look at this. The problem comes with the "add strict parsing errors when parsing XML documents" commit. Flipping that causes a bunch of tests to start failing for weird reasons having to do with tree construction in htmltodom.js. The last commit modifies the web platform tests manifest so that it only runs the tests which have shown failures.

What's weird is that the failures are due to stuff that should have nothing to do with strict vs. non-strict XML parsing. It's something about how nodes are emitted and what's on the stack at that time. I figure since you plugged in sax in the first place maybe you had some ideas.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -18,6 +18,7 @@ test/level2
 test/level3
 test/sizzle
 test/web-platform-tests/tests
+test/web-platform-tests/to-upstream/domparsing/DOMParser-dont-upstream.html
 test/window/files
 test/window/frame.js
 test/window/script.js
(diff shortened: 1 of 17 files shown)


domenic:
> the fact that HTML entities were unknown I think? Pretty sure it errors out on those by default.

It's actually correct to error out on HTML entities; that's what browsers do for strict XHTML.

However, a bunch of level1 tests failed due to throwing on *unknown* entities, which some of the level1 tests use. I fixed that with https://github.com/tmpvar/jsdom/pull/1535/commits/cb4c8587734797efbbdb5c92beee7fb99f0f7524 and now everything is mostly passing.

I'd definitely appreciate if you had the time to look in to the node stack stuff that's causing failures.

PR #1545, opened 2016-07-02
Named access nested context revision

### LINKED ISSUE #1532: Name access on the window object for iframes

matthewp:
This updates the "named access on the window object" feature to only
return the nested browsing context (the `contentWindow`) when the
browsing context name is set. Since the browsing context name is not set
by the `id` property, the iframe element should be returned when the id
is the name being used.

In summary:

* `id`: return the element.
* `name`: return the WindowProxy.

This is based on findings from a Firefox issue:
https://bugzilla.mozilla.org/show_bug.cgi?id=1281604

Closes #1532

### DIFF
--- a/lib/jsdom/living/named-properties-window.js
+++ b/lib/jsdom/living/named-properties-window.js
@@ -58,7 +58,8 @@ function namedPropertyResolver(HTMLCollection, window, name, values) {
   for (let i = 0; i < length; ++i) {
     const node = objects[i];
 
-    if ("contentWindow" in node && !hasOwnProp.call(node, "contentWindow")) {
+    if ("contentWindow" in node && !hasOwnProp.call(node, "contentWindow") &&
+       node.getAttribute("name") === name) {
       return node.contentWindow;
     }
   }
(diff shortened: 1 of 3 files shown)


matthewp:
Looks like this breaks a different test, investigating...

matthewp:
Yeah, another test was depending on id being used in this way, changing it to name fixes the test.

domenic:
Great stuff! Merged as 7182006a4c8cdffb45f7c30456430aab296a9768. Thanks very much!!

PR #1547, opened 2016-07-03
Implement the cloning steps for input elements

### LINKED ISSUE #946: jQuery .val() method on a cloned form element doesn't update its value attribute
I had a weird problem using zombie.js for a functional test of some production cloning script. It seems that using jQuery's `val()` on a cloned, unattached element doesn't work as it would in-browser. Here's a stripped-down scenario to reproduce it:

```html
<html><body>
<form>
	<input type="text" id="sourceField" name="field" value="should not expose this">
</form>

<script>
(function($) {
	var $template = $('#sourceField').clone();
	$template.val('');
	var $new = $template.clone();
	// in production code, would rename and append here
	console.log($new.prop('name') + ': ' + $new.val());
})(jQuery);
</script>
</body></html>
```

Running this in a browser console, you see:

    field:

In zombie.js, the console shows:

    field: should not expose this

If I replace the `$template.val('');` with `$template.attr('value', '');`then the code works consistently in both. It seems that the `.val()` method doesn't update the value attribute of the cloned node, which allows its content to persist across further clone operations.

I see it using jQuery 1.8 through 1.11.1. Looking at the [jQuery code](https://github.com/jquery/jquery/blob/1.x-master/src/attributes/val.js) it seems that it should be a pretty straightforward assignment. I'd welcome someone with better knowledge of jsdom taking a quick look to see if it's an obvious bug in how it's handled.

matthewp:
This implements the cloning steps for input elements by copying over the
value, checkedness, dirty value, and dirty checkedness properties to the
cloned input.

Closes #946

### DIFF
--- a/lib/jsdom/living/nodes/HTMLInputElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLInputElement-impl.js
@@ -7,7 +7,9 @@ const Event = require("../generated/Event");
 const mapper = require("../../utils").mapper;
 const createHTMLCollection = require("../../living/html-collection").create;
 const DOMException = require("../../web-idl/DOMException");
-const domSymbolTree = require("../helpers/internal-constants").domSymbolTree;
+const internalConstants = require("../helpers/internal-constants");
+const domSymbolTree = internalConstants.domSymbolTree;
+const cloningSteps = internalConstants.cloningSteps;
 const closest = require("../helpers/traversal").closest;
 const isDisabled = require("../helpers/form-controls").isDisabled;
 
@@ -379,6 +381,13 @@ class HTMLInputElementImpl extends HTMLElementImpl {
       throw new DOMException(DOMException.INDEX_SIZE_ERR);
     }
     this.setAttribute("size", String(value));
+  }
+
+  [cloningSteps](copy, node) {
+    copy._value = node._value;
+    copy._checkedness = node._checkedness;
+    copy._dirtyValue = node._dirtyValue;
+    copy._dirtyCheckedness = node._dirtyCheckedness;
   }
 }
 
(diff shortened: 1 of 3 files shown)


test/living-html/htmlinputelement.js
```diff
@@ -148,3 +149,30 @@ exports["an input's parsed type attribute should be reflected in both its proper
...
+
+exports["cloning an input also clones its dirty value"] = t => {
+  const doc = jsdom.jsdom();
+  const input = doc.createElement("input");
+  input.value = "foo bar";
+
+  const copy = input.cloneNode();
+  const impl = idlUtils.implForWrapper(copy);
+
+  t.equal(impl._dirtyValue, true);
```
Sebmaster: Don't test private API. This might change. Instead test for the effects of dirtyness and add that as a wpt (if none exists for it).

domenic: It looks like _dirtyValue being true will impact whether setting the value attribute impacts the element's actual value (`.value`). Similarly _dirtyCheckedness will impact whether adding/removing the checked attribute affects the actual checkedness (`.checked`). Should be pretty reasonable to write tests for these that fail before propagating and succeed after.

Sebmaster: The section in the spec which describes the modified behaviour is [here](https://html.spec.whatwg.org/#attr-input-checked) and [here](https://html.spec.whatwg.org/#attr-input-value) btw.

matthewp: Thanks, I couldn't tell what dirty value/checkedness actually did, I get it now, will update and squash.

domenic:
Thank you! Merged as 9ade70ef6444e03e21d282fc1c8e35209e0a4b49; 9.4.1 will be released shortly with this.

PR #1583, opened 2016-08-10
Use strict Boolean comparisons for inline handler return values

### LINKED ISSUE #1577: preventDefault has different behavior with real browser
Hi Guys,

I'm encountering following issue:
`<a href="/test.html" id="demo">
  <span onclick="myFunction()">
    Click here
  </span>
</a>`

In browser, if you click on the span element, the click event bubbled to the anchor element is not defaultPrevented. While in jsdom (EventTarget-impl.js), it will invoke the E.preventDefault(). And the event bubbled is defaultPrevented.

Could someone explain about this behavior?

Cheers

dmethvin:
This avoids the event handler seeing an `undefined` as `false` and calling `preventDefault()`.

Fixes #1577

### DIFF
--- a/lib/jsdom/living/events/EventTarget-impl.js
+++ b/lib/jsdom/living/events/EventTarget-impl.js
@@ -252,10 +252,10 @@ function getListenerForInlineEventHandler(target, type) {
       }
 
       if (type === "mouseover" || isWindowError) {
-        if (returnValue===true) {
+        if (returnValue === true) {
           E.preventDefault();
         }
-      } else if (returnValue===false) {
+      } else if (returnValue === false) {
         E.preventDefault();
       }
     };
--- a/lib/jsdom/living/events/EventTarget-impl.js
+++ b/lib/jsdom/living/events/EventTarget-impl.js
@@ -252,10 +252,10 @@ function getListenerForInlineEventHandler(target, type) {
       }
 
       if (type === "mouseover" || isWindowError) {
-        if (returnValue) {
+        if (returnValue===true) {
           E.preventDefault();
         }
-      } else if (!returnValue) {
+      } else if (returnValue===false) {
         E.preventDefault();
       }
     };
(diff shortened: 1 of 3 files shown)


domenic:
Thanks! New tests need to be web-platform-test format though...

dmethvin:
Oh, then I didn't understand the note in contributing.  Would it be a new html file in `to-upstream/html`, then update `to-upstream.js`?

Sebmaster:
Yes, exactly.

domenic:
I wonder if enabling https://github.com/w3c/web-platform-tests/blob/master/html/webappapis/scripting/events/event-handler-processing-algorithm.html would suffice?

domenic:
Probably not, since it doesn't seem to test the case of returning *non*-booleans. But I guess that's the right path; adding a html/webappapis/scripting/events/event-handler-processing-algorithm-non-booleans.html would be what I'd do.

dmethvin:
I've rewritten the test as a web platform test but I can't figure out how to actually debug it inside the html file. Where do errors go?

domenic:
They should be printed to the console. Nothing will be printed if the test passes. You can run the to-upstream tests with `npm run test-tuwpt` I believe.

dmethvin:
Yeah i was able to run the tests `npm run test-tuwpt`, but getting a timeout locally just like Travis. It runs in the browser both locally and in Travis. If I put something clearly incorrect in the file like `barf()` it also times out so I'm sure it's something in the code throwing an error. I've pushed my WIP to this branch.

domenic:
OK, so whatever is going wrong it is not obvious... I'm trying to debug now. Will report back.

domenic:
The problem is the filename is wrong and it's hanging instead of telling you the error. Let me try to work on a fix to the test harness to not do that.

dmethvin:
Duh, yeah, that did the trick! I pushed the fix. Not sure how I managed to screw up a copy-pasta of the path. :see_no_evil:

domenic:
Fixing the underlying problem is hard because apparently we don't cause 404s to fail jsdom.env. I'm going to wait until newapi has a better way of approaching this... hopefully soon.

dmethvin:
Yeah, no problem. I was just barking up the wrong tree because I didn't see any output.

domenic:
Merged as e69834a3fa3d16c4a40e7299951f35fe17833511 and a 9.4.2 release will go out shortly! Thanks so much!!

PR #1585, opened 2016-08-12
fix getElementsByTagName

wacky6:
in latest standard, getElementsBy* should return `HTMLCollection`

reference: https://dom.spec.whatwg.org/#concept-getelementsbytagname

`createLiveNodeList` is removed because it is no longer used in the file.

### DIFF
--- a/lib/jsdom/living/html-collection.js
+++ b/lib/jsdom/living/html-collection.js
@@ -5,13 +5,28 @@ const idlUtils = require("./generated/utils");
 
 const privates = Symbol("HTMLCollection internal slots");
 
+// workaround for element keys that conflicts with property and method name
+const conflictKeys = {
+  length: 1,
+  item: 1,
+  namedItem: 1
+};
+
 class HTMLCollection {
   constructor(secret, element, query) {
     if (secret !== privates) {
       throw new TypeError("Invalid constructor");
     }
 
-    this[privates] = { element, query, snapshot: undefined, keys: [], length: 0, version: -1 };
+    this[privates] = {
+      element,
+      query,
+      snapshot: undefined,
+      version: -1,
+      keys: [],
+      length: 0,
+      conflictElements: {}    // elements with conflict keys
+    };
     updateHTMLCollection(this);
   }
 
@@ -28,6 +43,9 @@ class HTMLCollection {
   namedItem(name) {
     updateHTMLCollection(this);
 
+    if (name in conflictKeys) {
+      return this[privates].conflictElements[name] || null;
+    }
     if (Object.prototype.hasOwnProperty.call(this, name)) {
       return this[name];
     }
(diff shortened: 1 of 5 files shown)


domenic:
Great catch. We need tests for this though before we can merge! Maybe there are existing web platform tests to enable, or we can write a quick one.

I wonder if createLiveNodeList is still used at all? I guess that would be easy to check if I grep through the source.

wacky6:
Provided getElementsBy* tests seems be too strict. Enabling it gives 12 fails. (Prior to this fix, 15 fails). Should I add a test?

I did a search on `createLiveNodeList`, it is still needed by Node.childNodes. [DOM standard](https://dom.spec.whatwg.org/#dom-node-childnodes) says it should be a NodeList.

domenic:
> Provided getElementsBy* tests seems be too strict. Enabling it gives 12 fails. (Prior to this fix, 15 fails). Should I add a test?

Yeah, we've run into this before. Our solution in cases like that has been to copy and paste the test, remove the failures, and notate it as "don't upstream". See e.g. https://github.com/tmpvar/jsdom/blob/master/test/web-platform-tests/to-upstream/html/semantics/scripting-1/the-script-element/script-languages-dont-upstream.html and https://github.com/tmpvar/jsdom/blob/master/.eslintignore#L21. Let me know if that makes sense.

wacky6:
ok. will investigate it and failed tests suggested by travis-ci.

wacky6:
### Fix failed tests
Failed tests are caused by elements with `id='length'`, it conflicts with HTMLCollection.length property. 

I modified its implementation to be consistent with Chrome and Firefox.

```html
<input id="length"></input>
<input id="namedItem"></input>
```

```JavaScript
let els = document.getElementsByTagName('input')

els.length    // => 1
els.namedItem    // => function namedItem()
els.namedItem('length')    // => input#length
els.namedItem('namedItem')    // => input#namedItem
```

A better (neat) solution requires Proxy, which simplifies name/id lookup operations. (It could make HTMLCollection "live" without calling `updateHTMLCollection` in other part of code.

However, it requires node v6+ (without command line flags). At present, to me, using Proxy is not a good idea, it is likely to cause problem for many users.

Maybe this can be added to long-term refactoring list? It should be implemented when most users switch to node v6.


### Added test
I create a separate commit for HTMLCollection test.

Sebmaster:
> Maybe this can be added to long-term refactoring list? It should be implemented when most users switch to node v6.

It's on there. I have a branch which is supposed to implement the named properties object [in webidl2js](https://github.com/jsdom/webidl2js/tree/feature/named-properties-object), which we'll start out using on a window refactor to initially get some data on performance. The fear is that a proxy getter may disable all v8 speed-ups if it hits the proxy, which hopefully isn't that big of a deal on window.

It may speed up write operations by a lot though, so that'd be neat.

domenic:
So I went ahead and fixed the original issue, slightly differently, in https://github.com/tmpvar/jsdom/commit/4b80cd674418736fe4cde25414699440ece07d4c; as usual with jsdom, this spurred me down a very large path of fixing lots of things, so there's also https://github.com/tmpvar/jsdom/commit/4be8634f4121286454f6f559551ca06b0b967b68. But I still haven't fixed some of the things this PR takes care of about clobbering reserved properties, so I'll leave this open and maybe incorporate that work soon...

Issue #1501, opened 2016-05-24
Newly-imported dom/events/Event-dispatch-multiple-stopPropagation.html fails when enabled


domenic:
Interestingly it also fails in Chrome. @Sebmaster, mind taking a look?

Sebmaster:
The stop propagation flag is never cleared in the spec except in [initialize event](https://dom.spec.whatwg.org/#concept-event-initialize). That, in turn, is only reset as a default and in initEvent. Seems like we're working according to spec here?

domenic:
I'll take this over to the web-platform-tests repo...

PR #1589, opened 2016-08-20
Unset stopPropagationFlags due to latest DOM spec changes

Sebmaster:
See #1501 and nested issues for more details.

### DIFF
--- a/lib/jsdom/living/events/EventTarget-impl.js
+++ b/lib/jsdom/living/events/EventTarget-impl.js
@@ -147,6 +147,8 @@ class EventTargetImpl {
     }
 
     eventImpl._dispatchFlag = false;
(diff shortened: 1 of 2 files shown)


Sebmaster:
Hrm, seems like there were a lot of changes in XMLHttpRequest again :disappointed:

domenic:
You can comment out the failing tests for now I guess, unless they are easy to fix... /cc @nicolashenry

domenic:
I rolled the tests and took your change as part of the process. I fixed some XHR stuff, discovered a spec vs. tests issue for others (https://github.com/whatwg/xhr/issues/72), and had to turn off some other tests as they became too rigorous for us and started relying on features like URLSearchParams or ReadableStream. (One day maybe we can work on a test expectations file that has more granularity than the file level.)

PR #1591, opened 2016-08-29
Kill WPT server with SIGINT instead of SIGTERM

i8-pi:
After running `npm run test`, several python processes are left behind. This is because the WPT web server isn't shutdown properly.

Python doesn't register a default handler for SIGTERM and it doesn't run `__exit__()` methods of context managers when it gets that signal. Using SIGINT avoids this problem

I've tested this on Linux and Windows. It doesn't make a difference on Windows but it doesn't break anything either

### DIFF
--- a/test/web-platform-tests/run-web-platform-test.js
+++ b/test/web-platform-tests/run-web-platform-test.js
@@ -57,7 +57,7 @@ module.exports = function (testDir) {
     pollForServer(() => urlPrefix).then(serverHasStarted);
 
     process.on("exit", () => {
-      python.kill();
+      python.kill("SIGINT");
     });
   });
 


snuggs:
Well done on this one @i8-pi. For anyone looking the doc on optional `child.kill([signal])` is here. https://nodejs.org/api/child_process.html#child_process_class_childprocess

i8-pi:
This is what I see:
1. Run `npm run test`
2. When the WPT server has fully started up, look at the process list. There are 7 python processes
3. Kill the test or wait for it to complete
4. Look at the process list again. There are still 6 python processes remaining.

There should not be any python processes left behind after the tests have terminated

domenic:
Hmm, on Windows 10 this causes the tests to hang at the end for me...

domenic:
Ah, sorry, false alarm, I was just staring at a slow test. Merged as 178a3ad1722a968c2e84a8666f8c10b7330527b4 with a comment added. Thank you!!

PR #1610, opened 2016-09-22
Compatibility with ZoneJS (and Angular 2)

Swiip:
I'm really not sure about the code itself, as I'm new to the source of jsdom. It's more about opening the discussion for working on the Angular 2 compatibility and showing that the gap is really thin.

I'm not a huge fan of ZoneJS but I have to work with Angular 2 and jsdom is currently not able to make it works because of ZoneJS.

After some digging, I found that there is only two blocking points for jsdom to work with ZoneJS : `onreadystatechange` to be writable and events callback which seems to happend to be valued at `true` instead of a function.

This PR contains the fixes I use to make Angular 2 apps fully working. I used my branch on my project and it works.

It could be a workaround or a fix for https://github.com/tmpvar/jsdom/issues/1513

### DIFF
--- a/lib/jsdom/living/events/EventTarget-impl.js
+++ b/lib/jsdom/living/events/EventTarget-impl.js
@@ -237,7 +237,7 @@ const wrappedListener = Symbol("inline event listener wrapper");
 function getListenerForInlineEventHandler(target, type) {
   const callback = target["on" + type];
 
-  if (!callback) { // TODO event handlers: only check null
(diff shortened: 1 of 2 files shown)


domenic:
Thanks for the pull request, but I can't accept these fixes. They don't appear to be spec-compliant, and don't come with any tests.

I look forward to further research into the problem, in the form of small test cases (e.g. one targeted at onreadystatechange and one targeted at event handlers) that pass in browsers and fail in jsdom.

Issue #1609, opened 2016-09-21
Add addEventListener options argument support


GianlucaGuarini:
jsom does not support the addEventListener options argument https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

It would be great to see at least the support for the `once` flag

GianlucaGuarini:
@domenic are you already working on it? Otherwise I will be glad to help making a pull request

domenic:
A pull request would definitely be welcome.

GianlucaGuarini:
@domenic you got it ;) https://github.com/tmpvar/jsdom/pull/1618

domenic:
After https://github.com/tmpvar/jsdom/pull/1618, the remaining work is to add the passive flag. \o/!

PR #1618, opened 2016-10-03
Feature/1609

GianlucaGuarini:
Enable the events options, allowing the use of the `once` flag closing #1609


lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -24,8 +24,17 @@ class EventTargetImpl {
       return;
     }
 
-    if (capture === undefined) {
-      capture = false;
+    if (typeof options === "boolean") {
+      options = { capture: options };
+    } else {
+      if (!options) {
```
domenic: `if (options === undefined || options === null)`

lib/jsdom/living/events/EventTarget.idl
```diff
@@ -1,6 +1,12 @@
 [Exposed=(Window,Worker)]
 interface EventTarget {
-  void addEventListener(DOMString type, EventListener? callback, optional boolean capture = false);
-  void removeEventListener(DOMString type, EventListener? callback, optional boolean capture = false);
+  void addEventListener(DOMString type, EventListener? callback, optional Options options);
```
domenic: You need to ues the spec's IDL here: https://dom.spec.whatwg.org/#eventtarget

@Sebmaster, I don't suppose we support union types in webidl2js, do we?

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -24,8 +24,17 @@ class EventTargetImpl {
...
+        options = {};
+      }
+      options = {
+        capture: typeof options.capture === "boolean" ? options.capture : false,
```
domenic: typeof tests are incorrect here. You need to do `capture: Boolean(options.capture)`.

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -34,18 +43,18 @@ class EventTargetImpl {
 
     for (let i = 0; i < this._eventListeners[type].length; ++i) {
       const listener = this._eventListeners[type][i];
-      if (listener.capture === capture && listener.callback === callback) {
+      if (listener.callback === callback) {
```
domenic: This modification is incorrect. See step 4 here (and please add a test to make sure nobody ever removes this in the future) https://dom.spec.whatwg.org/#dom-eventtarget-addeventlistener

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -57,8 +66,16 @@ class EventTargetImpl {
...
-      capture = false;
+    if (typeof options === "boolean") {
+      options = { capture: options };
+    } else {
+      if (!options) {
+        options = {};
+      }
+      options = {
+        capture: typeof options.capture === "boolean" ? options.capture : false,
+        passive: typeof options.passive === "boolean" ? options.passive : false
```
domenic: Passive does not apply for removeEventListener.

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -197,13 +214,17 @@ function invokeEventListeners(listeners, target, eventImpl) {
     }
 
     const listener = handlers[i];
+    const { capture, once } = listener.options;
```
domenic: Unfortunately we have not yet done a major version bump to require Node v6, so we cannot yet use destructuring.

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -197,13 +214,17 @@ function invokeEventListeners(listeners, target, eventImpl) {
...
     if (listeners.indexOf(listener) === -1 ||
-        (eventImpl.eventPhase === Event.CAPTURING_PHASE && !listener.capture) ||
-        (eventImpl.eventPhase === Event.BUBBLING_PHASE && listener.capture)) {
+        (eventImpl.eventPhase === Event.CAPTURING_PHASE && !capture) ||
+        (eventImpl.eventPhase === Event.BUBBLING_PHASE && capture)) {
       continue;
     }
 
     try {
+      if (once) {
```
domenic: This should be outside the try/catch

test/web-platform-tests/index.js
```diff
@@ -204,6 +204,9 @@ describe("Web Platform Tests", () => {
     "dom/events/Event-type-empty.html",
     "dom/events/EventListener-handleEvent.html",
     "dom/events/EventTarget-addEventListener.html",
+/*    "dom/events/EventListenerOptions-capture.html",*/
```
domenic: Why are these commented out? Capture especially we should support.

test/web-platform-tests/index.js
```diff
@@ -204,6 +204,9 @@ describe("Web Platform Tests", () => {
...
+    "dom/events/AddEventListenerOptions-once.html",
+/*    "dom/events/AddEventListenerOptions-passive.html",*/
```
domenic: We should either support passive and pass the tests, or not have any code for it.

lib/jsdom/living/events/EventTarget.idl
Sebmaster: It does not :(

domenic: It'll just pass through as if it were "any", right? Or will it blow up.

Sebmaster: It will just assume it's any.

domenic: Good, then yeah the rest of my code review comments (which are essentially about better implementing the bindings manually) still hold.

GianlucaGuarini:
@domenic thanks for your review I will update asap

GianlucaGuarini:
@domenic My last patch passes all the web platform tests. I would like to support also the `passive` flag but for that I would prefer another pull request

domenic:
Cool! Can you remove or comment out everything related to passive then in this commit?

GianlucaGuarini:
@domenic I already did https://github.com/tmpvar/jsdom/pull/1618/commits/ad7e8dab273d10cfc870646faf37316969e62d15#diff-a72f2a511def81e4948fd908e640a0deR201

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -12,62 +12,58 @@ class EventTargetImpl {
...
-  addEventListener(type, callback, capture) {
+  addEventListener(type, callback, options) {
     // webidl2js currently can't handle neither optional arguments nor callback interfaces
     if (callback === undefined || callback === null) {
       callback = null;
     } else if (typeof callback !== "object" && typeof callback !== "function") {
       throw new TypeError("Only undefined, null, an object, or a function are allowed for the callback parameter");
     }
 
+    options = normalizeEventHandlerOptions(options, ["capture", "once", "passive"]);
```
domenic: Remove passive from here

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -167,7 +163,10 @@ function invokeInlineListeners(object, event) {
 
     // Will be falsy for windows that have closed
     if (document && (!object.nodeName || document.implementation._hasFeature("ProcessExternalResources", "script"))) {
-      invokeEventListeners([{ callback: inlineListener }], object, event);
+      invokeEventListeners([{
+        callback: inlineListener,
+        options: normalizeEventHandlerOptions(false, ["capture", "once", "passive"])
```
domenic: remove passive from here

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -255,6 +241,48 @@ function invokeEventListeners(listeners, target, eventImpl) {
...
+    // somehow according to the specs users could pass also other properties as options
+    additionalOptions = Object.keys(options).filter(k => !defaultBoolKeys.includes(k)).reduce((obj, key) => {
+      if (!keysBlackList.includes(key)) {
+        obj[key] = options[key];
+      }
+
+      return obj;
+    }, {});
+  } catch (e) {
+    // should the errors be simply swallowed here?!
```
GianlucaGuarini: @domenic how should I handle the error here?

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -234,6 +241,48 @@ function invokeEventListeners(listeners, target, eventImpl) {
...
+ * @param   {Object} options         - user defined options
+ * @param   {Array} defaultBoolKeys  - boolean properties that should belong to the options object
+ * @param   {Array} keysBlackList    - options not allowed (optional)
+ * @returns {Object} object containing at least the  "defaultBoolKeys"
+ */
+function normalizeEventHandlerOptions(options, defaultBoolKeys, keysBlackList) {
+  // normalize the options argument in order to get always an object
+  if (options === null || options === undefined) {
+    options = {};
+  } else if (typeof options !== "object") {
```
domenic: This is incorrect. This only applies if the typeof is bool.

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -234,6 +241,48 @@ function invokeEventListeners(listeners, target, eventImpl) {
...
+    options = { capture: options };
+  }
+
+  let defaultOptions;
+  let additionalOptions;
+  keysBlackList = keysBlackList || [];
+
+  try {
```
domenic: This whole section is way overcomplicated. You should just write the following:

```js
const returnValue = {};

if (typeof options === "bool") {
  returnValue.capture = Boolean(options);
  return returnValue;
}

for (const key of defaultBoolKeys) {
  returnValue[key] = Boolean(options[key]);
}

return returnValue;
```

Don't catch any errors; don't bother with a blacklist; none of that. Users cannot pass other properties as options.

lib/jsdom/living/events/EventTarget.idl
```diff
@@ -1,6 +1,19 @@
...
+callback interface EventListener {
+  void handleEvent(Event event);
+};
+
+dictionary EventListenerOptions{
+  boolean capture = false;
+};
+
+dictionary AddEventListenerOptions: EventListenerOptions {
+  boolean passive = false;
```
domenic: Comment out passive here

lib/jsdom/living/events/EventTarget-impl.js
GianlucaGuarini: It's complicate because tests and specs are a bit weird. Somehow you should be able to [pass custom options values](https://github.com/w3c/web-platform-tests/blob/eae188126a21a593715ba2626cc059144d6f0978/dom/events/EventListenerOptions-capture.html#L54). If i don't get them tests will fail

domenic: That test is specifically testing that any custom option values are *not* gotten. An error is thrown if they ever are gotten.

lib/jsdom/living/events/EventTarget-impl.js
domenic: Errors should be rethrown, not caught---otherwise test failures, like the assert_unreached, will not surface correctly!

GianlucaGuarini:
@domenic thanks for the review you are right..I have simplified the code a lot thanks for the hint

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -234,6 +241,27 @@ function invokeEventListeners(listeners, target, eventImpl) {
...
+/**
+ * Normalize the event listeners options argument in order to get always a valid options object
+ * @param   {Object} options         - user defined options
+ * @param   {Array} defaultBoolKeys  - boolean properties that should belong to the options object
+ * @returns {Object} object containing at least the "defaultBoolKeys"
+ */
+function normalizeEventHandlerOptions(options, defaultBoolKeys) {
+  const returnValue = {};
+
+  if (typeof options === "boolean" || options === null || typeof options !== "object") {
```
domenic: This is not correct. Boolean needs to be handled separately, and then there needs to be a separate test for null/undefined, not for non-object. Here is a test case:

```js
Number.prototype.once = true;
const result = normalizeEventHandlerOptions(5, ["capture", "once"]);
```

This should return `{ capture: false, once: true }`, not (per the current implementation) `{ capture: true }`.

domenic: Yeah, agreed. You can add one to to-upstream if you're feeling ambitious, but definitely not required :)

GianlucaGuarini: ah wait I guess it's not necessary https://jsfiddle.net/d6n1dbrx/

GianlucaGuarini: > Yeah, agreed. You can add one to to-upstream if you're feeling ambitious, but definitely not required :)

I have also other important projects to maintain ;)

GianlucaGuarini:
Ok with my lastest patch everything should be fine and spec compliant.

GianlucaGuarini:
@domenic thanks for merging it ✌️

domenic:
Thanks for writing it! Sorry for taking so long on the merge. A release should be going out within the next hour :)

PR #1621, opened 2016-10-04
Browsing context container focus behavior

### LINKED ISSUE #1619: Iframes (browsing context containers) are focusable and should be focused if iframe child is focused
### Spec

https://html.spec.whatwg.org/multipage/interaction.html#focusing-steps (the 5th entry under focusing step 1)
> If new focus target is a browsing context container with a non-null nested browsing context
> Let new focus target be the browsing context container's nested browsing context's active document.

https://www.w3.org/TR/html5/editing.html#focus
> When a child browsing context is focused, its browsing context container must also have focus.

### Test

https://github.com/w3c/web-platform-tests/blob/master/html/editing/focus/document-level-focus-apis/document-level-apis.html

### Required changes

In `lib/jsdom/living/helpers/focusing.js`:
- `exports.isFocusableAreaElement` should return true iframe element. Could use something like:

```
  if (elImpl instanceof elImpl._ownerDocument._defaultView.HTMLIFrameElement) {
    return true;
  }
```

(using the defaultView's HTMLIFrameElement constructor means that if nested browsing contexts are ever made separate / sandboxed, the check will still work)

In `lib/jsdom/living/nodes/HTMLElement-impl.js`:
- `focus` needs to recursively check `ownerDocument.defaultView.parent` and invoke `.focus()` on any parent `HTMLIFrameElement`:
```
    if (this._ownerDocument._defaultView._frameElement) {
      this._ownerDocument._defaultView._frameElement.focus();
    }
```
- `blur` shouldn’t need to change, I don’t think. When a focusable element inside an iframe is blurred, the iframe element itself (the browsing context container) is still the `activeElement` of its parent.

acusti:
Fixes #1619

### DIFF
--- a/lib/jsdom/living/helpers/focusing.js
+++ b/lib/jsdom/living/helpers/focusing.js
@@ -19,6 +19,10 @@ exports.isFocusableAreaElement = elImpl => {
     return true;
   }
 
+  if (elImpl.nodeName === "IFRAME") {
+    return true;
+  }
+
   if (!Number.isNaN(parseInt(elImpl.getAttribute("tabindex")))) {
     return true;
   }
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -373,6 +373,10 @@ class DocumentImpl extends NodeImpl {
     return this.body;
   }
 
+  hasFocus() {
+    return Boolean(this._lastFocusedElement);
+  }
+
   toString() {
     return "[object HTMLDocument]";
   }
--- a/lib/jsdom/living/nodes/Document.idl
+++ b/lib/jsdom/living/nodes/Document.idl
@@ -74,7 +74,7 @@ partial /*sealed*/ interface Document {
   // user interaction
   readonly attribute WindowProxy? defaultView;
   readonly attribute Element? activeElement;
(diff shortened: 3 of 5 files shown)


acusti:
The actual implementation of `hasFocus()` is not dependent on whether there’s an `activeElement` or not, but rather on whether the actual browsing context (or a focusable area inside it, including a nested browsing context) is focused by the user. From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/hasFocus):
> For example, an active element within a popup window that is not the foreground does not have focus.

Or, if you open developer tools and click somewhere inside it, the current document and any nested browsing contexts will no longer have focus.

Here’s the [“has focus steps” in the spec](https://html.spec.whatwg.org/multipage/interaction.html#has-focus-steps)

I thought that in the context of jsdom, checking for an activeElement might be enough, considering that it is a headless browser. But if that isn’t the case, I’d be happy to remove `hasFocus()`. It will only mean that I will need to extract the other tests from `html/editing/focus/document-level-focus-apis/document-level-apis.html` and create a new test file to just check those. In that case, should I make the new test file as `test/living-dom/document-level-apis.js`?

domenic:
Sorry for the delay in responding. This is great as-is, however :). Merging and doing a release now!

PR #1628, opened 2016-10-15
Correct certain behaviors of File and Blob classes

TimothyGu:
The `lastModified` property has been moved from `Blob` to `File`, as that's where it [belongs][2].

Also changed was the semantics of the property, in case `lastModified` was not provided as part of the `options` dictionary. The specification dictates that if not provided, that property must be assigned to the value of `Date.now()` at the time when the constructor is called (see [`File` constructor][3]).

The missing [`isClosed` getter][4] for `Blob` has also been added.

[1]: https://w3c.github.io/FileAPI/#blob
[2]: https://w3c.github.io/FileAPI/#ref-for-dfn-lastModified-1
[3]: https://w3c.github.io/FileAPI/#file-constructor
[4]: https://w3c.github.io/FileAPI/#ref-for-dfn-isClosed-1


domenic:
Great fixes, thank you! Can you either find or write some web platform tests for them, per contributing.md?

TimothyGu:
@domenic, alright, added one for `isClosed` in the `to-upstream` directory. The issue with `lastModified` property seems to be too careless a bug to be put into the entire Web platform test corpus.

domenic:
Hmm I'd really prefer to have a test for all changed code; after all, if I could make this bug in my implementation, so could anyone else. Just using setTimeout to check that .lastModified is of the creation time, not the current time, would be great.

TimothyGu:
@domenic, so my test is basically a modified version of a test that is already present in web-platform-tests. For consistency within the existing file, I wrote it in ES5 syntax, which fails linting. Is there a preferred way to resolve such a situation? Should I remove all existing tests and just put this new one in to-upstream, or should I convert the rest of the file to ES2015 as well?

The same question also applies to the `lastModified` test I am working on.

Sebmaster:
Write in ES6 for us, we'll need to do the ES5 conversion when we upstream anyways, kinda.

Issue #1559, opened 2016-07-09
MouseEvent properties are not initialized correctly when using .click() method


moll:
Hey,

I'm not sure what any spec says about that, but from my limited testing of browsers, synthesized click events come with a `detail` property of `0`, differentiating them from real clicks that have it at 1 or higher. The `button` and `buttons` property are `0`, too, indicating the left mouse button.

[JSDOM's HTMLElement.prototype.click implementation](https://github.com/tmpvar/jsdom/blob/94ce8c4ed79f866528cd464875635cfaa3ae84cb/lib/jsdom/living/nodes/HTMLElement-impl.js#L67) has `detail` and other properties as `undefined`.

domenic:
Hmm. Because [the default value for the dictionary](https://github.com/tmpvar/jsdom/blob/13b2326b57ec621c4b75a64966ea23d89d8fce99/lib/jsdom/living/events/UIEvent.idl#L9) is zero, it should be 0 in all click events created by default. Do you have a small 10-line test case showing that this is not the case in jsdom?

moll:
For some reason, the following throws for me on JSDOM v9.4.1 with Node v4.3.0:

```js
var Jsdom = require("jsdom")
var assert = require("assert")

var reporter = Jsdom.createVirtualConsole()
reporter.on("jsdomError", function(err) { throw err })
var document = Jsdom.jsdom(undefined, {virtualConsole: reporter})
var window = document.defaultView

document.body.addEventListener("click", function(ev) {
  assert(ev.detail === 0)
})

document.body.click()
```

domenic:
@Sebmaster, this is an interesting bug arising from the webidl2js bindings. As-is, the line

```js
    const event = MouseEvent.createImpl(["click", { bubbles: true, cancelable: true }], {});
```

never triggers the actual generated MouseEvent constructor, and thus never gets all the dictionary defaulting goodness. Any ideas on the best fix? Maybe just call the real constructor, then use implForWrapper? Although even that is failing, hmm... I'll upload a PR.

moll:
Just to confirm liveliness, I believe v9.12 has not fixed this yet. My workaround since last year for those in need:

```js
var HTMLElement = require("jsdom/lib/jsdom/living").HTMLElement
var MouseEvent = require("jsdom/lib/jsdom/living").MouseEvent

HTMLElement.prototype.click = function() {
  this.dispatchEvent(new MouseEvent("click", {
[...]

PR #1629, opened 2016-10-16
Attempt at fixing #1559

domenic:
Even with this fix (which is not super-satisfying, since the impl side should ideally only ever have to deal with impls) it looks like only the basic Event values and the MouseEventInit values are getting correctly initialized; the rest are undefined. @Sebmaster, thoughts?


domenic:
We ended up with a very similar problem in the specs, resulting in https://dom.spec.whatwg.org/#concept-event-create. So I should implement that :)

PR #1630, opened 2016-10-16
Defer loading "request" module

zertosh:
`request` is `jsdom`'s largest dependency, accounting for ~20% of its load time. This PR adds a utility to defer requiring the `request` module until it's actually used - reducing load times by ~90ms. For many use cases (like [React](https://facebook.github.io/react/docs/test-utils.html) tests using [Jest](https://facebook.github.io/jest/)), XHR support is not needed (so neither is `request`).

```
# before
$ for i in {1..5}; do node -p "s=Date.now();require('./');Date.now()-s"; done
480
488
478
479
477

# after
$ for i in {1..5}; do node -p "s=Date.now();require('./');Date.now()-s"; done
395
398
381
385
390
```

cc: @cpojer

### DIFF
--- a/lib/jsdom/browser/resource-loader.js
+++ b/lib/jsdom/browser/resource-loader.js
@@ -3,9 +3,8 @@
 const parseContentType = require("content-type-parser");
 const sniffHTMLEncoding = require("html-encoding-sniffer");
 const whatwgEncoding = require("whatwg-encoding");
-const parseDataUrl = require("../utils").parseDataUrl;
+const utils = require("../utils");
 const fs = require("fs");
-const request = require("request");
 const documentBaseURLSerialized = require("../living/helpers/document-base-url").documentBaseURLSerialized;
 const NODE_TYPE = require("../living/node-type");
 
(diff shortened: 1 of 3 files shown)


domenic:
Unfortunately we can't do this, because it causes a 100 ms+ jank on the event loop when that `require` is encountered. 100 ms janks are sucky-but-acceptable during startup, but they are completly unacceptable at runtime, as they can block a server from serving any HTTP requests.

I wish there were a better solution here, like some kind of async require. But we can't just do sync require at runtime.

domenic:
It does look like there are a number of "async require" packages on npm that might be worth investigating. However we'll need to be sure that they don't break the browserified version of jsdom, which could be tricky.

zertosh:
Runtime jank is a very legitimate concern. I couldn't find any comparable libs, and rolling custom code here doesn't seem like the right solution. The ideal solution would be to optimize `request` directly since that's used everywhere - though I'm not sure how receptive they'd be to optimizations.

Thanks for the quick response!

wtgtybhertgeghgtwtg:
What if you switched to a smaller package?  I think `got` has about a fourth of the dependencies `request` does.

domenic:
Yeah, a smaller package would be really nice, but we currently depend on many request features. The ones I know about for sure is all the agent stuff we expose; redirect following; proxy support; GZIP decompression; and cookie jars (with tough-cookie integration). On the other hand I know request also has a lot of code for stuff we don't need, like OAuth signing, UNIX domain sockets, and HAR. I also think we might use the multipart support for XHR but I am not sure.

Maybe over time we can work to ensure we have 100% test coverage of all the features we use of request, and that request is appropriately hidden behind a facade (probably an evolution of resource loader), and at that point swapping out the backend would be easy.

PR #1639, opened 2016-10-29
Fix EventListener once handling

i8-pi:
Don't use array indexes from snapshot since we're modifying in-loop

A minor bug in how "once" event handlers are removed. I've cloned and modified a web platform test to trigger the problem.


test/web-platform-tests/to-upstream/dom/events/AddEventListenerOptions-once.html
```diff
@@ -0,0 +1,44 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>AddEventListenerOptions.once</title>
+<link rel="author" title="Xidorn Quan" href="https://www.upsuper.org">
```
domenic: Can you update to be your name and mailto: or website? (Or just remove the line if you'd prefer)

i8-pi:
Fixed

Issue #1644, opened 2016-11-01
split out FileAPI polyfill into a new module


isonmad:
This'd mean dropping using blobSymbols in a few places to get at the internal node Buffer.

While moving it out of here, should blobToBuffer be renamed to blobToNodeBuffer to be ultra-clear on what it does, as opposed to ArrayBuffer?

domenic:
The prerequisite for this work is moving these classes to the webidl2js infrastructure. That's really necessary before we move them out of the repo and thus out of the easy refactoring path.

I think I tried this once, but got stuck for some reason. The work is in https://github.com/tmpvar/jsdom/tree/filereader-webidl2js.

domenic:
It looks like Blob might be doable using https://github.com/tmpvar/jsdom/commit/fc2f27ec363b48281974807e64d5df7006c5616b but FileReader has some problems according to the commit message there.

For an example of how to move something to a separate module while using webidl2js, but still integrating with jsdom, see jsdom/whatwg-url.

domenic:
jsdom has gone in the opposite direction, preferring to keep everything in a single repo.

PR #1645, opened 2016-11-01
Encapsulate Blob details; fix FileReader

isonmad:
Prerequisite for #1644.


lib/jsdom/living/file-reader.js
```diff
@@ -69,7 +69,7 @@ module.exports = function createFileReader(window) {
     if (self.readyState === self.LOADING) {
       throw new DOMException(DOMException.INVALID_STATE_ERR);
     }
-    if (file[blobSymbols.closed]) {
+    if (file.isClosed) {
```
domenic: This is not correct, since authors can override `isClosed` using `Object.defineProperty`.

lib/jsdom/living/xmlhttprequest.js
```diff
@@ -540,8 +540,8 @@ module.exports = function createXMLHttpRequest(window) {
             // TODO content type; what is the form boundary?
           } else if (body instanceof Blob) {
             flag.body = body[blobSymbols.buffer];
-            if (body[blobSymbols.type] !== "") {
-              contentType = body[blobSymbols.type];
+            if (body.type !== "") {
```
domenic: Again, not correct.

lib/jsdom/living/file.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 
 const fileSymbols = require("./file-symbols");
-const Blob = require("./blob");
+const { Blob } = require("./blob");
```
domenic: jsdom is not yet requiring Node v6, so destructuring is not allowed yet.

lib/jsdom/living/xhr/FormData-impl.js
```diff
@@ -170,9 +169,9 @@ function constructTheFormDataSet(form, submitter) {
 }
 
 function isBlob(obj) {
-  return obj !== null && typeof obj === "object" && blobSymbols.buffer in obj;
+  return obj instanceof Blob;
```
domenic: This is not correct, as `Object.create(Blob.prototype)` now passes, whereas it should not.

lib/jsdom/living/file-reader.js
```diff
@@ -66,9 +66,14 @@ module.exports = function createFileReader(window) {
   });
 
   function readFile(self, file, format, encoding) {
+    if (file instanceof Blob !== true) {
```
domenic: This should be checking for a symbol, instead of instanceof.

lib/jsdom/living/xhr/FormData-impl.js
isonmad: Even using symbols, it means `Object.assign({},new Blob())` would pass. Is perfect encapsulation like the browser does even possible or worth shooting for here, that it's worth sacrificing trying to stick to a public api? There's always one weird trick that bypasses things.

domenic: That's not true; Object.assign doesn't copy symbols. You really have to work at it. Right now we're using symbols as the best balance between performance and unforgeability; in the future maybe we can use real private state.

isonmad: > That's not true

But it is. I just tested it and [the docs agree.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)

domenic:
Superceded by https://github.com/tmpvar/jsdom/commit/f77cc1e3a9eb38fb7db6affa5a9c3d2acca1b351

PR #1650, opened 2016-11-08
Rewrite Blob implementation

TimothyGu:
Don't let the diffstat fool you: there are a lot of things changed by this PR. In fact, there are so many changes, I feel a `<details>` is justified.

This PR isn't exactly completed. New Web Platform Tests still need to be written for many things, but especially

- `size` when the blob is closed
- `slice()` with negative arguments
- inheritance of type (not inherited) and readability state (inherited) in `slice()`

A lot of other changes are mainly just routine Web IDL compliance improvements.

<details><summary>Detailed changelog</summary>

#### General
For functions, I now declare all the optional function arguments inside the parentheses with default values of `undefined` if not specified. This is easier to read (for me anyway), while keeping the function arity intact.

- More closely followed step-by-step correspondence with the spec
  - That does indeed mean that some variable names are changed to the ones used in the spec
- Extraneous `this instanceof Blob` test dropped
  - The test is implicit for Harmony classes
- Any iterable can now be used as `blobParts`
  - https://heycam.github.io/webidl/#es-to-sequence
- Switch on element in `blobParts` is reordered
  - Previously the branch for `Buffer` was dead code as `ArrayBuffer.isView(buffer)` is always true
  - Now the test for `Buffer` is in first, followed by the order dictated by the spec (except for `USVString`, which is a catch-all case)
  - https://w3c.github.io/FileAPI/#constructorBlob
- The conversion from ES `String` to IDL `USVString` is enforced using webidl-conversions
  - https://w3c.github.io/FileAPI/#typedefdef-blobpart
- Special checks for `RegExp` and `Date` for `options` are removed
  - As they've been removed from the spec
  - https://heycam.github.io/webidl/#es-to-dictionary
- The check for non-ASCII-printable characters is refactored into a separate function so that it can be used in `slice`
  - The regex is also simplified

- Return 0 when the blob is closed.

- Negative `start` and `end` are now supported
  - https://w3c.github.io/FileAPI/#dfn-start
  - https://w3c.github.io/FileAPI/#dfn-end
- `start` and `end` are now explicitly converted to number and clamped to `long long`
  - Some nonsensical code now works as expected (e.g. `slice('abc') === slice(0)`)
  - https://w3c.github.io/FileAPI/#ref-for-dfn-slice-1
- Do not inherit type from parent blob
  - https://w3c.github.io/FileAPI/#dfn-contentTypeBlob
- `type` is now directly filled in `slice` instead of going through the constructor
  - https://w3c.github.io/FileAPI/#dfn-contentTypeBlob
- Inherit readability state from parent
  - https://w3c.github.io/FileAPI/#ref-for-readabilityState-8

</details>

### DIFF
--- a/lib/jsdom/living/blob.js
+++ b/lib/jsdom/living/blob.js
@@ -1,37 +1,55 @@
 "use strict";
 
+const conversions = require("webidl-conversions");
 const blobSymbols = require("./blob-symbols");
 
+function getType(type) {
+  if (type === null || type === undefined) {
+    return "";
+  }
+
+  type = conversions.DOMString(type); // eslint-disable-line new-cap
+  if (/[^\u0020-\u007E]/.test(type)) {
+    type = "";
+  }
+  return type.toLowerCase();
+}
+
 module.exports = class Blob {
-  constructor() {
-    if (!(this instanceof Blob)) {
-      throw new TypeError("DOM object constructor cannot be called as a function.");
+  constructor(blobParts = undefined, options = undefined) {
+    this[blobSymbols.closed] = false;
+    this[blobSymbols.type] = "";
+
+    if (arguments.length === 0) {
+      this[blobSymbols.buffer] = new Buffer(0);
+      return;
     }
-    const parts = arguments[0];
-    const properties = arguments[1];
-    if (arguments.length > 0) {
-      if (!parts || typeof parts !== "object" || parts instanceof Date || parts instanceof RegExp) {
-        throw new TypeError("Blob parts must be objects that are not Dates or RegExps");
-      }
+
+    if (!Array.isArray(blobParts) && blobParts[Symbol.iterator]) {
+      blobParts = Array.from(blobParts);
+    }
+    if (!Array.isArray(blobParts)) {
+      throw new TypeError("Blob parts must be an iterable object");
     }
 
     const buffers = [];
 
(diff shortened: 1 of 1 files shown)


domenic:
Sorry for the delay in this! I am able to work on it now.

A couple things come to mind:

- Are any of these changes testable? It looks like several of them should be, and we should be adding tests to make sure they don't regress. See https://github.com/tmpvar/jsdom/blob/master/Contributing.md#tests

- I wish we didn't support `Buffer`, actually; that's not part of the web platform. Can we remove that and make things more to-spec?

TimothyGu:
@domenic, alternatively, we can use webidl2js which would solve a lot of the issues in a better, more automated way. Is there a tutorial on how to convert a module to become IDL-based?

Regarding tests, yes, as I've written in the PR description, they can be written fairly easily. I held off on that in order to get the behaviors right first.

domenic:
Yeah, it would be much nicer if we could move to webidl2js. We haven't written a tutorial yet, although that's on my short-term todo list. I started in https://github.com/tmpvar/jsdom/commit/fc2f27ec363b48281974807e64d5df7006c5616b but got stuck for reasons in that commit message. But I think those reasons don't apply if you *only* try to port Blob...

TimothyGu:
Superseded by #1681.

PR #1651, opened 2016-11-10
Check for client.abort function before executing.

graemeworthy:
line is now identical to the open() method, on #435 in the same file.

### DIFF
--- a/lib/jsdom/living/xmlhttprequest.js
+++ b/lib/jsdom/living/xmlhttprequest.js
@@ -362,7 +362,7 @@ module.exports = function createXMLHttpRequest(window) {
       properties.timeoutFn = null;
       properties.timeoutStart = 0;
       const client = properties.client;
-      if (client) {
+      if (client && typeof client.abort === "function") {
         client.abort();
       }
       if (!(this.readyState === XMLHttpRequest.UNSENT ||


domenic:
Can you provide a test case that fails when this check is not present?

domenic:
Closing due to lack of response, but happy to reopen and merge if we can add a test for this.

eyko:
This blows up for me in a test (Jest) with an url variable that was undefined (so: `$.get(undefined).then(...)`), but the same goes for any other falsy url values.

```
  ● Test suite failed to run
    TypeError: client.abort is not a function

      at XMLHttpRequest.abort (node_modules/jsdom/lib/jsdom/living/xmlhttprequest.js:369:16)
      at Object.abort (node_modules/jsdom/lib/jsdom/living/xhr-utils.js:315:13)
      at RequestManager.close (node_modules/jsdom/lib/jsdom/living/nodes/Document-impl.js:146:21)
      at Window.close (node_modules/jsdom/lib/jsdom/browser/Window.js:362:29)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:169:7)
```

Firing up the debugger shows that client is an EventEmitter but does not have an `abort` fn. I did notice an abort fn under `_events` though.

screen shot 2017-06-22 at 20 43 12

domenic:
Please open a new issue following the issue template; in particular pay attention to the instructions about a minimal example, not involving other libraries like Jest.

PR #1661, opened 2016-11-22
Using HTTP headers from preflight response

fiftin:
Solution for https://github.com/tmpvar/jsdom/issues/1659

### DIFF
--- a/lib/jsdom/living/xhr-utils.js
+++ b/lib/jsdom/living/xhr-utils.js
@@ -278,7 +278,21 @@ exports.createClient = function createClient(xhr) {
     preflightClient.on("response", resp => {
       if (resp.statusCode >= 200 && resp.statusCode <= 299) {
         const realClient = doRequest();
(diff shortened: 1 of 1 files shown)


domenic:
I'm going to merge this despite it not having tests, since testing CORS is really hard :-/.

domenic:
Oh, actually, we can't merge this, since the addition of User-Agent to simple headers is against spec: https://fetch.spec.whatwg.org/#cors-safelisted-request-header

domenic:
Closing per https://github.com/tmpvar/jsdom/issues/1659#issuecomment-267842178

mbroadst:
@domenic it appears that part of this PR _should_ indeed be merged in. We're running jsdom tests against our local server, and failing miserably on CORS-related issues because the preflight headers aren't being forwarded to the request. I'm not sure whether adding `"user-agent"` to `simpleHeaders` is a good idea, but the second part of the commit does the job for us

Issue #1662, opened 2016-11-23
XHR throws before issuing CORS application/json POST request with basic authorization


dunnock:
XMLHttpRequest when POSTing not empty body to other site (CORS) with authorization header (supposedly with any not simple header) throws exception even not sending request to the server:

```
    TypeError: Cannot convert undefined or null to object

      at Object.getRequestHeader (node_modules/react-scripts/node_modules/jsdom/lib/jsdom/living/xhr-utils.js:19:23)
      at setDispatchProgressEvents (node_modules/react-scripts/node_modules/jsdom/lib/jsdom/living/xmlhttprequest.js:908:38)
      at XMLHttpRequest.send (node_modules/react-scripts/node_modules/jsdom/lib/jsdom/living/xmlhttprequest.js:700:11)
```

The issue is that on xmlhttprequest.js:908 'client' is EventEmitter, while code expected request client.

Following commit reproduces and fixes the issue 1e72e4f08f7f2228de860e9f3a78ae8e47360840

Drawback is it still won't provide progress on data upload, while fixing exception. I suppose proper fix would be to store final headers somewhere in the xhr object, while keeping properties.client only as EventEmitter..

PR #1663, opened 2016-11-23
bugfix: #1662 XHR throws before issuing CORS application/json POST re…

dunnock:
Solution to #1662:

XMLHttpRequest when POSTing not empty body to other site (CORS) with authorization header (supposedly with any not simple header) throws exception even not sending request to the server. This kind of requests might be used to communicate with local middleware via REST with authorization, especially when system is built of micro services.

```
    TypeError: Cannot convert undefined or null to object

      at Object.getRequestHeader (node_modules/react-scripts/node_modules/jsdom/lib/jsdom/living/xhr-utils.js:19:23)
      at setDispatchProgressEvents (node_modules/react-scripts/node_modules/jsdom/lib/jsdom/living/xmlhttprequest.js:908:38)
      at XMLHttpRequest.send (node_modules/react-scripts/node_modules/jsdom/lib/jsdom/living/xmlhttprequest.js:700:11)
```

The issue is that on xmlhttprequest.js:908 'client' is EventEmitter, while code expected request client.

Solution still won't provide progress on data upload for CORS POST, but will fix exception and enable such kind of requests. I suppose proper fix would be to store final headers somewhere in the xhr object, while keeping properties.client only as EventEmitter..

to-upstream test is pointing to hosted w3c-test.org since local python test server is not running while upstream tests executing.

### DIFF
--- a/lib/jsdom/living/xmlhttprequest.js
+++ b/lib/jsdom/living/xmlhttprequest.js
@@ -905,7 +905,7 @@ module.exports = function createXMLHttpRequest(window) {
 
     let total = 0;
     let lengthComputable = false;
-    const length = parseInt(xhrUtils.getRequestHeader(client.headers, "content-length"));
+    const length = client.headers && parseInt(xhrUtils.getRequestHeader(client.headers, "content-length"));
     if (length) {
       total = length;
       lengthComputable = true;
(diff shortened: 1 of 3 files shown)


domenic:
Thanks, this is really great! Ideally we'd be running the to-upstream tests under the Python server, but for now this is probably the way to go...

PR #1666, opened 2016-12-01
Submitting form elements should log errors if and only if not cancelled

rolftimmermans:
This logic was backwards, throwing errors when `event.preventDefault()` was called on a form submit event:

```
Error: Not implemented: HTMLFormElement.prototype.submit
```

When `event.preventDefault()` was **not** called, no error was logged.


domenic:
Great find and thanks especially for the tests!

PR #1674, opened 2016-12-13
Update whatwg-url to 4.0

danez:
The two methods removed in 4.0 are not used in jsdom.


domenic:
Thanks! Updated to 4.1.0 in a separate commit :)

PR #1676, opened 2016-12-16
Support `insertAdjacentHTML` without a parent element

alexlafroscia:
I was trying to get [Glimmer](https://github.com/tildeio/glimmer) running in JSDOM and ran into an issue where it attempts to use `insertAdjacentHTML` with an element that has no parent element. After digging into the code I was that JSDOM was set up to always error if the given element does not have a parent.

I did some research and it seems like `insertAdjacentHTML` should only require a parent element when inserting `beforebeing` or `afterend`, since that actually puts the text before or after the element that the method is called on.  However, `afterbegin` and `beforeend` do not require a parent, because the text is placed within the element itself.

The docs from MDN can be found [here](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML)

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -426,7 +426,7 @@ class ElementImpl extends NodeImpl {
   insertAdjacentHTML(position, text) {
     position = position.toLowerCase();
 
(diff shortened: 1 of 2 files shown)


domenic:
Thanks! I'll touch this up and turn it into something mergeable. Unfortunately your change doesn't actually work, since you are testing `text` instead of `position`.

PR #1681, opened 2016-12-19
Port File API to Web IDL

TimothyGu:
This is an alternative to #1650. The same changes in #1650 are applied to Blob, which is ported to Web IDL.

The issue raised in fc2f27ec363b48281974807e64d5df7006c5616b, namely:

> This whole branch of `FileReader` etc. doesn't work because `EventTarget`s depend on an `_ownerDocument` to dispatch inline event listeners. In general we need a reference from any given `EventTarget` to its creator window, but that goes against jsdom's whole architecture

is worked around with using a proxy-like method in 720b1021cf0b6a4aee90cee68005669c160918b2. It ain't pretty, but it works, and it brings about other standard compliance-related benefits.

Still no tests for now.


TimothyGu:
Ideally, we should add a custom attribute to webidl2js to make it generate not a class, but rather a class factory that looks like the one added in 720b1021cf0b6a4aee90cee68005669c160918b2. For now, however, this should work.

lib/jsdom/living/file-api/File.idl
```diff
@@ -0,0 +1,12 @@
+[Constructor(sequence<BlobPart> fileBits,
+            [EnsureUTF16] DOMString fileName,
```
domenic: This is a spec bug; should be https://github.com/w3c/FileAPI/issues/9

lib/jsdom/living/file-api/File-impl.js
```diff
@@ -0,0 +1,15 @@
...
+
+exports.implementation = class FileImpl extends BlobImpl {
+  constructor(args, privateData) {
+    const fileBits = args[0];
+    const fileName = args[1];
+    const options = args[2];
+    super([fileBits, options], privateData);
+
+    this.name = fileName.replace(/\//g, ":");
+    this.lastModified = "lastModified" in options ? options.lastModified : Date.now();
```
domenic: I am not sure but I think it might always be present due to the dictionary normalization webidl2js performs. Tests would be good.

lib/jsdom/browser/Window.js
```diff
@@ -261,7 +260,26 @@ function Window(options) {
     return result;
   };
 
-  this.FileReader = createFileReader(this);
+  // this will have to make do until webidl2js gets proper support
+  const origFileReader = this.FileReader;
+  function FileReader() {
+    origFileReader.call(this);
+
+    Object.defineProperty(this, "_ownerDocument", {
```
domenic: Add a comment explaining that our event dispatch depends on this

snuggs:
Great work @TimothyGu @domenic. @TimothyGu  you maintain Pugjs? I'm not only a user...also an evangelist (of pug & jsdom) 😎

domenic:
@TimothyGu if I can beg your forgiveness for the latency on all this, would you mind finishing this up? webidl2js v5.1.0 with your new feature should be published ASAP (as soon as @Sebmaster wakes up probably). So the remaining changes I think are:

- USVString spec bug
- Use new WebIDL2JSFactory feature
- Add or find web platform tests for the normative fixes you've made (besides just the IDL compliance stuff)

I'm happy to help with any of this (e.g. the tests), just let me know.

TimothyGu:
@domenic, done.

domenic:
@TimothyGu any thoughts on the failing test?

```
Domenic@Revan MINGW64 ~/Dropbox/GitHub/tmpvar/jsdom (TimothyGu-filereader-webidl2js)
$ npm run test-tuwpt -- --fgrep formdata-constructor

> jsdom@9.9.1 test-tuwpt C:\Users\Domenic\Dropbox\GitHub\tmpvar\jsdom
> mocha test/web-platform-tests/to-upstream.js "--fgrep" "formdata-constructor"



  Local tests in Web Platform Test format (to-upstream)
    1) XMLHttpRequest/formdata-constructor.html


  0 passing (159ms)
  1 failing

  1) Local tests in Web Platform Test format (to-upstream) XMLHttpRequest/formdata-constructor.html:
     Error: Failed in "test that FormData is correctly constructed from the form data set":
assert_array_equals: file-1 lengths differ, expected 1 got 2

    at Test.test (http://127.0.0.1:55942/XMLHttpRequest/formdata-constructor.html:51:5)
```

TimothyGu:
> any thoughts on the failing test?

Seems like it was an oversight in 85e7e9b30566116a625aaa09841efb7a45ef148a, as we cannot pass the constructor arguments into `Array`. However, simply reverting would trigger ESLint's no useless argument rule, and removal of the arguments would trigger a useless-constructor error in ESLint (since the body would look like `constructor() { super(); }`). Therefore, I added 0942d20 which should trick ESLint into accepting it.

domenic:
Ah of course, I should have realized it would have been one of my follow-up commits that introduced the error. \o/!

Issue #1636, opened 2016-10-26
Window.close windowToClean is undefined


tsirolnik:
While calling window.close I get the following error (sometimes).

``` js
/home/xxxxx/node_modules/jsdom/lib/jsdom/browser/Window.js:290
      for (let i = 0; i < windowToClean.length; i++) {

TypeError: Cannot read property 'length' of undefined
    at windowCleaner (/home/xxxxx/node_modules/jsdom/lib/jsdom/browser/Window.js:290:40)
    at windowCleaner (/home/xxxxx/node_modules/jsdom/lib/jsdom/browser/Window.js:291:9)
    at Window.close (/home/xxxxx/node_modules/jsdom/lib/jsdom/browser/Window.js:298:6)
```

I've wrapped window.close with the following (I still get the error that I've pasted) 
So, close is not being called twice. - 

``` js
let close = window.close;
window.closed = false;
window.close = function() {
  if(!window.closed) {
    window.closed = true;
    close.bind(window)();
  }
}
```

domenic:
A standalone repro case would be very helpful.

tsirolnik:
This is an issue as the full code is being used on production. 
But, for the case, I'm using a browser a [tiny browser emulator based on jsdom](https://github.com/tsirolnik/undead), this issue happens when I'm creating a lot of instances (for testing purposes) and when they are being closed on different times.

PR #1694, opened 2016-12-28
Fixing late window close error

tsirolnik:
I'm getting the following error on a production server of mine, and the only fix that seems to work is the simple implemented workaround.
`
TypeError: Cannot read property 'length' of undefined
  at windowCleaner (/app/node_modules/jsdom/lib/jsdom/browser/Window.js:290:40)
  at windowCleaner (/app/node_modules/jsdom/lib/jsdom/browser/Window.js:291:9)
  at Window.close (/app/current/node_modules/jsdom/lib/jsdom/browser/Window.js:298:6)
`

It seems to happen due to late living references to window trying to call close on it, close being closed twice or something like that. This a simple solution that won't cause any problem.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -287,6 +287,9 @@ function Window(options) {
     // Recursively close child frame windows, then ourselves.
     const currentWindow = this;
     (function windowCleaner(windowToClean) {
+      if(!windowToClean) {
+        return;
+      }
       for (let i = 0; i < windowToClean.length; i++) {
         windowCleaner(windowToClean[i]);
       }


dmethvin:
Are you doing something like what was described in #1636 maybe?

tsirolnik:
@dmethvin  Not anymore, that's why I'm still annoyed by this problem. I'm already even checked whether I've called close before or not.

As you can see I have this problem since October, and since then I couldn't figure out what's the source of the problem and it seems like this workaround can fix it.

domenic:
We really need tests before we can accept this, to ensure we understand the source of the problem and do not regress it in the future.

tsirolnik:
I understand, I wish I could've created a case for that problem but it happens only in production (Same node version, same package version and etc) and it seems like a kind of race condition or something like that

tsirolnik:
@domenic  Mind adding a snippet of an example test for this case? I want to write them by myself to help finishing this pull request

domenic:
@tsirolnik well, I can't know how to reproduce the situation, so the important thing is you providing some code that fails before this patch and succeeds after. But to turn that into a test, I'd put it somewhere in https://github.com/tmpvar/jsdom/blob/master/test/jsdom/misc.js , following that style.

domenic:
Closing due to lack of repro case. Happy to reopen if we can get one.

liiza:
@domenic 
We were able to reproduce this calling `window.close()` on dom that had an iframe. We are using version `^9.12.0` of jsdom.

I try to come up with more detailed reproduce instructions.

attaab:
@liiza i am not able to reproduce this calling window.close() on dom that had an iframe. I am using version ^9.12.0 of jsdom too. what do you think i can do?

Issue #1711, opened 2017-01-23
clientWidth value is undefined instead of 0


alistairjcbrown:
## Problem
The value retrieved for the `clientWidth` property is `undefined` in JSDOM. This does not mirror the expected behaviour as found in browser implementations. It is also contrary to details provided in another issue:

> It does not yet perform layouting however, for example node.clientWidth will always be 0
~ https://github.com/tmpvar/jsdom/issues/1274

The corresponding spec: 
> The clientWidth attribute must run these steps:
>
> 1. If the element has no associated CSS layout box or if the CSS layout box is inline, return zero.
>
~ https://drafts.csswg.org/cssom-view/#dom-element-clientwidth

__Note:__ The same could also be said for `clientTop`, `clientBottom` and `clientHeight`, which should return 0

## Data
The snippets below show getting the client width value for newly created elements which are not part of the DOM.

### JSDOM
```js
import jsdom from 'jsdom';
const document = jsdom.jsdom('<!doctype html><html><body></body></html>', { url: 'http://localhost' });

const myDiv = document.createElement('div');
console.log('myDiv.clientWidth', myDiv.clientWidth);
// myDiv.clientWidth undefined
```

### Chrome (55.0.2883.95 (64-bit))
```js
const myDiv = document.createElement('div');
console.log('myDiv.clientWidth', myDiv.clientWidth);
// myDiv.clientWidth 0
```

### Firefox (50.0)
```js
const myDiv = document.createElement('div');
console.log('myDiv.clientWidth', myDiv.clientWidth);
// myDiv.clientWidth 0
```

### Safari (9.1.3 (9537.86.7.8))
```js
const myDiv = document.createElement('div');
console.log('myDiv.clientWidth', myDiv.clientWidth);
// myDiv.clientWidth 0
```

PR #1712, opened 2017-01-23
Set 0 defaults for element client and scroll properties

alistairjcbrown:
The value retrieved for client properties (`clientWidth` `clientTop`, `clientBottom` and `clientHeight`) and scroll properties (`scrollWidth`, `scrollHeight`, `scrollTop` and `scrollLeft `) are `undefined` in JSDOM. This does not mirror the expected behaviour as found in browser implementations.

The corresponding spec for client properties: 
> The clientWidth attribute must run these steps:
>
> 1. If the element has no associated CSS layout box or if the CSS layout box is inline, return zero.
>
~ https://drafts.csswg.org/cssom-view/#dom-element-clientwidth

The same is also written for `clientTop`, `clientBottom` and `clientHeight`, which too should return 0.
There are more details logged in the issue #1711.

The corresponding spec for scroll properties:
> The scrollTop attribute, on getting, must return the result of running these steps:
>
> 2. If document is not the active document, return zero and terminate these steps.
>
~ https://drafts.csswg.org/cssom-view/#dom-element-scrolltop

The same is also written for `scrollWidth`, `scrollHeight` and `scrollLeft `, which too should return 0.

This PR change updates the `Element` implementation to provide `0` values for these eight properties, and adds a new test file for testing "layouting" to confirm that these properties, and the existing `getClientRects` and `getBoundingClientRect` functions, return zero as default.


alistairjcbrown:
The first two Travis jobs failed, but these failures seem unrelated to the change in this PR (the third job passed). I'm going to close and re-open this PR to trigger a new build and see if the same errors occur.

alistairjcbrown:
Same error on second build - not sure if expected or if due to a change in this PR?

alistairjcbrown:
I've updated this PR to also cover the scroll properties which were not being exposed as zero values.

test/living-dom/node-layouting.js
```diff
@@ -0,0 +1,98 @@
+"use strict";
```
domenic: As noted in the contributing.md, please do not add tests in this format. Write web-platform-tests instead.

alistairjcbrown: @domenic Ah, thanks and apologies. I've rewritten these as web platform tests and I'll push them up now - please let me know if I've missed anything.

alistairjcbrown:
@domenic Thanks for the feedback! I've corrected the help URLs, phrasing and moved to a new folder.

I've also updated the `getClientRects` implementation to return an empty array, and updated the test I'd added to check for that. Checking Firefox and Chrome, the list provided by `getClientRects` has an `item` function on it. I haven't implemented that as part of this.

Let me know if there's any other work that's needed for this PR 👍

fkling:
This breaks one of my tests where I set `document.clientWidth` to a specific value:

> TypeError: Cannot set property clientWidth of [object Object] which has only a getter

It seems I also cannot overwrite the getter with `Object.defineProperty`. Any suggestion what to do here? Is it possible to set `document.clientWidth` some other way (doesn't look like it)? If not, should/can we allow this?

PR #1734, opened 2017-02-11
Enable FormData being iterable

### LINKED ISSUE #1671: FormData is missing some methods
If I'm understanding this correctly, looking at this file:

https://github.com/tmpvar/jsdom/blob/a6acac4e9dec4f859fff22676fb4e9eaa9139787/lib/jsdom/living/xhr/FormData.idl

Seems like FormData supports _append_, _delete_, _get_, _getAll_, _has_ and _set_ methods. But, according to this reference:

https://developer.mozilla.org/en-US/docs/Web/API/FormData

It should also support _entries_, _keys_ and _values_. Seems like these methods are also part of living standard. Is it planned to add them in some future iteration?

domenic:
Now that this is supported in webidl2js, we can use it here. Closes #1671.

/cc @TimothyGu \o/

### DIFF
--- a/lib/jsdom/living/xhr/FormData-impl.js
+++ b/lib/jsdom/living/xhr/FormData-impl.js
@@ -69,7 +69,8 @@ exports.implementation = class FormDataImpl {
 
   * [Symbol.iterator]() {
     for (const entry of this._entries) {
-      yield [entry.name, entry.value];
+      const wrapper = idlUtils.wrapperForImpl(entry);
+      yield [wrapper.name, wrapper.value];
     }
   }
 };
--- a/lib/jsdom/living/xhr/FormData-impl.js
+++ b/lib/jsdom/living/xhr/FormData-impl.js
@@ -66,6 +66,12 @@ exports.implementation = class FormDataImpl {
       this._entries.push(entry);
     }
   }
+
+  * [Symbol.iterator]() {
+    for (const entry of this._entries) {
+      yield [entry.name, entry.value];
+    }
+  }
 };
 
 function createAnEntry(name, value, filename) {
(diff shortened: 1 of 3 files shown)


lib/jsdom/living/xhr/FormData-impl.js
```diff
@@ -66,6 +66,12 @@ exports.implementation = class FormDataImpl {
       this._entries.push(entry);
     }
   }
+
+  * [Symbol.iterator]() {
+    for (const entry of this._entries) {
+      yield [entry.name, entry.value];
```
TimothyGu: You might have to manually get the value's wrapper, since webidl2js doesn't try to do that yet. Once that's fixed in webidl2js, we can probably just `return this._entries.entries()` here

domenic:
I'm not sure why this is failing on Node v4; this syntax should be supported. Maybe just another reason to move to Node v6 though...

lib/jsdom/living/xhr/FormData-impl.js
```diff
@@ -66,6 +66,13 @@ exports.implementation = class FormDataImpl {
       this._entries.push(entry);
     }
   }
+
+  * [Symbol.iterator]() {
+    for (const entry of this._entries) {
+      const wrapper = idlUtils.wrapperForImpl(entry);
+      yield [wrapper.name, wrapper.value];
+    }
```
TimothyGu: Hmm... I thought you might have to do 

```js
for (const entry of this._entries) {
  yield [entry.name, idlUtils.wrapperForImpl(entry.value)];
}
```
?

domenic: Oh, right, that makes more sense... I guess this really needs tests.

PR #1741, opened 2017-02-21
New API draft

domenic:
@Sebmaster @Joris-van-der-Wel @TimothyGu would love your thoughts on this. Not quite ready to merge, but ready enough for review.

I've had this cooking for a long time. Unfortunately I think that my dream of getting a perfect new API out and completely killing the old API is not achievable in one stroke. The two main barriers are:

- Designing a new resource loader / request options system. I have some notes and a direction on this, but it's a big undertaking
- Converting all the tests to not rely on the old API (and ideally converting as many as possible to web platform tests)

So this weekend I worked toward getting something ready that could be usable by most consumers, with the idea being that we keep the old API around for a while until we've gotten through the above, and everyone can use it.

As such, the things to do before merging are:

- Add encoding-sniffing to fromURL/fromFile; maybe even making the JSDOM constructor accept a Buffer/Uint8Array. This is not too hard and it'd be sad to lose this feature.
- Do a big rename, so that newapi1 becomes the default, and the old API moves to e.g. `require("jsdom/oldapi")` or similar.
- Move "jsdom vs. PhantomJS" from the readme into a wiki page and link to it.

But I'd love to get some review on the code, and especially the docs.

Note that we'll also start requiring Node v6 as part of this breaking change.


domenic:
Karma is stuck on Chrome 48 I guess? That's not great...

lib/newapi1.js
```diff
@@ -0,0 +1,305 @@
...
+        headers: {
+          "User-Agent": options.userAgent,
+          Referer: options.referrer,
+          Accept: "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
+          "Accept-Language": "en"
+        },
+        jar: wrapCookieJarForRequest(options.cookieJar)
+      };
+
+      return request(url, requestOptions).then(res => {
```
Sebmaster: Just return and start a new outer-scope `.then`?

I guess that'd mean pulling out options into the outer scope which makes them throw?

lib/newapi1.js
```diff
@@ -0,0 +1,305 @@
...
+
+  if (normalized.contentType === undefined) {
+    const extname = path.extname(filename);
+    if (extname === ".xhtml" || extname === ".xml") {
+      normalized.contentType = "application/xhtml+xml";
+    }
+  }
+
+  if (normalized.url === undefined) {
+    normalized.url = new URL("file:" + path.resolve(filename));
```
Sebmaster: Wow. URL supports this for Win and Linux. This is amazing!

lib/newapi1.js
```diff
@@ -0,0 +1,305 @@
...
+    if (transformed.runScripts !== "dangerously" && transformed.runScripts !== "outside-only") {
+      throw new RangeError(`runScripts must be undefined, "dangerously", or "outside-only"`);
+    }
+  }
+
+  if (options.beforeParse !== undefined) {
+    transformed.beforeParse = options.beforeParse;
+  }
+
+  // concurrentNodeIterators?? deferClose?? parser??
```
Sebmaster: Let's just drop `deferClose`? If someone needs it they can just create an empty instance and then write their own HTML to it. It's a quite involved use-case and I'm not sure we need to maintain the current simplicity necessarily.

Parser shouldn't be configurable... ever anymore?

lib/newapi1.js
```diff
@@ -0,0 +1,305 @@
...
+
+  return transformed;
+}
+
+function normalizeHTML(html) {
+  if (html === undefined) {
```
Sebmaster: Check for null as well?

TimothyGu:
I'm not familiar enough with the inner workings of JSDOM to give a constructive code review, but for the issue of resource loader, is it possible to have a service worker-like mechanism? It isn't necessary to implement a cache or anything like that, and `Request` and `Response` should be made less opaque than they are in the browser, but it seems like a good framework to build upon?

domenic:
Yeah, something service worker-like would be the dream, definitely. Maybe just implementing the Request/Response classes would be a good start, although, I guess that gets tricky as it then drags along ReadableStream and everything...

lib/newapi1.js
domenic: Yeah, no great solution here until async methods arrive, it seems.

lib/newapi1.js
domenic: Nah, only undefined should trigger default behavior in JS APIs. I'll add a test explicitly that `null` becomes the string `"null"`.

domenic:
Given https://nodejs.org/en/blog/release/v7.6.0/ (i.e., Node v7.6.0 comes with async/await) I am seriously considering making Node v7.6.0 a baseline requirement for jsdom v10. What do people think?

lib/newapi1.js
```diff
@@ -0,0 +1,319 @@
...
+
+  return transformed;
+}
+
+function normalizeHTML(html = "") {
+  let encoding = "UTF-8";
+
+  if (ArrayBuffer.isView(html)) {
+    html = html.buffer;
+  }
```
TimothyGu: This doesn't work if the view has a `bufferOffset`. Check `Blob`'s [implementation](https://github.com/tmpvar/jsdom/blob/79f5522c13a2b900768183ca425ba90f90f9708d/lib/jsdom/living/file-api/Blob-impl.js#L32) for proper handling.

package.json
```diff
@@ -28,7 +28,9 @@
     "html-encoding-sniffer": "^1.0.1",
     "nwmatcher": ">= 1.3.9 < 2.0.0",
     "parse5": "^1.5.1",
+    "pn": "^1.0.0",
     "request": "^2.79.0",
+    "request-promise-native": "^1.0.3",
```
wtgtybhertgeghgtwtg: Why a `request` variant?  I thought [`jsdom` had perf problems with it](https://github.com/tmpvar/jsdom/pull/1630).

domenic: That's unrelated to this pull request.

domenic:
Rebased on 9.12.0. Left to do:

- Get all tests passing in all environments; we should be close.
- Move "jsdom vs. PhantomJS" from the readme into a wiki page and link to it.
- Ensure fragments are preserved in redirects for fromURL(), similar to #1762.
- Make JSOM.fragment() invocations share a document for speed.
- Test how subresource loading works by default; this is currently a coverage gap and I don't even know how it's configured to work.
  - Maybe by default we should not load any subresources.
  - If so, how should we allow opt-in to subresource loading? A future-proof variant might be `resourceLoader: jsdom.ResourceLoader.all`, where in the future `jsdom.ResourceLoader` becomes a constructible thing you can use to set policy. That's kind of verbose though; unsure what to do here.

domenic:
Latest round of commits leaves us with only one task: figuring out how to allow opt-in to loading all resources. I think it's important to allow that to be done easily on day 1 of newapi. Currently I am thinking a `resources: "all"` option. In the future you'll be able to do `resources: new jsdom.ResourceLoader(...)` to customize the process. But the special values `"all"` and `undefined` will give particular behavior (load all resources and load no resources).

domenic:
This should be ready to merge. I probably won't do so until next weekend, or at least later in the week. In the meantime if anyone wants to do code review, especially of the readme, that'd be swell :). (Ignore all the test file renames/changes except the new ones introduced in test/api...)

PR #1743, opened 2017-02-21
Allow null in options

alan-mroczek:
Allow passing null to options and assign empty object if so.


domenic:
`undefined` is the correct way to signify a missing argument in JavaScript and trigger default behavior, not `null`.

PR #1744, opened 2017-02-21
add localAddress option

graduss:


domenic:
I don't think we want to add more configuration to our resource loading in the old API without a very good reason. Can you explain this use case in more detail, why it is common enough to extend the jsdom public API, and why it must be accomplished through the addition of a new API?

graduss:
My server has some interfaces and this option is needed for select which the interface should be using.

domenic:
OK. I don't think we'll add this then, as it sounds very specific to one use case and not a widespread need. Unless you can explain in more detail why this is so important.

PR #1751, opened 2017-03-01
Parse colSpan/rowSpan attributes as integers in getters

aweary:
Resolves https://github.com/tmpvar/jsdom/issues/1745

Also, fixes an issue where `colSpan` would return `'0'` when after the value was set to an invalid value (`td.colSpan = "foo"`);

### DIFF
--- a/lib/jsdom/living/nodes/HTMLTableCellElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLTableCellElement-impl.js
@@ -34,8 +34,8 @@ class HTMLTableCellImpl extends HTMLElementImpl {
   }
 
   get rowSpan() {
-    const value = this.getAttribute("rowspan");
-    return value === null ? 1 : value;
+    const value = parseInt(this.getAttribute("rowspan"));
+    return value === null || value === 0 ? 1 : value;
   }
 
   set rowSpan(V) {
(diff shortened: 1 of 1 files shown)


aweary:
I've also found an issue where setting `colSpan` to a negative integer will cause what appears to be an overflow error:

```js
var td = document.createElement('td');
td.colSpan = -1;
td.getAttribute('colspan');
// 4294967295
```
It should set the attribute to `'0'` . I'm not sure where in the codebase integer attributes are being parsed, but I'm assuming this would affect more than just `colSpan`

domenic:
Thanks for working on this! However, we need a test before we can merge it. See contributing.md for the guidelines there.

domenic:
What should be implemented here is https://html.spec.whatwg.org/#reflect, in particular the paragraph

> If a reflecting IDL attribute has an unsigned integer type (unsigned long) then, on getting, the content attribute must be parsed according to the rules for parsing non-negative integers, and if that is successful, and the value is in the range 0 to 2147483647 inclusive, the resulting value must be returned. If, on the other hand, it fails or returns an out of range value, or if the attribute is absent, the default value must be returned instead, or 0 if there is no default value. On setting, first, if the new value is in the range 0 to 2147483647, then let n be the new value, otherwise let n be the default value, or 0 if there is no default value; then, n must be converted to the shortest possible string representing the number as a valid non-negative integer and that string must be used as the new content attribute value.

(where the default value is 1)

domenic:
Closing due to inactivity and lack of tests.

PR #1754, opened 2017-03-04
Enable support for canvas-prebuilt package

asturur:
canvas-prebuilt is a package from @chearon https://github.com/chearon/node-canvas-prebuilt, contributor of node-canvas.

Is easier to install for most of the situation and is same as canvas, but prebuilt.

Is a drop in replacement with a different name, i made this PR to enable the us of this package in jsdom.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ addons:
 
 matrix:
   include:
-  - node_js: 4
+    - node_js: 4
       env: TEST_SUITE=node-canvas-prebuilt
       script: "export CXX=g++-4.8 && npm i canvas-prebuilt && npm test"
       addons:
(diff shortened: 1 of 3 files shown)


asturur:
@domenic  any thought about this?

asturur:
@tmpvar, @Sebmaster anyone?

domenic:
Sorry, we're all busy and mostly work on jsdom on weekends. I wasn't able to do so last weekend.

asturur:
no worries! I ll wait the weekend.

domenic:
This is really great stuff! Thanks especially for adding a CI test configuration for it.

I was wondering if we should just always include canvas-prebuilt, but it looks like according to https://www.npmjs.com/package/canvas-prebuilt that would not be great for every platform, so the approach you went with here makes sense to me.

asturur:
Super! this will ease also fabricjs installations.

PR #1756, opened 2017-03-04
Blob toString no longer returns '[object Blob]'

woodb:
With the merging of tmpvar/jsdom#1681, it looks like [`toString` was dropped](https://github.com/tmpvar/jsdom/pull/1681/files#diff-2b1f271ea8bbc0be11ca9f49d84564b3L70), causing [certain code that checks for Blob types using `toString`](https://github.com/mzabriskie/axios/blob/master/lib/utils.js#L117-L125) to fail, as it's currently returning `[object Object]`.

I've submitted a failing test in this PR to demonstrate.

### DIFF
--- a/test/web-platform-tests/to-upstream/XMLHttpRequest/formdata-set-blob.html
+++ b/test/web-platform-tests/to-upstream/XMLHttpRequest/formdata-set-blob.html
@@ -41,6 +41,14 @@ test(() => {
 }, "blob with custom name");
 
 test(() => {
+  formData.set("blob-4", new Blob());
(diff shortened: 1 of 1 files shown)


domenic:
Thanks for finding this. I think the best thing to do here will be to wait until our next major release, when we drop Node.js v4 support in favor of Node.js v6. Then we can use `Symbol.toStringTag` and fix this once and for all, for all jsdom classes.

domenic:
Let's roll this into https://github.com/tmpvar/jsdom/issues/1809

Issue #1759, opened 2017-03-07
Option constructor support


NAlexPear:
As far as I can tell, `jsdom` does not support the [Option constructor](https://html.spec.whatwg.org/#dom-option) as expected. The `document.createElement()` version seems to work fine, and is generally my preference anyway, but both should be available according to the WHATWG spec.

My expectation would be that the `Option` constructor would be available directly from the `window`, as it is in Chrome and Firefox latest. 

To test:
```
var window = jsdom.jsdom().defaultView;

console.log( window.Option ); // undefined, expect HTMLOptionElement constructor
```

I suspect that this issue might be fixed in #1723, as well, but I wanted to make sure that this was documented or addressed!

domenic:
Good catch. This would be a fairly easy thing to fix as a PR if you're interested. You'd insert the code aronud https://github.com/tmpvar/jsdom/blob/2161116160f81d93c2891e9290fc74d6b5f6188b/lib/jsdom/browser/Window.js#L187 based on the spec at https://html.spec.whatwg.org/#the-option-element and https://html.spec.whatwg.org/#dom-option. You'd add to-upstream web platform tests as well.

PR #1760, opened 2017-03-08
Add Option element constructor

NAlexPear:
This adds an Option constructor that adheres to the [current spec for Option constructors](https://html.spec.whatwg.org/#dom-option), available through `jsdom`'s `window`/`defaultView`. 

This has been tested against the specs included in this PR, as well as the real-life use case that led me to raise #1759. 

Please let me know if the location of the tests added to the to-upstream directory are organized correctly.


domenic:
Thank you! This isn't quite correct in a few ways, but instead of back-and-forthing on code review, I'll just merge this and do a follow-up commit correcting the issues :)

PR #1762, opened 2017-03-11
Preserve window.location.hash across redirects

josephfrazier:
When `jsdom.env` is passed a url that does a 302 redirect, any hash on the
url is lost. You can check browser behavior here:
https://httpbin.org/redirect-to?url=https://example.com#hash

<strike>So far, this pull request consists only of a change in the
tests that demonstrates this issue. However, I think I'm making some
progress towards a fix, so hopefully I'll be able to update this once I
find one.</strike>

This change ensures that the url fragment is not lost when a redirect happens.

---

Relevant Specifications (found via [stack overflow](https://stackoverflow.com/questions/2286402/url-fragment-and-302-redirects)):

* https://www.w3.org/TR/cuap#uri

> Suppose that a user requests the resource at http://www.w3.org/TR/WD-ruby/#changes and the server redirects the user agent to http://www.w3.org/TR/ruby/. Before fetching that latter URI, the browser should append the fragment identifier #changes to it: http://www.w3.org/TR/ruby/#changes.

* https://tools.ietf.org/html/rfc7231#section-7.1.2

> If the Location value provided in a 3xx (Redirection) response does
not have a fragment component, a user agent MUST process the
redirection as if the value inherits the fragment component of the
URI reference used to generate the request target (i.e., the
redirection inherits the original reference's fragment, if any).

### DIFF
--- a/lib/jsdom.js
+++ b/lib/jsdom.js
@@ -252,6 +252,8 @@ exports.env = exports.jsdom.env = function () {
       agentOptions: config.agentOptions
     };
 
+    const fragment = whatwgURL.parseURL(config.url).fragment;
+
     return resourceLoader.download(config.url, options, (err, responseText, res) => {
       if (err) {
         reportInitError(err, config);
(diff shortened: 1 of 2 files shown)


domenic:
Very interesting find. We're going to be changing the API soon (see #1741), but I'll be sure to keep this in that work as well.

PR #1763, opened 2017-03-11
Bug: stopImmediatePropagation not working

Sebmaster:
The `stopImmediatePropagation` event seems to be [implemented in jsdom](https://github.com/tmpvar/jsdom/blob/578e2cb3f16c760f26e737299a071ff78ec84589/lib/jsdom/living/events/Event-impl.js#L51). It should stop executing listeners for the same element, but it seems this isn't working as expected in jsdom (this assertion is ok in the browser):

```js

jsdom.env('<html><body></body></html>', (err, window) => {
  let i = 0;
  const body = window.document.body;
  body.addEventListener('test', evt => {
    i++;
    evt.stopImmediatePropagation();
  });
  body.addEventListener('test', evt => {
    i++;
  });
  const event = new window.Event('test');
  body.dispatchEvent(event);
  assert(i === 1);
});
```

On the other hand, `stopPropagation` works fine:

```js

jsdom.env('<html><body><div></div></body></html>', (err, window) => {
  let i = 0;
  const body = window.document.body;
  const div = body.querySelector('div');
  body.addEventListener('test', evt => {
    i++;
  });
  div.addEventListener('test', evt => {
    i++;
    evt.stopPropagation();
  });
  const event = new window.Event('test');
  div.dispatchEvent(event);
  assert(i === 1);
});
```

### DIFF
--- a/lib/jsdom/living/events/Event-impl.js
+++ b/lib/jsdom/living/events/Event-impl.js
@@ -50,7 +50,7 @@ class EventImpl {
 
   stopImmediatePropagation() {
     this._stopPropagationFlag = true;
-    this._stopImmediatePropagation = true;
+    this._stopImmediatePropagationFlag = true;
   }
 
   preventDefault() {
--- a/test/web-platform-tests/to-upstream.js
+++ b/test/web-platform-tests/to-upstream.js
@@ -26,6 +26,7 @@ describe("Local tests in Web Platform Test format (to-upstream)", () => {
     "dom/attributes-are-not-nodes.html",
     "dom/collections/HTMLCollection-iterator.html",
     "dom/events/AddEventListenerOptions-once.html",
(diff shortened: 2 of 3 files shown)


Sebmaster:
Added a fix to this PR. Weird that wpt does not test this...

webpro:
Thanks @Sebmaster, I can confirm it works as expected now.

domenic:
Well, we need tests before we can merge this... I'll take care of it.

webpro:
Thanks guys, I've happily switched from PhantomJS to jsdom for testing [domtastic](https://github.com/webpro/DOMtastic). It has [quite some tests](https://github.com/webpro/DOMtastic/tree/master/test/spec), so testing this dom/event lib with jsdom turned out pretty great: awesome job!

PR #1784, opened 2017-03-22
Implement and test Element.prototype.closest

### LINKED ISSUE #1555: Support for Element.closest()
Hey folks, I was wondering whether there were plans to support [`Element.closest()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) in jsdom?

apexskier:
Resolves #1555

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -519,6 +519,20 @@ ElementImpl.prototype.matches = memoizeQuery(function (selectors) {
 
 ElementImpl.prototype.webkitMatchesSelector = ElementImpl.prototype.matches;
 
+ElementImpl.prototype.closest = memoizeQuery(function (selectors) {
+  // https://dom.spec.whatwg.org/#dom-element-closest
+  let el = this;
+
+  while (el.nodeType === NODE_TYPE.ELEMENT_NODE) {
+    if (ElementImpl.prototype.matches.call(el, selectors)) {
+      return el;
+    }
+    el = el.parentNode;
+  }
+
+  return null;
+});
+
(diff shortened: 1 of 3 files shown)


Sebmaster:
Please don't implement tests in the old test format.

Either look for a pre-existing test in the web-platform-tests and enable it, or implement it in that format in the to-upstream directory.

apexskier:
Sorry, I was following the `matches` test pattern. It looks like a closest test [already exists](https://github.com/w3c/web-platform-tests/blob/c75cae4d1fcd0ed472b9110b14f6a8b5e2ce2759/dom/nodes/Element-closest.html), though, so I'll try enabling it.

apexskier:
Okay, I tested enabling the ["dom/nodes/Element-closest.html"](https://github.com/w3c/web-platform-tests/blob/c75cae4d1fcd0ed472b9110b14f6a8b5e2ce2759/dom/nodes/Element-closest.html) test, but it fails because jsdom's matcher implementation isn't complete (["dom/nodes/Element-matches.html"](https://github.com/w3c/web-platform-tests/blob/c75cae4d1fcd0ed472b9110b14f6a8b5e2ce2759/dom/nodes/Element-matches.html) also fails). In the interest of enabling the `closest` method, I'd like to merge this PR and refactor the `matches` method (which this depends on) to enable both tests in the future.

domenic:
We won't merge things that add more tests in the old format. If you could create some new tests in the to-upstream directory for closest(), that would be great. You can even base them off of the existing one that we can't get to pass; see others named -dont-upstream in that directory for an example of what we've done in the past.

domenic:
Closing due to lack of response; looks like #1951 is attempting to pick this up.

Issue #1779, opened 2017-03-21
document.implementation.createHTMLDocument behaves differently than browser


gavinwahl:
draft-js has this code in it (paraphrased):

```javascript
function getSafeBodyFromHTML(html) {
  doc = document.implementation.createHTMLDocument('foo');
  doc.documentElement.innerHTML = html;
  return doc.getElementsByTagName('body')[0];
}
```

In Chrome 56 and Firefox 45, `getSafeBodyFromHTML('')` returns an HTMLBodyElement. In jsdom 9.12.0, it returns undefined. `getSafeBodyFromHTML('foo')` does return an HTMLBodyElement like it's supposed to.

You can reproduce this in node with this code:

```javascript
> var jsdom = require('jsdom').jsdom;
> var document = jsdom();
> function getSafeBodyFromHTML(html) {
...   doc = document.implementation.createHTMLDocument('foo');
...   doc.documentElement.innerHTML = html;
...   return doc.getElementsByTagName('body')[0];
... }
> getSafeBodyFromHTML('')
undefined
> getSafeBodyFromHTML('asdf')
HTMLBodyElement {}
```

domenic:
Just wanted to say great catch, and thanks for the very small repro (this is a model of what we like to see, no external frameworks or anything). That makes it much more likely we can address this soon, although I am in Tokyo for the next couple weeks so my weekends are not as free for open source as usual. A PR would be much appreciated if you want to help dig into it.

domenic:
Fixed by https://github.com/tmpvar/jsdom/pull/1787

PR #1787, opened 2017-03-28
Handle setting innerHTML to the empty string

Zirro:
I have looked at issue #1779, and the problem seems to have been in the innerHTML implementation. The setInnerHTML function includes a specific if-condition rejecting the empty string as input, while browsers seem to handle it. With the if-condition removed the output has been consistent with the way browsers behave, during my limited testing.

I am naturally a bit wary about simply removing a condition since someone must've put it there for a reason in the first place, but perhaps it no longer applies as the codebase has changed. Regardless, I'd appreciate if someone knowing more about the intricacies of innerHTML would chime in.

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -50,12 +50,10 @@ function setInnerHTML(document, node, html) {
     clearChildNodes(node);
   }
 
-  if (html !== "") {
-    if (node.nodeName === "#document") {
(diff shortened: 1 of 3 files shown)


domenic:
This is great. I have sufficient confidence in our test coverage that I'm happy to accept this. My guess is that the empty string check was copied from `jsdom.env` where we treat empty strings specially?

I'll merge this and do a release when I have some free time not at work, so at worst by Saturday.

PR #1790, opened 2017-03-29
[fix] window.origin doesn't exists (closes #1789)

### LINKED ISSUE #1789: window.origin is undefined - was it supposed to be window.location.origin?
I've been trying to get some of my previous karma tests to pass in node using jsdom and I finally found my issue. The file `/lib/jsdom/living/post-message.js` executes this in line 19:

```javascript
if (targetOrigin !== "*" && targetOrigin !== this.origin) {
```

Now keep in mind that `this` is the object that calls `postMessage` which means that we are dealing with the window object. When I change the second condition to

```javascript
targetOrigin !== this.location.origin
```

It works fine. I was trying to create a PR but I can't seem to be able to run the post-message tests to see if my fix would create any issues.

I stumbled upon https://html.spec.whatwg.org/#crossDocumentMessages and if you search for `location.origin` it will advice you to use `window.origin` which is why probably the code is written as it is. But consider this, if I try to see `window.origin` in chrome i get undefined.

screen shot 2017-03-28 at 6 44 13 pm

Would this work?

  ```javascript
targetOrigin !== (this.origin || this.location.origin)
```

### PR DESCRIPTION
**Pull request by jmlopez-rod**
As stated in the issue, `window.origin` doesn’t exist. This change
allows us to post messages where the origin is something other than `*`.

### DIFF
--- a/lib/jsdom/living/post-message.js
+++ b/lib/jsdom/living/post-message.js
@@ -21,11 +21,11 @@ module.exports = function (message, targetOrigin) {
   }
 
   // TODO: event.source - requires reference to source window
-  // TODO: event.origin - requires reference to source window
   // TODO: event.ports
   // TODO: event.data - structured clone message - requires cloning DOM nodes
   const event = new this.MessageEvent("message", {
-    data: message
+    data: message,
+    origin: targetOrigin
   });
 
   event.initEvent("message", false, false);
--- a/lib/jsdom/living/post-message.js
+++ b/lib/jsdom/living/post-message.js
@@ -21,11 +21,11 @@ module.exports = function (message, targetOrigin) {
   }
 
   // TODO: event.source - requires reference to source window
+  // TODO: event.origin - requires reference to source window
   // TODO: event.ports
   // TODO: event.data - structured clone message - requires cloning DOM nodes
   const event = new this.MessageEvent("message", {
-    data: message,
-    origin: targetOrigin
+    data: message
   });
 
   event.initEvent("message", false, false);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/jsdom/living/post-message.js [hunk 1]
```diff
@@ -21,11 +21,11 @@ module.exports = function (message, targetOrigin) {
   }
 
   // TODO: event.source - requires reference to source window
-  // TODO: event.origin - requires reference to source window
   // TODO: event.ports
   // TODO: event.data - structured clone message - requires cloning DOM nodes
   const event = new this.MessageEvent("message", {
-    data: message
+    data: message,
+    origin: targetOrigin
```
domenic: This is not correct---the target origin is not the same as the source origin. Please remove.

kentongray: an origin is needed to have the same behavior and in all cases it should be the same as the targetOrigin (except *), should we do localhost in that case?

kentongray: Just to add more detail, simple test case you can try in browser: 
```
window.addEventListener('message', m => console.log(m.origin))
window.postMessage('test', 'https://github.com')

> https://github.com

window.postMessage('test', '*')

> https://github.com
```

domenic: Please read the spec. The origin property of the event is completely unrelated to targetOrigin. IT's the source origin. The whole point of postMessage is cross-origin messaging.

If you want to introduce origin, you're going to need to start importing and passing more web platform tests. I think it'll be easier not to make this unrelated change, introducing a new feature, here in this PR, which is about fixing a typo.

jmlopez-rod: I have reverted the commit. Any idea how to pass in the origin when we register the event handler? I guess I'll just open up an issue and let someone more experienced handle it.

PR #1800, opened 2017-04-04
fix(xhr-utils): forward approved preflight headers to request

### LINKED ISSUE #1799: CORS woes
TL;DR: I think there are two bugs with jsdom presently:
  * preflight responses are not incorporated into subsequent XHR requests
  * you can't easily change the document origin

Hey,
I'm trying to overcome some issues using fetch with jsdom and CORS. [This code](https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/xmlhttprequest.js#L868-L873) is being tripped, because the origin of the created XHR is `null`, and therefore must validate headers, which then fails because preflight options are not incorporated in the subsequent GET (this would be corrected by using part of [this PR](https://github.com/tmpvar/jsdom/pull/1661)). As a result, since the `Authorization` header exists (which was explicitly allowed by the server in the preflight `OPTIONS`, but not removed from the unsafe headers list), the attempt to fetch fails outright.

It did seem like `jsdom.changeURL` would actually change the origin, but it doesn't appear to do so consistently once it gets into the XHR setup code (which is a bit hard to follow):
```
const window = jsdom.jsdom('<!doctype html><html><body></body></html>').defaultView;
jsdom.changeURL(window, 'http://localhost:9090');
global.window = window;
console.log(window.document.origin);  // prints 'http://localhost:9090'
```
however, by the time we get [here](https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/xmlhttprequest.js#L129) the value is once again `null` mysteriously. 

Are there any suggested ways to get around these issues? Would you accept a PR to utilize preflight responses for XHR?

Thank you.

mbroadst:
fixes #1799

### DIFF
--- a/lib/jsdom/living/xhr-utils.js
+++ b/lib/jsdom/living/xhr-utils.js
@@ -39,12 +39,25 @@ function updateRequestHeader(requestHeaders, header, newValue) {
   }
 }
 
+function mergeHeaders(lhs, rhs) {
+  const rhsParts = rhs.split(",");
+  const lhsParts = lhs.split(",");
+  return rhsParts.concat(lhsParts.filter(p => rhsParts.indexOf(p) < 0)).join(",");
+}
+
 const simpleMethods = new Set(["GET", "HEAD", "POST"]);
 const simpleHeaders = new Set(["accept", "accept-language", "content-language", "content-type"]);
+const preflightHeaders = new Set([
+  "access-control-expose-headers",
+  "access-control-allow-headers",
+  "access-control-allow-credentials",
+  "access-control-allow-origin"
+]);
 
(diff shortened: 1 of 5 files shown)


mbroadst:
Admittedly the test could be far more generalized for CORS scenarios, but at least its a start 😄

domenic:
Thanks so much for this work! I've rebased it on top of the latest changes and will try to get it merged soon; sorry for my absence.

Note that ideally we'd be testing this with web-platform-tests, but our infrastructure right now doesn't allow writing to-upstream WPTs that need server-side support, so this style of test is the best we can do right now. :+1:

mbroadst:
:+1:

Issue #1786, opened 2017-03-28
The implementation of properties.client.on("data", chunk) in xmlhttprequest.js may cause performance problem in some cases.


skygon:
Currently, I have encountered some performance problem when using jsdom to load some plain data. Here is some code snippet about how I load data:
``` javascript
eventEmitter.on('doSomething', doSomething);

jsdom.env({
  url : http://somewhere.com,
    features : {
        FetchExternalResources : ['script'],
        ProcessExternalResources : ['script']
    },
    done : function (error, window) {
      eventEmitter.emit('doSomething', window);
    }
});

function doSomething(w){
   /*
   Here will invoke function from window, like w.fetchData().  
   */
    w.fetchData();
}
```
The function fetchData from http://somewhere.com basically just uses jquery to load a CSV file. like :
```javascript
var uri = "./files/test.csv";
$.get(uri,function(response){
   /*omit implementation*/
}
```
In my test, loading a 150MB csv file will cost about **6 minutes** and use **1.5GB** memory. By using the  node profiling options, --prof and --prof-process, I got the following information:
```
 ticks    parent    name
41381 89.3% C:\Program Files\nodejs\node.exe
37253 90.0%      C:\Program Files\nodejs\node.exe
31658 85.0%           LazyCompile: ~Buffer.concat buffer.js:293:25
29548 93.3%                 LazyCompile: *properties.client.on.chunk node_modules\jsdom\lib\jsdom\living\xmlhttprequest.js:879:34
```
The above section told us that in the active ticks of node.exe, Buffer.concat costs about 85% ticks,  which happens at ```jsdom\lib\jsdom\living\xmlhttprequest.js:879```.

Here is the code snippet  around line 879 in xmlhttprequest.js:
```javascript
properties.client.on("data", chunk => {
    properties.responseBuffer = Buffer.concat([properties.responseBuffer, chunk]);
    properties.responseCache = null;
```	

Every time chunk comes,  properties.responseBuffer will be replaced with a new buffer, which is composed by the old properties.responseBuffer and the new arrived chunk. And the old responseBuffer will be left as so called 'garbage'. In order to test how frequently Buffer.concat is invoked, I added the following debug log into the above code:
```javascript
properties.client.on("data", chunk => {
    properties.responseBuffer = Buffer.concat([properties.responseBuffer, chunk]);
    //debug log. properties.count is defined previously and initialized to 0.
    properties.count += 1; 
    console.log("Times of buffer.concat: ", properties.count);
[...]

PR #1806, opened 2017-04-10
pre-allocate buffer in xmlhttprequest.js

skygon:
This PR is associate to #1786

### DIFF
--- a/lib/jsdom/living/xmlhttprequest.js
+++ b/lib/jsdom/living/xmlhttprequest.js
@@ -150,7 +150,9 @@ module.exports = function createXMLHttpRequest(window) {
         error: "",
         uploadComplete: true,
         abortError: false,
-        cookieJar: this._ownerDocument._cookieJar
+        cookieJar: this._ownerDocument._cookieJar,
+        bufferStepSize: 1 * 1024 * 1024, // pre-allocate buffer increase step size. init value is 1MB
+        totalReceivedChunkSize: 0
       };
       this.onreadystatechange = null;
     }
@@ -186,6 +188,10 @@ module.exports = function createXMLHttpRequest(window) {
         return properties.responseCache;
       }
       let res = "";
+
+      const responseBuffer = !properties.responseBuffer ? null :
+            properties.responseBuffer.slice(0, properties.totalReceivedChunkSize);
+
       switch (this.responseType) {
         case "":
         case "text": {
@@ -193,18 +199,18 @@ module.exports = function createXMLHttpRequest(window) {
           break;
         }
         case "arraybuffer": {
-          if (!properties.responseBuffer) {
+          if (!responseBuffer) {
             return null;
           }
-          res = (new Uint8Array(properties.responseBuffer)).buffer;
+          res = (new Uint8Array(responseBuffer)).buffer;
           break;
         }
         case "blob": {
-          if (!properties.responseBuffer) {
+          if (!responseBuffer) {
             return null;
           }
           const contentType = getContentType(this);
-          res = Blob.create([[new Uint8Array(properties.responseBuffer)], {
+          res = Blob.create([[new Uint8Array(responseBuffer)], {
             type: contentType && contentType.toString() || ""
           }]);
           break;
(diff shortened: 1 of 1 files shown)


lib/jsdom/living/xmlhttprequest.js
```diff
@@ -186,25 +188,29 @@ module.exports = function createXMLHttpRequest(window) {
         return properties.responseCache;
       }
       let res = "";
+
+      const responseBuffer = (!properties.responseBuffer) ? null :
```
domenic: No unnecessary parens. Here and below.

I am surprised ESLint did not catch this.

lib/jsdom/living/xmlhttprequest.js
```diff
@@ -246,7 +252,10 @@ module.exports = function createXMLHttpRequest(window) {
       if (properties.responseTextCache) {
         return properties.responseTextCache;
       }
-      const responseBuffer = properties.responseBuffer;
+      // const responseBuffer = properties.responseBuffer;
```
domenic: Please don't comment out code. Delete it instead. Here and below.

lib/jsdom/living/xmlhttprequest.js
```diff
@@ -679,6 +691,8 @@ module.exports = function createXMLHttpRequest(window) {
         const client = xhrUtils.createClient(this);
 
         properties.client = client;
+        // on new cline, reset totalReceivedChunkSize
```
domenic: "cline" = "client"?

lib/jsdom/living/xmlhttprequest.js
```diff
@@ -905,7 +919,9 @@ module.exports = function createXMLHttpRequest(window) {
       progressObj.loaded = 0;
       progressObj.lengthComputable = true;
     }
-    properties.responseBuffer = new Buffer(0);
+    // properties.responseBuffer = new Buffer(0);
+    // pre-allocate buffer.
+    properties.responseBuffer = Buffer.alloc(properties.bufferStepSize);
```
domenic: Since this uses Buffer.alloc, it will have to wait until v10.0.0, which is when we drop Node.js v4 support. That is fine; my next jsdom release day will release v10.0.0.

lib/jsdom/living/xmlhttprequest.js
skygon: Removed.

lib/jsdom/living/xmlhttprequest.js
skygon: Deleted.

lib/jsdom/living/xmlhttprequest.js
skygon: My bad, it's a typo.

lib/jsdom/living/xmlhttprequest.js
skygon: Sounds great! May I ask the release time of v10.0.0 in your roadmap? thanks

skygon:
ci check still has 1 failed at XMLHttpRequest/responsetype.html, node.js v7.2.1, say set responseType timeout. Does this fail really matters?  I failed to figure out why this timeout happens...

domenic:
Yes, that's a pretty crucial test; it needs to pass or we cannot accept this.

lib/jsdom/living/xmlhttprequest.js
domenic: We don't have a schedule or roadmap, since we work on this in our free time.

skygon:
All 1300+ tests pass except **XMLHttpRequest/responsetype.html**. And I use the unchanged version of jsdom on my own machine (directly git clone from *https://github.com/tmpvar/jsdom.git*) and run test-wpt again, this case still failed with the same error result.  @domenic Have you ever seen such kind of error case?

domenic:
I have not. But if you're saying it's failing before the change too, then yeah, it's probably not the fault of this PR. I'll do my best to investigate this weekend. With any luck you won't need to change anything and I'll just be able to merge after fixing whatever's wrong there.

PR #1807, opened 2017-04-12
URL POST,PUT. e.t.c METHODs

molokovskikh:
Added option for request like this https://github.com/request/request#requestoptions-callback


domenic:
Thanks, but we're not interested in extending the jsdom.env API any further. If you want to make requests and feed them to jsdom.jsdom or jsdom.env yourself, you're welcome to do so.

PR #1810, opened 2017-04-17
Fix HTMLHyperlinkElementUtils pathname setter to update the href

makana:
Fixes the issue that changing the pathname of an HTMLAnchorElement
didn't update the href.


domenic:
Thanks! Can you add a test, so that we can accept this? See contributing.md for instructions.

makana:
Hi @domenic ,

sorry for not adding a proper test.

I added the url/url-setter web platform test and fixed several more issues.
I'm not entirely sure about the removed javascript scheme check in the hash setter, but that's what the tests expect and how the browsers behave.

The test passes, Travis failed with time outs.

domenic:
Thanks, that's really great work. I'll try to merge this as soon as possible.

A bonus would be if you could check that your changes match https://html.spec.whatwg.org/multipage/semantics.html#api-for-a-and-area-elements. I know we haven't kept up to date with all those changes.

makana:
Hi @domenic 

according to the specs this is the result.

PR #1811, opened 2017-04-18
canvas-prebuilt package is currently broken

danriti:
After attempting to use `jsdom` and `canvas-prebuilt`, I ran into several failing test cases where a canvas element was not found in a project I am currently working on. After some digging, it turns out that:

1.) https://github.com/tmpvar/jsdom/pull/1754 added support for `canvas-prebuilt`, but it's exception handling was a "catch all", so it was catching *all* exceptions, not just import exceptions.
2.) `canvas-prebuilt` (as of version `1.4.0`) had a broken Linux build https://github.com/chearon/node-canvas-prebuilt/issues/8#issuecomment-294982947 (this is now fixed)
3.) `canvas-prebuilt` requires `GLIBC_2.17`, which is not available in `precise`. Thus, this would require upgrading the Travis CI build to use `trusty` which I experimented in this [commit](https://travis-ci.org/tmpvar/jsdom/builds/223976247) and caused some of the tests to fail. Additionally, I can imagine this being a change that the `jsdom` maintainers might not be to thrilled about, so I am hesitant to go further unless they thought it was a good idea.

So this PR makes some changes which ensures that `canvas` and `canvas-prebuilt` integration works as expected. Thus, the build will continue to be broken for this PR until the `canvas-prebuilt` package is fixed.

Example of problem with `canvas-prebuilt`:

https://travis-ci.org/tmpvar/jsdom/builds/223577476

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,16 @@ matrix:
           - www2.web-platform.test
           - xn--n8j6ds53lwwkrqhv28a.web-platform.test
           - xn--lve-6lad.web-platform.test
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-4.8
+            - libcairo2-dev
+            - libjpeg8-dev
+            - libpango1.0-dev
+            - libgif-dev
+            - build-essential
     - node_js: "7.2.1"
       env: TEST_SUITE=node-canvas
       script: "export CXX=g++-4.8 && npm i canvas && npm test"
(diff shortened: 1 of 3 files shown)


domenic:
Did this ever get resolved in any way? Should we maybe stop recommending canvas-prebuilt?

danriti:
@domenic i believe the maintainer of `canvas-prebuilt` has reverted the problem that broke the build:

https://github.com/chearon/node-canvas-prebuilt/issues/8#issuecomment-295736937

PR #1813, opened 2017-04-22
Implement value sanitization algorithms for several input types

Zirro:
This PR implements several sanitization algorithms used on the value of `input` elements. The types supported in this PR are:

* color
* email
* password
* search
* tel
* text
* url

It should be easy to extend the code with more types in the future. I've enabled the WPT-files where all tests pass as of these changes (as well as `search_input.html` which was already passing). Several tests for the `email` and `text` types are also passing now, but not all of them since they make use of other unimplemented functionality.

However, I've discovered that it breaks one of the `to-upstream` tests:
`XMLHttpRequest/formdata-constructor.html`
`Error: Failed in "test that FormData is correctly constructed from the form data set": 
assert_equals: input.value accepts \n followed by \r expected "value\n\r" but got "value"`

The test seems to be expecting newlines to be retained, while the spec is saying `The value sanitization algorithm is as follows: Strip newlines from the value` and `To strip newlines
from a string, remove any U+000A LF and U+000D CR code points from the string` for text inputs.

Is this a problem in my code or with the test?

Finally, I decided to use destructuring assignment in one file, under the assumption that the new API (along with its Node v6 requirement) is about to be merged. I hope that's alright.


domenic:
Great stuff, thank you!!

It looks like that is a problem with the test; please feel free to fix it! It seems like textarea might be kind of broken in that test too, but no need to fix that unless it's a problem also.

Zirro:
This should be ready for merging now if it looks good to you.

PR #1822, opened 2017-04-26
Make Attr extend Node again

### LINKED ISSUE #1641: Make Attr inherit from Node again
The inheritance of `Attr` from `Node` was [added back](https://github.com/whatwg/dom/issues/102). Please support (although `nodeName` exists as an alias, `nodeType` is not available, etc.).

Robbert:
This is my proposal for the approach to fix #1641 by extending `Attr` from `Node` again, so `attr.nodeType` exists again etcetera. snuggs requested to send a pull request so it would be easier to review my changes and judge the approach.

I am aware that several unit tests are now failing, and I plan to spend time on fixing those when the general idea of this fix is okay with you all.


lib/jsdom/living/attributes.js
```diff
@@ -59,15 +58,15 @@ NamedNodeMap.prototype.getNamedItemNS = function (namespace, localName) {
 };
 
 NamedNodeMap.prototype.setNamedItem = function (attr) {
-  if (!attrGenerated.is(attr)) {
+  if (!this[INTERNAL].element._ownerDocument._isNodeType(attr, NODE_TYPE.ATTRIBUTE_NODE)) {
```
domenic: Why this change? The old code should still work.

lib/jsdom/living/attributes.js
```diff
@@ -375,7 +374,9 @@ exports.setAttributeValue = function (element, localName, value, prefix, namespa
 
   const attribute = exports.getAttributeByNameNS(element, namespace, localName);
   if (attribute === null) {
-    const newAttribute = attrGenerated.createImpl([], { namespace, namespacePrefix: prefix, localName, value });
+    const newAttribute = element._ownerDocument._createNode(NODE_TYPE.ATTRIBUTE_NODE, [], {
```
domenic: Again, the old code should still work here. There's no advantage to using _createNode (here and below).

Robbert:
The reason was the whole cyclic-dependency problem you also ran into earlier:

- https://github.com/tmpvar/jsdom/pull/1247#issuecomment-149060470
- https://github.com/tmpvar/jsdom/issues/1641#issuecomment-274614405

domenic:
Oh, I see :(. OK, let me reconsider...

TimothyGu:
The circular dependency problem should be gone after https://github.com/tmpvar/jsdom/pull/2052.

snuggs:
@TimothyGu thanks a bunch with the info on #2052 . That should help out with https://github.com/tmpvar/jsdom/issues/1641

@Robbert what do we do next to move forward on this?

PR #1829, opened 2017-04-27
Implement HTMLSelectElement.selectedOptions

Zirro:
Implements: https://html.spec.whatwg.org/multipage/forms.html#dom-select-selectedoptions

Includes a to-upstream test as one did not exist in the testsuite.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLSelectElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLSelectElement-impl.js
@@ -90,6 +90,12 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
     return this.options.length;
   }
 
+  get selectedOptions() {
+    return createHTMLCollection(this, () => domSymbolTree.treeToArray(this, { filter(node) {
+      return node._localName === "option" && node._selectedness === true;
+    } }));
+  }
+
   get selectedIndex() {
     return Array.prototype.reduceRight.call(this.options, (prev, option, i) => {
       option = idlUtils.implForWrapper(option);
--- a/test/web-platform-tests/to-upstream/html/semantics/forms/the-select-element/select-selectedOptions.html
+++ b/test/web-platform-tests/to-upstream/html/semantics/forms/the-select-element/select-selectedOptions.html
@@ -1,0 +1,95 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>HTMLSelectElement.selectedOptions</title>
+<link rel="help" href="https://html.spec.whatwg.org/multipage/forms.html#dom-select-selectedoptions">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+</head>
+<body>
+<div id="log"></div>
+
+<select id="select-none-selected">
+  <option>One</option>
+  <option>Two</option>
+  <option>Three</option>
+</select>
+
+<select id="select-one-selected">
+  <option>One</option>
+  <option selected>Two</option>
+  <option>Three</option>
+</select>
+
+<select multiple id="multiple-select-none-selected">
+  <option>One</option>
+  <option>Two</option>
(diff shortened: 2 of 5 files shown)


lib/jsdom/living/nodes/HTMLSelectElement-impl.js
```diff
@@ -90,6 +90,12 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
     return this.options.length;
   }
 
+  get selectedOptions() {
+    return createHTMLCollection(this, () => domSymbolTree.treeToArray(this, { filter(node) {
```
Sebmaster: Does `createHTMLCollection` ensure that `[SameObject]` works? I think we might have to cache this object in a private var @domenic?

Zirro:
@domenic Thanks! I pushed a minor change to give the CI another spin. The single failure looks sporadic.

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
domenic: Oh wow, I totally skipped over the [SameObject]. You're right; we should store it in a private var (and probably create it lazily). And add tests for that!!

Zirro: This is probably beyond my current knowledge about jsdom. I appreciate if you're willing to make this change.

domenic: Can do, but it's not so bad. Basically just `return this._selectedOptions || (this._selectedOptions = ...);`, plus a test that el.selectedOptions === el.selectedOptions (and maybe some tests that changing selection also preseves identity).

Zirro: Oh, I see. If you don't mind waiting until the weekend, I can give it a try.

Zirro:
The [SameObject] support will be resolved with https://github.com/jsdom/webidl2js/pull/37, and I've added a test which should pass with those changes.

test/web-platform-tests/to-upstream/html/semantics/forms/the-select-element/select-selectedOptions.html
```diff
@@ -0,0 +1,109 @@
...
+</select>
+
+<script>
+"use strict";
+
+test(() => {
+  const select = document.getElementById("select-none-selected");
+
+  assert_equals(select.selectedOptions.item(0), select.children[0]);
+  assert_equals(select.selectedOptions.length, 1);
```
domenic: I am pretty sure you can make this more idiomatic by using `assert_array_equals(select.selectedOptions, [select.children[0]]);`

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
domenic: Can you update the package.json for webidl2js v6.1.0? You may need to rebase.

test/web-platform-tests/to-upstream/html/semantics/forms/the-select-element/select-selectedOptions.html
```diff
@@ -0,0 +1,109 @@
...
+}, ".selectedOptions without the 'multiple' attribute but " +
+   "more than one selected option should return the last one");
+
+test(() => {
+  const select = document.getElementById("select-same-object");
+  const selectAgain = document.getElementById("select-same-object");
+
+  assert_equals(select.selectedOptions, selectAgain.selectedOptions);
+
+}, ".selectedOptions should always return the same value - [SameObject]");
```
domenic: Can you add a test that even if you modify the selection, e.g. by setting `select.selectedOptions[0].selected = false`, the object stays the same?

Zirro:
Done. I replaced some of the .item() checks, but decided to keep a few as well in order to test that the value is/behaves like a HTMLCollection.

PR #1845, opened 2017-05-10
Update web platform tests

Zirro:
I've updated the web platform tests, and have started resolving issues that are causing 25 new test failures. I've made the following changes so far:

  * Change rules for DOMTokenList/set replacement according to spec update - https://github.com/whatwg/infra/pull/126

  * Don't set attribute when using classList.remove() if it doesn't exist already (appeared as a result of https://github.com/w3c/web-platform-tests/pull/5823)

  * Throw InvalidCharacterError instead of NamespaceError - https://github.com/whatwg/dom/pull/423

  * Do nothing rather than throw when using input.select() on types where it does not apply - https://github.com/whatwg/html/pull/2327

  * Update Web IDL for Event and CustomEvent to make all but the first argument optional - https://github.com/w3c/web-platform-tests/pull/5043

  * Remove Blob.close() and Blob.isClosed() methods - https://github.com/w3c/FileAPI/pull/68

Several failing tests remain, though most seem to come down to changes in header casing. I'll continue to work on them when I have time, but feel free to review the current progress.


Sebmaster:
This is missing updates to the other events: CustomEvent, ErrorEvent, Event, HashChangeEvent, MessageEvent and PopStateEvent changed.

Zirro:
I had intended to keep this PR focused on the changes directly related to updating the web platform tests, but I can make the other changes in this PR as well.

Sebmaster:
I'm pretty sure we'll miss those changes if we don't keep them related to the originating spec changes, so that's why I'd like that in here.

Zirro:
Down to three failing tests with the following changes:

  * Don't lowercase headers in responses - https://github.com/w3c/web-platform-tests/pull/5008

  * Lowercase headers for `getAllResponseHeaders()` - https://github.com/whatwg/xhr/pull/130

  * Allow non-GET methods for data URLs - https://github.com/w3c/web-platform-tests/issues/3738

  * Add space after comma between header values - https://github.com/w3c/web-platform-tests/pull/5008

  * Allow redirect after CORS preflight - https://github.com/w3c/web-platform-tests/pull/5107

  * Allow redirect without userinfo - https://github.com/w3c/web-platform-tests/pull/5121

I've also disabled a few tests - two because of missing features, and one which I believe depends on how jsdom, request and tough-cookie work together. If the `Cookie` header is not set explicitly when passed to request, it uses a lowercase one.

I would appreciate a little help with the `location-stringifier.html` test, which seems to require [special rules](https://github.com/whatwg/html/pull/2294).

domenic:
Amazing work; thank you so much. I'll try my best to help with the remaining failures today after I go through the rest of the pull request backlog here and on webidl2js.

Zirro:
Thanks! I won't be around for the next 15 hours or so, but feel free to merge and work on this if you want to. I should mention that some of the `to-port-to-wpts`-tests will need updates as well.

domenic:
I think I managed to get them all, although I have to sleep now so I can't stick around to see if Travis agrees with my assessment... I didn't touch the to-upstream tests, so I guess probably there is still some work to do.

Zirro:
Splendid! I've updated the old tests to comply with the latest spec, and everything seems to be passing now.

PR #1847, opened 2017-05-12
Enable to change content HTML by `reconfigure` method

sttk:
I've modified `JSDOM.reconfigure` method to enable to change the content HTML with its second parameter.

I try to simulate a window's behavior on node.js, and this modification will be useful for this purpose.


domenic:
Sorry, this is not a change we're interested in. reconfigure is just for changing things that can't be changed through normal means. You can replace the HTML through normal means, or just create a new JSDOM.

PR #1849, opened 2017-05-13
Upgrade parse5 to 3.0.2

### LINKED ISSUE #1778: Parse5 'endTagInBody' causes stack overflow when loading certain documents
Possible to upgrade to a later version of Parse5? I see this is using an older version.

Zirro:
This PR upgrades parse5 to version 3.0.2, and includes changes to affected tests and documentation. Since the ambitious changes in #1316 have not seen progress for a while, I figured that it'd be better to do this with minimal changes than to remain on such an old version.

I removed three old tests in the misc.js file which were testing the serialization of obsolete DOCTYPEs. From what I've gathered, parse5 removed support for these as they are no longer recognised in the spec.

In addition to the parser changes that have been made, there are also breaking changes to the API's output of `.nodeLocation`. I've updated the tests and documentation to reflect this.

This should resolve #1778 and any other issues which have been due to the old version 🎉

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,9 +223,9 @@ const textNode = pEl.firstChild;
 const imgEl = document.querySelector("img");
 
 console.log(dom.nodeLocation(bodyEl));   // null; it's not in the source
-console.log(dom.nodeLocation(pEl));      // { start: 0, end: 39, startTag: ..., endTag: ... }
-console.log(dom.nodeLocation(textNode)); // { start: 3, end: 13 }
-console.log(dom.nodeLocation(imgEl));    // { start: 13, end: 32 }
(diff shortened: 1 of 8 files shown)


domenic:
Great stuff!! I guess this will have to be a major version bump because of the change to nodeLocation... Maybe we should also take the opportunity to remove the parser option, as we've been meaning to for a while.

papandreou:
We could patch `nodeLocation` to alias `startOffset` and `endOffset` as `start` and `end` so it can be released in 10.x?

domenic:
Nah, let's just release 11.0.0; version numbers are cheap.

PR #1857, opened 2017-05-16
Use Set instead of Array for HTMLCollection keys

Zirro:
I am seeing a significant performance improvement parsing large documents with this change. When testing on the markup of the full HTML Spec, I saw an average time of 17 seconds drop to 5 seconds.


Sebmaster:
I love the change! Let's do more of these! 😆

PR #1858, opened 2017-05-17
Delete webkit prefixes in default CSS

flaviut:
See tmpvar/jsdom#1560.

In particular, this helps because it takes out all the selectors that would cause exceptions to be thrown: webkit prefixes in selectors and case-insensitive attribute selectors.

I've tested this patch by running the test that runs slowly with breakpoints set on "all exceptions". There were no exceptions arising from <https://github.com/dperini/nwmatcher/blob/29a1cdc284e330ec6aa38e31c2f02e755a536381/src/nwmatcher-noqsa.js#L765>.

Even though this deletes code, this adds functionality. For example, the rule `-input[type="hidden" i]` would have been ignored before, because nwmatcher doesn't know how to handle ` i`. It would now apply (although type would have to be "hidden", and not "HiDdEn" or "HIDDEN").

This single change improves performance in one of my tests from 28s to 8s.

I'm not really sure how to run the benchmarks. `node ./benchmark/runner` appears to do nothing.

### DIFF
--- a/lib/jsdom/browser/default-stylesheet.js
+++ b/lib/jsdom/browser/default-stylesheet.js
@@ -124,13 +124,10 @@ q {
     display: inline
 }
 
-q:before {
-    content: open-quote;
-}
-
-q:after {
-    content: close-quote;
-}
+/* nwmatcher does not support ::before and ::after, so we can't render q
+correctly: https://html.spec.whatwg.org/multipage/rendering.html#phrasing-content-3
+TODO: add q::before and q::after selectors
+*/
 
 center {
     display: block;
@@ -425,7 +422,8 @@ input, textarea, keygen, select, button {
     text-align: start;
 }
 
-input[type="hidden" i] {
+/* TODO: Add " i" to attribute matchers to support case-insensitive matching */
+input[type="hidden"] {
     display: none
 }
 
@@ -439,7 +437,7 @@ input {
     cursor: auto;
 }
 
-input[type="search" i] {
+input[type="search"] {
     -webkit-appearance: searchfield;
     box-sizing: border-box;
 }
@@ -462,37 +460,37 @@ textarea {
     word-wrap: break-word;
 }
 
-input[type="password" i] {
+input[type="password"] {
     -webkit-text-security: disc !important;
 }
 
-input[type="hidden" i], input[type="image" i], input[type="file" i] {
+input[type="hidden"], input[type="image"], input[type="file"] {
     -webkit-appearance: initial;
     padding: initial;
     background-color: initial;
     border: initial;
 }
 
-input[type="file" i] {
+input[type="file"] {
     align-items: baseline;
     color: inherit;
     text-align: start !important;
 }
 
-input[type="radio" i], input[type="checkbox" i] {
+input[type="radio"], input[type="checkbox"] {
     margin: 3px 0.5ex;
     padding: initial;
     background-color: initial;
     border: initial;
 }
 
-input[type="button" i], input[type="submit" i], input[type="reset" i] {
+input[type="button"], input[type="submit"], input[type="reset"] {
     -webkit-appearance: push-button;
     -webkit-user-select: none;
     white-space: pre
 }
 
-input[type="button" i], input[type="submit" i], input[type="reset" i], button {
+input[type="button"], input[type="submit"], input[type="reset"], button {
     align-items: flex-start;
     text-align: center;
     cursor: default;
@@ -503,7 +501,7 @@ input[type="button" i], input[type="submit" i], input[type="reset" i], button {
     box-sizing: border-box
 }
 
-input[type="range" i] {
+input[type="range"] {
     -webkit-appearance: slider-horizontal;
     padding: initial;
     border: initial;
@@ -511,17 +509,17 @@ input[type="range" i] {
     color: #909090;
 }
 
-input[type="button" i]:disabled, input[type="submit" i]:disabled, input[type="reset" i]:disabled,
+input[type="button"]:disabled, input[type="submit"]:disabled, input[type="reset"]:disabled,
 button:disabled, select:disabled, keygen:disabled, optgroup:disabled, option:disabled,
 select[disabled]>option {
     color: GrayText
 }
 
-input[type="button" i]:active, input[type="submit" i]:active, input[type="reset" i]:active, button:active {
+input[type="button"]:active, input[type="submit"]:active, input[type="reset"]:active, button:active {
     border-style: inset
 }
 
-input[type="button" i]:active:disabled, input[type="submit" i]:active:disabled, input[type="reset" i]:active:disabled, button:active:disabled {
+input[type="button"]:active:disabled, input[type="submit"]:active:disabled, input[type="reset"]:active:disabled, button:active:disabled {
     border-style: outset
 }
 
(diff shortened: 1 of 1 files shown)


lib/jsdom/browser/default-stylesheet.js
```diff
@@ -30,8 +30,6 @@ module.exports = `
  *
  */
 
-@namespace "http://www.w3.org/1999/xhtml";
```
domenic: This should be kept per spec; did it cause problems with our implementation?

lib/jsdom/browser/default-stylesheet.js
```diff
@@ -128,14 +123,6 @@ q {
     display: inline
 }
 
-q:before {
-    content: open-quote;
```
domenic: Why are these removed?

lib/jsdom/browser/default-stylesheet.js
```diff
@@ -173,36 +160,12 @@ h1 {
...
 }
 
-:-webkit-any(article,aside,nav,section) h1 {
+article h1, aside h1, nav h1, section h1 {
     font-size: 1.5em;
     -webkit-margin-before: 0.83__qem;
     -webkit-margin-after: 0.83em;
 }
 
-:-webkit-any(article,aside,nav,section) :-webkit-any(article,aside,nav,section) h1 {
```
domenic: Will nwmatcher work with `:matches()` instead of `:-webkit-any()`?

If not, we should add a comment here explaining that we are going against spec for now.

lib/jsdom/browser/default-stylesheet.js
```diff
@@ -423,7 +386,7 @@ input, textarea, keygen, select, button {
     text-align: start;
 }
 
-input[type="hidden" i] {
+input[type="hidden"] {
```
domenic: We need to add a TODO comment for these cases where we go against spec

lib/jsdom/browser/default-stylesheet.js
```diff
@@ -437,95 +400,15 @@ input {
...
-    opacity: 0;
-    pointer-events: none;
-}
-
-input:enabled:read-write:-webkit-any(:focus,:hover)::-webkit-inner-spin-button {
-    opacity: 1;
-    pointer-events: auto;
-}
-
 keygen, select {
```
domenic: Maybe just remove keygen while you're here

dperini:
Could it be that muting the exception helps in these cases so that execution could continue ?
By default 'nwmatcher' is configured to throw exceptions for CSS syntax errors, however this behaviour can be changed both before execution and at runtime by issuing a:

    NW.Dom.configure({ VERBOSITY: false });

I recently suggested the same for a different parsing problem with similar requirements.
Hope i did correctly understand the OP and the problem posed above.

I will also have an attempt to implement the ' i ' flag for insensitive attributes matching.

domenic:
@dperini, that's a great point. When parsing CSS stylesheets we should definitely mute such errors. I think we currently try/catch them, causing the slowness @flaviut refers to in his OP.

lib/jsdom/browser/default-stylesheet.js
flaviut: nwmatcher [does not support `:before` or `:after`](https://github.com/dperini/nwmatcher/wiki/CSS-supported-selectors#content-pseudo-element-selectors-not-supported)

lib/jsdom/browser/default-stylesheet.js
flaviut: Nope, `:matches()` is CSS4, and [nwmatcher only supports CSS3](https://github.com/dperini/nwmatcher/wiki/CSS-supported-selectors#content-pseudo-element-selectors-not-supported):

>The list has been built upon current W3C Selectors Level 3, but differs in that it contains some deprecated specification that we needed to temporarily maintain.

Going through the source code confirms that `:matches()` is not implemented.

> If not, we should add a comment here explaining that we are going against spec for now.

Done!

lib/jsdom/browser/default-stylesheet.js
flaviut: It did, see https://gist.github.com/flaviut/9485daa073065e5944d769fd6900138f#file-errors-txt-L1

I've restored it though. It seems like a bug that could be fixed in another way.

lib/jsdom/browser/default-stylesheet.js
flaviut: Done!

lib/jsdom/browser/default-stylesheet.js
flaviut: I'd rather not -- kegen is still in the spec, and it doesn't break anything.

domenic: keygen was removed from the spec in January: https://gist.github.com/domenic/ec8b0fc8ab45f39403dd

flaviut:
Thanks for the quick feedback, I really appreciate it!

`NW.Dom.configure({ VERBOSITY: false });` won't work for me, unless there is some way to mute the messages. It prints a crazy amount of stuff to the console:

```console
$ jest MyTest > out
<snip>
$ wc -l out
131136 out
```

lib/jsdom/browser/default-stylesheet.js
flaviut: Ok then, removed.

dperini:
@flaviut what if I would add an extra configuration flag to also mute console errors ?
Would this fix your issue ?

I am thinking to name the flag `LOGERRORS` and assign `true` to it as the default value to maintain backward compatibility. If this helps, the change would be trivial and could happen immediately.
/cc @domenic

flaviut:
@dperini It absolutely would solve my issue. Actually, that what I think I'm going to do for now as a temporary workaround--deleting those three lines in node_modules is much easier that applying this patch.

lib/jsdom/browser/default-stylesheet.js
```diff
@@ -128,13 +124,10 @@ q {
...
 
-q:before {
-    content: open-quote;
-}
-
-q:after {
-    content: close-quote;
-}
+/* nwmatcher does not support :before and :after, so we can't render q
+correctly: https://www.w3.org/TR/html5/rendering.html#phrasing-content-0
```
domenic: Wrong spec; you want https://html.spec.whatwg.org/multipage/ not the W3C fork. I can fix while merging.

dperini:
@flaviut 
the requested change was released with version 1.4.0 (the. flag name is LOGERRORS).
Setting this flag to `false` will mute all the errors that were previously logged to the console.

lib/jsdom/browser/default-stylesheet.js
dperini: @domenic 
now nwmatcher does support the :after and :before pseudo-elements and a few others too, the supports is obviously limited to their parsing (grammar) without throwing exceptions.

For these pseudo-elements, the `select()` method will return an empty node set (empty array) while the `match()` method will always return `false`.

dperini: @domenic I have added the case insensitive flag for attributes, the `i` flag from the Selectors Level 4 specs have a look at current branch, also the new `nwsapi`will be ready to ship sometime next week with all the bells a whistles, I will also be able to port `:matches` to current branch.

PR #1869, opened 2017-05-22
Update nwmatcher dependency

AStoker:
Resolves: https://github.com/tmpvar/jsdom/issues/1740
Due to a bug in the nwmatcher (https://github.com/dperini/nwmatcher/issues/103), query selectors with single characters would return an error. Update nwmatcher minimum version to include bug fix.


domenic:
Thanks! I rolled this into https://github.com/tmpvar/jsdom/pull/1907.

PR #1872, opened 2017-05-29
WIP: Initial work for Custom Elements support

tbranyen:
Changes here paired with https://github.com/jsdom/webidl2js/pull/44 will
enable `./test-web-components.js` to work. You will need to `npm
install` before running `node ./test-web-components.js`.

### DIFF
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -384,7 +384,7 @@ class DocumentImpl extends NodeImpl {
     let Constructor = null;
 
     if (Constructor = this._global.window.customElements.get(name)) {
-      return new Constructor();
+      return idlUtils.tryImplForWrapper(new Constructor());
     }
     // https://dom.spec.whatwg.org/#concept-element-interface
     // TODO: eventually we should re-write the element-builder system to be namespace aware, but for now it is not.
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -381,6 +381,11 @@ class DocumentImpl extends NodeImpl {
   }
 
   _createElementWithCorrectElementInterface(name, namespace) {
+    let Constructor = null;
+
+    if (Constructor = this._global.window.customElements.get(name)) {
+      return new Constructor();
+    }
     // https://dom.spec.whatwg.org/#concept-element-interface
     // TODO: eventually we should re-write the element-builder system to be namespace aware, but for now it is not.
     const builder = this._elementBuilders[name.toLowerCase()] || this._defaultElementBuilder.bind(this);
--- a/lib/jsdom/living/nodes/HTMLElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLElement-impl.js
@@ -29,6 +29,20 @@ class HTMLElementImpl extends ElementImpl {
     });
   }
 
+  attachShadow(options) {
+    switch (options.mode) {
+      case 'open': {
+        const instance = idlUtils.wrapperForImpl(this);
+        instance.shadowRoot = instance.ownerDocument.createDocumentFragment();
+      }
+
+      case 'closed': {
+        const instance = idlUtils.wrapperForImpl(this);
+        return instance.ownerDocument.createDocumentFragment();
+      }
+    }
+  }
+
   // Add default event behavior (click link to navigate, click button to submit
   // form, etc). We start by wrapping dispatchEvent so we can forward events to
   // the element's default functions (only events that did not incur
--- a/lib/jsdom/living/nodes/HTMLElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLElement-impl.js
@@ -13,6 +13,7 @@ class HTMLElementImpl extends ElementImpl {
   constructor(args, privateData) {
     super(args, privateData);
     this._initGlobalEvents();
+    this._core = this._core || global.window;
 
     this._tabIndex = 0;
 
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -422,6 +422,19 @@ function Window(options) {
   };
 
   ///// PUBLIC DATA PROPERTIES (TODO: should be getters)
+  this.customElements = new class CustomElementsRegistry {
+    get(tagName) {
+      return this[idlUtils.registry].get(tagName);
+    }
+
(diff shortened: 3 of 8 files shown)


domenic:
Please be sure to follow the HTML and DOM Standards when making these changes. For example they do not look up properties on the instance, but instead store them at definition time.

This is a big ambitious project, but it sounds like you're interested in sinking some time into it, so that's great! Just please follow the spec, and use official web platform tests, not ad-hoc ones.

tbranyen:
Awesome @domenic. I'll work on spec-compliance and cleaning up my implementation. At least I wanted to make sure it was even do-able and what the path of integration might look like. So I didn't take into account standards or doing things the-right-way.

That said, I'm a bit stuck on how to get the associated `defaultView` from the HTMLElement constructor or `document.createElement`. Lets say you have code that looks like this:

``` js
customElements.define('hello-world', class extends HTMLElement {
  constructor() {
    this.innerHTML = `<b>Hello world</b>`;
  }
})

document.createElement('hello-world');
```

How can we make the `HTMLElement-impl` and `Document-impl` aware of where `customElements` are defined? Assuming I tie everything in with the IDL and let the generator scaffolding out the source, is there any clean way to reference the `customElements` registry from the output? I'm doing `global.window` currently which sucks since end users need to do that `global.window = window`.

I'll check the `HTMLConstructor` specification and see if it lends any clues as to how it finds the associated `window`.

tbranyen:
Hrm:

> Let registry be the current global object's CustomElementRegistry object.

> Similarly, the current global object is the global object of the current Realm Record.

> Before it is evaluated, all ECMAScript code must be associated with a realm. Conceptually, a realm consists of a set of intrinsic objects, an ECMAScript global environment, all of the ECMAScript code that is loaded within the scope of that global environment, and other associated state and resources.

Sounds like the `vm.createContext(window)` code, `vm == realm`?

Sebmaster:
> Sounds like the vm.createContext(window) code, vm == realm?

Sadly, that doesn't work in jsdom because we don't define HTMLElement in the VM context.

`document.createElement` is easy, since it goes through our own code. Anything where the on-page code can directly `new` up an interface directly, we have to cheat in some way. We mostly do it by wrapping the constructor. The code that [implements Window](https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/browser/Window.js) contains a lot of these tricks. See the definition of Image and XMLHttpRequest for example.

snuggs:
This is awesome @tbranyen

durandj:
I don't suppose this PR is still active is it? I was really hoping to use this for testing my web components.

tbranyen:
@durandj you should be able to use jsdom-wc (from npm) to test certain features. I will be taking another stab at this soon. Real life in the way ATM though.

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -381,6 +381,11 @@ class DocumentImpl extends NodeImpl {
   }
 
   _createElementWithCorrectElementInterface(name, namespace) {
+    let Constructor = null;
+
+    if (Constructor = this._global.window.customElements.get(name)) {
+      return idlUtils.tryImplForWrapper(new Constructor());
```
majo44: @domenic 
Hi, I trying to test similar implementation by web-platform-tests, and I have strange problem.
If I will run in node code like 
```
import {JSDOM} from 'jsdom-wc';
const dom = new JSDOM(`<!DOCTYPE html>`);
global.HTMLElement = dom.window.HTMLElement;
global.window = dom.window;
global.document = dom.window.document;
class A extends HTMLElement {
    constructor() {
        super();
        console.log('AAAA');
    }
}
new A();
```
Everything works fine.
But if I want to run similar code from web-platform-tests I have a error: 
```
TypeError: Illegal constructor
    at CustomElement.HTMLElement (...\jsdom\living\generated\HTMLElement.js:14:10)
    at ...
    at Object.<anonymous> (http://w3c-test.org/custom-elements/attribute-changed-callback.html:6:31)
```
I putted similar code to create-jsdom.js just before `window.shimTest = () => {` and I see similar result.
The 'window.HTMLElement' is a function from ...\jsdom\living\generated\HTMLElement.js so calling of new is throwing same exception. 
In most of wpt there is no direct call of the elements constructors (instead of this document.creatElement is called) that why I think this is not occurs on other tests. But for custom-elements we have to call the constructor directly, and I do not know why on wpt this is no working. Any suggestion ?

TimothyGu: @majo44 We do not support jsdom-wc.

majo44: @TimothyGu 
Hi, I do not ask you support jsdom-wc. Forget about the previous question it was my mistake (this is expected behavior which fallow spec).

lib/jsdom/living/nodes/HTMLElement-impl.js
```diff
@@ -28,6 +29,20 @@ class HTMLElementImpl extends ElementImpl {
     });
   }
 
+  attachShadow(options) {
+    switch (options.mode) {
```
durandj: This appears to be missing the `delegateFocus` option that the [spec](https://www.w3.org/TR/shadow-dom/#dfn-delegates-focus) allows for to allow focus changes through the shadow root.

caridy:
@tbranyen do you have plans or time to continue pushing for this?

tbranyen:
@caridy at this point not really, my implementation with jsdom-wc works pretty well and if others want to improve on that I'd be happy to push out PRs, but I don't have the time to go through the IDL/specification dance to get this implemented properly to land.

mraerino:
I'd be willing to pick up on this.
@domenic do you recommend continuing the work done by @tbranyen or rather start from scratch?

caridy:
@mraerino it seems that @TimothyGu and @domenic are in agreement on how to move forward based on https://github.com/tmpvar/jsdom/issues/1030#issuecomment-333994158. We can also help once we have a clear direction on how to move forward. Can you work on a concrete plan?

TimothyGu:
Work-in-progress that may be of interest:

- https://github.com/TimothyGu/webidl2js/tree/bundle (webidl2js part is fully working)
- https://github.com/TimothyGu/jsdom/tree/vm-hah (need to generate a browserify bundle manually; many places not working)

mraerino:
Ok, so let's continue this in #1030

domenic:
Let's close this and roll future work into #1030. Excited to see such progress!

Issue #1874, opened 2017-05-31
React issue onScroll


hammer65:
### Basic info:

7.1.0
10.1.0

This warning appears because of the following in React code but is only supposed to trigger for old versions of IE. unfortunately it triggers for JSDOM because of what JSDOM returns in this scenario.
*called here*
https://github.com/facebook/react/pull/6678/files#diff-5574f655d491348f422bca600ff6711dR221
*ran here and fails here*
https://github.com/facebook/react/blob/b1768b5a48d1f82e4ef4150e0036c5f846d3758a/src/renderers/dom/shared/utils/isEventSupported.js#L51

### Minimal reproduction case

```javascript
  const { window: win } = new JSDOM(
  '<!doctype html><html><body></body></html>',
  {
    runscripts: 'dangerously',
    beforeParse: (window) => {
      const eventName = 'scroll';
      const element = window.document.createElement('div');
      element.setAttribute(eventName, 'return;');
      const isSupported = typeof element[eventName] === 'function';
      console.log('####', typeof element[eventName]);
    }
  }
);
```
Console will show `undefined` not `function`

### How does similar code behave in browsers?

no errors in browser

domenic:
`eventName` should be `onscroll`, not `scroll`, right?

hammer65:
correct. Not camel case

hammer65:
Okay my mistake but then I have no idea where this error is coming from

Zirro:
You've written `runscripts` rather than `runScripts`. Replacing it (and `scroll` with `onscroll`) resolves the issue in your reproduction case.

domenic:
Great find, thanks @Zirro!

PR #1876, opened 2017-06-02
Throw error when the API receives an unknown option

Zirro:
This could spare users from some confusion and time spent debugging issues like #1874.

**EDIT**: It has occurred to me that the words valid/invalid may be more suitable than known/unknown in this context. I can make the switch if you'd prefer that.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -249,6 +249,28 @@ function transformOptions(options, encoding) {
     beforeParse() { }
   };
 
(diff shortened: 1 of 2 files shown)


domenic:
I'd rather not do this. It's extremely un-idiomatic for JavaScript. If people need this kind of validation, TypeScript or other tools are probably the right approach.

Zirro:
Sure, that's reasonable.

Issue #1893, opened 2017-06-21
TypeError: Cannot read property '_location' of null


Enverbalalic:
### Basic info:

NodeJS version: v6.10.3
jsdom version: jsdom@11.0.0

### Minimal reproduction case

```js
const jsdom = require("jsdom");
const { JSDOM } = jsdom;
const window = (new JSDOM(``, { runScripts: "outside-only" })).window;

window.setTimeout(function() {
        window.close();
        window.setTimeout(function () {
                window.location.href = "hello";
        }, 300);
}, 1000);
```
In browsers after window.close(); all pending timeouts are cleared and new ones are not set up.

domenic:
Great repro code, thanks! I will investigate this as soon as I am able to.

PR #1894, opened 2017-06-21
Don't add timer if 'this' or 'this._document' are undefined or null

Enverbalalic:
Fix for [#1893](https://github.com/tmpvar/jsdom/issues/1893)

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -527,7 +527,7 @@ function matchesDontThrow(el, selector) {
 }
 
 function startTimer(window, startFn, stopFn, timerId, callback, ms, args) {
-  if (this && this._document) {
+  if (window && window._document) {
     if (typeof callback !== "function") {
       const code = String(callback);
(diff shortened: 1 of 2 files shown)


domenic:
Thanks, but this will need tests before we can accept it.

Since window.close() is somewhat of a specialty feature, those tests will be a bit special. Probably the best thing to do would be to copy https://github.com/tmpvar/jsdom/blob/master/test/api/basics.js into a new file named, maybe, from-outside.js? With the top-level describe being `"Test cases only possible to test from the outside"`.

Enverbalalic:
I'm not as familiar with this project as you are, I'm still able to reproduce the same issue with the same code even if I remove "{ runScripts: "outside-only" }" if you think that that's what's causing this.

domenic:
Heh, I can see the confusion. What I meant is that this can only be *tested* from the outside. You could not write a .html file that triggered this bug.

However it actually doesn't rely on running scripts at all, so yeah, the runScripts: "outside-only" option is unrelated.

Enverbalalic:
So after running the tests the previous pull request was wrong. This one should be ok, complete with a test that confirms the behaviour.

PR #1896, opened 2017-06-26
fix(package.json): add engines field

jedwards1211:
it should be there, because there is code in `lib/api.js` that doesn't work on versions below 6.


domenic:
Thanks, but I prefer not do engine-sniffing, but instead feature detection. This entry, e.g., causes problems on chakra core or browser engines.

Issue #1903, opened 2017-06-28
Navigation approach


ForbesLindesay:
I'm working on making navigation and form post work in jsdom, so I can upstream things instead of adding them to zombie.js.  You can see what I've done so far in #1902.

The next step is to implement the hook for loading pages that get navigated to, and including them in the history.  What this means is that the JSDOM object may ultimately get new window objects, as the navigation occurs.  I think this is the simplest approach.  My suggestion is that there should also be a hook `onNewWindow` that is called when a link is opened in a new tab.  This hook will be given a reference to a separate JSDOM instance.

What this means is that JSDOM changes from representing a single loaded window, to being equivalent to a single tab within a modern, tabbed browser.  I don't think this notion of history can be simply made external, while keeping jsdom's implementation of the history and hash-change APIs.

domenic:
Amazing. This is exactly what we've dreamed of. Having the JSDOM object change what it represents makes perfect sense to me.

The ideal solution would be to implement the WindowProxy so that the window is === itself before and after navigation. I think that can be layered on top though, later.

Can you say more about what the onNewWindow hook would be useful for?

ForbesLindesay:
If you imagine a implementing a full tabbed browser in JavaScript, by rendering a JSDOM dom to canvas or something: onNewWindow means that a new tab should be opened, displaying the contents of the new dom.  e.g. if you rendered the following page using jsdom:

```html
<a href="http://example.com" target="__blank">Clicking me opens a new tab</a>
```

ForbesLindesay:
I've started looking at this, but the spec is really hard to follow because there doesn't seem to be a class in jsdom called "Browsing Context".  I think currently all that functionality is tangled up with the window object.  Navigating across pages will mean multiple windows per browsing context, so I think it needs to be a separate object.

Is the right approach to make this a class and put it in `lib/jsdom/browser/BrowsingContext`?  I can try and create it according to `https://html.spec.whatwg.org/#windows` but I'm not 100% clear what each step of each algorithm involves.

domenic:
That sounds right, yeah. The interesting part will be hooking it up to iframes. Also note that browsing contexts are 1:1 with WindowProxy objects.

And yeah, the spec is known to be pretty confusing. Anne and I are working to fix it but it's a long-term project. I'm at breakfast right now but will be in #jsdom IRC later today if you want more real-time discussion.

domenic:
[...]

PR #1909, opened 2017-07-03
Begin working on creating a BrowsingContext

ForbesLindesay:
This is a work in progress, not ready to merge, but I’d appreciate assistance and pointers as I’m getting started.

Which objects in jsdom are representing the Window vs. WindowProxy?

Does “get the [[Foo]] Internal property of X” translate to approximately:

```js
idlUtils.implForWrapper(X)._foo
```

### DIFF
--- a/lib/jsdom/browser/BrowsingContext.js
+++ b/lib/jsdom/browser/BrowsingContext.js
@@ -1,0 +1,58 @@
+"use strict";
+
+const idlUtils = require("../living/generated/utils");
+
+// https://html.spec.whatwg.org/#windows
+class BrowsingContext {
+  // readonly windowProxy: WindowProxy
+  // readonly sessionHistory: SessionHistory
+
+  // create a new browsing context
+  constructor() {
+    this.windowProxy = new WindowProxy();
+    this.sessionHistory = new SessionHistory();
+
+    // 2. Call the JavaScript InitializeHostDefinedRealm() abstract operation
+    const window = new Window();
+    // TODO: InitializeHostDefinedRealm({globalObject: window, globalThisValue: this.windowProxy, relmExecutionContext: new JavaScriptExecutionContext})
+
+
+    this.windowProxy = windowProxy;
+    // https://html.spec.whatwg.org/#session-history
+    this._sessionHistory = [{
+      document: idlUtils.implForWrapper(this._document),
+      url: idlUtils.implForWrapper(this._document)._URL,
+      stateObject: null
+    }];
+    this._currentSessionHistoryEntryIndex = 0;
+  }
+
+  // A browsing context's active document is its WindowProxy object's [[Window]] internal slot value's associated Document.
+  getActiveDocument() {
+    return this.windowProxy.document;
+  }
+  setActiveDocument(document, window) {
+    if (!window) {
+      // If window is not given, let window be document's relevant global object.
+      window = idlUtils.implForWrapper(document)._global;
+    }
+    // Set browsingContext's WindowProxy object's [[Window]] internal slot value to window.
+    // Set window's associated Document to document.
+    // Set window's relevant settings object's execution ready flag.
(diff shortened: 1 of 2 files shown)


lib/jsdom/browser/BrowsingContext.js
```diff
@@ -0,0 +1,58 @@
...
+  // readonly sessionHistory: SessionHistory
+
+  // create a new browsing context
+  constructor() {
+    this.windowProxy = new WindowProxy();
+    this.sessionHistory = new SessionHistory();
+
+    // 2. Call the JavaScript InitializeHostDefinedRealm() abstract operation
+    const window = new Window();
+    // TODO: InitializeHostDefinedRealm({globalObject: window, globalThisValue: this.windowProxy, relmExecutionContext: new JavaScriptExecutionContext})
```
domenic: I don't think this TODO is very actionable.

Creating a new realm is basically the vm.createContext() business, which the Window constructor takes care of. I'd maybe revise this to

```
// 2. Call ...
// In the current implementation, the Window() constructor takes care of creating a new realm via the vm module
```

lib/jsdom/browser/BrowsingContext.js
```diff
@@ -0,0 +1,58 @@
+"use strict";
+
+const idlUtils = require("../living/generated/utils");
+
+// https://html.spec.whatwg.org/#windows
+class BrowsingContext {
+  // readonly windowProxy: WindowProxy
+  // readonly sessionHistory: SessionHistory
```
domenic: Not a big fan of these myself

lib/jsdom/browser/BrowsingContext.js
```diff
@@ -0,0 +1,58 @@
...
+    // TODO: InitializeHostDefinedRealm({globalObject: window, globalThisValue: this.windowProxy, relmExecutionContext: new JavaScriptExecutionContext})
+
+
+    this.windowProxy = windowProxy;
+    // https://html.spec.whatwg.org/#session-history
+    this._sessionHistory = [{
+      document: idlUtils.implForWrapper(this._document),
+      url: idlUtils.implForWrapper(this._document)._URL,
+      stateObject: null
+    }];
```
domenic: So currently all the history-tracking is done by the History object, which hangs off of the window. I guess we need to disentangle them so that session history survives navigation. Then History will be an interface into a SessionHistory.

lib/jsdom/browser/BrowsingContext.js
```diff
@@ -0,0 +1,58 @@
...
+const idlUtils = require("../living/generated/utils");
+
+// https://html.spec.whatwg.org/#windows
+class BrowsingContext {
+  // readonly windowProxy: WindowProxy
+  // readonly sessionHistory: SessionHistory
+
+  // create a new browsing context
+  constructor() {
+    this.windowProxy = new WindowProxy();
```
domenic: If we don't implement WindowProxy for now, we could just say `this.windowProxy = window` with a comment about how it's an interim solution, but the intent is that the public API presented by BrowsingContext can be reused later even when we implement a proper WindowProxy.

domenic (commented):
> Which objects in jsdom are representing the Window vs. WindowProxy?

jsdom does not have WindowProxy at all right now. This is basically because WindowProxy is only relevant when you navigate, and even then only when you navigate a frame. It helps enforce that `iframe.contentWindow` stays the same object (a `WindowProxy`) even though the underlying global (a `Window`) completely changes.

I think it'll be important eventually, although for now I'd be OK with breaking that equality to get things off the ground, especially since (as noted in #1903) I am unsure about the performance implications. However, not having WindowProxy will probably make it very hard to pass a number of web platform tests related to navigation :-/.

> Does “get the [[Foo]] Internal property of X” translate to approximately:

Yes, that sounds perfect.

PR #1912, opened 2017-07-04
Define a session-history class

ForbesLindesay:
This is a fairly minimal session history implementation.  Next Step is
to add some ability for SessionHistory to report when the `window`
object has changed.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -94,7 +94,7 @@ function Window(options) {
     }
   });
   // https://html.spec.whatwg.org/#session-history
-  this._sessionHistory = options.sessionHistory || new SessionHistory({
+  this._sessionHistory = new SessionHistory({
     document: idlUtils.implForWrapper(this._document),
     url: idlUtils.implForWrapper(this._document)._URL,
     stateObject: null
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -19,6 +19,7 @@ const Document = require("../living/generated/Document");
 const Navigator = require("../living/generated/Navigator");
 const reportException = require("../living/helpers/runtime-script-errors");
 const { matchesDontThrow } = require("../living/helpers/selectors");
+const SessionHistory = require("../living/window/SessionHistory");
 const { contextifyWindow } = require("./documentfeatures.js");
 
 const GlobalEventHandlersImpl = require("../living/nodes/GlobalEventHandlers-impl").implementation;
@@ -93,12 +94,11 @@ function Window(options) {
     }
   });
   // https://html.spec.whatwg.org/#session-history
-  this._sessionHistory = [{
+  this._sessionHistory = options.sessionHistory || new SessionHistory({
     document: idlUtils.implForWrapper(this._document),
     url: idlUtils.implForWrapper(this._document)._URL,
     stateObject: null
-  }];
-  this._currentSessionHistoryEntryIndex = 0;
+  });
 
   // TODO NEWAPI can remove this
   if (options.virtualConsole) {
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -98,7 +98,7 @@ function Window(options) {
     document: idlUtils.implForWrapper(this._document),
     url: idlUtils.implForWrapper(this._document)._URL,
     stateObject: null
-  });
+  }, this);
 
   // TODO NEWAPI can remove this
   if (options.virtualConsole) {
--- a/lib/jsdom/living/window/History-impl.js
+++ b/lib/jsdom/living/window/History-impl.js
@@ -2,7 +2,6 @@
 const DOMException = require("../../web-idl/DOMException.js");
 const documentBaseURLSerialized = require("../helpers/document-base-url.js").documentBaseURLSerialized;
 const parseURLToResultingURLRecord = require("../helpers/document-base-url.js").parseURLToResultingURLRecord;
-const traverseHistory = require("./navigation.js").traverseHistory;
 
 exports.implementation = class HistoryImpl {
   constructor(args, privateData) {
@@ -11,8 +10,6 @@ exports.implementation = class HistoryImpl {
     this._actAsIfLocationReloadCalled = privateData.actAsIfLocationReloadCalled;
     this._state = null;
     this._latestEntry = null;
-
-    this._historyTraversalQueue = new Set();
   }
 
   _guardAgainstInactiveDocuments() {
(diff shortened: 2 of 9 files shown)


lib/jsdom/browser/Window.js
```diff
@@ -93,12 +94,11 @@ function Window(options) {
     }
   });
   // https://html.spec.whatwg.org/#session-history
-  this._sessionHistory = [{
+  this._sessionHistory = options.sessionHistory || new SessionHistory({
```
domenic: This shouldn't be an option passed in, at least for now, I think. (That would require threading through api.js and the readme and such.)

lib/jsdom/living/window/SessionHistory.js
```diff
@@ -0,0 +1,60 @@
+"use strict";
+const traverseHistory = require("./navigation.js").traverseHistory;
+
+// https://html.spec.whatwg.org/#session-history
+class SessionHistory {
+  constructor(initialEntry, window) {
+    this._window = window;
+    this._historyTraversalQueue = new Set();
+    this.entries = [initialEntry];
+    this.currentIndex = 0;
```
domenic: Do you think it'd be a good idea to encapsulate entries/currentIndex entirely? It seems a bit strange that sometimes you go through sessionHistory's properties (e.g. length, currentEntry, removeALlEntriesAfterCurrentEntry) and sometimes you directly manipulate the entries array.

Also, below you use an underscore in currentIndex. It seems bad that we don't have any tests that would catch that :-/.

lib/jsdom/living/window/History-impl.js
```diff
@@ -11,8 +10,6 @@ exports.implementation = class HistoryImpl {
     this._actAsIfLocationReloadCalled = privateData.actAsIfLocationReloadCalled;
     this._state = null;
     this._latestEntry = null;
```
domenic: _state and _latestEntry seem like prime candidates for moving into SessionHistory, although I haven't looked at the spec that much recently.

ForbesLindesay:
I have most of the tests passing again.  I think the failing test might just be too slow, since it's running in 40s on stable node, and times out after 60s on node v6.  That feels well within margin of error.

lib/jsdom/browser/Window.js
ForbesLindesay: This is intended as an internal option.  It'll be used when we do real navigation to create a new window with the same session history as an existing window.  Do you think there's a better way of expressing that?

lib/jsdom/living/window/History-impl.js
ForbesLindesay: I agree, I'll move them in.

lib/jsdom/living/window/SessionHistory.js
ForbesLindesay: I'm trying to encapsulate everything, but using language as close as possible to the spec.  e.g. "Remove all entries after the current entry" is pretty much exactly what it says in the spec in several of the algorithms.  I'll have a go at encapsulating a few more things.

Regarding `currentIndex` vs. `_currentIndex` I'm assuming you would be anti adding something like flow or typescript to this project on account of it requiring a build step?  It would catch this kind of basic mistake, and would have made it much easier to approach the project with the subtle distinction between implementations and idl wrappers actually enforced by some kind of type system. My personal preference would be to gradually convert to typescript, but I'm happy to carry on with plain js assuming that's your preference.

ForbesLindesay:
I've added a test for handling of fragments, especially when using replace.  This should exercise much more of the "SessionHistory" class than was currently tested.  I believe that for true cross-page navigation, we will need to use API tests as well as maybe something using IFrames, but for fragment navigation, the web platform tests work well, and helped me spot a couple of bugs.

lib/jsdom/browser/Window.js
domenic: Well, I would leave out the option for now, until we actually have something that uses it.

lib/jsdom/living/window/SessionHistory.js
domenic: Yeah, no build step please. I kind of thought at least Flow could catch these things itself without requiring it to be integrated?

test/web-platform-tests/to-upstream/html/browsers/browsing-the-web/navigating-across-documents/fragment-replace.html
```diff
@@ -0,0 +1,60 @@
+<!DOCTYPE html>
```
domenic: This test doesn't pass in browsers. (Save it locally and add `https://w3c-test.org/` in front of the script URLs.)

I think this is because if you do multiple window.location.assign()s in a row, only the last one has any impact? You can confirm by logging (and asserting against) `window.history.length` in various places.

test/web-platform-tests/to-upstream/html/browsers/browsing-the-web/navigating-across-documents/fragment-replace.html
```diff
@@ -0,0 +1,60 @@
...
+        t.step_timeout(() => {
+          assert_equals(window.location.hash, "#cat", "After Forward");
+          assert_equals(window.history.length, historyLengthBefore, "replacing does not change history length");
+          t.done();
+        }, 0);
+      }, 0);
+    }, 0);
+  }, 0);
+
+  // assert_equals(history.length === 1, undefined, "Before");
```
domenic: Why are these commented out?

domenic:
So the remaining blocker here is test coverage. I've left some comments on the test added. I'm also wondering if we can try passing the various tests in https://github.com/w3c/web-platform-tests/tree/master/html/browsers/browsing-the-web/scroll-to-fragid, or get an accounting of why we aren't passing them yet.

Other potential tests, although most of them I don't expect us to pass:

- https://github.com/w3c/web-platform-tests/blob/master/html/browsers/browsing-the-web/history-traversal/
- https://github.com/w3c/web-platform-tests/tree/master/html/browsers/history

All of these can be enabled by editing test/web-platform-tests/index.js

ForbesLindesay:
I've enabled all the web platform tests that I could easily make pass.  I updated the test I added so it passes in real browsers.  The main issue seemed to be that real browsers implement the optional part of the spec where calling `location.assign` before the page has finished loading counts as a `replace`, to stop pages breaking the back button by navigating immediately.

test/web-platform-tests/index.js
```diff
@@ -514,5 +525,12 @@ describe("Web Platform Tests", () => {
     "progress-events/tests/submissions/Samsung/firing-events-http-content-length.html",
     "progress-events/tests/submissions/Samsung/firing-events-http-no-content-length.html"
   ]
+  .filter(str => {
```
domenic: This shouldn't be necessary; we can use Mocha's built-in `--fgrep`.

test/web-platform-tests/to-upstream/html/browsers/browsing-the-web/navigating-across-documents/fragment-replace.html
```diff
@@ -11,50 +11,47 @@
 setup({ allow_uncaught_exception: false });
```
domenic: Interesting, this test now fails in Edge and Firefox, but not Chrome. If you are able to do the work of tracking down exactly why, that would be awesome (and we can maybe straighten it out in the HTML spec). But if not that's fine too.

domenic: OK, per my meandering in https://github.com/whatwg/html/pull/2869 I figured this out, so this is all good.

test/web-platform-tests/index.js
domenic: See also https://github.com/tmpvar/jsdom/blob/master/Contributing.md#web-platform-feature-tests

PR #1915, opened 2017-07-08
Forms

ForbesLindesay:
Support navigation triggered by forms.  At the moment, this is only for when the method is “get”.  I think implementing “post” is fairly high priority, but “dialog” is less urgent.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLFormElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLFormElement-impl.js
@@ -77,7 +77,7 @@ class HTMLFormElementImpl extends HTMLElementImpl {
     if (!flags.submittedFromSubmit) {
       const ev = document.createEvent("HTMLEvents");
       ev.initEvent("submit", true, true);
-      if (this.dispatchEvent(ev)) {
+      if (!this.dispatchEvent(ev)) {
         return;
       }
     }
--- a/lib/jsdom/living/nodes/HTMLFormElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLFormElement-impl.js
@@ -102,7 +102,6 @@ class HTMLFormElementImpl extends HTMLElementImpl {
     if (method === 'dialog') {
       notImplemented("dialog method on forms", window);
     }
-    console.dir({method, action});
     switch (scheme) {
       case 'http':
       case 'https':
@@ -148,7 +147,6 @@ class HTMLFormElementImpl extends HTMLElementImpl {
       }
       form._plannedNavigation = window.setTimeout(() => {
         form._plannedNavigation = null;
-        console.log({destination});
         navigate(targetWindow, destination, {});
       }, 0);
     }
--- a/lib/jsdom/living/nodes/HTMLFormElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLFormElement-impl.js
@@ -7,6 +7,12 @@ const { domSymbolTree } = require("../helpers/internal-constants");
 const createHTMLCollection = require("../../living/html-collection").create;
 const notImplemented = require("../../browser/not-implemented");
 const { reflectURLAttribute } = require("../../utils");
+const { constructTheFormDataSet } = require("../xhr/FormData-impl");
+const { parseURLToResultingURLRecord } = require("../helpers/document-base-url.js");
+const { chooseBrowsingContext, navigate } = require("../window/navigation");
+
+// TODO: this needs to be moved to the public API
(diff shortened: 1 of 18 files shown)


lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -53,19 +59,102 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
+          mutateActionURL();
+        } else {
+          getActionURL();
+        }
+        break;
+      default:
+        notImplemented("url scheme " + scheme, window);
+    }
+    // https://html.spec.whatwg.org/#submit-mutate-action
+    function mutateActionURL() {
```
domenic: Would be ideal to avoid nested functions where possible, moving these to top-level of the module or class and passing in parameters as appropriate

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -53,19 +59,102 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
+      }
+      form._plannedNavigation = window.setTimeout(() => {
+        form._plannedNavigation = null;
+        console.log({destination});
+        navigate(targetWindow, destination, {});
+      }, 0);
+    }
+  }
+  _getAttributeForSubmission(name, submitter) {
+    return submitter.hasAttribute('form' + name) ? submitter.getAttribute('form' + name) : this.getAttribute(name);
```
domenic: Somewhat nicer to write as

```js
const formAttribute = submitter.getAttribute("form" + name);
return formAttribute === null ? this.getAttribute(name) : formAttribute;
```

lib/jsdom/living/window/navigation.js
```diff
@@ -21,6 +21,23 @@ exports.evaluateJavaScriptURL = (window, urlRecord) => {
   return undefined;
 };
 
+// https://html.spec.whatwg.org/#the-rules-for-choosing-a-browsing-context-given-a-browsing-context-name
+exports.chooseBrowsingContext = (currentWindow, name) => {
```
domenic: Can we use this in navigation from a elements too?

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -7,6 +7,12 @@ const { domSymbolTree } = require("../helpers/internal-constants");
 const createHTMLCollection = require("../../living/html-collection").create;
 const notImplemented = require("../../browser/not-implemented");
 const { reflectURLAttribute } = require("../../utils");
+const { constructTheFormDataSet } = require("../xhr/FormData-impl");
```
domenic: Ideally let's move this to `../xhr/helpers/formdata.js` or similar (i.e. -impl classes should not generally contain helper utilities)

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -53,19 +59,102 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
     }
-    const ev = this._ownerDocument.createEvent("HTMLEvents");
-    ev.initEvent("submit", true, true);
-    if (this.dispatchEvent(ev)) {
-      this.submit();
+
+    const form = this;
+    const document = this._ownerDocument;
+    const window = document._defaultView;
+
```
domenic: Check for a browsing context by checking if `window` is truthy.

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -53,19 +59,102 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
+    // NOT IMPLEMENTED: validate the form
+
+    if (!flags.submittedFromSubmit) {
+      const ev = document.createEvent("HTMLEvents");
+      ev.initEvent("submit", true, true);
```
domenic: The modern way to do this is

```js
const ev = Event.createImpl(["submit", { bubbles: true, cancelable: true }], {});
```

domenic (commented):
Lots of potential tests to enable or find out why they're not working here:

-  https://github.com/w3c/web-platform-tests/tree/master/html/semantics/forms/the-form-element form-action-*
- https://github.com/w3c/web-platform-tests/tree/master/html/browsers/browsing-the-web/navigating-across-documents navigation-unload-form-submit
- https://github.com/w3c/web-platform-tests/tree/master/html/semantics/forms/form-submission-0

jacobbogers:
Wow this is an oldy, what is the status? any way I can help out?

PR #1917, opened 2017-07-12
Undefined input parameter in getRequestHeader cause TypeError exception

skygon:
Object.keys will throw TypeError exception if requestHeaders is undefined. Need to check the input parameter firstly.


domenic:
This needs a test before we can accept it. Please see CONTRIBUTING.md.

TimothyGu:
I think we should fix the callers of `getRequestHeader()` that pass it `undefined` instead of `getRequestHeader()`.

domenic:
Closing due to lack of tests.

PR #1920, opened 2017-07-20
Correct changing readyState and script running order

sttk:
JSDOM is different from Web browsers about changing `document.readyState` and the order of running scripts.

This pr is for modifying this issue.

A sample HTML:

```
<!DOCTYPE html>
<html>
<head>
<meta charset="utf-8"/>
<title>script test</title>
<script>
console.log("script-0 (readyState = " + document.readyState + ")");
  console.log("window#load-0 (readyState = " + document.readyState + ")");
});
document.addEventListener("DOMContentLoaded", () => {
  console.log("document#DOMContentLoaded (readyState = " + document.readyState + ")");
});
function loaded() {
  console.log("body#onload (readyState = " + document.readyState + ")");
}
</script>
<body onload="loaded()">
<script>
console.log("script-1 (readyState = " + document.readyState + ")");
  console.log("window#load-1 (readyState = " + document.readyState + ")");
});
</script>
</body>
<script>
console.log("script-2 (readyState = " + document.readyState + ")");
  console.log("window#load-2 (readyState = " + document.readyState + ")");
});
</script>
</html>
```

The result by Chrome, Firefox, Edge, IE is as follows:

```
script-0 (readyState = loading)
script-1 (readyState = loading)
script-2 (readyState = loading)
document#DOMContentLoaded (readyState = interactive)
window#load-0 (readyState = complete)
body#onload (readyState = complete)
window#load-1 (readyState = complete)
window#load-2 (readyState = complete)
```

The result by JSDOM before modified is as follows:

```
script-0 (readyState = loading)
script-1 (readyState = loading)
script-2 (readyState = loading)
document#DOMContentLoaded (readyState = complete)
body#onload (readyState = complete)
window#load-0 (readyState = complete)
window#load-1 (readyState = complete)
window#load-2 (readyState = complete)
```


domenic:
Thanks! However, this does not follow the guidelines in CONTRIBUTING.md for introducing new tests.

Also, I am not happy with the code structuring here which modifies the global event handlers implementation to add a special case. You should be able to make it work without that.

domenic:
Sorry, maybe I wasn't clear. This should be tested using web-platform-tests (which we can then run against browsers), not mocha tests. And this should not impact the general event handler implementation (onxxx) at all; it should instead use the same eventing mechanism as anything else. There is nothing in the spec that corresponds to the special-case algorithm you are introducing for onload, so we shouldn't put anything in jsdom.

In general, please try to implement the spec.

sttk:
@domenic
> I am not happy with the code structuring here which modifies the global event handlers implementation to add a special case.

Oops! I pushed shameful code. I removed modifications from `GlobalEventHandlers-impl.js` and modified only `HTMLBodyElement-impl.js`.

> this does not follow the guidelines in CONTRIBUTING.md for introducing new tests.

I moved new test into `test/api`.

According to CONTRIBUTING.md, it may be good to place in `to-upstream` dir in the form of a HTML file.
However, I tried it but the result of `npm run test-tuwpt` was failure because the order of running `body.onload` handler was not changed correctly and same with before this modification.

Moreover, the result of `npm run test-karma` is also failure because `body.onload` handler does not run, (This behavior is seen from before this modification.)

I don't know the causes of these differences by tests, but it's OK for me that this runs correctly at least on Node.js, not on browser.

sttk:
@domenic 
> There is nothing in the spec that corresponds to the special-case algorithm you are introducing for onload, so we shouldn't put anything in jsdom.

I understand about the order of running scripts, and it's OK to cancel the modification in `HTMLBodyElement-impl.js`

Should I cancel the modification about `document.readyState`, too?
Changing it to "interactive" once before "complete" is written in [https://html.spec.whatwg.org/multipage/dom.html#current-document-readiness](https://html.spec.whatwg.org/multipage/dom.html#current-document-readiness).

sttk:
This issue which should be solved by this pr consists of five sub issues:

1. In `setChild` function of `htmltodom.js`, loading script files or evaluating inline scripts of all script elements, which are in `_attach` methods, are executed after parsing all HTML elements with their attributes, including event handlers.

    Therefore, `onload` of body is evaluated before all script element contents.

2. In `_attach` method of `HTMLScriptElement-impl.js`, evaluating inline script is executed asynchronously.

    It should be executed immediately after parsing script element.

3. In `_attach` method of `HTMLScriptElement-impl.js`, loading external script file is executed asynchronously, like specified `defer` attribute.

    It should be executed immediately after parsing script element if neither `async` nor `defer` are specified. 

4. In `vm-shim.js`, `vm.runInContent` does not deal with variables and functions declared at top level as globals, for example:

    ```js
    x = 2;  // global
    var y = 2; // not global
    function z() {} // not global
    ```

    Therefore, even if html contains `<body onload="z()">...`, the function `z` is not executed because `z` is not found.

    This causes on karma.

5. `DOMContentLoaded` event is occured after loading resources, such as script and css, and the value of `document.readyState` is changed to "completed".

    That event should be occured after parsing html and before loading resources. In addition, `document.readyState` should be changed to "interactive" before that event is occured.

---

Refs.

- [4.12.1 The script element - HTML Living Standard](https://html.spec.whatwg.org/multipage/scripting.html#the-script-element)
- [4.3.1 The script element - HTML5
A vocabulary and associated APIs for HTML and XHTML](https://dev.w3.org/html5/spec-preview/the-script-element.html#the-script-element)
- [3.1.2 Resource metadata management - HTML Living Standard](https://html.spec.whatwg.org/multipage/dom.html#dom-document-readystate)

sttk:
Still in the middle of modifications.

sttk:
I've modified the codes about the above 1,2,3,5 for this issue with new RequestQueue.
I defered the 4 because it is the issue of `acorn-globals`.

PR #1921, opened 2017-07-27
Allow beforeParse() to modify the HTML that will be parsed

justlep:
Parsing large HTML documents is resource-intensive, so sometimes it makes sense to limit parsing to only relevant parts of a document or parse nothing at all. For example if
(a) you're interested only in a fraction of the content ("region of interest")
(b) you can tell by a simple keyword-scan whether the document contains the wanted information at all.

With my modification, beforeParse() will be called with an additional, second parameter which is the html that would "normally" go straight to the parser. beforeParse() can now return a modified version of that html (or any other string, including empty ones) that will then be parsed instead. If it returns no string, parsing happens as usual, so the modification won't break any existing code.

I've included tests and a respective section in the README.

Depending on the document, I've seen performance increases of almost a second on a Pi3, so I really hope this becomes a feature of jsdom :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,30 @@ const dom = new JSDOM(`<p>Hello</p>`, {
 });
 ```
 
-This is especially useful if you are wanting to modify the environment in some way, for example adding shims for web platform APIs jsdom does not support.
+This is especially useful if you are wanting to modify the environment in some way, for example adding shims for web platform APIs jsdom does not support. 
+
+Futhermore, `beforeParse` allows you to **modify** the HTML that will be fed into the parser. This is particularly useful for dealing with large HTML documents when you are interested in small portions of their content only. Stripping the HTML down to a "region of interest" can speed up parsing significantly.
+
+```js
+const ROI_START_MARKER = '<!-- region of interest -->';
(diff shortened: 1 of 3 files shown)


domenic:
Thanks, but I don't see the value of this feature. Just pass the preprocessed string to jsdom directly, instead of passing the wrong string, letting jsdom create a window, having jsdom ask you to fix the string to be the right string, and then going back to jsdom code.

justlep:
I agree that the feature doesn't make sense when creating a JSDOM instance with a predefined html snippet. 

However, my usecase in mind is JSDOM.forUrl() (see my example code snippet in README.md). When you parse e.g. a typical phpbb forum and just want to parse the thread list, chances are that the page is 100kb while the relevant threadlist html is just 20k. Killing the "noise" around before parsing makes a huge difference here.

domenic:
Ok. In that case, I will close this, referring you to this paragraph from the readme:

> The initial request is not infinitely customizable to the same extent as is possible in a package like request; fromURL() is meant to be a convenience API for the majority of cases. If you need greater control over the initial request, you should perform it yourself, and then use the JSDOM constructor manually.

PR #1929, opened 2017-08-05
Add several obsolete no-op methods

Zirro:
This pull request adds some required no-op methods which are part of the 'Obsolete features' section of the HTML specification.

Specifically, `clear()`, `captureEvents()` and `releaseEvents()` on document:
https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-clear

And the latter two again on window:
https://html.spec.whatwg.org/multipage/obsolete.html#dom-window-captureevents

It also implements the [`External`](https://html.spec.whatwg.org/multipage/obsolete.html#dom-external) interface used for `window.external`, no-ops for the two SearchProvider methods that go with it and a to-upstream test for this.

Finally, I took the opportunity to remove some unimplemented IDL for the alternative stylesheets API which has been [removed](https://drafts.csswg.org/cssom/#changes-from-5-december-2013) from the CSSOM specification.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -16,6 +16,7 @@ const idlUtils = require("../living/generated/utils");
 const createXMLHttpRequest = require("../living/xmlhttprequest");
 const createFileReader = require("../living/generated/FileReader").createInterface;
 const Document = require("../living/generated/Document");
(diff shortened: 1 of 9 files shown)


domenic:
Awesome stuff! Thanks especially for the tests; they're obviously the hardest part!

Issue #1932, opened 2017-08-07
id="undefined" shadows actual undefined variable


TimothyGu:
### Basic info:

- **Node.js version:** v8.2.1
- **jsdom version:** v11.1.0

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const options = {
  runScripts: 'outside-only'
};
const dom = new JSDOM(`
  <div id="undefined"></div>
`, options);

console.log(dom.window.eval('undefined'));
// Prints "HTMLDivElement {}"
```

(RunKit: https://runkit.com/timothygu/jsdom-issue-1932)

### How does similar code behave in browsers?

Prints `undefined`.

TimothyGu:
Potentially a Node.js issue:

```js
> vm.runInNewContext('undefined', { undefined: 123 })
123
```

PR #1933, opened 2017-08-07
Only track properties if it's undefined in proxy

TimothyGu:
Fixes: #1932


domenic:
Sad about our flaky test suite, but thanks so much for this fix and test <3

PR #1934, opened 2017-08-07
Support IDL getters/setters/deleters

TimothyGu:
Unfortunately Proxy performance isn't there yet for me to port `NamedNodeMap`/`Element#attributes` over (ecmarkup compilation goes from 18s to 2min), but it is enough to make some other features possible (including the hotly wanted `dataset` and later `localStorage` and `sessionStorage`).

Requires https://github.com/jsdom/webidl2js/pull/48


lib/jsdom/living/nodes/HTMLSelectElement-impl.js
```diff
@@ -81,33 +99,49 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
...
-    return this.options.length;
+  get options() {
+    return this._options;
   }
 
   get selectedOptions() {
-    return createHTMLCollection(this, () => domSymbolTree.treeToArray(this, { filter(node) {
-      return node._localName === "option" && node._selectedness === true;
-    } }));
+    if (this._selectedOptions !== null) {
```
Zirro: No need to handle `[SameObject]` manually as webidl2js maintains its own cache.

Zirro:
Great work! I just figured I'd mention that there are a few test failures in the to-upstream tests and old tests (`npm run test-old`) that you may want to take a look at.

lib/jsdom/living/nodes/NodeList-impl.js
```diff
@@ -0,0 +1,43 @@
...
+  get length() {
+    this._update();
+    return this._list.length;
+  }
+  item(index) {
+    this._update();
+    return this._list[index] || null;
+  }
+  _update() {
+    if (this._isLive) {
```
domenic: Nit: bailing out early seems nicer

lib/jsdom/living/nodes/NodeList-impl.js
```diff
@@ -0,0 +1,43 @@
...
+    if (privateData.nodes) {
+      this._list = [...privateData.nodes];
+      this._isLive = false;
+    } else {
+      this._list = [];
+      this._isLive = true;
+      this._version = -1;
+      this._element = privateData.element;
+      this._query = privateData.query;
+      this._update();
```
domenic: Hmm, I had naively hoped we wouldn't have to do all this caching and stuff when we could just trigger on the appropriate property access. But that seems pretty dumb now; it'd slow things down way too much. Oh well. This is a nice simplification anyway.

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
```diff
@@ -75,24 +99,34 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
...
-  get options() {
-    return HTMLOptionsCollection.createImpl([], {
-      element: this,
-      query: () => descendantsByHTMLLocalName(this, "option")
-    });
+  get _displaySize() {
+    if (this.hasAttribute("size")) {
+      const attr = this.getAttribute("size");
+      // We don't allow hexadecimal numbers here.
+      // eslint-disable-next-line radix
```
domenic: Wow, today I learned about that special case. Fun times.

test/web-platform-tests/index.js
```diff
@@ -145,6 +146,14 @@ describe("Web Platform Tests", () => {
     "html/dom/dynamic-markup-insertion/document-writeln/document.writeln-02.html",
     "html/dom/dynamic-markup-insertion/document-writeln/document.writeln-03.html",
     "html/dom/elements/global-attributes/classlist-nonstring.html",
+    "html/dom/elements/global-attributes/custom-attrs.html",
+    "html/dom/elements/global-attributes/data_unicode_attr.html",
+    // "html/dom/elements/global-attributes/dataset.html", // we only support HTML elements
+    "html/dom/elements/global-attributes/dataset-delete.html",
+    "html/dom/elements/global-attributes/dataset-enumeration.html",
+    "html/dom/elements/global-attributes/dataset-get.html",
+    // "html/dom/elements/global-attributes/dataset-prototype.html", // tests Object.prototype
```
domenic: Is this about vm issues, or...?

lib/jsdom/browser/domtohtml.js
```diff
@@ -8,7 +8,7 @@ exports.domToHtml = function (iterable) {
   let ret = "";
   for (const node of iterable) {
     if (node.nodeType === NODE_TYPE.DOCUMENT_NODE) {
-      ret += parse5.serialize(node, {
+      ret += parse5.serialize(idlUtils.tryWrapperForImpl(node), {
```
domenic: Can you explain why this is necessary in the commit message? What changed, or were we buggy before?

test/web-platform-tests/index.js
TimothyGu: Yeah, the dataset object's prototype is the Object.prototype from another <del>universe</del> context, not the window one.

lib/jsdom/browser/domtohtml.js
TimothyGu: Sure. domToHtml accepts an iterable of nodes, but before this change we used both wrappers and impls against it. It worked before, but now that certain impls are not array-like (only their Proxy wrappers are) parse5 throws on certain nodes. I'm just making sure only wrappers are passed to parse5.

I also looked into changing the callers of domToHtml to pass wrappers, but for some reason I forgot it seems to be less doable than just changing domToHtml.

test/web-platform-tests/index.js
domenic: Great, maybe just expand the comment so we know in teh future

lib/jsdom/browser/domtohtml.js
domenic: I see, interesting. Just include that in the commit message please :). Maybe also a TODO comment about fixing the callers.

domenic:
FYI my plan is:

- Merge #1950 
- Update this PR with my nits
- Make sure it passes all tests when run against webidl2js master (especially worth checking the failures @Zirro noted)
- When we're sure webidl2js new features work for this PR, release it (minor version)
- Update this PR's version reference for webidl2js
- Merge it

Zirro:
Having looked at the latest commits, rebasing on https://github.com/tmpvar/jsdom/commit/3743b423cb45657401eef3475a07fd99e015b2e2 will probably resolve the failures. The cause seems to have been a change that was unintentionally included here in addition to that commit.

domenic:
Thanks so much @TimothyGu; this is amazing, herculean work!

StefanoDeVuono:
Can you publish this to the npm registry?

PR #1950, opened 2017-08-20
Roll WPT; lots of resulting updates

domenic:
Ironically, I was mainly doing this because I was looking at all the "TODO uncomment when this test is fixed" in the WPT manifest. However, instead I ended up discovering lots of new WPT bugs and commenting out the resulting tests: https://github.com/w3c/web-platform-tests/issues/6940 https://github.com/w3c/web-platform-tests/issues/6941 https://github.com/w3c/web-platform-tests/issues/6942

Along the way I converted a decent amount of stuff to webidl2js, so that's always nice.

Will merge if CI likes it; no need to review.


domenic:
NodeIterator rewrite is blocked on https://github.com/whatwg/dom/issues/496. If that doesn't get resolved quickly, maybe I should back that out and just mark the new test as failing...

PR #1951, opened 2017-08-20
Element.prototype.closest

caub:
continues #1784

web platform tests fails for .closest on `:invalid` selector (throw SyntaxError)

Also the former .matches fails the wpt, SyntaxError on a `>>` selector, apparently it shouldn't throw for them (`:invalid` is a valid css-selector, so maybe .matches parser doesn't handle it)

seems it is an option https://github.com/dperini/nwmatcher/wiki/CSS-supported-selectors#webforms-and-html5-support-optional-external-add-on  `USE_HTML5` option (going to check if it's enabled)


domenic:
I think the real problem here is that it's not using `:scope`, which is necessary to get correct results for things like `> *`. I believe @dperini was planning on working on that at some point, but until we have that, I don't think we can implement closest().

dperini:
The implementation details are not settled yet across browsers, I have been thinking about it but differences between current implementations pushed me to delay the addition of the `:scope` pseudo-class selector. However, since I am currently working on a new completely rewritten version I will consider its inclusion.

The following two links are about the new pseudo-class and the related implications.

https://developers.google.com/web/updates/2013/03/What-s-the-CSS-scope-pseudo-class-for
https://developer.mozilla.org/en-US/docs/Web/CSS/:scope

test/web-platform-tests/to-run.yaml
```diff
@@ -169,7 +169,7 @@ Element-childElementCount-dynamic-add-xhtml.xhtml: [fail, Unknown]
 Element-childElementCount-dynamic-remove-xhtml.xhtml: [fail, Unknown]
 Element-childElementCount-nochild-xhtml.xhtml: [fail, Unknown]
 Element-childElementCount-xhtml.xhtml: [fail, Unknown]
-Element-closest.html: [fail, Not implemented]
+Element-closest.html: [fail, Unknown]
```
caub: Failed in "Element.closest with context node 'test11' and selector ':invalid'":

dperini:
Nice addition 👍 ,

FYI and as a tip, this is what I currently have in the new rewrite of `nwmatcher` (to be released soon as `nwsapi`):

    // equivalent of w3c 'closest' method
    ancestor =
      function _closest(selector, element) {
        while (element) {
          if (match(selector, element)) break;
          element = element.parentElement;
        }
        return element || null;
      },  

by using `parentElement` I skip the `nodeType` check for when the walk-up reaches the `document` node (Node.DOCUMENT_NODE == 9). Performances should be sightly better.
`null` will be returned when no closest is found and the walk-up reaches the top `Element` (**document.documentElement**) or when `undefined` is passed as the `element` argument.

I also found exhaustive test for the `.closest` method here:

    https://github.com/w3c/web-platform-tests/blob/master/dom/nodes/Element-closest.html

but maybe they are the same you are already using (30 tests).

caub:
Thanks, yes it's better

Great, looking forward to try `nwsapi`

caub:
@dperini "The implementation details are not settled yet across browsers" I think `<style scoped>` and `:scope` are more or less abandoned (I could be wrong), only implemented in Firefox, and web-components/shadow-dom are supposed to replace that functionality

domenic:
:scope is very much alive. It is unrelated to scoped stylesheets.

dperini:
@caub thank you for the notice about the ':scope' pseudo-class.
Yes in fact I haven't seen any move on ':scope' recently and I don't like spreading <style scoped> scattered any place in the body of the document. I believe it is an HTML degradation not an improvement. The pseudo-class for itself may be handy in specific cases.
However somebody is playing with that pseudo-class and if possible I want to help them too, at least make it so we avoid throwing errors in existing document using it.

caub:
Oh, sorry like Domenic said I was wrong about `:scope`

The wpt tests still fails with the 'old' nwmatcher at [[*|title]](https://github.com/w3c/web-platform-tests/blob/master/dom/nodes/selectors.js#L82)  https://jsfiddle.net/crl/ad3dc5of, but it looks fine with nwsapi https://jsfiddle.net/crl/ad3dc5of/1 so maybe it could be worth starting to integrate it

folmert:
Any chance this can be merged & released soon? I get `TypeError: e.target.closest is not a function` in my tests...

dperini:
@folmert Can you provide the test showing the issue with "closest()" ?

TimothyGu:
@dperini It's not a bug in nwsapi, but rather in jsdom. We just haven't implemented `.closest()` yet.

dperini:
@TimothyGu thank you for the heads up.
Should I publish "nwsapi" on npm to make testing and integration easier ?
Is there something else I can do to speed up the switch to "nwsapi" ?

caub:
@dperini Yes on npm, it would be nice, I'll update this branch as soon as it's ready

Zirro:
@dperini Publishing nwsapi on npm would be great, but unfortunately we would still need to fix https://github.com/jsdom/jsdom/pull/2058 before we can use it. None of us have had time yet, it seems.

ilyazub:
@Zirro #2058 has been merged month ago. What else should be done before this PR?

caub:
note: [nwsapi already provides closest](https://github.com/dperini/nwsapi/blob/master/src/nwsapi.js#L1245) we might use it, even if it's small

Zirro:
Yes, now that we are using nwsapi we can expose this.

@caub, do you want to refactor this pull request to use the `ancestor` method from nwsapi instead? Otherwise I can see if I have time to look at it over the weekend.

caub:
yes done @Zirro

dperini:
Should I rename something to better fit the needs ?

caub:
@dperini `.closestAncestor`, nah :) no need to change that, the exposed one is great

lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -421,6 +421,20 @@ ElementImpl.prototype.matches = memoizeQuery(function (selectors) {
 
 ElementImpl.prototype.webkitMatchesSelector = ElementImpl.prototype.matches;
 
+ElementImpl.prototype.closest = memoizeQuery(function (selectors) {
```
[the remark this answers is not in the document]
domenic: This comment still holds. We shouldn't memoize closest().

test/web-platform-tests/to-run.yaml
domenic: @dperini @Zirro I thought we supported `div > :scope` now?

Zirro: @domenic `:scope` is supported in general, but the `div > :scope` test hits a bug in the implementation. It was not fixed earlier due to being confused with the last test which fails in all browsers - see https://github.com/web-platform-tests/wpt/issues/11316.

I've spoken to @dperini and it should be fixed in the next release of nwsapi.

domenic: Ah OK. @caub, we can either sit tight and wait for the next nwsapi release, or we can copy over the test and comment out the failing one for now---see other "dont-upstream.html" examples.

dperini:
@domenic
I already have a fix for that, the problem is that I currently use `.contains()` in the `:scope` resolver  and it's implementation seems to maintain the original IE buggy behavior since:

    e.contains(e) // true but should be false

so I am going to use `compareDocumentPosition()` in place of `.contains()`.

Should I make a new release just for that or can I wait other patches I am working on ?

Zirro:
@dperini Nice, I see that you've committed the fix. It would be good if you could make a release with this now, just in case the other changes cause unexpected issues.

dperini:
@Zirro 
Done ! I released version 2.0.2 and published on npm.

caub:
Perfect, it passes all tests, except the one all browsers fail on (`:has`) I tried adding a `to-upstream/dom/nodes/Element-closest-dont-upstream.html` with that test commented, but I'm missing something since it doesn't help to pass the tests yet

```
$n test-apt --fgrep=dom/nodes
 1 failing

  1) web-platform-tests dom/nodes Element-closest.html:
     Failed in "Element.closest with context node 'test4' and selector ':has(> :scope)'": 
unknown pseudo-class selector ':has(>:scope)'
```

domenic:
Thank you @caub for your epic patience!!

PR #1952, opened 2017-08-20
Do not trigger click on a disabled labeled control

### LINKED ISSUE #1925: Clicking <label> causes disabled checkbox to become checked
### Basic info:

- **Node.js version:** 6.11.1
- **jsdom version:** 11.1.0

It appears that if you trigger a click on a `<label>` that wraps a disabled `<input type="checkbox">`, jsdom will change the `checked` property. This does not occur if you trigger the click on the `<input>` directly. This is not the behavior in Chrome.

### Minimal reproduction case

```js

const dom = new JSDOM('\
  <!DOCTYPE html>\
  <html>\
    <body>\
      <label><input type="checkbox"></label>\
    </body>\
  </html>\
');

const label = dom.window.document.querySelector('label');
const input = dom.window.document.querySelector('input');

label.click();
console.log('expecting true, actual', input.checked); // actual = true

input.click();
console.log('expecting false, actual', input.checked); // actual = false

input.disabled = true;

input.click();
console.log('expecting false, actual', input.checked); // actual = false

label.click();
console.log('expecting false, actual', input.checked); // actual = true (UNEXPECTED)
```

Output:

```
expecting true, actual true
expecting false, actual false
expecting false, actual false
expecting false, actual true (UNEXPECTED)
```

In OSX/Chrome 59:

```
expecting true, actual true
expecting false, actual false
expecting false, actual false
expecting false, actual false
```

https://jsfiddle.net/cja7cpng/

schreifels:
Fixes #1925 

I opted to test the specific incorrect behavior I discovered (where the labeled control is a checkbox). Alternatively, I could simply test that the "click" event is never dispatched. Let me know if that more generic test would be preferable.

### DIFF
--- a/test/web-platform-tests/to-upstream/html/semantics/forms/the-label-element/disabled-labeled-control.html
+++ b/test/web-platform-tests/to-upstream/html/semantics/forms/the-label-element/disabled-labeled-control.html
@@ -1,0 +1,93 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Label event handling with disabled labeled control</title>
+
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<body>
+
+<label id="label1">
+  <input id="input1" type="text" disabled>
+</label>
+
+<label id="label2" for="input2"></label>
+<input id="input2" type="text" disabled>
+
+<label id="label3">
+  <input id="input3" type="checkbox" disabled>
+</label>
+
+<label id="label4" for="input4"></label>
+<input id="input4" type="checkbox" disabled>
+
+<script>
+"use strict";
+
+test(() => {
+  const label = document.getElementById("label1");
+  const input = document.getElementById("input1");
+
+  let clickReceived = false;
+  input.addEventListener("click", () => {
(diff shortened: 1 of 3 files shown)


test/web-platform-tests/to-upstream/html/semantics/forms/the-label-element/disabled-labeled-control.html
```diff
@@ -0,0 +1,46 @@
...
+  assert_true(checkbox1.disabled);
+  assert_false(checkbox1.checked);
+  label1.click();
+  assert_false(checkbox1.checked);
+
+  checkbox1.disabled = false;
+  label1.click();
+  assert_true(checkbox1.checked);
+
+  const label2 = document.getElementById("label2");
```
domenic: Can you separate this into two `test()`s? They are testing distinct things.

schreifels:
@domenic Sounds good, thanks! I updated the PR with the changes you requested. Let me know if there is anything else.

snuggs:
I wish i would have known this last month @schreifels. :-/ Can't get that hour back. My former self says thanks for this. <3

PR #1953, opened 2017-08-21
Convert DOMTokenList to IDL + related goodies

TimothyGu:
See also https://github.com/tmpvar/jsdom/issues/1129.


lib/jsdom/living/nodes/HTMLTableCellElement-impl.js
```diff
@@ -34,19 +29,49 @@ class HTMLTableCellImpl extends HTMLElementImpl {
...
+      return -1;
+    }
+
+    return tr.cells.indexOf(this);
+  }
+
+  get scope() {
+    // TODO: This logic should only apply to <th>, not <td>, since only <th> defines the scope attribute and its states.
+    // But Firefox and Chrome alike seem to apply it to <td> as well.
+    // https://html.spec.whatwg.org/#attr-th-scope
```
TimothyGu: Correct me if I'm wrong about this of course, but IMO this should be true:

```js
// From the test I added
const th = document.createElement("th");
th.scope = "row";
console.assert(th.scope === "row");

const td = document.createElement("td");
td.scope = "row";
console.assert(td.scope === "");
```

In Firefox and Chrome only the first case is true. In the second case `td.scope` equates to `"row"` even though `td` elements should not support the `scope` attribute.

TimothyGu:
To clarify, this requires https://github.com/jsdom/webidl2js/pull/65. https://github.com/jsdom/webidl2js/pull/66 would be good too but not absolutely required for the WPTs I enabled. The `iteration` test does test values (and is how I discovered the bug) but was unfortunately disabled for unrelated reasons.

lib/jsdom/living/nodes/DOMTokenList-impl.js
```diff
@@ -0,0 +1,197 @@
...
+
+  replace(token, newToken) {
+    validateTokens(token, newToken);
+    this._syncWithElement();
+
+    let seen = false;
+
+    for (let i = 0; i < this._tokenSet.length;) {
+      if (this._tokenSet[i] === token || this._tokenSet[i] === newToken) {
+        if (seen === false) {
```
domenic: Just use !seen

lib/jsdom/living/nodes/DOMTokenList-impl.js
```diff
@@ -0,0 +1,197 @@
...
+    return false;
+  }
+
+  replace(token, newToken) {
+    validateTokens(token, newToken);
+    this._syncWithElement();
+
+    let seen = false;
+
+    for (let i = 0; i < this._tokenSet.length;) {
```
domenic: This kind of complexity makes me wonder if a separate file/class for Infra's ordered set might be a good idea.

lib/jsdom/living/nodes/HTMLFrameElement-impl.js
```diff
@@ -27,7 +27,7 @@ function loadFrame(frame) {
   // https://html.spec.whatwg.org/#process-the-iframe-attributes
   let url;
   const srcAttribute = getAttributeValue(frame, "src");
-  if (srcAttribute === null || srcAttribute === "") {
+  if (srcAttribute === "") {
```
domenic: Why this change?

lib/jsdom/living/nodes/HTMLTableCellElement-impl.js
```diff
@@ -3,29 +3,24 @@
...
-      return -1;
-    }
-
-    return tr.cells.indexOf(this);
-  }
-
   get colSpan() {
     const value = this.getAttribute("colspan");
-    return value === null ? 1 : value;
+    if (value === null) {
```
domenic: Given the duplicate colSpan/rowSpan code, we should factor this out, so that it's something like `return reflectedAttributeWithinRange(this.getAttribute("colspan"), 1, 1000)`

lib/jsdom/living/helpers/traversal.js
```diff
@@ -4,9 +4,9 @@ const { DOCUMENT_NODE } = require("../node-type");
 
 // All these operate on and return impls, not wrappers!
 
-exports.closest = (e, localName) => {
+exports.closest = (e, localName, namespaceURI = undefined) => {
```
domenic: Is there actually any time we use this that it shouldn't be the HTML namespace?

lib/jsdom/living/nodes/HTMLTableCellElement-impl.js
domenic: What's the TODO here? The spec for HTMLTableCellElement's scope is that it reflects the scope content attribute. Chrome and Firefox seem to be working as intended.

I guess the spec could add a note that even though the scope content attribute is defined for th, it is reflected on tds anyway. But this implementation seems correct.

test/web-platform-tests/to-upstream/html/semantics/tabular-data/the-td-element/class-hierarchy.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE HTML>
```
domenic: This is kind of redundant with the updates done at https://github.com/w3c/web-platform-tests/commit/5b6aa3d108eeb306a326465e87447b53dee68a13 but given that we can't run partial tests I guess it is the best we're going to do...

test/web-platform-tests/to-upstream/html/semantics/tabular-data/the-th-element/scope.html
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE HTML>
```
domenic: Also redundant with existing reflection tests but probably worth keeping given our limitations.

lib/jsdom/living/helpers/traversal.js
TimothyGu: Yes, like the following WPT that'll get enabled by this PR:

https://github.com/w3c/web-platform-tests/blob/be8b3bbb1017519fa27bd90b8a50bf0b5b33aedf/html/semantics/tabular-data/attributes-common-to-td-and-th-elements/cellIndex.html#L27-L33

domenic: Right, but, shouldn't all callers of closest() always pass the HTML NS?

lib/jsdom/living/nodes/HTMLTableCellElement-impl.js
TimothyGu: > I guess the spec could add a note that even though the scope content attribute is defined for th, it is reflected on tds anyway.

That is exactly the bug. Even though the `scope` IDL attribute is defined for td, it defers to the "known values" for the specific element. Its known values are not defined for td (only for th), so the default known values of the empty set should be used for td. Yet browsers still apply the known values of `scope` for th to td.

lib/jsdom/living/nodes/HTMLFrameElement-impl.js
TimothyGu: `getAttributeValue` will return the empty string if the attribute doesn't exist, per [spec](https://dom.spec.whatwg.org/#concept-element-attributes-get-value).

lib/jsdom/living/nodes/HTMLTableCellElement-impl.js
domenic: Well, known values are really for the attribute, not the element. At least that's the intent.

lib/jsdom/living/helpers/traversal.js
TimothyGu: Do you mean the default value should be the HTML NS or am I understanding wrong?

domenic: Well, it should just always check HTML NS, and never allow checking for anything else, I think.

lib/jsdom/living/nodes/HTMLTableCellElement-impl.js
TimothyGu: > known values are really for the attribute, not the element

Indeed, but `td` element doesn't have a `scope` attribute...

----

Alright I'll change it. Do note though that this part of spec can be seen as fairly ambiguous.

lib/jsdom/living/nodes/DOMTokenList-impl.js
```diff
@@ -0,0 +1,197 @@
...
+  hasAttributeByName
+} = require("../attributes.js");
+
+function validateTokens(...tokens) {
+  for (const token of tokens) {
+    if (token === "") {
+      throw new DOMException(DOMException.SYNTAX_ERR, "The token provided must not be empty.");
+    }
+  }
+  for (const token of tokens) {
```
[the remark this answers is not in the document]
TimothyGu: That's why `add` calls validateTokens always with a single argument :p

lib/jsdom/living/helpers/traversal.js
TimothyGu: Didn't seem to break anything, so done.

lib/jsdom/living/index.js
```diff
@@ -58,7 +59,6 @@ require("../level3/xpath")(exports);
 
 // These are OK but need migration to webidl2js eventually.
```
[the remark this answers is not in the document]
TimothyGu: Moved NamedNodeMap down here, so it's plural again ;)

TimothyGu:
Comments addressed. webidl2js changes published. Tests should be green.

lib/jsdom/living/helpers/strings.js
```diff
@@ -1,5 +1,15 @@
 "use strict";
 
+// https://infra.spec.whatwg.org/#ascii-lowercase
+exports.asciiLowercase = s => {
+  return s.replace(/[A-Z]/g, l => l.toLowerCase());
+};
+
+// https://infra.spec.whatwg.org/#ascii-uppercase
+exports.asciiUppercase = s => {
```
Sebmaster: Unused?

Sebmaster:
Got through everything, rest looks good to me.

lib/jsdom/living/nodes/HTMLTableCellElement-impl.js
```diff
@@ -3,50 +3,63 @@
...
+    if (tr === null) {
+      return -1;
+    }
+
+    return tr.cells.indexOf(this);
+  }
+
+  get scope() {
+    const value = this.getAttribute("scope");
+    if (value === "row" || value === "col" || value === "rowgroup" || value === "colgroup") {
```
Sebmaster: This should match [ascii case insensitive](https://html.spec.whatwg.org/#enumerated-attribute).

lib/jsdom/living/helpers/ordered-set.js
```diff
@@ -0,0 +1,80 @@
+"use strict";
+
+// Only use this class if a Set cannot be used, e.g. when "replace" operation is needed, since there's no way to replace
+// an element while keep the relative order using a Set, only remove and then add something at the end.
+
+module.exports = class OrderedSet extends Array {
```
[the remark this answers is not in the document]
TimothyGu: The intention is for the user of this class not to use regular Array methods such as `push()`. Wrapping the array would also make `set[idx]` impossible.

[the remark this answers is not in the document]
TimothyGu: I'd rather optimize the algorithms if only found to be beneficial and necessary.

lib/jsdom/living/helpers/strings.js
TimothyGu: Indeed, but it makes sense to have an upper case counterpart to lower case, no?

lib/jsdom/living/helpers/ordered-set.js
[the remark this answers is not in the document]
domenic: If the intention is not to expose regular Array methods, this should not inherit from Array, but wrap it. It doesn't appear any of the client code uses indexed access either, so ideally that shouldn't be exposed too.

lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -115,8 +114,8 @@ class ElementImpl extends NodeImpl {
     }
 
     // update classList
-    if (name === "class") {
-      resetDOMTokenList(this.classList, value);
+    if (name === "class" && this._classList !== undefined) {
+      this._classList._attrModified();
```
domenic: It looks like _attrModified shouldn't be "private". Picture a world where we move all our underscore-prefixed stuff to actual private fields/methods, and this doesn't work.

TimothyGu: Changed.

PR #1956, opened 2017-08-22
Enable document.forms test

Zirro:
I gave the disabled tests another spin considering recent changes. Thanks to the hard work of @TimothyGu, it seems HTMLCollection is wonky no more :-)


TimothyGu:
Always happy to see tests get enabled!

domenic:
Looking forward to my new test runner where we get a failing build if a previously-failing test starts passing :). I should probably open an issue outlining that...

PR #1964, opened 2017-08-30
Support Canvas 2.x

zbjornson:
Fixes Automattic/node-canvas#978

I'm not sure if there's a good way to add tests for this; would have to test against two different versions of node-canvas somehow.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLCanvasElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLCanvasElement-impl.js
@@ -15,11 +15,7 @@ class HTMLCanvasElementImpl extends HTMLElementImpl {
 
   _getCanvas() {
     if (Canvas && !this._canvas) {
-      if (Canvas.createCanvas) { // node-canvas 2.x
-        this._canvas = Canvas.createCanvas(this.width, this.height);
-      } else { // node-canvas 0.x, 1.x
-        this._canvas = new Canvas(this.width, this.height);
-      }
+      this._canvas = Canvas.createCanvas(this.width, this.height);
     }
     return this._canvas;
   }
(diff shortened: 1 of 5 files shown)


zbjornson:
Oh, I didn't realize that the CI matrix already includes node-canvas and node-canvas-prebuilt as separate scenarios. Do you want me to add node-canvas@2.0.0-alpha.4? (node-canvas-prebuilt hasn't been updated yet with the breaking 2.x change that this addresses.)

asturur:
i really need this!
since i m bound to work on jsdom 9 for node 4 support, do you think i can backport it to jsdom9 and making a 9.13 release without forking?

zbjornson:
I don't quite understand your question, or know if you're asking me vs. the maintainers, but this patch makes it compatible with both 1.x and 2.x, so could be back ported to any version as necessary.

asturur:
i was generally asking if a backport of this would be accepted to 9.x, being 9.x the latest available for node4.

TimothyGu:
@asturur We (jsdom maintainers) are not interested in maintaining old versions of jsdom. If you would like Node.js 4.x support, we urge you to maintain your fork.

lib/jsdom/living/nodes/HTMLCanvasElement-impl.js
```diff
@@ -69,6 +73,7 @@ class HTMLCanvasElementImpl extends HTMLElementImpl {
     const canvas = this._getCanvas();
     if (canvas) {
       let stream;
+      // TODO: use toBuffer instead: faster and less code.
```
TimothyGu: Is this something that could be done easily? If so I wonder if it might be easier to just do so in this PR?

TimothyGu:
@zbjornson Regarding tests, just adding node-canvas@2 is fine. I vaguely remember some issues users had with node-canvas-prebuilt, but given that we only test that package against Node.js 6.x I think we can leave it alone for now.

asturur:
not even a small small patch like this? a quick 9.13 version?

I really not like the idea of deciding between:
no node4 for fabricjs
no canvas 2.x for fabricjs
no official jsdom for fabricjs

every option seems bad to me.

TimothyGu:
- Node.js 4.x is in LTS
- node-canvas 2.x is still in alpha
- jsdom 9.x isn't "official jsdom" :)

domenic:
In general I'd prefer to not merge this until node-canvas 2.x is released final so that we know we won't be making more changes.

At that point we may want to drop support for 1.x.

zbjornson:
Sounds good; I'll update this PR when node-canvas 2.x is out of alpha.

aggrosoft:
So canvas-2 is out of alpha, any plans to continue on this?

zbjornson:
Yep, I'll update it in my morning (Pacific time).

zbjornson:
@TimothyGu @domenic this is ready for review again.

The first commit adds 2.x compat while maintaining 1.x compat. The second commit drops 1.x compat. They could be squashed, or the 2nd one could be dropped, depending on your preference. If the second commit is kept, the placeholder version in the readme needs to be replaced. There is [so much fixed](https://github.com/Automattic/node-canvas/blob/master/CHANGELOG.md#200) in 2.x that I encourage you to drop 1.x support.

The third commit removes the canvas-prebuilt suite. Canvas 2.x uses prebuilds by default, so it's redundant. We could remove the `apt:` section that installs the Canvas dependencies, but since your test suite uses `yarn install canvas` (with no version spec), your builds would break if a new version of Canvas is published before the corresponding prebuild is finished (an [issue we need to fix](https://github.com/node-gfx/node-canvas-prebuilt/issues/57)).

Ailrun:
Stable version of canvas@2 is released

domenic:
I'll pull this in and we'll do a new major release. Woohoo!

asturur:
do you plan to leave node 6 in?

domenic:
Node v6 hasn't been supported for many releases.

asturur:
i think is my fault i m stuck at jsdom 9 yet. T.T

Issue #1958, opened 2017-08-23
New WPT runner design


domenic:
_Edit: most of this was taken care of in #1966. What remains is summarized in https://github.com/tmpvar/jsdom/issues/1958#issuecomment-333435727._

Here are the problems I want to solve:

- Our manifest format is a JavaScript array with comments. It's not easy to read, modify, or write
- We have to list a lot of files individually. That's kind of dumb, when we mostly want to run a whole directory.
- When new WPT get added to a directory for which we believe we implement, we don't notice and enable them. They might pass; they might reveal bugs, or spec changes; we just don't really know. We only notice if a WPT roll uncovers changed or deleted files
- If we start passing a previously-failing test, we don't notice, and thus we lose the regression protection.
- We cannot partially pass a test. If a stricter assertion gets added to a test which we previously passed, we now have to disable it.

Here's the plan:

- A new manifest format. It's directory-based, and you list the failures and timeouts. I think it's going to be to-run-2.yaml from https://gist.github.com/domenic/abaeb197600dad644d3ea8ee8a3f21a4
  - to-run-3.yaml doesn't work, I believe, because we need to distinguish between timeouts and failures
  - Thoughts welcome on whether to-run.yaml might be better
  - I'm not sure what to do with timeouts; running them every time to verify they still time out kind of sucks. For now I'm thinking don't run them, but maybe have a special command that does that we can manually try.
  - This is derived from WPT's manifest tooling which allows you to get a list of all test URLs (including .any.js-derived ones), excluding support files
- An option in the manifest format to list a list of failing assertions in the test, instead of just saying "fail". I'm less sure how this would look. We'd need it to be easily input, e.g. by outputting the failure text from the test runner for easy copying and pasting.

The first part of this is mostly working on my local machine, actually. The biggest work is just converting all the lines from our big old JS array into appropriate directory-scoped YAML documents, especially since some new tests have been added to our directories along the way.

Would love any feedback the contributors have!

Zirro:
This looks like a great improvement over the current system. While it's just a small detail among many good things, as someone who's had to balance quotes or insert missing commas one time too many I'm glad you went with YAML for the manifest ;-)

I am currently looking for a WPT-testing solution for [css-object-model](https://github.com/Zirro/css-object-model) that I can keep within the repo rather than having to run it through the separate jsdom repository, without necessarily having to duplicate the whole setup.

[...]

PR #1966, opened 2017-09-01
[WIP] new WPT runner

domenic:
@Zirro requested that I upload this so he could help with it, so here it is. Please note the MANIFEST.json generation step.

Here are the todos:

- [x] Port existing list of tests to this format, by finding all the directories we plan to pass, and adding exception entries for all the failing tests. Some of these might already exist as commented-out tests inside the existing manifest, but probably some do not. Note that timeouts should be noted as "timeout", not "fail", for future work.
- [x] Generate and commit a manifest. Currently I did `wpt manifest` inside the `web-platform-tests/tests/` directory, which generates a .gitignored MANIFEST.json in that directory. I think for our purposes we should generate it inside the `web-platform-tests/` directory and commit it to the repository, then make updating it part of the WPT roll process.
- [x] Speaking of which, we should document the WPT roll process a bit better. Maybe a README.md in the web-platform-tests directory would work well.
- [x] Update the to-upstream test runner to no longer use a terrible manifest format as well, but instead just do readdir recursive or something.
- [x] Delete the old runner and replace all relevant package.json scripts, CONTRIBUTING.md references, etc.

And here are some nice-to-haves, not necessary before merging:

- [x] Add some sanity-check errors, e.g. if a directory in to-run.yaml doesn't exist, or if a filename specified there doesn't exist. Maybe also if the directory ends with a / or the filename starts with a /. Maybe also check that each entry is an array whose first element is either the string "fail" or "timeout". (Mostly done)
- [ ] Run failing tests (but not timeout tests) and confirm that they fail.
- [x] Change fail/timeout tests to have a different name, instead of just the test file path.


test/web-platform-tests/to-run.yaml
```diff
@@ -0,0 +1,21 @@
...
+HTMLCollection-as-proto-length-get-throws.html: [fail, https://github.com/jsdom/webidl2js/issues/62]
+HTMLCollection-supported-property-indices.html: [fail, Named versus indexed properties trickiness]
+HTMLCollection-supported-property-names.html: [fail, Names vs. ids priority and effect on enumeration order]
+domstringmap-supported-property-names.html: [fail, DOMStringMap not implemented]
+
+---
+
+DIR: dom/events
+
+AddEventListenerOptions-passive.html: [fail, https://github.com/tmpvar/jsdom/issues/1609]
```
Zirro: Could we place URLs to issues in a third (optional) index of the array? Keeping the short error descriptions and URLs separate could potentially be useful if we want to do more with the output in the future, and might help keep track of which failures are covered by existing bugs.

domenic: Seems reasonable. Assuming yaml supports syntax like `[fail, , https://...]`? I.e. omitting array elements.

TimothyGu:
Link: #1958.

test/web-platform-tests/to-run.yaml
Zirro: It seems like it does, though I figure that all failures ought to have a short description (with or without a link) to make it easier to tell the cause just from reading the list.

domenic: I guess so, just seems a bit annoying to have to duplicate the information. But I suppose it won't be that bad.

test/web-platform-tests/to-run.yaml
```diff
@@ -0,0 +1,21 @@
+DIR: dom/collections
+
+HTMLCollection-as-proto-length-get-throws.html: [fail, https://github.com/jsdom/webidl2js/issues/62]
+HTMLCollection-supported-property-indices.html: [fail, Named versus indexed properties trickiness]
+HTMLCollection-supported-property-names.html: [fail, Names vs. ids priority and effect on enumeration order]
+domstringmap-supported-property-names.html: [fail, DOMStringMap not implemented]
```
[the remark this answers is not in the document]
domenic: Less indentation seems nice, but maybe that version is fine...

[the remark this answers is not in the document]
TimothyGu: That's good enough of a reason for me.

test/web-platform-tests/to-run.yaml
```diff
@@ -14,6 +14,105 @@ url/**: [fail, blob URLs not implemented]
...
+send-entity-body-get-head.htm: [fail, hard to get Node to not send Content-Length]
+send-entity-body-none.htm: [fail, hard to get Node to not send Content-Length]
+send-redirect-infinite-sync.htm: [fail, the test seems broken locally]
+send-redirect-infinite.htm: [fail, the test seems broken locally]
+send-redirect-post-upload.htm: [fail, Unknown]
+send-redirect-to-cors.htm: [fail, request module remove content-type header on redirect]
+send-redirect.htm: [fail, request module remove content-type header on redirect]
+send-response-event-order.htm: [fail, why no readystatechange for headers received? Test seems wrong maybe?]
+send-sync-response-event-order.htm: [fail, not sure; sync is special]
+send-usp.html: [fail, needs URLSearchParams implementation]
```
TimothyGu: We have a URLSearchParams impl already

Zirro:
I noticed that a few comments were missing from the new version. I believe [this](https://gist.github.com/Zirro/3dca89e6130aebcda7ca283722f77ec0) covers all of them.

domenic:
This is ready for review; I think CI will pass with my newest tricks.

test/web-platform-tests/to-run.yaml
```diff
@@ -0,0 +1,658 @@
...
+redirect-preflight-2.htm: [fail, preflight should also be done before redirected requests but request module redirects cannot be paused while doing preflight]
+remote-origin.htm: [fail, postMessage event does not contain source]
+response-headers.htm: [fail, I don't find a spec about combining same value response headers; also https://github.com/tmpvar/jsdom/issues/1833]
+simple-requests.htm: [fail, maybe https://github.com/tmpvar/jsdom/issues/1833, but fails locally too]
+
+---
+
+DIR: dom/collections
+
+HTMLCollection-as-proto-length-get-throws.html: [fail,  Brand checks not strict enough, https://github.com/jsdom/webidl2js/issues/62]
```
Zirro: Extra space before "Brand".

test/web-platform-tests/to-run.yaml
```diff
@@ -0,0 +1,658 @@
...
+
+DIR: html/semantics/scripting-1/the-script-element
+
+async_001.htm: [timeout, Unknown]
+async_002.htm: [timeout, Unknown]
+async_003.htm: [timeout, Unknown]
+async_004.htm: [timeout, Unknown]
+async_005.htm: [timeout, Unknown]
+async_006.htm: [timeout, Unknown]
+async_007.htm: [needs-node8]
```
Zirro: It might be helpful to include the reason for why a newer Node-version is required in the second field, in case more reasons appear in future versions.

Zirro:
Maybe we can avoid keeping a ~20MB file in the main repository? I imagine that it could grow the repo size/history quite a bit as it changes. Since we already make use of them, perhaps we could keep it in a submodule where we force push updates to not retain history?

domenic:
I don't see the advantage of that. jsdom is already large, especially counting its submodule. Adding more submodule pain for unclear gain doesn't seem like a good tradeoff to me.

Zirro:
I wasn't aware that there was any pain attached to using submodules, but if you're not feeling concerned about this I suppose that's fine.

Contributing.md
```diff
@@ -39,17 +39,19 @@ First you'll want to `npm install`. Then, configure your system to run the web p
...
+All tests for web platform features (as opposed to features of jsdom itself, such as the `JSDOM()` constructor) should be in [web-platform-tests](https://github.com/w3c/web-platform-tests) format. We have some infrastructure for running these directly against jsdom documents. So ideally, when contributing a bugfix or new feature, you can browser the web-platform-tests repository and find the test covering your work, and then just enable it in [the `to-run.yaml` file](https://github.com/tmpvar/jsdom/blob/master/test/web-platform-tests/to-run.yaml). These tests are HTML files which use a special library called [testharness.js](http://testthewebforward.org/docs/testharness-library.html) to report their results.
```
Zirro: Not introduced in this commit, but "you can browser the web-platform-tests repository" could do with one "r" less.

test/README.md
```diff
@@ -10,9 +10,7 @@ The jsdom project uses two different test frameworks:
 For working on new tests:
 
 - Tests for web platform features go in the [`web-platform-tests/to-upstream`](./web-platform-tests/to-upstream/) subdirectory. This should be the majority of new work.
-- Tests for the jsdom API itself go on the [`api`](./api/) subdirectory.
-
-If you add a new test file, edit [`index.js`](./index.js) to include it.
+- Tests for the jsdom API itself go on the [`api`](./api/) subdirectory. If you add a new test file, edit [`index.js`](./index.js) to include it.
```
Zirro: on -> in?

PR #1969, opened 2017-09-02
Fix wrong script tests

Sebmaster:
Our tests relied on the fact that scripts ran when the whole document
was loaded into memory. This is not guaranteed in a streaming parser.

Let's not test jsdom implementation specific behaviour if it goes
against expected web platform behaviour.

### DIFF
--- a/test/old-api/env.js
+++ b/test/old-api/env.js
@@ -215,6 +215,7 @@ describe("jsdom/env", () => {
   specify("with configurable resource loader", { async: true }, t => {
     env({
       html: "<!DOCTYPE html><html><head><script src='foo.js'></script></head><body></body></html>",
+      url: "http://example.org/",
       resourceLoader(resource, callback) {
         callback(null, "window.resourceLoaderWasOverridden = true;");
       },
--- a/test/old-api/inside-worker-smoke-tests.js
+++ b/test/old-api/inside-worker-smoke-tests.js
@@ -55,9 +55,9 @@ describe("jsdom/inside-worker-smoke-tests", () => {
   });
 
   specify("execute scripts referring to global built-ins (GH-1175)", { async: true }, t => {
-    const document = jsdom.jsdom(`<!DOCTYPE html><script>
+    const document = jsdom.jsdom(`<!DOCTYPE html><body><script>
       document.body.textContent = Error.name + window.Object.name + NaN + ("undefined" in window);
-    </script>`);
+    </script></body>`);
 
     assert.strictEqual(document.body.textContent, "ErrorObjectNaNtrue");
     t.done();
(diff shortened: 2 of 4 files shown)


test/old-api/env.js
```diff
@@ -215,6 +215,7 @@ describe("jsdom/env", () => {
   specify("with configurable resource loader", { async: true }, t => {
     env({
       html: "<!DOCTYPE html><html><head><script src='foo.js'></script></head><body></body></html>",
+      url: "http://example.org/",
```
domenic: What does this fix?

test/old-api/inside-worker-smoke-tests.js
```diff
@@ -55,9 +55,9 @@ describe("jsdom/inside-worker-smoke-tests", () => {
   });
 
   specify("execute scripts referring to global built-ins (GH-1175)", { async: true }, t => {
-    const document = jsdom.jsdom(`<!DOCTYPE html><script>
+    const document = jsdom.jsdom(`<!DOCTYPE html><body><script>
       document.body.textContent = Error.name + window.Object.name + NaN + ("undefined" in window);
-    </script>`);
+    </script></body>`);
```
domenic: End `</body>` is optional :). Fine to keep it though.

test/to-port-to-wpts/misc2.js
```diff
@@ -974,10 +975,10 @@ describe("jsdom/miscellaneous", () => {
     });
 
     specify("ensure_resources_can_be_skipped_via_options_features", { async: true }, t => {
-      const html = `<html><head><script src="./files/hello.js"></script>
-                 <script src="./files/nyan.js"></script></head>
+      const html = `<html><head></head>
                  <body><span id="test">hello from html</span><span id="cat">
-                 hello from cat</body></html>`;
+                 hello from cat</body><script src="./files/hello.js"></script>
```
domenic: scripts should be before `</body>`, ideally.

test/old-api/env.js
Sebmaster: url default is about:blank, so foo.js can't be resolved, so the custom resource loader isn't executed.

test/to-port-to-wpts/misc2.js
Sebmaster: Fixed

test/old-api/env.js
domenic: How was this working before? :-/

Sebmaster: 🌠 resource loader!

domenic: But, what changed?

Is this in anticipation of other changes you have locally or something, that will break resolving `foo.js` against `about:blank`?

Sebmaster: These were all changes which I initially developed for streaming with parse5. Just trying to take what I can get from that PR and make it easier in the future, then close it out and maybe restart it down the road again from mostly scratch.

PR #1971, opened 2017-09-03
Make element builders namespace aware & add/fix caption methods

Sebmaster:
I haven't found exact details on when to build `HTMLUnknownElement` vs `Element` in the spec. Chrome constructs HTMLUnknownElements if the namespace is xhtml, and Elements otherwise. Should we add that as well?


domenic:
I think what you want for HTMLUnknownElement vs. Element is https://html.spec.whatwg.org/#elements-in-the-dom:element-interface . (Nothing in the HTML namespace should fall back to Element.)

domenic:
Lots of single-quotes seem to have snuck in (I suggest installing VSCode's ESLint extension!) but the code LGTM once CI passes. Hooray!

Sebmaster:
So we want to emulate chrome's behaviour then? That is: changing non-html createElementNSs to return Element instead?

domenic:
Well, so far we only implement the HTML namespace, so yeah, that seems right. If we ever implemented SVG or MathML (haha, MathML), the "element interface" algorithm would gain extra branches, but until then, yeah, Element is the fallback (per https://dom.spec.whatwg.org/#concept-element-interface ).

PR #1972, opened 2017-09-03
Implement the details, picture and marquee elements

Zirro:
This pull request adds implementations for the `<details>` and `<picture>` elements, as well as a stub for `<marquee>`.

The `<details>` "toggle" event is implemented, though it is missing one step involving a [task queue](https://html.spec.whatwg.org/#details-notification-task-steps) which I wasn't entirely sure how to approach. It uses the Event constructor, as it seems to be needed to set isTrusted.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLDetailsElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLDetailsElement-impl.js
@@ -1,0 +1,28 @@
+"use strict";
+
+const Event = require("../generated/Event.js");
+
+const HTMLElementImpl = require("./HTMLElement-impl").implementation;
+
+class HTMLDetailsElementImpl extends HTMLElementImpl {
+  _dispatchToggleEvent() {
+    const event = Event.createImpl(
+      ["toggle",
+      { bubbles: false, cancelable: false }],
+      { isTrusted: true }
+    );
+    this._dispatch(event);
+  }
+
+  _attrModified(name, value, oldValue) {
+    super._attrModified(name, value, oldValue);
+
+    if (name === "open") {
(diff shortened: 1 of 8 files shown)


domenic:
"queue task" in jsdom is just going to be `setTimeout(, 0)`.

So to implement this we'll want to store the return value of that setTimeout, and null it out when the function starts. Then we can add an if check that prevents dispatching the event if the stored timeout ID is non-null.

For marquee, I think you need to set up the event handlers by using code similar to https://github.com/tmpvar/jsdom/blob/0b4b8e0f548d1c6568c5902b3b6773b11f839e29/lib/jsdom/living/nodes/GlobalEventHandlers-impl.js#L87 . I think maybe because HTMLElement already mixes in globalEventHandlers, only the createEventAccessor will be what you need, not the rest of the stuff. I guess the event handlers aren't tested anywhere :-/

Zirro:
@domenic Thanks for the advise! I have implemented it now, and all tests except for one is passing in [toggleEvent.html](https://github.com/w3c/web-platform-tests/blob/master/html/semantics/interactive-elements/the-details-element/toggleEvent.html). I've not yet determined what is needed to pass the failing test, but interestingly this particular test is [failing in Chrome and Safari](http://wpt.fyi/html/semantics/interactive-elements/the-details-element/toggleEvent.html) as well.

It is tempting to dismiss the test itself as being incorrect, but since Firefox does pass it I suppose this will require some further investigation. I'll be busy for the next few days, so if anyone else would like to take a look it would be appreciated :-)

As for the details of `<marquee>`, I'm content to leave it as a stub for now. I'd imagine that there are other areas of jsdom where my time would be better spent ;-)

PR #1973, opened 2017-09-04
Only append inline handlers for an element once

### LINKED ISSUE #1948: Inline event handlers still not quite right
Two symptoms:

- [ ] inline-event-handler-ordering.html doesn't pass (even after a WPT roll to fix a bug that used to be present in it)
- [x] EventTarget-impl.js still has inline event handler stuff, but that really shouldn't be necessary anymore (fixed in https://github.com/tmpvar/jsdom/pull/1950)

To investigate after I disentangle myself from my current mess.

Sebmaster:
We set the raw event body to null if compiling the source fails,
which previously led to us re-appending when we set another inline
attribute.

Fixes #1948.

### DIFF
--- a/lib/jsdom/living/helpers/create-event-accessor.js
+++ b/lib/jsdom/living/helpers/create-event-accessor.js
@@ -58,11 +58,13 @@ exports.setupForSimpleEventAccessors = (prototype, events) => {
   };
 
   prototype._setEventHandlerFor = function (event, handler) {
-    if (!this._eventHandlers) {
+    if (!this._registeredHandlers) {
+      this._registeredHandlers = new Set();
       this._eventHandlers = Object.create(null);
     }
 
-    if (!this._eventHandlers[event] && handler !== null) {
+    if (!this._registeredHandlers.has(event) && handler !== null) {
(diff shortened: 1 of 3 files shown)


Sebmaster:
Re event-handler-spec-example:

`eventHandlerArgCoercion` is boxing the primitives to null, which then causes the handler to not get appended and so the ordering is wrong. Haven't checked the spec yet if we should be doing that this way.

Sebmaster:
Okay, fixed. `eventHandlerArgCoercion` wasn't at fault, I just got confused with the test failures.

Basically, when accessing the IDL attribute (`el.onclick`), we'd compile. If that failed, we'd set the internal eventHandlers tracker to null, thereby allowing scripts to re-add the inline handler in the handler list.

domenic:
Do you see a way to do this without an additional Set, e.g. using `in` checks or `undefined`?

If not, LGTM and feel free to merge :)

Sebmaster:
No, sadly has to be this way since we only want to mark in cases where handler !== null, but _eventHandlers[event] is always set.

PR #1975, opened 2017-09-06
Use domexception module

### LINKED ISSUE #1939: Use jsdom/domexception
Newly published: https://github.com/jsdom/domexception

This will require a decent bit of work as currently we use an incorrect DOMException constructor all over the place.

Sebmaster:
Split this up into three commits for easier reviewing. First two commits were largely automated search&replaces, last commit are manual rewrites.

Fixes #1939.


TimothyGu:
Hmm, I was thinking of using a codemod for this, but this works nevertheless. I'll try to take a look at this.

domenic:
As I go through this I notice lots of ways the code could be cleaner. E.g. using template literals or less temporary variables or being consistent about whether we end error messages with periods. But they are all pre-existing problems so I think they should stay out of scope and we should just review to check that there are no regressions.

Sebmaster:
@TimothyGu This is amazing!

Squashed all commits together, I think we should be good.

PR #1976, opened 2017-09-06
Enable more wpt tests from table and anchor categories

Sebmaster:
Sadly jsdom/webidl2js#67 is a thing so our anchor stringifier doesn't actually work and we had no other tests for it...

### DIFF
--- a/test/web-platform-tests/index.js
+++ b/test/web-platform-tests/index.js
@@ -245,6 +245,7 @@ describe("Web Platform Tests", () => {
     "html/semantics/scripting-1/the-template-element/template-element/template-descendant-body.html",
     "html/semantics/scripting-1/the-template-element/template-element/template-descendant-frameset.html",
     "html/semantics/scripting-1/the-template-element/template-element/template-descendant-head.html",
+    "html/semantics/tabular-data/the-caption-element/caption_001.html",
     "html/semantics/tabular-data/the-table-element/caption-methods.html",
     "html/semantics/tabular-data/the-table-element/createTBody.html",
     "html/semantics/tabular-data/the-table-element/delete-caption.html",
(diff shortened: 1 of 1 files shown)


TimothyGu:
Using 

- A version of webidl2js with https://github.com/jsdom/webidl2js/pull/68 applied
- A patch for now-removed `idlUtils.mixin`: http://sprunge.us/LaFJ

the anchor stringifier test can be enabled.

TimothyGu:
Another thing about this PR is that `html/semantics/tabular-data/attributes-common-to-td-and-th-elements/cellIndex.html` is currently failing. I've already fixed the bug in https://github.com/tmpvar/jsdom/pull/1953 (in particular 5aa307258741f2d103971e8aa94d7285f736444c), and rebased that commit to this one.

I've put all my work into https://github.com/TimothyGu/jsdom/tree/test/enable-more-wpt, which includes

- The webidl2js updating patch I mentioned earlier
- Enabling the `<a>` stringifier test
- Fixing `html/semantics/tabular-data/attributes-common-to-td-and-th-elements/cellIndex.html`
- Removing nonexistent `html/semantics/tabular-data/the-col-element/caption_001.html` from manifest

After all those changes (and https://github.com/jsdom/webidl2js/pull/68), this PR should be ready to merge.

test/web-platform-tests/index.js
```diff
@@ -176,6 +176,7 @@ describe("Web Platform Tests", () => {
     "html/editing/focus/sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-default-value.html",
     // "html/infrastructure/urls/terminology-0/document-base-url.html", // we don't support srcdoc <base> correctly
     "html/obsolete/requirements-for-implementations/other-elements-attributes-and-apis/nothing.html",
+    "html/semantics/disabled-elements/disabledElement.html",
```
Zirro: This test already exists further down in the list.

Sebmaster:
> Removing nonexistent html/semantics/tabular-data/the-col-element/caption_001.html from manifest

Removed.

> Fixing html/semantics/tabular-data/attributes-common-to-td-and-th-elements/cellIndex.html

Removed this for now, let's leave this to the other PR.

test/web-platform-tests/index.js
Sebmaster: Ah, ordered lexically, not by folders. We should sort this whole list in general, the order really sucks currently.

PR #1978, opened 2017-09-11
Use correct interfaces for many elements

Zirro:
This pull requests assigns the HTMLElement interface to the following elements:

`acronym`
`basefont`
`big`
`center`
`nobr`
`noembed`
`noframes`
`noscript`
`plaintext`
`rb`
`rtc`
`strike`
`sub`
`summary`
`sup`
`tt`

...as well as the HTMLPreElement interface to `listing` and `xmp` as [mentioned](https://html.spec.whatwg.org/multipage/dom.html#elements-in-the-dom%3Aelement-interface) in the specification.

I have included a test which looks at the interfaces being used, but further work will be needed before it fully passes as noted in the associated comment. I've also simplified the code used to initialise the elements, moving the list of HTMLElements to register-elements.js.

### DIFF
--- a/lib/jsdom/living/register-elements.js
+++ b/lib/jsdom/living/register-elements.js
@@ -4,6 +4,8 @@
 const DocumentImpl = require("./nodes/Document-impl.js");
 
 const mappings = {
+  // https://html.spec.whatwg.org/multipage/dom.html#elements-in-the-dom%3Aelement-interface
+  // https://html.spec.whatwg.org/multipage/indices.html#elements-3
   "http://www.w3.org/1999/xhtml": {
     HTMLElement: {
       file: require("./generated/HTMLElement.js"),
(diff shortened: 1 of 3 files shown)


lib/jsdom/living/register-elements.js
```diff
@@ -7,7 +7,58 @@ const mappings = {
   "http://www.w3.org/1999/xhtml": {
     HTMLElement: {
       file: require("./generated/HTMLElement.js"),
-      tags: []
+      tags: [
```
Sebmaster: Is there a full list of these centralized somewhere, or are we picking these through the whole HTML spec?

Sebmaster: Ah, there's an [elements](https://html.spec.whatwg.org/#elements-3) section which has them all.

PR #1979, opened 2017-09-16
Fix FocusEventImpl inheritance

### LINKED ISSUE #1959: Regression v11.1.0 -> v11.2.0
### Basic info:

- **Node.js version:** v6.11.2
- **jsdom version:** v11.2.0

### Minimal reproduction case

```sh
git clone git@github.com:callemall/material-ui.git
cd material-ui
git checkout a1a17995202db7f669451ad41690fc5e126a770c
yarn
yarn test:unit
```

or even better: https://github.com/callemall/material-ui/commit/a1a17995202db7f669451ad41690fc5e126a770c

> Error: Uncaught [TypeError: Illegal invocation]
    at reportException (/tmp/material-ui/node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:63:24)
    at invokeEventListeners (/tmp/material-ui/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:211:9)
    at HTMLLIElementImpl._dispatch (/tmp/material-ui/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:111:7)
    at Object.exports.fireFocusEventWithTargetAdjustment (/tmp/material-ui/node_modules/jsdom/lib/jsdom/living/helpers/focusing.js:73:10)
    at HTMLLIElementImpl.focus (/tmp/material-ui/node_modules/jsdom/lib/jsdom/living/nodes/HTMLElement-impl.js:65:14)
    at HTMLLIElement.focus (/tmp/material-ui/node_modules/jsdom/lib/jsdom/living/generated/HTMLElement.js:49:21)
    at Object.Menu._this.handleEnter [as onEnter] (/tmp/material-ui/src/Menu/Menu.js:113:27)
    at Object.Popover._this.handleEnter [as onEnter] (/tmp/material-ui/src/internal/Popover.js:228:18)
    at Object.Grow._this.handleEnter [as onEnter] (/tmp/material-ui/src/transitions/Grow.js:81:18)
    at Transition.performEnter (/tmp/material-ui/src/internal/Transition.js:235:15)
    at Transition.componentDidMount (/tmp/material-ui/src/internal/Transition.js:167:12)

It's working on [IE11, Firefox 45, Safari 10.1, Edge 14, Chrome 49](https://circleci.com/gh/callemall/material-ui/6201?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), jsdom v11.1.0 but [fails with jsdom v11.2.0](https://circleci.com/gh/callemall/material-ui/6203#tests/containers/0).

TimothyGu:
This is rebased on https://github.com/tmpvar/jsdom/pull/1953 to avoid conflict.

Fixes #1959.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -65,6 +65,7 @@ function Window(options) {
 
   // vm initialization is defered until script processing is activated
   this._globalProxy = this;
+  Object.defineProperty(idlUtils.implForWrapper(this), idlUtils.wrapperSymbol, { get: () => this._globalProxy });
 
   this.__timers = Object.create(null);
 
--- a/lib/jsdom/living/events/CompositionEvent-impl.js
+++ b/lib/jsdom/living/events/CompositionEvent-impl.js
@@ -1,0 +1,18 @@
+"use strict";
+
+const UIEventImpl = require("./UIEvent-impl").implementation;
+
+class CompositionEventImpl extends UIEventImpl {
+  initCompositionEvent(type, bubbles, cancelable, view, data) {
+    if (this._dispatchFlag) {
+      return;
+    }
+
+    this.initUIEvent(type, bubbles, cancelable, view, 0);
+    this.data = data;
+  }
+}
+
+module.exports = {
+  implementation: CompositionEventImpl
+};
--- a/lib/jsdom/living/events/CompositionEvent.webidl
+++ b/lib/jsdom/living/events/CompositionEvent.webidl
@@ -1,0 +1,19 @@
+// https://w3c.github.io/uievents/#interface-compositionevent
+[Constructor(DOMString type, optional CompositionEventInit eventInitDict), Exposed=Window]
+interface CompositionEvent : UIEvent {
+  readonly attribute DOMString data;
+};
+
+dictionary CompositionEventInit : UIEventInit {
+  DOMString data = "";
+};
+
+// https://github.com/w3c/uievents/issues/134
+partial interface CompositionEvent {
+  // Originally introduced (and deprecated) in this specification
+  void initCompositionEvent(DOMString typeArg,
+                            optional boolean bubblesArg = false,
+                            optional boolean cancelableArg = false,
+                            optional Window? viewArg = null,
+                            optional DOMString dataArg = "");
+};
--- a/lib/jsdom/living/events/FocusEvent-impl.js
+++ b/lib/jsdom/living/events/FocusEvent-impl.js
@@ -1,4 +1,4 @@
 "use strict";
-const EventImpl = require("./Event-impl").implementation;
+const UIEventImpl = require("./UIEvent-impl").implementation;
 
-exports.implementation = class FocusEventImpl extends EventImpl { };
+exports.implementation = class FocusEventImpl extends UIEventImpl {};
--- a/lib/jsdom/living/events/KeyboardEvent-impl.js
+++ b/lib/jsdom/living/events/KeyboardEvent-impl.js
@@ -3,16 +3,18 @@
 const UIEventImpl = require("./UIEvent-impl").implementation;
 
 class KeyboardEventImpl extends UIEventImpl {
-  initKeyboardEvent(type, bubbles, cancelable, view, key, location, modifiersList, repeat, locale) {
+  initKeyboardEvent(type, bubbles, cancelable, view, key, location, ctrlKey, altKey, shiftKey, metaKey) {
     if (this._dispatchFlag) {
       return;
     }
 
-    this.initUIEvent(type, bubbles, cancelable, view, key);
+    this.initUIEvent(type, bubbles, cancelable, view, 0);
+    this.key = key;
     this.location = location;
-    this.modifiersList = modifiersList;
-    this.repeat = repeat;
-    this.locale = locale;
+    this.ctrlKey = ctrlKey;
+    this.altKey = altKey;
+    this.shiftKey = shiftKey;
+    this.metaKey = metaKey;
   }
 }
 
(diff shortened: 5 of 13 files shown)


TimothyGu:
I haven't quite thought of a way to test this. I mean we can just test the specific property manually, but WPT has idlharness (or some other mechanism) that if I understand correctly should test this automatedly.

Sebmaster:
Hrm, disappointing that the Event tests do not test all props of an event. I think a regression test would work? I'm fine with it though, copy-paste issue that no one's actually going to make if they were changing code.

Zirro:
This feels like a problem which we could discover automatically by validating the implementation files against their IDL. Either as part of the webidl2js-conversion process or perhaps through an eslint plugin which looks at the accompanying IDL file to make sure that both files agree on the pieces they have in common.

In case I'd have time to work on this later on - what do you think? Should it be part of webidl2js or would it be better as an eslint plugin?

TimothyGu:
@Zirro 

> In case I'd have time to work on this later on - what do you think? Should it be part of webidl2js or would it be better as an eslint plugin?

That's wonderful! In fact I discovered the entries in 3574ac85f55f7824f9945d195d380a1c860186b5 through some hacking with the included IDL files. However, my approach was fairly crude, adding console.log's to webidl2js internals to print the necessary informaion in JSON, and then manually consuming that information through the Node.js REPL.

I would say either option would work. ESLint would probably create a better-looking output, while integrating it with webidl2js would make it possible to reuse the impl paths and other algorithms already in webidl2js.

If you decide to integrate it with idl conversion, there are two paths I can think of, each with its own caveats:

- AST-based: parse the JS impl file and try to statically analyze the impl class's name and hierarchy. While Acorn and its included AST walker make this easy, this feels somewhat like reinventing the wheel with ESLint however.
- Executing the impl code: check `require(implFile).implementation.name` and `Object.getPrototypeOf(require(implFile).implementation.prototype).constructor.name`. Care must be taken to separate validation from generation due to circular dependency issues. This might also be awkward for WIP interfaces that haven't been fully implemented.

Also, I'd prefer it if the validation itself could be done in jsdom rather than webidl2js. After all, many conventions (such as the 'Impl' suffix) are purely jsdom-internal conventions.

Another option is to integrate ESLint and idl conversion: run ESLint through its API after idl conversion with a set of custom rules generated dynamically during the conversion.

Zirro:
@TimothyGu Ah, sorry, I ended up deleting my original comment after coming to the realisation that web platform tests ought to be the best way to catch those issues, after all. The problem with them is usually that we fail unrelated parts as well, though.

That said, I'm still happy to work on this if we come up with a case where a validation/linting solution would be more useful than web platform tests.

For anyone else coming across this in the future, my original comment was:

> This feels like a problem which we could discover automatically by validating the implementation files against their IDL. Either as part of the webidl2js-conversion process or perhaps through an eslint plugin which looks at the accompanying IDL file to make sure that both files agree on the pieces they have in common.
In case I'd have time to work on this later on - what do you think? Should it be part of webidl2js or would it be better as an eslint plugin?

TimothyGu:
> if we come up with a case where a validation/linting solution would be more useful than web platform tests

Well, *nothing* is better than comprehensive tests. However, because of the current restrictions with the testing system, specifically the lack of support for partially passing test, I'm certain that a static analysis solution would still be very beneficial, if it is not very complex and only to check the very basics (e.g. settable attribute has a setter on impl, class inheritance is correct, etc.). Not that it is necessary by any means: this kind of error is rare and I didn't find anything like this issue in the audit I did.

Zirro:
Okay, I might give an ESLint-based solution a try some day since we are already using it and it provides several useful methods.

> Another option is to integrate ESLint and idl conversion: run ESLint through its API after idl conversion with a set of custom rules generated dynamically during the conversion.

Could you give an example of how this would work? Thanks for all the suggestions!

This looks like the right web platform tests for this particular issue, btw: https://github.com/w3c/web-platform-tests/blob/56ea52b8668ceee63ec196c59dd11ad7797e4f0d/dom/events/Event-subclasses-constructors.html - Perhaps worth adding to the manifest but behind a comment for now?

TimothyGu:
> This looks like the right web platform tests for this particular issue, btw: https://github.com/w3c/web-platform-tests/blob/master/dom/events/Event-subclasses-constructors.html - Perhaps worth adding to the manifest but behind a comment for now?

Done. Fixed a drive-by issue when adding it as a comment: b59484d7075dd76667ef8ab01e797b1b393ab904

TimothyGu:
Update: also fixed the remaining issues so that Event-subclasses-constructor.html now actually passes! \0/

Sebmaster:
Seems like there is a lint failure and test failures as well 😞

TimothyGu:
@Sebmaster should hopefully be fixed.

lib/jsdom/living/nodes/HTMLTableCellElement-impl.js
```diff
@@ -2,54 +2,74 @@
...
+const { closest } = require("../helpers/traversal");
 
-  get cellIndex() {
-    const tr = closest(this, "tr");
-    if (tr === null) {
-      return -1;
-    }
-
-    return tr.cells.indexOf(this);
+function reflectedAttributeClampedToRange(attrValue, min, max, defaultValue = 0) {
```
Zirro: This looks like it could be useful in other locations. Move to the helpers directory, perhaps? Though eventually we should add `[Reflect]` varieties to webidl2js, I suppose.

Zirro:
@TimothyGu Awesome work! Great to see the other parts of the test implemented as well.

TimothyGu:
Rebased!

domenic:
@TimothyGu are all of these commits tested by the test enabled in the last commit? Or could I remove any of them and still pass that test?

TimothyGu:
@domenic they are all tested by the test. I added Composition- and WheelEvent just so that the test can pass.

domenic:
Yeah, I'm more specifically wondering if e.g. the Window impl wrapper thing, or the original FocusEventImpl thing, is tested by that test. But it sounds like yes, so that's good.

I think you need to rename to .webidl for the new classes for CI to pass.

TimothyGu:
Oops, done.

PR #1982, opened 2017-09-20
downstream dependency change breaks backwards compatibility

eandersonpgi:
request 2.82.0 to 2.82.1 made a breaking change in dependencies and as a result jsdom will no longer work with pre v4


TimothyGu:
That's very unfortunate :/ But this PR LGTM.

Also see https://github.com/request/request/issues/2772 and https://github.com/request/request/issues/2774.

Zirro:
But that note is still correct... It says "work with Node.js v4", not pre v4 which are the releases affected by this.

domenic:
Yeah this doesn't seem accurate. Let me close this, per @Zirro's comment. If we've misunderstood, happy to reopen.

eandersonpgi:
my mistake - I slightly misread that line while trying to track down the problems yesterday.

PR #1985, opened 2017-09-22
Updating tough-cookie due to security fix.

karlnorling:
This addresses:
```
The tough-cookie module is vulnerable to regular expression denial of service. Input of around 50k characters is required for a slow down of around 2 seconds.

Unless node was compiled using the -DHTTPMAXHEADER_SIZE= option the default header max length is 80kb so the impact of the ReDoS is limited to around 7.3 seconds of blocking.

At the time of writing all version <=2.3.2 are vulnerable
```

- See: https://www.versioneye.com/Node.JS/tough-cookie/2.3.2
  - Also tough-cookie:
    - https://github.com/salesforce/tough-cookie/pull/97
    - https://github.com/salesforce/tough-cookie/pull/92


kartikm:
When will be this included in the release?

Sebmaster:
> When will be this included in the release?

There is no action necessary from our side. Just run `npm update`.

karlnorling:
@Sebmaster can you clarify? This PR is not needed? I'm trying to run `npm update` on project that has `jsDom: 11.2.0` as a dependency and I'm getting `tough-cookie: 2.3.2`

Sebmaster:
> I'm running npm update on project that has jsDom: 11.2.0 as a dependency and I'm getting tough-cookie: 2.3.2

Ugh, yeah. I can repro that.

If you remove the jsdom version increase I can merge this.

karlnorling:
@Sebmaster pushed!

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "jsdom",
-  "version": "11.2.0",
+  "version": "11.2.1",
```
domenic: Please don't change the version of jsdom

karlnorling: @domenic done

domenic:
And now the package-lock.json is not updated, so nobody will actually see this update... I'll just do the update myself.

domenic:
Oh, I see, tough-cookie was already updated in package-lock.json. So this was not a necessary pull request after all. Still, might as well.

PR #1989, opened 2017-09-26
Speed up sync XHR result parsing

Zirro:
I figured that rather than using JSON.parse()'s reviver function, it might be more efficient to let JSON.parse() do its thing and modify the resulting object afterwards.

This seems to cut the time spent parsing the result from the other process from an average of 650ms to 60ms. It should help with https://github.com/tmpvar/jsdom/issues/1833, and might allow us to enable some timeout-tests :-)

### DIFF
--- a/lib/jsdom/living/xhr-sync-worker.js
+++ b/lib/jsdom/living/xhr-sync-worker.js
@@ -16,15 +16,15 @@ process.stdin.on("data", chunk => {
 
 process.stdin.on("end", () => {
   const buffer = Buffer.concat(chunks);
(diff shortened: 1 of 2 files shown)


Sebmaster:
9 => 7 minute travis runs is 👌

Zirro:
Turns out that a JSON.parse()-reviver was used in `xhr-sync-worker.js` as well. Replacing it seems to have improved performance further.

Sebmaster:
Wow. Another ~2 mins shaved off. This change is amazing. Can you get us another 2 minutes? 😀

Issue #1422, opened 2016-03-21
Add mocks for element scrolling methods


just-boris:
Element prototype has several methods to scroll it, but those methods are commented in [the sources](https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/nodes/Element.idl#L56-L63).

Why did you do this? For example [window.scrollTo](https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/browser/Window.js#L382) is a function that does nothing, just prints warning in console. I think that it would be okay for `Element.prototype.scroll()` method as well.

domenic:
PR welcome

M3kH:
How far is this issue to be close?

thenikso:
I also needed this, I ended up doing something like this:

```javascript
const el = dom.window.document.querySelector(
    '.myScrollingElement',
  );
el.scrollTop = 12345;
el.dispatchEvent(new dom.window.Event('scroll'));
```

jordantomax:
This affecting me to. I've tried many variations of to override it, such as:

```
Object.defineProperty(window.Element.prototype, 'scrollIntoView', {
  writable: true,
  value: jest.fn()
})
```

But I either get an error that I can't override, or it does nothing. Am I doing something wrong?

jessethomson:
I know this is old, but are you all still open to a PR to change this? 

If scrollX, scrollY, etc are all hardcoded to 0, it seems like it makes sense that `scroll`, `scrollBy`, and `scrollTo` functions on the window object are just stubs. Let me know!

https://github.com/jsdom/jsdom/pull/2626

domenic:
We need to tackle this problem comprehensively. Let's roll discussions into https://github.com/jsdom/jsdom/issues/2751.

PR #1994, opened 2017-09-30
Add {request,cancel}AnimationFrame

### LINKED ISSUE #1963: window.requestAnimationFrame is not defined
### Basic info:

- **Node.js version:** v8.0.0
- **jsdom version:** 11.2.0

Using jsdom with [jsdom-global](https://github.com/rstacruz/jsdom-global) to test a program that uses `requestAnimationFrame`. On first test I got
```
requestAnimationFrame is not defined
```
Then I added the `window` namespace and still got that it was not defined. Is there no implementation for it on jsdom and is there a workaround?

SimenB:
~This seems deceptively straightforward...~ It was :P

I'm unsure how to test this. The readme in `test/api/README.md` says that the web-platform tests should cover this. Am I supposed to pull in some extra tests?

My current "test" is running this script:

```js
const { JSDOM } = require('./');
const { window } = new JSDOM();

window.requestAnimationFrame(console.log);
const id = window.requestAnimationFrame(console.log);
window.cancelAnimationFrame(id);
```

It outputs e.g. `1752.195617`.

We can add https://npmjs.com/raf if a battel-tested implementation is wanted.

Fixes #1963

EDIT: Ah, figured it out, I think. Activated some more tests. One of them fails, though (commented it out), anything I can do about that?

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -11,6 +11,7 @@ const cssom = require("cssom");
 const postMessage = require("../living/post-message");
 const DOMException = require("domexception");
 const { btoa, atob } = require("abab");
+const hrtime = require("browser-process-hrtime");
 const idlUtils = require("../living/generated/utils");
 const createXMLHttpRequest = require("../living/xmlhttprequest");
 const createFileReader = require("../living/generated/FileReader").createInterface;
@@ -43,7 +44,7 @@ dom.Window = Window;
 function Window(options) {
   EventTarget.setup(this);
 
-  const windowInitialized = process.hrtime();
+  const windowInitialized = hrtime();
 
   this._initGlobalEvents();
 
@@ -208,7 +209,7 @@ function Window(options) {
 
   if (this._pretendToBeVisual) {
     this.requestAnimationFrame = fn => {
-      const hr = process.hrtime(windowInitialized);
+      const hr = hrtime(windowInitialized);
       const hrInNano = hr[0] * 1e9 + hr[1];
       const fps = 1000 / 60;
       const hrInMicro = hrInNano / 1e6;
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -11,7 +11,6 @@ const cssom = require("cssom");
 const postMessage = require("../living/post-message");
 const DOMException = require("domexception");
 const { btoa, atob } = require("abab");
-const hrtime = require("browser-process-hrtime");
 const idlUtils = require("../living/generated/utils");
 const createXMLHttpRequest = require("../living/xmlhttprequest");
 const createFileReader = require("../living/generated/FileReader").createInterface;
@@ -44,7 +43,7 @@ dom.Window = Window;
 function Window(options) {
   EventTarget.setup(this);
 
-  const windowInitialized = hrtime();
+  const windowInitialized = process.hrtime();
 
   this._initGlobalEvents();
 
@@ -209,10 +208,9 @@ function Window(options) {
 
   if (this._pretendToBeVisual) {
     this.requestAnimationFrame = fn => {
-      const hr = hrtime(windowInitialized);
-      const hrInNano = hr[0] * 1e9 + hr[1];
+      const hr = process.hrtime(windowInitialized);
+      const hrInMicro = hr[0] * 1e3 + hr[1] / 1e6;
       const fps = 1000 / 60;
-      const hrInMicro = hrInNano / 1e6;
 
       return startTimer(
         window,
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -43,6 +43,8 @@ dom.Window = Window;
 function Window(options) {
   EventTarget.setup(this);
 
+  const windowInitialized = process.hrtime();
+
   this._initGlobalEvents();
 
   const window = this;
@@ -62,11 +64,12 @@ function Window(options) {
 
   ///// PRIVATE DATA PROPERTIES
 
-  // vm initialization is defered until script processing is activated
+  // vm initialization is deferred until script processing is activated
   this._globalProxy = this;
   Object.defineProperty(idlUtils.implForWrapper(this), idlUtils.wrapperSymbol, { get: () => this._globalProxy });
 
-  this.__timers = Object.create(null);
+  const timers = Object.create(null);
+  const animationFrameCallbacks = Object.create(null);
 
   // List options explicitly to be clear which are passed through
   this._document = Document.create([], {
@@ -182,24 +185,49 @@ function Window(options) {
   ///// METHODS
 
   let latestTimerId = 0;
+  let latestAnimationFrameCallbackId = 0;
 
   this.setTimeout = function (fn, ms) {
     const args = [];
     for (let i = 2; i < arguments.length; ++i) {
       args[i - 2] = arguments[i];
     }
-    return startTimer(window, setTimeout, clearTimeout, ++latestTimerId, fn, ms, args);
+    return startTimer(window, setTimeout, clearTimeout, ++latestTimerId, fn, ms, timers, args);
   };
   this.setInterval = function (fn, ms) {
     const args = [];
     for (let i = 2; i < arguments.length; ++i) {
       args[i - 2] = arguments[i];
     }
-    return startTimer(window, setInterval, clearInterval, ++latestTimerId, fn, ms, args);
+    return startTimer(window, setInterval, clearInterval, ++latestTimerId, fn, ms, timers, args);
   };
-  this.clearInterval = stopTimer.bind(this, window);
-  this.clearTimeout = stopTimer.bind(this, window);
-  this.__stopAllTimers = stopAllTimers.bind(this, window);
+  this.requestAnimationFrame = fn => {
+    const hr = process.hrtime(windowInitialized);
+    const hrInNano = hr[0] * 1e9 + hr[1];
+    const fps = 1000 / 60;
+    const hrInMicro = hrInNano / 1e6;
+
+    return startTimer(
+      window,
+      setTimeout,
+      clearTimeout,
+      ++latestAnimationFrameCallbackId,
+      fn,
+      fps,
+      animationFrameCallbacks,
+      [hrInMicro]
+    );
+  };
+  this.clearInterval = stopTimer.bind(this, timers);
+  this.clearTimeout = stopTimer.bind(this, timers);
+  this.cancelAnimationFrame = stopTimer.bind(this, animationFrameCallbacks);
+  this.__stopAllTimers = function () {
+    stopAllTimers(timers);
+    stopAllTimers(animationFrameCallbacks);
+
+    latestTimerId = 0;
+    latestAnimationFrameCallbackId = 0;
+  };
 
   function Option(text, value, defaultSelected, selected) {
     if (text === undefined) {
@@ -377,7 +405,7 @@ function Window(options) {
       delete this._document;
     }
 
-    stopAllTimers(currentWindow);
+    this.__stopAllTimers();
   };
 
   this.getComputedStyle = function (node) {
@@ -527,7 +555,7 @@ Object.defineProperty(Window.prototype, Symbol.toStringTag, {
   configurable: true
 });
 
-function startTimer(window, startFn, stopFn, timerId, callback, ms, args) {
+function startTimer(window, startFn, stopFn, timerId, callback, ms, timerStorage, args) {
   if (!window || !window._document) {
     return undefined;
   }
@@ -546,24 +574,23 @@ function startTimer(window, startFn, stopFn, timerId, callback, ms, args) {
   };
 
   const res = startFn(callback, ms);
-  window.__timers[timerId] = [res, stopFn];
+  timerStorage[timerId] = [res, stopFn];
   return timerId;
 }
 
-function stopTimer(window, id) {
-  const timer = window.__timers[id];
+function stopTimer(timerStorage, id) {
+  const timer = timerStorage[id];
   if (timer) {
     // Need to .call() with undefined to ensure the thisArg is not timer itself
     timer[1].call(undefined, timer[0]);
-    delete window.__timers[id];
+    delete timerStorage[id];
   }
 }
 
-function stopAllTimers(window) {
-  Object.keys(window.__timers).forEach(key => {
-    const timer = window.__timers[key];
+function stopAllTimers(timers) {
+  Object.keys(timers).forEach(key => {
+    const timer = timers[key];
     // Need to .call() with undefined to ensure the thisArg is not timer itself
     timer[1].call(undefined, timer[0]);
   });
-  window.__timers = Object.create(null);
 }
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -226,11 +226,6 @@ function Window(options) {
       );
     };
     this.cancelAnimationFrame = stopTimer.bind(this, animationFrameCallbacks);
-  } else {
-    // TODO: Should we throw to tell users about `pretendToBeVisual`?
-    // Also, should we somehow allow this to be polyfilled byt code using feature detection?
-    this.requestAnimationFrame = () => {};
-    this.cancelAnimationFrame = () => {};
   }
 
   this.__stopAllTimers = function () {
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -69,8 +69,8 @@ function Window(options) {
   this._globalProxy = this;
   Object.defineProperty(idlUtils.implForWrapper(this), idlUtils.wrapperSymbol, { get: () => this._globalProxy });
 
-  const timers = Object.create(null);
-  const animationFrameCallbacks = Object.create(null);
+  let timers = Object.create(null);
+  let animationFrameCallbacks = Object.create(null);
 
   // List options explicitly to be clear which are passed through
   this._document = Document.create([], {
@@ -239,6 +239,9 @@ function Window(options) {
 
     latestTimerId = 0;
     latestAnimationFrameCallbackId = 0;
+
+    timers = Object.create(null);
+    animationFrameCallbacks = Object.create(null);
   };
 
   function Option(text, value, defaultSelected, selected) {
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -11,7 +11,6 @@ const cssom = require("cssom");
 const postMessage = require("../living/post-message");
 const DOMException = require("domexception");
 const { btoa, atob } = require("abab");
-const hrtime = require("browser-process-hrtime");
 const idlUtils = require("../living/generated/utils");
 const createXMLHttpRequest = require("../living/xmlhttprequest");
 const createFileReader = require("../living/generated/FileReader").createInterface;
@@ -22,6 +21,10 @@ const reportException = require("../living/helpers/runtime-script-errors");
 const { matchesDontThrow } = require("../living/helpers/selectors");
 const SessionHistory = require("../living/window/SessionHistory");
 const { contextifyWindow } = require("./documentfeatures.js");
+
+// Browserify's process implementation doesn't have hrtime, and this package is small so not much of a burden for
+// Node.js users.
+const hrtime = require("browser-process-hrtime");
 
 const GlobalEventHandlersImpl = require("../living/nodes/GlobalEventHandlers-impl").implementation;
 const WindowEventHandlersImpl = require("../living/nodes/WindowEventHandlers-impl").implementation;
@@ -210,9 +213,8 @@ function Window(options) {
   if (this._pretendToBeVisual) {
     this.requestAnimationFrame = fn => {
       const hr = hrtime(windowInitialized);
-      const hrInNano = hr[0] * 1e9 + hr[1];
+      const hrInMicro = hr[0] * 1e3 + hr[1] / 1e6;
       const fps = 1000 / 60;
-      const hrInMicro = hrInNano / 1e6;
 
       return startTimer(
         window,
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -130,6 +130,8 @@ function Window(options) {
   // HTMLFrameElement implementation.
   this._length = 0;
 
+  this._pretendToBeVisual = options.pretendToBeVisual;
+
   ///// GETTERS
 
   const external = External.create();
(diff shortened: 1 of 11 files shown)


lib/jsdom/browser/Window.js
```diff
@@ -198,8 +199,10 @@ function Window(options) {
     }
     return startTimer(window, setInterval, clearInterval, ++latestTimerId, fn, ms, args);
   };
+  this.requestAnimationFrame = fn => this.setTimeout(fn, 1000 / 60, performanceNow());
```
domenic: Hmm I don't think performanceNow() is correct. Per https://w3c.github.io/hr-time/#time-origin-0 it should be when the jsdom window is created, whereas this package uses "the reported time is relative to the time the current Node process has started". These are not the same.

I think instead you need to record the appropriate (high-res) time in the Window constructor.

domenic: The use of setTimeout and clearTimeout is not correct here. It means that calling setTimeout then requestAnimationFrame will cause the latter to return 1, not 0. Similarly, it means that using clearTimeout or clearInterval, we can clear animation frame callbacks. Neither of these is per spec. See https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#animation-frames and the "animation frame callback identifier" business.

It sounds like the test suite is lacking here, so we should add new to-upstream tests for those scenarios.

test/web-platform-tests/index.js
```diff
@@ -301,6 +301,11 @@ describe("Web Platform Tests", () => {
     "html/syntax/parsing/template/clearing-the-stack-back-to-a-given-context/clearing-stack-back-to-a-table-context.html",
     "html/syntax/parsing/template/clearing-the-stack-back-to-a-given-context/clearing-stack-back-to-a-table-row-context.html",
     // "html/syntax/parsing/template/creating-an-element-for-the-token/template-owner-document.html", // flaky on CI; possibly due to how we load iframes
+    "html/webappapis/animation-frames/callback-exception.html",
+    // "html/webappapis/animation-frames/callback-invoked.html", document.hidden should be false, but is true
```
domenic: This test points to a real problem, which is that we now have a contradiction within jsdom. jsdom signals itself as a "headless" browser, which does not display things visually. Thus hidden is true. But for such browsers, the spec says that requestAnimationFrame should not be called. So I am not sure what to do here.

lib/jsdom/browser/Window.js
SimenB: Should I duplicate the code for `setTimeout` changing `latestTimerId`, and implement another `stopTimer`, to be it's own thing?

test/web-platform-tests/index.js
SimenB: IMO the spec is being overly strict, but as I have no real experience with specs and all the edge cases they have to cover, I'll of course defer to you.

But I would like to avoid having to polyfill this for projects with tests run in a browser environment 🙂

lib/jsdom/browser/Window.js
SimenB: Good point!

domenic: Yeah, I guess so, or use some abstraction that allows determining which timer variable to update maybe?

test/web-platform-tests/index.js
SimenB: What I as a developer find more natural is "a headless browser should have it's callback invoked immediately, not at what matches 60 FPS (or the browser's refreshrate)" or something along those lines.

domenic: I mean, no animations can run in a headless browser, so I think the spec makes sense.

domenic: My current thinking is maybe we need a new top-level jsdom option, like `pretendToBeVisual: true`, which opts you out of document.hidden and opts you in to requestAnimationFrame. Curious what other contributors think.

(And also better names. Although it needs to convey that this is "pretend" or "fake" or something, since we *don't* run e.g. CSS animations or do layout or paint or any other visual stuff.)

lib/jsdom/browser/Window.js
SimenB: Updtaed

test/web-platform-tests/index.js
SimenB: I like that idea.
Does it have to be done in this PR, or is this blocked on a new option?

domenic: I think we need such an option before we can add requestAnimationFrame. But, you probably shouldn't spend much time on it before we give the other contributors and team members a chance to wake up and weigh in. Maybe they'll have a brilliant alternate suggestion we should do instead.

lib/jsdom/browser/Window.js
```diff
@@ -183,23 +186,42 @@ function Window(options) {
...
   };
   this.setInterval = function (fn, ms) {
     const args = [];
     for (let i = 2; i < arguments.length; ++i) {
       args[i - 2] = arguments[i];
     }
-    return startTimer(window, setInterval, clearInterval, ++latestTimerId, fn, ms, args);
+    return startTimer(window, setInterval, clearInterval, ++latestTimerId, fn, ms, window.__timers, args);
+  };
+  this.requestAnimationFrame = fn => {
```
domenic: Can you add a test to the to-upstream web platform tests that tests non-interference between setTimeout/setInterval and requestAnimationFrame? Basically:

- requestAnimationFrame's return value starts from zero, no matter how many setTimeouts/setIntervals happen before it.
- clearTimeout does not clear a requestAnimationFrame
- clearInterval does not clear a requestAnimationFrame
- cancelAnimationFrame does not clear a timeout
- cancelAnimationFrame does not clear an interval

SimenB: Done (I think 🙂)

test/web-platform-tests/to-upstream/html/webappapis/animation-frames/no-interference-timers.html
```diff
@@ -0,0 +1,53 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>AnimationTiming Test: does not interfere with timers</title>
+<link rel="author" title="Intel" href="http://www.intel.com">
+<link rel="help" href="https://html.spec.whatwg.org/multipage/#dom-window-requestanimationframe">
```
SimenB: no idea what metadata should be here. I just copied a test from the existent animation-frame tests

test/web-platform-tests/index.js
Zirro: I'm positive towards the idea of an option, as long as the README makes the limitations of this mode clear. It should probably also affect `document.visibilityState`.

In the longer term this could potentially be used to resolve issues like #1422, with `window.scroll()` logging to the console by default but pretending to scroll a viewport when this option is enabled. We may want to make the value of the option an object or a string (akin to `runScripts`) to enable different "levels" of pretending, in this case.

lib/jsdom/browser/Window.js
```diff
@@ -67,6 +69,7 @@ function Window(options) {
   this._globalProxy = this;
 
   this.__timers = Object.create(null);
+  this.__animationFrameCallbacks = Object.create(null);
```
TimothyGu: Does this have to go on the Window object? Can it not be just a variable?

SimenB: I just copied `__timers`. Should I change both?

TimothyGu: I'm guessing there is a reason why `__timers` is still there. @Sebmaster/@domenic?

Sebmaster: It's used (or should be used) to stop all timers on window close, I think?

TimothyGu: Yeah but if we bring `stopAllTimers()` function into the Window constructor, it will have access to these variables as well w/o leaking them to the outside environment. If there are no more reasons, I think that should be done.

domenic: Well, I'm not sure it's a great idea to add that closure (and all the other closures that currently reference window.__timers) into every window instance. I think the current setup is OK-ish.

lib/jsdom/browser/Window.js
```diff
@@ -183,23 +186,42 @@ function Window(options) {
...
+      window.__animationFrameCallbacks,
+      [hrInMicro]
+    );
   };
-  this.clearInterval = stopTimer.bind(this, window);
-  this.clearTimeout = stopTimer.bind(this, window);
+  this.clearInterval = stopTimer.bind(this, window.__timers);
+  this.clearTimeout = stopTimer.bind(this, window.__timers);
+  this.cancelAnimationFrame = stopTimer.bind(this, window.__animationFrameCallbacks);
   this.__stopAllTimers = stopAllTimers.bind(this, window);
```
TimothyGu: You need to cancel all animation frame callbacks `window.close()` as well.

lib/jsdom/browser/Window.js
TimothyGu: @domenic Right now, that function is the only one that uses `__timers` that is not already under the Window constructor. It would also be a move towards consistency to move it under the Window constructor as well.

lib/jsdom/browser/Window.js
SimenB: Ah, I was unsure of the purpose here. Will do, thanks!

lib/jsdom/browser/Window.js
SimenB: I pushed a commit which removed both from `this`. PTAL 🙂

SimenB:
I merged in master (to see if I introduced any linting errors after upgrading eslint), and `package-lock` got quite a large change (fsevents added, so I'm guessing current lockfile was generated on linux/windows).

Too bad npm differs between platforms when it comes to optionalDependencies. I thought they'd fixed that in 5... I don't think fsevents will be removed from the file now, though

TimothyGu:
@SimenB You should rebase on instead of merge from `master` for linear history.

SimenB:
I skipped it to not bog down #1963 with references. I see that I didn't reference it in any commit though, so can rebase instead (I see there's conflicts now anyway)

domenic:
I've been making a bit of a mess of things out from under you this weekend, so I can take care of the rebasing.

@TimothyGu @Sebmaster do you have thoughts on the thread at https://github.com/tmpvar/jsdom/pull/1994#discussion_r142004952 ? I think that's the real blocker for moving forward with this.

test/web-platform-tests/index.js
Sebmaster: We're hitting this more often nowadays when implementing new APIs, I do think it's time for a new config option...

domenic:
OK, I rebased, and for good measure made some tweaks to the new test you wrote.

It sounds like a new option is the way to go. This will be a bit tricky. Adding a new option is not really meant to be easy, and I'm still a bit hesitant; the old API had way too many options, and I don't want to revisit that world too quickly. But oh well. Here's the process.

- Decide on a name. So far `pretendToBeVisual: true` (defaults false) is my best candidate.
- Edit lib/jsdom/api.js to thread this option through from the JSDOM constructor to the Window, storing it as a `_pretendToBeVisual` property (which is always present, and either true or false; convert any incoming value via `Boolean(value)`).
- Edit lib/jsdom/old-api.js to do the same, at least for jsdom.env. This doesn't have to be pretty, or tested, or documented, but unfortunately for now we're running web platform tests through jsdom.env, so it needs to happen.
- Edit lib/jsdom/living/nodes/Document-impl.js so that the `hidden` and `visibilityState` getters look up `this._defaultView && this._defaultView._pretendToBeVisual`. If that's true, return false/visible.
- Edit Window.js so that requestAnimationFrame/cancelAnimationFrame are no-ops unless that option is set.
- Edit test/api/options.js with a new describe() block. It should test:
  - By default, document.hidden is false, document.visibilityState is prerender, and after, say, 500 ms, requestAnimationFrame never calls its callback
  - When the option is set to true, the reverse of all those apply
- Edit run-single-wpt.js's jsdom.env call to pass pretendToBeVisual: true
- Describe the option in the readme. I think it will deserve a separate subsection next to "Executing scripts".

I can help with some of this, but likely not this weekend. Let me know which of these you feel up for and which you'd rather leave to me.

SimenB:
Great, thanks!

>Let me know which of these you feel up for and which you'd rather leave to me.

It's bedtime for me now, but most of them seems pretty doable. I'll come back tomorrow with a go at it

SimenB:
Ok, pushed a commit adding the option.

2 things:
1. I added a TODO for behavior when not pretending to be visual.
2. I was unable to run the karma tests because of `process.hrtime` not being defined in the browser. Adding a polyfill is fine, but browser tests still failed with this error:
  ```
Chrome 61.0.3163 (Mac OS X 10.13.0) ERROR
  Uncaught Error: Cannot find module '/test/web-platform-tests/wpt-config.json'
  at /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/node_modules/browser-pack/_prelude.js:1:0 <- /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/0f948202e4b638c34a34dcff7aa992af.browserify:1
```

I ran the new tests with mocha, `yarn mocha test/api/options.js`

File: test/web-platform-tests/to-upstream/html/webappapis/animation-frames/no-interference-timers.html [hunk 7]
```diff
@@ -32,7 +32,7 @@
   assert_equals(firstRafId, 1);
   assert_equals(rafId1, 2);
   assert_equals(rafId2, 3);
-}, "Animation frame IDs must be unaffected by timer IDs, and vice-versa");
+}, "Animation frame IDs must be unaffected by timer IDs");
```
SimenB: we only check that raf is not impacted by timers now

Zirro:
> I was unable to run the karma tests because of process.hrtime not being defined in the browser.

It looks like the default [`process`-shim](https://github.com/defunctzombie/node-process) used in browserify lacks hrtime, needing [this separate module](https://github.com/kumavis/browser-process-hrtime) to be included as well. We may either have to work around that by also checking for `performance.now()`, or add a note to the README for browserify users and consider this a breaking change. I'm not sure about the preferred solution here.

> jsdom by default will do no rendering, layout, animation, rendering or css. However, you can opt-in to a subset of features which enables some of it.

I'm a bit concerned that this wording might give the opposite impression of what we intend. I think it could be more clear that it enables none of those features (though you can exclude CSS, which it does process to some degree), but only changes a few properties concerning the browser window visibility, along with adding an imitation of `requestAnimationFrame()`.

SimenB:
WRT `hrtime` in the browser, I can work around it by using a simple wrapper which works in both browser and node.

```diff
diff --git i/lib/jsdom/browser/Window.js w/lib/jsdom/browser/Window.js
index 737a997f..2de5fd78 100644
--- i/lib/jsdom/browser/Window.js
+++ w/lib/jsdom/browser/Window.js
@@ -11,6 +11,7 @@ const cssom = require("cssom");
 const postMessage = require("../living/post-message");
 const DOMException = require("domexception");
 const { btoa, atob } = require("abab");
+const hrtime = require("browser-process-hrtime");
 const idlUtils = require("../living/generated/utils");
 const createXMLHttpRequest = require("../living/xmlhttprequest");
 const createFileReader = require("../living/generated/FileReader").createInterface;
@@ -43,7 +44,7 @@ dom.Window = Window;
 function Window(options) {
   EventTarget.setup(this);
 
-  const windowInitialized = process.hrtime();
+  const windowInitialized = hrtime();
 
   this._initGlobalEvents();
 
@@ -208,7 +209,7 @@ function Window(options) {
 
   if (this._pretendToBeVisual) {
     this.requestAnimationFrame = fn => {
-      const hr = process.hrtime(windowInitialized);
+      const hr = hrtime(windowInitialized);
       const hrInNano = hr[0] * 1e9 + hr[1];
       const fps = 1000 / 60;
       const hrInMicro = hrInNano / 1e6;
diff --git i/package-lock.json w/package-lock.json
index 7075716d..45032de6 100644
--- i/package-lock.json
+++ w/package-lock.json
@@ -532,6 +532,11 @@
         "umd": "3.0.1"
       }
     },
+    "browser-process-hrtime": {
+      "version": "0.1.2",
+      "resolved": "https://registry.npmjs.org/browser-process-hrtime/-/browser-process-hrtime-0.1.2.tgz",
+      "integrity": "sha1-Ql1opY00R/AqBKqJQYf86K+Le44="
+    },
     "browser-resolve": {
       "version": "1.11.2",
       "resolved": "https://registry.npmjs.org/browser-resolve/-/browser-resolve-1.11.2.tgz",
index 9e169494..8909ad78 100644
--- i/package.json
+++ w/package.json
@@ -21,6 +21,7 @@
     "acorn": "^5.1.2",
     "acorn-globals": "^4.0.0",
     "array-equal": "^1.0.0",
+    "browser-process-hrtime": "^0.1.2",
     "content-type-parser": "^1.0.1",
     "cssom": ">= 0.3.2 < 0.4.0",
     "cssstyle": ">= 0.2.37 < 0.3.0",
```

As mentioned though, tests still fail because it's unable to load a json file. Am I doing something weird, or is something missing from yesterday's 19183e86c3aea12fe0c3eb5e49c9d4a4f1cd841f?

>I'm a bit concerned that this wording might give the opposite impression of what we intend.

I'd love some exact wording, I doubt I'll be able to convey what you guys want here 🙂

TimothyGu:
For the browser issues, see https://github.com/tmpvar/jsdom/issues/1998. https://github.com/tmpvar/jsdom/pull/2000 fixes it.

Zirro:
> I'd love some exact wording...

While I don't speak for everyone as I am just a frequent contributor, I'd suggest making the title "Pretend to render" rather than "Visual mode" (I think I'd prefer naming the option `pretendToRender` instead of `pretendToBeVisual` as well, but I'll let @domenic decide that one).

As for the text, I'm thinking:

> jsdom does not have the capability to render graphical content, and will act like a headless browser by default. It provides hints to web pages through APIs such as `document.hidden` that their content is not visible.
> 
> When the `pretendToBeVisual` option is set to `true`, jsdom will pretend that it is rendering and displaying content. It does this by changing the values of `document.hidden` and `document.visibilityState`. It will also provide an imitation of the `window.requestAnimationFrame()` / `window.cancelAnimationFrame()` methods, normally only available for content being rendered.
>
> Note that using this option could lead to unpredictable behaviour in code that changes action depending on content visibility.

Feel free to provide feedback on this.

SimenB:
Rebased on master and added a commit so that it works in the browser. I have a single failing test, but it fails on master as well.

```
Chrome 61.0.3163 (Mac OS X 10.13.0) API: runScripts constructor option <script>s and eval() should execute <script>s with correct location when set to "dangerously" and includeNodeLocations FAILED
        AssertionError: Unspecified AssertionError
            at VirtualConsole.virtualConsole.on.err (/var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/test/api/options-run-scripts.js:46:0 <- /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/b7206e9876400913581586fe6f6fa4fe.browserify:188467:13)
            at VirtualConsole.EventEmitter.emit (/var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/node_modules/events/events.js:81:0 <- /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/b7206e9876400913581586fe6f6fa4fe.browserify:113855:17)
            at reportException (/var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/lib/jsdom/living/helpers/runtime-script-errors.js:70:0 <- /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/b7206e9876400913581586fe6f6fa4fe.browserify:48579:28)
            at processJavaScript (/var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/lib/jsdom/living/nodes/HTMLScriptElement-impl.js:142:0 <- /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/b7206e9876400913581586fe6f6fa4fe.browserify:54380:7)
            at HTMLScriptElementImpl._eval (/var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/lib/jsdom/living/nodes/HTMLScriptElement-impl.js:65:0 <- /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/b7206e9876400913581586fe6f6fa4fe.browserify:54303:7)
            at /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/lib/jsdom/browser/resource-loader.js:31:0 <- /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/b7206e9876400913581586fe6f6fa4fe.browserify:2181:22
            at Object.check (/var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/lib/jsdom/living/nodes/Document-impl.js:90:0 <- /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/b7206e9876400913581586fe6f6fa4fe.browserify:50172:11)
            at /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/lib/jsdom/living/nodes/Document-impl.js:109:0 <- /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/b7206e9876400913581586fe6f6fa4fe.browserify:50191:12
            at HTMLScriptElementImpl._attach (/var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/lib/jsdom/living/nodes/HTMLScriptElement-impl.js:42:0 <- /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/b7206e9876400913581586fe6f6fa4fe.browserify:54280:72)
            at HTMLBodyElementImpl._attach (/var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/lib/jsdom/living/nodes/Node-impl.js:281:0 <- /var/folders/58/x7ryp52n7mxbtkld6lr0bqg00000gn/T/b7206e9876400913581586fe6f6fa4fe.browserify:55772:15)
Chrome 61.0.3163 (Mac OS X 10.13.0): Executed 1263 of 2504 (1 FAILED) (skipped 1241) (6.457 secs / 5.561 secs)
```

TimothyGu:
That'll be https://github.com/tmpvar/jsdom/issues/2004… but hey, Travis CI was green for one commit, at least!

lib/jsdom/browser/Window.js
```diff
@@ -182,24 +188,58 @@ function Window(options) {
...
+        fn,
+        fps,
+        animationFrameCallbacks,
+        [hrInMicro]
+      );
+    };
+    this.cancelAnimationFrame = stopTimer.bind(this, animationFrameCallbacks);
+  } else {
+    // TODO: Should we throw to tell users about `pretendToBeVisual`?
+    // Also, should we somehow allow this to be polyfilled byt code using feature detection?
```
SimenB: I'd prefer if we just don't define them at all rather than no-op. I'm guessing this will lead to really hard-to-debug behavior for users, where the callback silently never gets called, and it's probably non-obvious why.

Maybe we can add a getter which prints a warning or something?

SimenB:
Hmm, lots of tests timed out, so tried removing the `test/web-platform-tests/tests` change

SimenB:
@domenic anything you want me to change here?

domenic:
Haven't had much time for jsdom recently; will try to get to it this weekend.

lib/jsdom/browser/Window.js
TimothyGu: I think not having the functions at all seems better.

lib/jsdom/browser/Window.js
```diff
@@ -182,24 +188,58 @@ function Window(options) {
...
+    this.requestAnimationFrame = () => {};
+    this.cancelAnimationFrame = () => {};
+  }
+
+  this.__stopAllTimers = function () {
+    stopAllTimers(timers);
+    stopAllTimers(animationFrameCallbacks);
```
TimothyGu: It used to be that one can continue creating timers after a `__stopAllTimers` call. It would not be the case any more, since `timers` and `animationFrameCallbacks` are no longer reset. I think they should be.

lib/jsdom/browser/Window.js
```diff
@@ -377,7 +417,7 @@ function Window(options) {
       delete this._document;
     }
 
-    stopAllTimers(currentWindow);
+    this.__stopAllTimers();
```
TimothyGu: Is monkey-patching a concern?

Sebmaster: Not really. We've basically always "reserved" underscore props, it's nowadays taken mostly care of by webidl2js, but for a long time, that's just how it always was. I hope one day it'll be window's turn.

lib/jsdom/browser/Window.js
SimenB: Oh, good catch! Did not mean to remove that

lib/jsdom/browser/Window.js
SimenB: Cool, I removed them

lib/jsdom/browser/Window.js
```diff
@@ -182,24 +188,56 @@ function Window(options) {
...
+      const fps = 1000 / 60;
+      const hrInMicro = hrInNano / 1e6;
+
+      return startTimer(
+        window,
+        setTimeout,
+        clearTimeout,
+        ++latestAnimationFrameCallbackId,
+        fn,
+        fps,
```
domenic: It occurs to me that while in a browser if you did something like:

rAF
wait 6 milliseconds
rAF
wait 6 milliseconds
rAF

then you'd probably get all callbacks called together, instead of at now + 16 ms, now + 22 ms, now + 28 ms. But, we can leave it like this, and see if anyone complains...

SimenB: I've tried to test this in Chrome, but at least using `setTimeout` I'm unable to have them fire at once.

```js
requestAnimationFrame((t)=>{console.log('hello', t, performance.now())});
setTimeout(() => requestAnimationFrame((t)=>{console.log('hello delayed', t, performance.now())}), 1);
```

Prints e.g.

```
hello 2098569.228999957 2098571.895
hello delayed 2098585.90299997 2098587.875
```

Running them multiple times never have them fire at once

PR #2000, opened 2017-09-30
Fix browser testing

### LINKED ISSUE #1998: New WPT runner breaks browser CI
This wasn't caught because PRs can't trigger browser CI, I guess?

Maybe we should switch back to having branch + PR builds on travis, so as to catch this sort of thing.

Hopefully this will be easy enough to debug and fix locally... we'll see.

This needs to get fixed ASAP so we can have a nice green CI again. Unfortunately I'm kind of losing steam for the weekend. Help appreciated, if anyone's up for it, but I totally understand not wanting to clean up my mess.

TimothyGu:
Fixes #1998.

### DIFF
--- a/test/karma-webworker.conf.js
+++ b/test/karma-webworker.conf.js
@@ -41,6 +41,9 @@ module.exports = config => {
       debug: true,
       configure(bundle) {
         bundle.ignore("fs");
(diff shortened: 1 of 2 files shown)


domenic:
Thank you so, so much!

PR #2003, opened 2017-10-01
Use correct line number for <script> if possible

TimothyGu:
This makes debugging WPT drastically easier. With the new API, this requires `includeNodeLocations` option to work.

Column is not tracked since it works a bit differently:

```
> vm.runInContext('throw new Error()', ctx, { columnOffset: 4 })
evalmachine.<anonymous>:1
throw new Error()
    ^

Error
    at evalmachine.<anonymous>:1:11
```


domenic:
Ah, this is so great! I've been meaning to do this for a while but I always forget about it after my annoyance goes away. I'll merge after a CI re-run.

domenic:
There's no need for column, right? I guess maybe on the first line if you do `<script>some stuff</script>` (with no newline after `<script>`), but that's quite an edge case...

TimothyGu:
I mean, preferably, especially for minified HTML. But for our purposes (making debugging WPT bearable) I'm okay with not doing that at the moment.

TimothyGu:
The issue about column number is fixed in https://github.com/nodejs/node/pull/15771.

PR #2010, opened 2017-10-11
Rethrow exception in timeout callback.

H1Gdev:
Application(e.g.Jest) watches uncaught Exception in callback function.(use **uncaughtException** event)
So rethrow exception in timeout callback.


domenic:
This is against spec.

H1Gdev:
Thanks.
Is there any way to catch Exception ?

domenic:
Just like in a browser: `window.addEventListener("error", ...)`

H1Gdev:
Thanks.
I will try in this way.

PR #2011, opened 2017-10-11
A Better SVG implementation

TimothyGu:
Supersedes #1445, #1946. Implemented entirely according to spec.

Fixes #1423, #1528, #1938, #2001.

<del>Needs tests.</del>

PR #2014, opened 2017-10-13
fix #1986

### LINKED ISSUE #1986: querySelector within a fragment doesn't return SVG elements
### Basic info:

- **Node.js version: 6.11.2**
- **jsdom version: 11.2.0**

### Minimal reproduction case

```js
var {JSDOM} = require("jsdom")
var frag = JSDOM.fragment(`<svg></svg>`);
console.log(frag.querySelector("svg"));
console.log(frag.querySelector("*"));
```

`frag.querySelector("svg")` returns `null`, while the wildcard selector correctly returns the SVG element.

caub:


package.json
```diff
@@ -27,7 +27,7 @@
     "domexception": "^1.0.0",
     "escodegen": "^1.9.0",
     "html-encoding-sniffer": "^1.0.1",
-    "nwmatcher": "^1.4.1",
+    "nwmatcher": "^1.4.3",
```
TimothyGu: You need to update yarn.lock as well.

caub: thanks, done

domenic:
This needs tests before we can merge it.

caub:
I'd need #2011 merged first, the test doesn't pass yet without

domenic:
Please read CONTRIBUTING.md; that is not an appropriate test for a feature like this.

caub:
Ok, from what I understand it should be a wpt test, so I'll search which to enable, thanks

caub:
For adding a test, I added `Element.matches`

if fails differently than before (it was failing on a `:invalid`) selector
```
Failed in "In-document Element.matches: Attribute presence selector, matching title attribute, case insensitivity (with no refNodes): #attr-presence [*|TiTlE]": 
The string "#attr-presence [*|TiTlE]", is not a valid CSS selector

SyntaxError: The string "#attr-presence [*|TiTlE]", is not a valid CSS selector
 ...
```

domenic:
Yes, a web platform test in to-upstream would be good.

caub:
I added a stupid test, but well I think I should remove it, Timothy already covers #1986 case in https://github.com/tmpvar/jsdom/blob/master/test/web-platform-tests/to-upstream/svg/element-svg.html#L19

domenic:
I mean, we need a test that fails without this change, and passes with it. If Timothy's test covers it already, then this change isn't needed, it sounds like, and we should close this pull request?

caub:
on master branch I get the issue only for fragments, like #1986 described:
```sh
~/dev/jsdom$ node
> var {JSDOM} = require('.')
> var frag = JSDOM.fragment(`<svg></svg>`);
> frag.querySelector("svg")
null
> var frag = JSDOM.fragment(`<div><svg><foreignObject/></svg></div>`);
> var div=frag.firstChild
> div.querySelector('svg')
null
> var {window: {document}}=new JSDOM('<div><svg><foreignObject/></svg></div>')
> var div = document.querySelector('div')
> div.querySelector('svg')
SVGSVGElement {}
```
The SVG PR adds those new SVG classes instead of HTMLUnknownElement.

I don't understand well, but this PR only fixes it for fragments, so I'll need more classic unit tests

domenic:
No, please don't do that. If we can't figure out how this is exposed to web code, we have a bigger bug.

caub:
using jsdom ^11.1.0

```js
var frag1 = JSDOM.fragment(`<svg><foreignObject/></svg>`);
var frag2 = JSDOM.fragment(`<div><svg><foreignObject/></svg></div>`);
frag1.querySelector('svg') // null
frag1.querySelector('*') // HTMLUnknownElement

frag2.querySelector('svg') // null
frag.querySelector('div').querySelector('svg') // HTMLUnknownElement
```

domenic:
I managed to reproduce the issue in a web platform test as follows:

```html
<!DOCTYPE html>
<meta charset="utf-8">
<title>querySelector on template fragments with SVG elements</title>

<template id="template1"><div></div></template>
<template id="template2"><svg></svg></template>
<template id="template3"><div><svg></svg></div></template>

<script>
"use strict";

test(() => {
  const fragment = document.querySelector("#template1").content;
  assert_not_equals(fragment.querySelector("div"), null);
}, "querySelector works on template contents fragments with HTML elements (sanity check)");

test(() => {
  const fragment = document.querySelector("#template2").content;
  assert_not_equals(fragment.querySelector("svg"), null);
}, "querySelector works on template contents fragments with SVG elements");

test(() => {
  const fragment = document.querySelector("#template2").content;
  assert_not_equals(fragment.firstChild.querySelector("svg"), null);
}, "querySelector works on template contents fragments with nested SVG elements");
</script>
```

I will push this to this branch so we can merge.

PR #2016, opened 2017-10-15
Update to new webidl2js

TimothyGu:
- Overloads
- Enumerations
- Update some IDL fragments from specs

Needs https://github.com/jsdom/webidl2js/pull/86

### DIFF
--- a/lib/jsdom/living/nodes/ChildNode.webidl
+++ b/lib/jsdom/living/nodes/ChildNode.webidl
@@ -1,10 +1,10 @@
 [NoInterfaceObject,
  Exposed=Window]
 interface ChildNode {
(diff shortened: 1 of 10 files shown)


domenic:
I'm surprised there aren't more locations where we use enumerations and overloads...

Zirro:
I've got another one in css-object-model - [`CSS.supports()`](https://drafts.csswg.org/css-conditional-3/#the-css-interface) - so I'm happy to see this get implemented.

domenic:
It looks like this causes a bunch of XHR test failures?

domenic:
Back to @TimothyGu to assess the XHR failures.

TimothyGu:
Test failures (should be) fixed.

PR #2022, opened 2017-10-21
Make tests pass on Node.js 9.x

TimothyGu:
This happens when the input string contains a NUL character.


domenic:
What a strange breaking change. Meh, happy to take anything to keep old-api chugging along.

PR #2023, opened 2017-10-21
Roll web-platform-tests

Zirro:
With many new tests already having been added since the last roll, I figured it'd be good to get this done. At this point in `to-run.yaml`, I've only listed tests which depend on unimplemented features or otherwise seem like they would require major changes to resolve.

~~Twelve~~ (...thirty-seven in Node v6 it would seem) failing tests remain, and I'd like to see whether these can be resolved with minor fixes before this is merged. I've already included one fix for [abort-during-headers-received.htm](https://github.com/w3c/web-platform-tests/blob/master/XMLHttpRequest/abort-during-headers-received.htm) which previously would get the headers through `getAllResponseHeaders()` even after aborting.

If anyone else feels like looking at the remaining tests as well, feel free to help out.


lib/jsdom/living/xmlhttprequest.js
```diff
@@ -1090,9 +1090,10 @@ module.exports = function createXMLHttpRequest(window) {
 
   function setResponseToNetworkError(xhr) {
     const properties = xhr[xhrSymbols.properties];
+    properties.responseCache = properties.responseTextCache = properties.responseXMLCache = null;
+    properties.responseHeaders = {};
     properties.status = 0;
     properties.statusText = "";
-    properties.responseCache = properties.responseTextCache = properties.responseXMLCache = null;
```
Zirro: Note: This line is just me taking the opportunity to sort the properties in alphabetical order. Let me know in case you'd like me to avoid this in the future.

Zirro:
I've added support for the [filename mode](https://html.spec.whatwg.org/multipage/input.html#dom-input-value-filename) of `<input>`'s value attribute, which is partially tested through [input-value-invalidstateerr.html](https://github.com/w3c/web-platform-tests/blob/master/html/semantics/forms/the-input-element/input-value-invalidstateerr.html).

While there is a [manual test](https://github.com/w3c/web-platform-tests/blob/master/html/semantics/forms/the-input-element/file-manual.html) covering the rest, we'll probably need a solution for https://github.com/tmpvar/jsdom/issues/1272 to be able to test this further.

I've also listed the execution-timing tests that were failing only in older Node versions.

domenic:
I disabled some more tests that I know we're not going to get to soon.

In general I think it makes more sense to just add fail, unknown and get things rolled, instead of blocking a roll on fixing all the issues discovered. Unless you're particularly excited about fixing the issues, of course.

Zirro:
> Unless you're particularly excited about fixing the issues, of course.

I actually am quite excited about fixing them, but don't always have the time. Feel free to disable any remaining tests for now, and I might take a new look at fixing them later on.

Zirro:
This is ready for review. I rolled a second time to keep up-to-date and diagnosed/fixed some more tests, while disabling those I have not yet gotten to. I might continue to fix issues while this pull request remains open if I have time, but you can merge it whenever you prefer now.

Zirro:
I've added implementations of `Node.isConnected` and `Node.isSameNode()`. The former because it is being used in a recently added `option-text-setter.html` test, and the latter as it appeared in the updated IDL and seemed easy to handle.

`Node.isConnected` passes most parts of its test, but fails two regarding `<iframe>`'s which are removed from the document. I'm not sure whether the `isConnected`-implementation will need more code in order to handle that case, or whether this is due to another issue in jsdom. I still think it can be merged like this initially, but any suggestions are welcome.

lib/jsdom/living/nodes/Node.webidl
```diff
@@ -15,9 +15,11 @@ interface Node : EventTarget {
   readonly attribute unsigned short nodeType;
   readonly attribute DOMString nodeName;
 
-  readonly attribute DOMString? baseURI;
+  readonly attribute USVString baseURI;
```
Zirro: Not sure if a `DOMString` -> `USVString` change is worth mentioning in the changelog, but highlighting this just in case.

snuggs:
@Zirro  define the term "Rolled" please. Never heard it before related to jsdom.

Thanks in Advance!

PR #2027, opened 2017-10-25
fix some cors header issues

ScottAlbertine:
This solves 3 issues:
1) access-control-allow-headers is checked on regular requests
2) preflight responses that are 200 but have access-control-allow-headers with a non-matching header don't actually prevent regular requests
3) all request headers were being sent on preflight requests, which is against the CORS standard


ScottAlbertine:
oops, I left an extra variable in there that's broken the build, I'll fix it real quick

lib/jsdom/living/xhr-utils.js
```diff
@@ -5,283 +5,379 @@ const fs = require("fs");
 const { URL } = require("whatwg-url");
 
 const utils = require("../utils");
+const DOMException = require("domexception");
 const xhrSymbols = require("./xmlhttprequest-symbols");
 
-function wrapCookieJarForRequest(cookieJar) {
```
domenic: Can you undo this change? This file should stay entirely per-window and per-window processing should stay in xhr.js, like requestErrorSteps etc.

In other words, this file really isn't supposed to be xhr-utils, it's xhr-utils-independent-of-any-window.

ScottAlbertine: I can't, actually, because xhr-utils doesn't use a XMLHttpRequest to make preflight (options) requests, it just uses a regular request to do it.  Because those preflight requests can fail (due to mismatched headers, which is the whole point of this change), I need to be able to call dispatchError (xhr-utils.js:61) which then needs to call requestErrorSteps (xhr-utils.js:109) which needs to instantiate new ProgressEvents (xhr-utils.js:128,129,133,134), which require having the window.  I tried leaving all those functions in xmlhttprequest.js, but because they're not public, I can't call them from xhr-utils.  Thus, I moved them in and made xhr-utils depend on a window.  I did have the option of moving the createClient (xhr-utils.js:146) into xmlhttprequest instead, as that's what calls the CORS preflight validation check, but doing so would have eliminated basically all of the code in xhr-utils, so I opted to spread the code between the two files evenly rather than lumping it all together.

I'm comfortable with organizing this however best fits your design and use-cases for the project, so let me know the general refactoring plan I should use and I'll reorganize it that way.

domenic: Can you leave the code that calls dispatchError inside xmlhttprequest.js, like it is today? And have xhr-utils.js just return a signal (e.g. false) to the calling code in xmlhttprequest.js, which it uses to call dispatchError?

In general, we should be able to separate out "pure logic" utilities, like validating CORS headers, into the utilities files, whereas the side effecting things, like dispatching an error event, go in the xmlhttprequest.js file.

ScottAlbertine: That would require xmlhttprequest to be the class that makes the preflight request, which it currently isn't (and arguably, shouldn't be, unless you want me to rebuild your entire CORS flow from scratch).  If you look on xhr-utils.js:377, you'll see that the actual validation of the preflight response occurs outside xmlhttprequest, in a handler inside xhr-utils's createClient function.  The only way I can send a signal into xmlhttprequest from there is if we expose a public method to do so, which I think violates the xmlhttprequest spec in the first place.  If it doesn't, then I can recode this entire thing to be way more efficient by moving most of the xhr-utils methods (the ones that take xhr as their first parameter) into xmlhttprequest in the first place, which would solve both our problems.  Which would you prefer?

domenic: I will try to analyze the substance of your comment when I'm not at work, but let me try to be clearer on the larger point: we need as much code to be window-agnostic as possible. Doing so avoids the code being re-created, re-compiled, and stored in memory for every jsdom Window.

The current code has managed to factor out a lot of logic into xhr-utils as Window-agnostic. Your pull request destroys this distinction, creating copies of the utils for every Window. The same would be true of your proposal to move everything into xmlhttprequest. So those are not good.

I'll try to work through your response later and keep suggesting ways of preserving this property in the context of your fix. But perhaps by considering this high-level goal you can figure out a solution.

ScottAlbertine: I've looked at the call chain and tried to refactor the code in a way to maximize window-agnostic code, and I've realized that the entire substance of my change is that createClient function in xhr-utils.js cannot be window agnostic anymore, because making it window agnostic was what was causing the issue I'm fixing in the first place.

Let me spell it out for you:
The problem, previously, was that invalid preflight requests were not stopping their main requests from being sent.  This is the whole point of CORS.  I've added code that (at least in the cases I need for my day job) stops the main requests from being sent if the preflight response is invalid.  As part of that stoppage, you need to send an Event to the window, so that it can let the developer know that their request is not only failing, but failing before it's sent.  Thus, the checking of the preflight response *cannot* be window agnostic.  Since this is done in createClient, there is no way to make createClient window agnostic anymore.  The only functions in xhr-utils that can still be window agnostic are: setResponseToNetworkError, mergeHeaders, updateRequestHeader, getRequestHeader, and wrapCookieJarForRequest.  Everything else calls something that calls new Event.  

I understand your desire for window agnostic code, but that code inherently cannot do CORS properly, which is what I'm trying to fix.  Please let me know when you have time to look through the code and see why this is the case, or if you'd like me to do an even more in-depth line-by-line walkthrough of why this has to be the case.

domenic: > As part of that stoppage, you need to send an Event to the window

That's the part that I don't follow. Can you perform the stoppage in the window-agnostic code, then send the event in the window-specific code?

> new Event

Creating events is window-agnostic in jsdom. See, for example, https://github.com/tmpvar/jsdom/blob/87e95931cb8ad1b509a96daa9f6f59987ed5c3cf/lib/jsdom/living/nodes/HTMLElement-impl.js#L96

ScottAlbertine: ohhhhh! ok, cool, yeah, if I can create events without a window, I can leave most of the code as it is in my change.  I didn't know you could do that.  I've done the fix, running the tests now.

ScottAlbertine: Ok, that .createImpl() isn't working for me.  Sadly, I'm relatively new in Javascript, I mostly write Java, so there's probably something obvious I'm missing here.  My import in xhr-utils.js looks like

`const Event = require('./generated/Event');`

and the actual usage looks like

`xhr.dispatchEvent(Event.createImpl("readystatechange"));`

where xhr is an XMLHttpRequest.  When I do that, though, I get the following test failure:
```
  1) jsdom/cookie Getting a file URL should not set any cookies:
     Uncaught TypeError: Failed to execute 'dispatchEvent' on 'EventTarget': parameter 1 is not of type 'Event'.
      at convert (lib/jsdom/living/generated/Event.js:307:11)
      at XMLHttpRequest.dispatchEvent (lib/jsdom/living/generated/EventTarget.js:141:13)
      at requestErrorSteps (lib/jsdom/living/xhr-utils.js:121:7)
      at dispatchError (lib/jsdom/living/xhr-utils.js:62:3)
      at Object.validCORSHeaders (lib/jsdom/living/xhr-utils.js:77:5)
      at receiveResponse (lib/jsdom/living/xmlhttprequest.js:837:21)
      at EventEmitter.client.on.res (lib/jsdom/living/xmlhttprequest.js:669:38)
      at process.nextTick (lib/jsdom/living/xhr-utils.js:203:35)
      at _combinedTickCallback (internal/process/next_tick.js:131:7)
      at process._tickCallback (internal/process/next_tick.js:180:9)
```

What am I doing wrong?

domenic: Ah, that's due to a weirdness of XHR and how it's different from jsdom. I think you can just do Event.create() instead and it'll work.

Zirro: Note that the `create()` and `createImpl()` methods take an array with the name of the event as the first index (as shown on the line Domenic linked to earlier), which I think might be the cause of that error. It might help with readability if you create the event and dispatch it on separate lines.

ScottAlbertine: I've fixed it, xhr-utils no longer depends on window, and I didn't have to refactor the utils code back into xmlhttprequest.  The travis build is having trouble though, but as you can see from the build logs, all the tests pass, and I fixed the linter issues that the first build had.  If someone could kick off another build for me, that'd be great.

ScottAlbertine:
Something weird happened to the last Travis CI build, I don't have permissions to re-run it, if someone could do that for me I'd appreciate it.

ScottAlbertine:
Hi, it's been 5 days since I've updated this, can someone comment or pull it?

PR #2029, opened 2017-10-26
Correct DOMParser parsing of HTML entities in XML

myabc:
To be consistent with browser implementations, `parseFromString` should not parse HTML character entities in XML documents.

node sax already provides a `strictEntities` option. This commit enables this option, ensuring only predefined XML character entities (`&amp;`, `&apos;`, `&gt;`, `&lt;`, and `&quot;`) are parsed.

See: https://github.com/isaacs/sax-js/tree/master#arguments
See: https://www.w3.org/TR/REC-xml/#sec-predefined-ent

---

Chrome 61:

correct_domparser_parsing_of_html_entities_in_xml_by_myabc_ _pull_request__2029_ _tmpvar_jsdom

### DIFF
--- a/lib/jsdom/browser/htmltodom.js
+++ b/lib/jsdom/browser/htmltodom.js
@@ -52,7 +52,7 @@ module.exports = class HTMLToDOM {
 
   _parseWithSax(html, fragment, element) {
     const SaxParser = this.parser.parser;
-    const parser = new SaxParser(/* strict = */true, { xmlns: true });
+    const parser = new SaxParser(/* strict = */true, { xmlns: true, strictEntities: true });
     parser.noscript = false;
     parser.looseCase = "toString";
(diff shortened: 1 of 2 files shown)


domenic:
Great catch, thanks. Can you add this to a separate file instead of the dont-upstream.html? That one is supposed to be just a copy of the existing one from upstream but with tests we don't pass commented out.

myabc:
@domenic Sure. Should that test file also live in the `test/web-platform-tests/to-upstream/domparsing/` directory?

domenic:
Yep!

domenic:
New tests should pass lint, not be ignored :)

myabc:
@domenic sorry, I didn't realise that ES5 syntax was ok!

myabc:
@domenic awesome! thanks

PR #2030, opened 2017-10-26
Teach DOMParser.parseFromString to parse XML CDATA

### LINKED ISSUE #1593: DOMParser.parseFromString removes CDATA from document
Greetings.

When creating XMLDocument using DOMParser from a string which contains CDATA, the CDATA is removed from the output.

Example:

``` js
console.log(new window.DOMParser().parseFromString('<main><sub><![CDATA[http://someurl.com]]></sub></main>','text/xml').documentElement.innerHTML);
```

output:

``` html
<sub></sub>
```

myabc:
Fixes #1593

We're using JSDOM for unit testing a content player that (ab)uses XML by embedding HTML within CDATA tags. Unfortunately all CDATA sections are ignored by JSDOM's implementation of `DOMParser`. Were we designing an XML-based content format today, we would probably opt for XML namespacing or escaping HTML.

### DIFF
--- a/lib/jsdom/browser/htmltodom.js
+++ b/lib/jsdom/browser/htmltodom.js
@@ -62,6 +62,12 @@ module.exports = class HTMLToDOM {
         data: text
       });
     };
+    parser.oncdata = cdata => {
+      setChild(this.core, openStack[openStack.length - 1], {
+        type: "cdata",
+        data: cdata
+      });
+    };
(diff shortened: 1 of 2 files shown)


domenic:
Simple and elegant; thank you again!

domenic:
I think this also helps our XHTML support, for what it's worth.

PR #2042, opened 2017-11-08
trying syntax: object literal with methods

chharvey:
testing a pull request. might add to it by editing other files in a similar manner


domenic:
This seems like a bad change; we don't need this.

PR #2045, opened 2017-11-11
Fix typo in Readme

ipmsteven:
I think what the doc actually mean is trying to set the location href not compare the href value
then I changed it to `window.location.href = "https://example.com/"`


snuggs:
@ipmsteven nice catch 👓

PR #2046, opened 2017-11-11
Editorial: Add installation instructions to README.md

snuggs:
Not certain if we want to add a `yarn` install notification relative to https://github.com/tmpvar/jsdom/pull/1999 but at least need an install section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,10 @@ The latest versions of jsdom require Node.js v6 or newer. (Versions of jsdom bel
 
 As of v10, jsdom has a new API (documented below). The old API is still supported for now; [see its documentation](./lib/old-api.md) for details.
 
+## Installation
+
+jsdom is available as the `jsdom` package on [npm](https://npmjs.com). It is also available via [yarn](https://yarnpkg.com).
+
 ## Basic usage
 
 ```js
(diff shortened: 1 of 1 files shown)


snuggs:
cc @TimothyGu

Zirro:
The line is currently missing the package name.

Regardless, I doubt the need for this as it is just the standard installation procedure. Anyone with enough experience to find this package useful will presumably also be familiar with how to install it. Not including this also avoids suggesting which package manager to use, or whether to install as a dependency or devDependency.

snuggs:
@Zirro understood. Just uncanny as this would officially be the first package without even a light installation section. Below is how React handles multiple installation procedures in less than a paragraph.

Nonetheless, we will still need the following commit introduced based on your recommendations https://github.com/tmpvar/jsdom/pull/2046/commits/131d28def36798afb9a8d3b431020305f423e670

Zirro:
Those mentions refer to packages in the npm Registry, which Yarn publishes to as well. The tool-specificity is only a problem in the context of the CLI.

That said, a sentence like the one used by React looks okay to me if someone else agrees that it is worth adding.

domenic:
I think @Zirro's got it right, and the readme is fine as-is. Let's close this.

PR #2052, opened 2017-11-17
Convert NamedNodeMap to IDL

TimothyGu:
This also switches the ownership of "attribute list" and the associated cache to Element, which is a more literal reading of the spec and simplifies things here in jsdom.


domenic:
Amazing stuff!! I'm surprised we don't get any new WPTs passing for this, but I couldn't find any relevant ones.

I wonder if this cleans up the circular dependency enough to make #1641 fixable again.

PR #2054, opened 2017-11-17
Update dependencies

TimothyGu:
Most updates are in-range updates to yarn.lock, though two major updates are included also. The change to test/api/methods.js is because of a bugfix in parse5.


domenic:
Shouldn't we also update package.json?

TimothyGu:
Definitely a good idea for parse5, though the others can be done optionally, since a new installation would pull those in anyway.

domenic:
Sure, it just seems weird to encourage previous installations that upgrade to get different dependency versions than the one's we're testing against.

TimothyGu:
package.json updated too.

PR #2055, opened 2017-11-18
Remove the <applet> element

Zirro:
`<applet>` was removed from the spec in https://github.com/whatwg/html/pull/1399.

The WPT [html/dom/historical.html](https://github.com/w3c/web-platform-tests/blob/master/html/dom/historical.html) tests this in various places, but remains disabled here as it uses `document.all` and hits a quirk in our `getComputedStyle()` implementation. I've modified or removed old tests to account for the changes as well.


domenic:
Why the changes to to-run.yaml? Did they gain coverage that showed applet was removed? I thought per your post they didn't, since we can't pass historical.html.

Zirro:
It made it possible to associate these changes with a test, and document the failure reason for future reference. While two of the subtests don't pass for the reasons mentioned, the others do now.

Looking ahead, I'm also interested in the reflection tests located at the same directory level.

domenic:
Hmm. I'd like to keep the file length of to-run.yaml a bit shorter, probably by using `**` patterns to block out subdirectories we don't intend to work on yet.

If we wanted to associate this change with a test, we'd add something to to-upstream that copies a subset of the historical.html test, and name it dont-upstream, like https://github.com/tmpvar/jsdom/blob/d6cba380fcb61ee39e26099900825982e22d8b64/test/web-platform-tests/to-upstream/domparsing/DOMParser-dont-upstream.html

Zirro:
Okay, I can add a dont-upstream test and revert the to-run.yaml changes tomorrow assuming I have time. Or you may do it if you want to merge this today.

PR #2058, opened 2017-11-19
Add DocumentFragment.prototype.getElementById

domenic:
Based on a patch by @Zirro from IRC, with my own tests added. However, this does *not* work, because the functions in Element-impl.js that update the `_ids` collection do not care about DocumentFragment containers.

I'm unsure how to proceed. I guess one route is to have totally separate implementations of getElementById for Document and DocumentFragment, with the DocumentFragment version doing a whole traversal. That seems relatively easy. Alternately we could try to develop some kind of caching system, and a notation of "root" i.e. furthest ancestor.

Thoughts welcome.

### DIFF
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -454,6 +454,12 @@ class DocumentImpl extends NodeImpl {
     this.write(...arguments, "\n");
   }
 
+  // This is implemented separately for Document (which has a _ids cache) and DocumentFragment (which does not).
+  getElementById(id) {
+    // Return the first element with this ID.
+    return this._ids[id] && this._ids[id].length > 0 ? this._ids[id][0] : null;
+  }
+
   get referrer() {
     return this._referrer || "";
   }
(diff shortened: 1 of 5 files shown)


Sebmaster:
> I guess one route is to have totally separate implementations of getElementById for Document and DocumentFragment, with the DocumentFragment version doing a whole traversal.

That'd be very okay with me, really. Implement first, speed up later.

However:

> However, this does not work, because the functions in Element-impl.js that update the _ids collection do not care about DocumentFragment containers.

Do we not have the elementAttached thing which we could use for that? Does that bubble to the documentfragment?

domenic:
Hmm, I kind of assumed attached only worked on documents, not document fragments. But I didn't check.

dperini:
@domenic a couple of notes on gEBID implementation... Nodes/Elements attached to a DocumentFragment always have their `isConnected` property set to 'false'. So the `isConnected` thing only works for real Documents (nodeType 9).

Also recall that the HTML5 spec allows for duplicated ID in the same document, thus I believe that would be true also for DocumentFragment (nodeType 11). Does 'jsdom' allow duplicated ID in the same document ?

domenic:
Updated with separate Document and DocumentFragment implementations. It seems to work. Anyone up for reviewing?

I'll try nwsapi on top of this next. BTW @dperini, sorry for not answering your question. Yes, just like in a browser, multiple elements can have the same ID in jsdom.

Sebmaster:
Commit msg has WIP still 😁

PR #2059, opened 2017-11-19
Implement AbortSignal

Sebmaster:
I'm really glad now we have Webidl2JSFactory.


lib/jsdom/browser/Window.js
```diff
@@ -361,6 +363,14 @@ function Window(options) {
     window: this
   }).interface;
 
+  const AbortSignalWrapper = createAbortSignal({
```
Sebmaster: Sadly we require window because EventTarget uses it to make sure we're attached to a document/not closed. We should fix that some day (same problem in XHR, I think), but ... yak-shaving.

PR #2064, opened 2017-11-21
<input>.value improvements

TimothyGu:
Fixes #2063.

PR #2065, opened 2017-11-21
Add a test for the input[type=file].value setter

Zirro:
This test will pass with the changes in https://github.com/tmpvar/jsdom/pull/2064. It also passes in Chrome and Safari, but fails in Firefox which throws a `SecurityError` rather than the `InvalidStateError` defined by the specification.


Zirro:
Closing in favour of #2064 which now includes this test.

PR #2073, opened 2017-12-03
Don't include fragment in data: URL body

Zirro:
A test for this was added in https://github.com/w3c/web-platform-tests/pull/8109.


domenic:
Nice quick fix, thanks. Note that we're going to want a fuller parser per https://github.com/whatwg/fetch/pull/579 when that gets finalized; right now it's blocked on https://github.com/whatwg/mimesniff/pull/36 / https://github.com/jsdom/content-type-parser/pull/8 which is almost finished. But let's definitely merge this in the meantime.

PR #2075, opened 2017-12-06
[WIP] Add WebSocket Support

### LINKED ISSUE #1195: Supported WebSocket implementations
There's code we want to be able to test with JSDom that uses websockets. There are a [lot of different implementations](http://stackoverflow.com/questions/16392260/which-websocket-library-to-use-with-node-js). Which would you suggest for use in contexts server side in JSDom + client side in the browser? I'm leaning towards [ws](https://github.com/websockets/ws).

thdxr:
Fixes #1195

Opening this PR to track progress.

### DIFF
--- a/lib/jsdom/living/websocket/WebSocket-impl.js
+++ b/lib/jsdom/living/websocket/WebSocket-impl.js
@@ -1,4 +1,20 @@
 "use strict";
+const { setupForSimpleEventAccessors } = require("../helpers/create-event-accessor");
 
-exports.implementation = class WebSocketImpl {
-};
+const READY_STATES = Object.freeze({
+  CONNECTING: 0,
+  OPEN: 1,
+  CLOSING: 2,
+  CLOSED: 3
+});
+
+const events = ["open", "error", "close"];
+
+class WebSocketImpl {
+  constructor() {
+    this.readyState = READY_STATES.EMPTY;
+  }
+}
+setupForSimpleEventAccessors(WebSocketImpl.prototype, events);
+
+exports.implementation = WebSocketImpl;
--- a/lib/jsdom/living/websocket/WebSocket-impl.js
+++ b/lib/jsdom/living/websocket/WebSocket-impl.js
@@ -1,0 +1,7 @@
+"use strict"
+
+exports.implementation = class WebSocketImpl {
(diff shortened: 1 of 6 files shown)


lib/jsdom/living/websocket/WebSocket-impl.js
```diff
@@ -0,0 +1,20 @@
...
+const READY_STATES = Object.freeze({
+  CONNECTING: 0,
+  OPEN: 1,
+  CLOSING: 2,
+  CLOSED: 3
+});
+
+const events = ["open", "error", "close"];
+
+class WebSocketImpl {
```
TimothyGu: This should `extends EventTargetImpl`, after importing `EventTargetImpl` from `require('../events/EventTarget-impl').implementation`.

lib/jsdom/living/websocket/WebSocket-impl.js
```diff
@@ -0,0 +1,20 @@
...
+  constructor() {
```
TimothyGu: This will have to be `constructor(args, privateData)`, and immediately after the declaration `super(args, privateData)` to initialize `EventTarget`.

lib/jsdom/living/websocket/WebSocket.webidl
```diff
@@ -0,0 +1,31 @@
+// https://html.spec.whatwg.org/multipage/web-sockets.html#the-websocket-interface
+
+enum BinaryType { "blob", "arraybuffer" };
+[Constructor(USVString url, optional (DOMString or sequence<DOMString>) protocols = []), Exposed=(Window,Worker)]
```
TimothyGu: It looks like this is going to have to contain `WebIDL2JSFactory` attribute as well, since EventTarget won't function w/o a `ownerDocument`, which is per-window.

Basically, instead of having a shared `WebSocket` interface in `living/index.js`, you would want to create a new interface in the Window constructor. Check out f77cc1e3a9eb38fb7db6affa5a9c3d2acca1b351 and https://github.com/jsdom/webidl2js#webidl2jsfactory.

TimothyGu:
This is superseded by https://github.com/tmpvar/jsdom/pull/2088.

PR #2076, opened 2017-12-07
Implement Web storage - localStorage, sessionStorage, StorageEvent

Zirro:
This is a work-in-progress implementation of the `Storage` interface, along with `localStorage`, `sessionStorage` and the `storageEvent`. It resolves https://github.com/tmpvar/jsdom/issues/1137.

All the tests for storing data are currently passing, but some of the event tests will require further work. I welcome feedback at this stage, but keep in mind that this is not yet complete.

(Also, thank you @TimothyGu for making this a breeze to implement through all your improvements to webidl2js!)

### DIFF
--- a/lib/jsdom/living/webstorage/Storage-impl.js
+++ b/lib/jsdom/living/webstorage/Storage-impl.js
@@ -1,0 +1,98 @@
+"use strict";
+
+const DOMException = require("domexception");
+const StorageEvent = require("../generated/StorageEvent");
+const idlUtils = require("../generated/utils");
+
+// https://html.spec.whatwg.org/multipage/webstorage.html#the-storage-interface
+class StorageImpl {
+  constructor(args, { associatedWindow, storageArea, url, type }) {
+    this._associatedWindow = associatedWindow;
+    this._items = storageArea;
+    this._url = url;
+    this._type = type;
+
+    // The spec suggests a default storage quota of 5 MB
+    this._quota = 5000;
+  }
+
+  _dispatchStorageEvent(key, oldValue, newValue) {
+    return this._associatedWindow._currentOriginData.windowsInSameOrigin
+      .filter(target => target !== this._associatedWindow)
+      .forEach(target => target.dispatchEvent(StorageEvent.create([
+        "storage",
+        {
+          bubbles: false,
+          cancelable: false,
+          key,
+          oldValue,
+          newValue,
+          url: this._url,
+          storageArea: target["_" + this._type]
+        }
+      ])));
+  }
+
+  get length() {
+    return this._items.size;
+  }
+
+  key(n) {
+    if (n >= this._items.size) {
+      return null;
+    }
+    return [...this._items.keys()][n];
+  }
+
+  getItem(key) {
+    if (this._items.has(key)) {
+      return this._items.get(key);
+    }
+    return null;
+  }
+
+  setItem(key, value) {
+    const oldValue = this._items.get(key) || null;
+
+    if (oldValue === value) {
+      return;
+    }
+
+    // Concatenate all keys and values to measure their size against the quota
+    let itemsConcat = key + value;
+    this._items.forEach((v, k) => {
+      itemsConcat += v + k;
+    });
+    if (Buffer.byteLength(itemsConcat) > this._quota) {
+      throw new DOMException(`The ${this._quota} byte storage quota has been exceeded.`, "QuotaExceededError");
+    }
+
+    setTimeout(this._dispatchStorageEvent.bind(this), 0, key, oldValue, value);
+
+    this._items.set(key, value);
+  }
+
+  removeItem(key) {
+    if (this._items.has(key)) {
+      setTimeout(this._dispatchStorageEvent.bind(this), 0, key, this._items.get(key), null);
+
+      this._items.delete(key);
+    }
+  }
+
+  clear() {
+    if (this._items.size > 0) {
+      setTimeout(this._dispatchStorageEvent.bind(this), 0, null, null, null);
+
+      this._items.clear();
+    }
+  }
+
+  get [idlUtils.supportedPropertyNames]() {
+    return this._items.keys();
+  }
+}
+
+module.exports = {
+  implementation: StorageImpl
+};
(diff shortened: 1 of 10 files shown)


lib/api.js
```diff
@@ -239,6 +239,7 @@ function transformOptions(options, encoding) {
       runScripts: undefined,
       encoding,
       pretendToBeVisual: false,
+      storageToDisk: false,
```
Sebmaster: I'd avoid tying us down to files as a storage provider. Maybe it'd make more sense to allow a way to listen for storage update events and just provide those with domain + js object.

Zirro: Hmm, this was my interpretation of Domenic's comment in the issue - "For now we can keep everything in-memory, although eventually we should provide a way of storing to disk if people want that".

If we want something more complex I think I'll focus on just the internal implementation for this pull request, and let us decide on this later.

domenic: Right, I'd say for the initial implementation, both local and session storage are in-memory, and in the future, we could allow providing some kind of storage provider interface (which could also be used for HTTP caches, maybe!).

lib/jsdom/living/webstorage/Storage-impl.js
```diff
@@ -0,0 +1,104 @@
...
+    if (oldValue === value) {
+      return;
+    }
+
+    const storageSize = Buffer.byteLength(JSON.stringify(convertMapToObject(this._items)));
+    if (storageSize > this._quota) {
+      throw new DOMException(`The ${this._quota} byte storage quota has been exceeded.`, "QuotaExceededError");
+    }
+
+    this._taskQueue = setTimeout(this._dispatchStorageEvent.bind(this), 0, key, oldValue, value);
```
domenic: _taskQueue is set, but never used?

lib/jsdom/living/webstorage/Storage-impl.js
```diff
@@ -0,0 +1,104 @@
...
+const fs = require("pn/fs");
+const DOMException = require("domexception");
+const StorageEvent = require("../generated/StorageEvent");
+const idlUtils = require("../generated/utils");
+const { convertMapToObject } = require("../../utils");
+
+// https://html.spec.whatwg.org/multipage/webstorage.html#the-storage-interface
+class StorageImpl {
+  constructor(args, privateData) {
+    this._items = new Map();
```
domenic: What are the advantages of using a map over an object, if you're going to just convert it to an object later?

Zirro: I originally used a Map before I saw that there could be a need for stringification, knowing that (at least historically) the property order has not been entirely [reliable](https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order).

I did try an object and found while they performed about the same, the code became less clean/consistent (given `Object.keys(x)` vs `x.size` / `delete x[y]` vs `x.delete(y)` with the object, so I've kept the Map. Now that I've removed the writeToDisk-ability, stringification is not as important and I've opted for string concatenation of the keys + values to determine the storage size instead.

Zirro:
Updated to remove the writeToDisk-ability until we've settled on an API, and use string concatenation rather than stringification to determine the size of the data being stored. Will address `_taskQueue` when I work on the event issues later.

Zirro:
This is now ready for review 🎉

The only part I've knowingly not implemented is the `url` property for `StorageEvent`. It is supposed to be the URL for the context from where the Storage object was manipulated. Since an `<iframe>` can share the same `Storage` object as its parent I don't see a clear way to distinguish which context was the cause of the change. I'd appreciate suggestions for how to handle it, but otherwise I think it would be okay to place this property behind a comment in the IDL for now.

It would appear that one of the tests is using `Object.values()` which isn't available in Node.js v6, thus causing CI to fail. I'll put it behind the `needs-node8` reason for my next push.

Zirro:
It doesn't make for the prettiest code, but I think I've finally nailed the logic for storage areas with separate objects. This also made it possible to implement the `url` property, allowing the last relevant tests to pass :-)

Zirro:
Updated to add condition for opaque origins to `sessionStorage` as well, in anticipation of the spec being updated for https://github.com/whatwg/html/issues/3283.

Zirro:
I had time to do some testing in detail today and found more cases not covered by the existing tests. These include making sure that data is shared between a parent and several iframes within the same origin, as well as checking that the `StorageEvent` is dispatched on all `window`'s except the one through which the storage object was manipulated.

I've fixed the former issue, getting some neat refactoring done in the process. The latter issue is only partially resolved however - manipulations done on a parent window will no longer cause the `StorageEvent` to be dispatched there, but manipulating a storage object in a frame will still result in a `StorageEvent` on the `window` of both frame and parent windows. More work needed there.

Considering this and some gaps in the test coverage I found earlier I'm now hesitant to merge this without writing several new tests, so this could take a bit longer to complete.

domenic:
So I believe in IRC we decided to rebase this even without the full coverage. Maybe a good intermediate step would be to file an issue on the WPT repo detailing the missing coverage and your ideas of how you might test it?

Zirro:
I've rebased this to take the latest commits into account. Will file an issue in the WPT repo if I have time this weekend.

lib/jsdom/living/webstorage/Storage-impl.js
```diff
@@ -0,0 +1,98 @@
...
+// https://html.spec.whatwg.org/multipage/webstorage.html#the-storage-interface
+class StorageImpl {
+  constructor(args, { associatedWindow, storageArea, url, type }) {
+    this._associatedWindow = associatedWindow;
+    this._items = storageArea;
+    this._url = url;
+    this._type = type;
+
+    // The spec suggests a default storage quota of 5 MB
+    this._quota = 5000;
```
fstoerkle: @Zirro 
As I understand, 5000 would suggest that the default storage quota is 5 KB... shouldn't that be 5_000_000 bytes?

agjs:
What's the status on this? I need the storage to mock some tests and wondering if this is gonna be merged and released soon? Thanks

yogeshthanvi:
I am facing the below issue anyone can help me out 
SecurityError: localStorage is not available for opaque origins

      at Window.get localStorage [as localStorage] (../node_modules/jsdom/lib/jsdom/browser/Window.js:257:15)
          at Array.forEach (<anonymous>)

Zirro:
See: https://github.com/jsdom/jsdom/issues/2304#issuecomment-408320484

PR #2079, opened 2017-12-12
replace content-type-parser with whatwg-mimetype

graingert:


domenic:
Any idea why this causes so many test failures?

domenic:
Done in https://github.com/jsdom/jsdom/commit/d5639655d32d45b6690c9ac30901136e650eb99a

PR #2081, opened 2017-12-12
Update all the Web IDL - Part Two

Zirro:
Updates for the remaining Web IDL after https://github.com/tmpvar/jsdom/pull/2053 handles the nodes/ directory. An exception is `Attr.webidl`, which I've left unchanged as it depends on the work from https://github.com/tmpvar/jsdom/pull/1822.

### DIFF
--- a/lib/jsdom/living/domparsing/DOMParser.webidl
+++ b/lib/jsdom/living/domparsing/DOMParser.webidl
@@ -1,7 +1,7 @@
-[Constructor]
+[Constructor,
+  Exposed=Window]
 interface DOMParser {
-    [NewObject]
-    Document parseFromString(DOMString str, SupportedType type);
+  [NewObject] Document parseFromString(DOMString str, SupportedType type);
 };
 
 enum SupportedType {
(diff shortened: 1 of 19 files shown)


Zirro:
Mainly changes to formatting in this one, but here are other changes of potential consequence:

- Event
  The `timeStamp` attribute uses DOMHighResTimeStamp instead of DOMTimeStamp

- UIEvent
  No longer has a `modifierOS` boolean

- FileReader
  `error` attribute changed from using `DOMError` to `DOMException`

- History
  Third argument of `pushState()` and `replaceState()` takes a USVString instead of DOMString

domenic:
The timeStamp change also came with a spec change that might be worth implementing: https://github.com/whatwg/dom/commit/912d58746bc4c67c2fd210124639f7c34991fce5

Zirro:
Okay, I believe I can do that this weekend.

Zirro:
Updated to revert the DOMTimeStamp -> DOMHighResTimeStamp change for now since it appears to require many changes to event constructors and the code leading up to them.

PR #2082, opened 2017-12-13
Various additions and corrections for window

Zirro:
These are a collection of changes I made while exploring the accuracy of our `window` implementation.

* Implement the `BarProp` interface and the `window` attributes that use it.

* Implement the `Screen` interface used for `window.screen`, which was previously just an object with two properties instead of the six attributes it actually contains. Tested in `cssom-view-window-screen-interface.html` and `Screen-pixelDepth-Screen-colorDepth001.html`.

* Implement `window.devicePixelRatio` with the default value `1`.

* Implement `window.status` which exists for historical reasons.

* Expose the `External` interface.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -18,6 +18,7 @@ const BarProp = require("../living/generated/BarProp");
 const Document = require("../living/generated/Document");
 const External = require("../living/generated/External");
 const Navigator = require("../living/generated/Navigator");
+const Screen = require("../living/generated/Screen");
 const createAbortController = require("../living/generated/AbortController").createInterface;
 const createAbortSignal = require("../living/generated/AbortSignal").createInterface;
 const reportException = require("../living/helpers/runtime-script-errors");
@@ -149,6 +150,7 @@ function Window(options) {
   const toolbar = BarProp.create();
   const external = External.create();
   const navigator = Navigator.create([], { userAgent: options.userAgent });
+  const screen = Screen.create();
 
   define(this, {
     get length() {
@@ -204,6 +206,9 @@ function Window(options) {
     },
     get toolbar() {
       return toolbar;
+    },
+    get screen() {
+      return screen;
     }
   });
 
@@ -547,10 +552,6 @@ function Window(options) {
     scrollY: 0,
     scrollTop: 0,
     scrollLeft: 0,
-    screen: {
-      width: 0,
-      height: 0
-    },
 
     alert: notImplementedMethod("window.alert"),
     blur: notImplementedMethod("window.blur"),
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -538,6 +538,7 @@ function Window(options) {
 
   define(this, {
     name: "nodejs",
+    devicePixelRatio: 1,
     innerWidth: 1024,
     innerHeight: 768,
     outerWidth: 1024,
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -538,6 +538,9 @@ function Window(options) {
 
   define(this, {
     name: "nodejs",
+    // Node v6 has issues (presumably in the vm module)
+    // which this property exposes through an XHR test
+    // status: "",
     devicePixelRatio: 1,
     innerWidth: 1024,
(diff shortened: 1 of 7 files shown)


lib/jsdom/living/window/Screen-impl.js
```diff
@@ -0,0 +1,28 @@
+"use strict";
+
+// https://drafts.csswg.org/cssom-view-1/#the-screen-interface
+exports.implementation = class ScreenImpl {
+  get availWidth() {
+    return 0;
```
TimothyGu: These will never get changed, so I'd rather just do something like https://github.com/TimothyGu/jsdom/blob/window-webidl2js/lib/jsdom/living/window/Screen-impl.js

```js
class ScreenImpl {}
ScreenImpl.prototype.availWidth = 0;
ScreenImpl.prototype.availHeight = 0;
ScreenImpl.prototype.width = 0;
ScreenImpl.prototype.height = 0;
ScreenImpl.prototype.colorDepth = 0;
ScreenImpl.prototype.pixelDepth = 0;
```

lib/jsdom/living/window/Screen-impl.js
```diff
@@ -0,0 +1,28 @@
+"use strict";
+
+// https://drafts.csswg.org/cssom-view-1/#the-screen-interface
```
TimothyGu: I think we usually have such URL comments in the IDL file. This is totally fine by me, but just wanted to know if that policy changed..

Zirro: I believe they're mainly used for partials within IDL files. That said, I would not mind making it standard practice to have them in both IDL and the implementation. They're often handy for looking up implementation details in the current context.

lib/jsdom/living/window/Screen-impl.js
Zirro: Oh, I wasn't aware of how far you've gotten on the new implementation of window. The refactored version looks great!

Regarding this file in particular, I wouldn't mind making this change. I usually prioritise being consistent across files, but the advantages of making an exception here are pretty clear.

Sebmaster: I'd also prefer the prototype version for static values (wherever possible).

lib/jsdom/living/window/Screen-impl.js
Sebmaster: I think we usually do IDL source link in the IDL file, implementation detail links in impl.

lib/jsdom/browser/Window.js
```diff
@@ -538,6 +538,7 @@ function Window(options) {
 
   define(this, {
     name: "nodejs",
+    devicePixelRatio: 1,
```
domenic: Note that I think these should all be getters (own-getters, but still getters). Happy to continue with them in their current state for now though.

lib/jsdom/living/window/Screen-impl.js
domenic: I kind of like the consistency, given that these aren't going to be commonly accessed. But either way.

domenic:
It looks like the addition of `status` is causing XHR test failures on Node v6 due to some vm stuff. We should add needs-node8 to those tests.

Zirro:
Updated to address feedback. I suppose we'll have to decide on a final getter vs prototype-approach later on.

> It looks like the addition of status is causing XHR test failures on Node v6 due to some vm stuff. We should add needs-node8 to those tests.

Since `window.status` isn't that important, I've decided place it behind a comment with a note until we stop supporting Node v6. After all, if it causes issues in a test there is a risk that it could break real uses too.

lib/jsdom/living/index.js
```diff
@@ -42,6 +42,7 @@ exports.WheelEvent = require("./generated/WheelEvent").interface;
 exports.EventTarget = require("./generated/EventTarget").interface;
 
 exports.BarProp = require("./generated/BarProp").interface;
+exports.External = require("./generated/External").interface;
```
domenic: So actually `External` is `[NoInterfaceObject]` so we should not do this. I guess we should try to make use of the `exposed` thingy some day.

TimothyGu: > I guess we should try to make use of the `exposed` thingy some day.

Already have a WIP :)

Issue #2050, opened 2017-11-17
New API resource loader design


domenic:
Continued from https://github.com/tmpvar/jsdom/issues/2041#issuecomment-345143720; /cc @molant

These are my notes from last time I thought about this problem:

### New design

- A new class, `jsdom.ResourceLoader`.
- It has one method, `fetch(url, { headers, ...other stuff maybe })`
  - `url` must be an absolute URL, not relative.
  - I'm not sure what type `url` should be. In general I prefer public APIs to be strings, [per convention](https://url.spec.whatwg.org/#url-apis-elsewhere), but this might lead to a lot of spurious conversions of string -> URL -> string -> back to URL inside this API. We'll see, but lean toward string...
  - I'm not sure how much the callers should have to manually set headers, e.g. Referer and Accept, or whether we should allow them to be inferred from a Window that is part of the other stuff.
- The method returns a `Promise<Buffer | null>`. `null` indicates skipping that resource load. Callers must take care of decoding to strings according to the appropriate specs. ResourceLoader does not concern itself with decoding.
- To make the fetch abortable, for the first round we just tack on `.abort()` on the returned promise. Later, we make a breaking change to add `AbortSignal`, when we implement that in jsdom.
- Similarly, over time we can make breaking changes to make this use more Fetch API-like stuff, e.g. maybe `headers` starts out Node.js style, but later uses the web platform `Headers` class.
- jsdom's `resources` option accepts a `ResourceLoader` in addition to the string `"usable"`. Perhaps `"usable"` triggers a specific built-in `ResourceLoader` instance? It'd be good to figure out how to build a `ResourceLoader` that emulates the behavior of `"usable"`; this may inform some of the `...other stuff maybe` above.
- To get custom resource loading, you subclass `ResourceLoader`, override `fetch()`, and use `super.fetch()` as appropriate.
- This would also be a good place to add support for customizing the options to `request`; this capability was lost in the new API. I am thinking that becomes a property, `requestOptions`, and the default `fetch()` consults `this.requestOptions`. Thus you can create a subclass that just overrides `requestOptions`. Currently the defaults I have written down (which we should double-check with jsdom's code) are `{ strictSSL: true, pool: { maxSockets: 6 }, agentOptions: { keepAlive: true, keepAliveMsecs: 115 * 1000 } }`, which are supposed to roughly match browser behavior.

### How to implement

[...]

PR #2083, opened 2017-12-14
Add `originalRequest` to the object returned by the resource loader

tmeasday:
I'm not sure if you are still interested in taking changes to what is a deprecated API, but hopefully this is simple enough to be uncontroversial.

The use case is to allow me to use the request object to track the progress of long-lived resource downloads (i.e. show a progress bar).


domenic:
Yeah, sorry, we don't want to add any stuff to the old API. It's good to hear of that use case though; we'll be sure to keep it in mind when designing the new resource loading API (#2050).

PR #2085, opened 2017-12-16
Give failing tests on Travis a second chance

Zirro:
Since we continue to see spurious failures on Travis for reasons which may be out of our control, it could help to give them a retry.


domenic:
It's sad this is necessary but I guess we should do it.

PR #2091, opened 2017-12-24
allow pass more option for request of fromURL

bluelovers:
let user without re clone full fromURL just only wanna add more option on request


TimothyGu:
From the [README](https://github.com/tmpvar/jsdom#fromurl):

> The initial request is not infinitely customizable to the same extent as is possible in a package like [request](https://www.npmjs.com/package/request); `fromURL()` is meant to be a convenience API for the majority of cases. If you need greater control over the initial request, you should perform it yourself, and then use the `JSDOM` constructor manually.

bluelovers:
i know that, but i still wanna can allow pass options
when we know what we doing

domenic:
You should use the `request()` package or similar then pass the result to `new JSDOM()`, as the README asks you to.

PR #2088, opened 2017-12-20
Implement WebSocket API

TimothyGu:
Depends on w3c/web-platform-tests#8746 to be merged and rolled before this could be applied.

Fixes #1195. Closes #2075.

Travis CI will fail due to the tests not being current.

PR #2092, opened 2017-12-25
Use new webidl2js bundle-entry.js feature

TimothyGu:
Uses the feature introduced in https://github.com/jsdom/webidl2js/pull/101. Skipping CI for now as the change has not yet been merged, much less published, in webidl2js.

[ci skip]


lib/jsdom/living/nodes/Document.webidl
```diff
@@ -1,5 +1,6 @@
 [Constructor,
- Exposed=Window]
+ Exposed=Window,
+ LegacyWindowAlias=HTMLDocument]
```
TimothyGu: See https://github.com/whatwg/dom/pull/551.

Zirro:
This looks like a great improvement over the previous code. The comment you added makes me wonder, though...

> the "core" here is the "extraDOM" object imported from living/index.js, not the core tied to a specific Window

Maybe we could take this as an opportunity to give these objects names that better clarify their purpose?

While I'm familiar enough with the codebase to know what they are, having two different `core`'s (and something named `extraDOM`) is probably not the most obvious naming for new contributors.

Perhaps the `core`-names are no longer appropriate for the code as it is currently structured at all?

TimothyGu:
@Zirro I wholeheartedly agree.

I plan on coming back to this after getting #2088 merged (review welcome!). I'll also think of how to integrate some good parts from my [TimothyGu/jsdom:`vm-hah`](https://github.com/TimothyGu/jsdom/tree/vm-hah) branch, as this change (and the whole `bundle-entry.js` concept) originated from that branch, though the full clean-globals-for-every-Window probably feature will probably come after this PR is merged.

Sebmaster:
So we're still sharing prototypes with this change, but have a proper window object in all class instances?

TimothyGu:
No, this doesn’t change the status quo regarding that: all prototypes are shared, most classes are window-agnostic. It only simplifies the code as it is. The passed `window` is for the few webidl2js-enabled interfaces that require `window` to be known. But yes, this API is what I expect how `window`-specific prototypes will eventually look like.

lib/jsdom/browser/Window.js
```diff
@@ -60,15 +57,27 @@ function Window(options) {
...
       configurable: true,
       writable: true,
-      value: dom[name]
+      value: extraDOM[name]
     });
+
+    this._core[name] = extraDOM[name];
   }
-  this._core = dom;
+  // TODO move this somewhere else.
```
domenic: Running this on every window initialization seems sad since it's all a shared object.

lib/jsdom/browser/Window.js
```diff
@@ -36,14 +33,14 @@ const WindowEventHandlersImpl = require("../living/nodes/WindowEventHandlers-imp
 // NB: the require() must be after assigning `module.exports` because this require() is circular
 // TODO: this above note might not even be true anymore... figure out the cycle and document it, or clean up.
 module.exports = Window;
-const dom = require("../living");
+const bootstrapDOM = require("../living/generated/bundle-entry").bootstrap;
+const extraDOM = require("../living");
```
domenic: I'd like us to carefully consider our names here. E.g. instead of "extraDOM", maybe "nonGeneratedInterfaces" or similar.

lib/jsdom/browser/Window.js
```diff
@@ -81,7 +90,7 @@ function Window(options) {
 
   // List options explicitly to be clear which are passed through
   this._document = Document.create([], {
-    core: dom,
+    core: this._core,
```
domenic: Shouldn't be necessary after #2098.

lib/jsdom/browser/Window.js
```diff
@@ -60,15 +57,27 @@ function Window(options) {
   ///// INTERFACES FROM THE DOM
   // TODO: consider a mode of some sort where these are not shared between all DOM instances
   // It'd be very memory-expensive in most cases, though.
-  for (const name in dom) {
+  this._core = bootstrapDOM("Window", window, { window });
```
domenic: We no longer store `_core` on Window after #2098. Not sure how that fits in here.

lib/jsdom/living/register-elements.js
```diff
@@ -355,6 +355,8 @@ const mappings = {
   }
 };
 
+// N.B. the "core" here is the "extraDOM" object imported from living/index.js, not the core tied to a specific Window.
+// Thus, it only contains the limited set of objects in living/index.js, but that is enough currently.
 module.exports = core => {
```
domenic: Let's just update the parameter name then, instead of creating a comment explaining why it's misleading.

lib/jsdom/living/nodes/Document.webidl
ExE-Boss: Note that this might no longer be correct because of https://github.com/whatwg/html/issues/4792.

ExE-Boss: Well whatwg/dom#551 was closed, so this change should be reverted.

PR #2094, opened 2017-12-26
Implement window.performance

Zirro:
This pull request implements the `Performance` interface and `window.performance` as defined by the [High Resolution Time](https://w3c.github.io/hr-time/#the-performance-interface) specification, with the exception of the default toJSON operation as it is not supported by webidl2js yet.

I use `Date.now()` to obtain the "time origin timestamp" as I am not aware of another way to obtain a reference to an absolute time without native code.

This resolves https://github.com/tmpvar/jsdom/issues/1510.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -2,6 +2,7 @@
 
 const webIDLConversions = require("webidl-conversions");
 const { CSSStyleDeclaration } = require("cssstyle");
+const { Performance: RawPerformance } = require("w3c-hr-time");
 const notImplemented = require("./not-implemented");
 const VirtualConsole = require("../virtual-console");
 const { define, mixin } = require("../utils");
@@ -20,6 +21,7 @@ const BarProp = require("../living/generated/BarProp");
 const Document = require("../living/generated/Document");
 const External = require("../living/generated/External");
 const Navigator = require("../living/generated/Navigator");
+const Performance = require("../living/generated/Performance");
 const Screen = require("../living/generated/Screen");
 const createAbortController = require("../living/generated/AbortController").createInterface;
 const createAbortSignal = require("../living/generated/AbortSignal").createInterface;
@@ -53,6 +55,7 @@ dom.Window = Window;
 function Window(options) {
   EventTarget.setup(this);
 
+  const rawPerformance = new RawPerformance();
   const windowInitialized = hrtime();
 
   this._initGlobalEvents();
@@ -150,6 +153,7 @@ function Window(options) {
   const toolbar = BarProp.create();
   const external = External.create();
   const navigator = Navigator.create([], { userAgent: options.userAgent });
+  const performance = Performance.create([], { rawPerformance });
   const screen = Screen.create();
 
   define(this, {
@@ -206,6 +210,9 @@ function Window(options) {
     },
     get toolbar() {
       return toolbar;
+    },
+    get performance() {
+      return performance;
     },
     get screen() {
       return screen;
--- a/lib/jsdom/living/hr-time/DOMHighResTimeStamp.webidl
+++ b/lib/jsdom/living/hr-time/DOMHighResTimeStamp.webidl
@@ -1,0 +1,1 @@
+typedef double DOMHighResTimeStamp;
--- a/lib/jsdom/living/hr-time/Performance-impl.js
+++ b/lib/jsdom/living/hr-time/Performance-impl.js
@@ -1,0 +1,25 @@
+"use strict";
+
+const EventTargetImpl = require("../events/EventTarget-impl").implementation;
+
+class PerformanceImpl extends EventTargetImpl {
+  constructor(args, privateData) {
+    super(args, privateData);
+
+    this._rawPerformance = privateData.rawPerformance;
+  }
+
+  now() {
+    return this._rawPerformance.now();
+  }
+
+  get timeOrigin() {
+    return this._rawPerformance.timeOrigin;
+  }
+
+  toJSON() {
+    return this._rawPerformance.toJSON();
+  }
+}
+
+exports.implementation = PerformanceImpl;
--- a/test/web-platform-tests/to-run.yaml
+++ b/test/web-platform-tests/to-run.yaml
@@ -198,6 +198,15 @@ xml-serialization.xhtml: [fail, Unknown]
 
 ---
 
+DIR: hr-time
+
+idlharness.html: [fail, ReferenceError - WebIDL2 is not defined]
+test_cross_frame_start.html: [fail, Not implemented]
+timeOrigin.html: [fail, Needs Worker implementation]
+window-worker-timeOrigin.window.html: [fail, Needs Worker implementation]
+
+---
+
 DIR: html/browsers/browsing-the-web/history-traversal
 
 001.html: [fail, Unknown]
--- a/lib/jsdom/living/hr-time/Performance.webidl
+++ b/lib/jsdom/living/hr-time/Performance.webidl
@@ -1,0 +1,6 @@
+[Exposed=(Window,Worker)]
+interface Performance : EventTarget {
+    DOMHighResTimeStamp now();
+    readonly attribute DOMHighResTimeStamp timeOrigin;
+    [Default] object              toJSON();
+};
--- a/lib/jsdom/living/index.js
+++ b/lib/jsdom/living/index.js
@@ -46,6 +46,7 @@ exports.BarProp = require("./generated/BarProp").interface;
 exports.Location = require("./generated/Location").interface;
 exports.History = require("./generated/History").interface;
 exports.Screen = require("./generated/Screen").interface;
+exports.Performance = require("./generated/Performance").interface;
 
 exports.Blob = require("./generated/Blob").interface;
 exports.File = require("./generated/File").interface;
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "sax": "^1.2.4",
     "symbol-tree": "^3.2.2",
     "tough-cookie": "^2.3.3",
+    "w3c-hr-time": "^1.0.0",
     "webidl-conversions": "^4.0.2",
     "whatwg-encoding": "^1.0.3",
     "whatwg-url": "^6.4.0",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "sax": "^1.2.4",
     "symbol-tree": "^3.2.2",
     "tough-cookie": "^2.3.3",
-    "w3c-hr-time": "^1.0.0",
+    "w3c-hr-time": "^1.0.1",
     "webidl-conversions": "^4.0.2",
     "whatwg-encoding": "^1.0.3",
     "whatwg-url": "^6.4.0",
--- a/scripts/webidl/convert.js
+++ b/scripts/webidl/convert.js
@@ -30,6 +30,7 @@ addDir("../../lib/jsdom/living/domparsing");
 addDir("../../lib/jsdom/living/svg");
 addDir("../../lib/jsdom/living/aborting");
 addDir("../../lib/jsdom/living/websockets");
+addDir("../../lib/jsdom/living/hr-time");
 
 
 const outputDir = path.resolve(__dirname, "../../lib/jsdom/living/generated/");
--- a/test/web-platform-tests/to-upstream/hr-time/timeOrigin-dont-upstream.html
+++ b/test/web-platform-tests/to-upstream/hr-time/timeOrigin-dont-upstream.html
@@ -1,0 +1,45 @@
+<!doctype html>
+<html>
+<head>
+<script src="/resources/testharness.js"></script>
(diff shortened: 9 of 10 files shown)


TimothyGu:
What's wrong with using `process.hrtime()` for `timeOrigin`?

Zirro:
If I've interpreted the specification correctly, `timeOrigin` relies on [Unix time](https://en.wikipedia.org/wiki/Unix_time) and thus uses absolute time. From what I can tell, `process.hrtime()` only works with relative time.

TimothyGu:
This probably better illustrates what I'm saying.

```js
// This t1 only needs to be calculated once, since process.hrtime() is guaranteed to be
// monotonically increasing, and can be used in all subsequent Window instances.
const t1 = getT1();

function getT1() {
  // Wait at most 1 millisecond, a negligible amount of time for initialization.
  let cur = Date.now();
  const next = cur + 1;
  while (cur !== next) {
    cur = Date.now();
  }

  // At this point cur "just" became equal to the next millisecond -- the unseen digits after cur
  // are approximately all 0, and cur is the closest to the actual value of the UNIX time.
  // Now, get the current global monotonic clock value:

  const [hrSec, hrNS] = process.hrtime();
  const globalMonotonicClockMS = hrSec * 1e3 + hrNS / 1e6;

  // Let |t1| be the DOMHighResTimeStamp representing the high resolution Unix time at
  // which the global monotonic clock is zero.
  // I.e., t1 + globalMonotonicClockMS = cur    =>    t1 = cur - globalMonotonicClockMS
  return cur - globalMonotonicClockMS;
}


function Window() {
  // ...

  // Let |t2| be the DOMHighResTimeStamp representing the high resolution time value
  // of the global monotonic clock at global's time origin.
  const [hrSec2, hrNS2] = process.hrtime();
  const t2 = hrSec * 1e3 + hrNS / 1e6;;

  // Return the sum of |t1| and |t2|.
  const timeOriginTimestamp = t1 + t2;

  // ...
}
```

The `getT1()` function is very stable (i.e. the returned value is pretty accurate). If we want to be very very accurate we could discard the first result of `getT1()` (let V8 compile the function first) and/or take the average of ten runs or so. 10ms is still very little time for one-time initialization.

```js
> getT1()
1514269990803.6875
> getT1()
1514269990803.937
> getT1()
1514269990803.9385
> getT1()
1514269990803.9387
> getT1()
1514269990803.9382
> getT1()
1514269990803.9385
> getT1()
1514269990803.9387
> getT1()
1514269990803.9387
> getT1()
1514269990803.9392
> getT1()
1514269990803.9392
> getT1()
1514269990803.9392
```

Regarding the accuracy: it's pretty accurate. Two sources of error in the process of calculating `t1` are the 1ms loop and the time it takes for `process.hrtime()` to execute. On my machine, V8 can execute anywhere between 8000-10000 iterations of the loop within 1ms on the initial run, and after TurboFan kicks in and JIT's the function it can execute 10000-18000 iterations. I.e., the difference between `cur` and the actual time when we get `process.hrtime()` would always be less than 0.0002ms. `process.hrtime()`, once warmed up, takes just a bit over 0.001ms to execute as measured by `process.hrtime(process.hrtime())[1] / 2 / 1e6`, and this difference dwarfs the 0.0002ms imprecision of the loop. A 0.001ms (i.e. 1 microsecond) precision of `t1` is pretty darn good, even when it is multiplied by two to account for the imprecision of calculating `t2`, when the spec says we should aim for 5 microseconds.

lib/jsdom/living/hr-time/DOMHighResTimeStamp.webidl
```diff
@@ -0,0 +1 @@
+typedef double DOMHighResTimeStamp;
```
TimothyGu: FYI this doesn't have to be in a file of its own, though it certainly can.

TimothyGu:
https://github.com/nodejs/node/issues/17893#issuecomment-354383863

> It assumes `Date.now()` and `process.hrtime()` are near-instantaneous, which they usually are but not always. If the system clock is emulated (virtualized), you'll miss the mark by many microseconds, maybe by even more than a millisecond.

:cry:

Zirro:
Given the recent discussions, how do we want to proceed with this? Should I implement your solution @TimothyGu, or keep the current one? Or should we wait until we have the accuracy needed from Node.js before we add support for `window.performance` at all?

TimothyGu:
For most of the cases, `process.hrtime()` and `Date.now()` *are* fast enough. It would be a shame to not allow using them when they are.

My recommendation is the following. It is pretty verbose, but should be as spec compliant as anything can be at this point.

<details><summary>Clock accuracy checker + rough Performance implementation</summary>

```js
// |t1| only needs to be calculated once, since process.hrtime() is guaranteed to be monotonically increasing, and can
// be used in all subsequent Window instances. However, if the clock is not fast enough, set |t1| to undefined to
// signify that we should use Date.now() to get the time origin timestamp with millisecond accuracy, per spec.
const t1 = (() => {
  // Warm up the function.
  testClockAccuracy();
  testClockAccuracy();
  if (!testClockAccuracy()) {
    return undefined;
  }

  // Warm up the function.
  getT1();
  getT1();
  getT1();
  let sum = 0;
  for (let i = 0; i < 5; i++) {
    sum += getT1();
  }
  return sum / 5;
})();

function toMS([sec, nanosec]) {
  return sec * 1e3 + nanosec / 1e6;
}

// Spec calls for 5-μs accuracy. Check that we have that in both process.hrtime() and Date.now().
function testClockAccuracy() {
  // Test process.hrtime() first. The check is simpler and more stable, and we use process.hrtime() to measure
  // Date.now()'s performance.
  const roundTrip = process.hrtime(process.hrtime());
  if (roundTrip[0] > 1 || roundTrip[1] > 5e3 * 2) {
    return false;
  }

  // Test Date.now() twice: first with a looser bound (10 μs) but with a smaller run time to filter out very bad
  // Date.now() performance, and then with a tighter bound (5 μs) to check we have the accuracy we need.
  let times;
  let cur;
  let start;
  let end;

  times = 100;
  start = process.hrtime();
  while (times-- > 0) {
    cur = Date.now();
  }
  end = process.hrtime(start);
  if ((end[0] * 1e9 + end[1]) > 1000000) {
    return false;
  }

  times = 10000;
  start = process.hrtime();
  while (times-- > 0) {
    cur = Date.now();
  }
  end = process.hrtime(start);
  if ((end[0] * 1e9 + end[1]) > 50000000) {
    return false;
  }

  return true;
}

function getT1() {
  const start = Date.now();
  let cur = start;
  while (cur === start) {
    cur = Date.now();
  }

  // At this point cur "just" became equal to the next millisecond -- the unseen digits after cur are approximately all
  // 0, and |cur| is the closest to the actual value of the UNIX time. Now, get the current global monotonic clock
  // value:
  const globalMonotonicClockMS = toMS(process.hrtime());

  // Let |t1| be the DOMHighResTimeStamp representing the high resolution Unix time at which the global monotonic clock
  // is zero.
  // I.e., t1 + globalMonotonicClockMS = cur    =>    t1 = cur - globalMonotonicClockMS
  return cur - globalMonotonicClockMS;
}

class Window {
  constructor() {
    // ...

    if (t1 !== undefined) {
      // Clock is accurate enough.

      // Used in performance.now().
      const timeOrigin = this._timeOrigin = process.hrtime();

      // Let |t2| be the DOMHighResTimeStamp representing the high resolution time value of the global monotonic clock
      // at global's time origin.
      const t2 = toMS(timeOrigin);

      // Return the sum of |t1| and |t2|.
      this._timeOriginTimestamp = t1 + t2;
    } else {
      // Clock isn't accurate enough. Use millisecond accuracy per spec.
      this._timeOrigin = this._timeOriginTimestamp = Date.now();
    }

    // ...
  }

  // Methods below should go into PerformanceImpl class. They are only in Window for easier testing.

  get timeOrigin() {
    return this._timeOriginTimestamp;
  }

  now() {
    if (t1 !== undefined) {
      return toMS(process.hrtime(this._timeOrigin));
    } else {
      return Date.now() - this._timeOriginTimestamp;
    }
  }
}
```
</details>

This code works well for both fast-clock and slow-clock systems, with a graceful fallback to `Date.now()` per [spec](https://w3c.github.io/hr-time/#dom-domhighrestimestamp):

> If the User Agent is unable to provide a time value accurate to 5 microseconds due to hardware or software constraints, the User Agent can represent a `DOMHighResTimeStamp` as a time in milliseconds accurate to a millisecond.

To emulate slower-clocked systems, I used the following code:

<details><summary>Emulation code</summary>

```js
const realHrtime = process.hrtime;

Date.now = delay(Date.now);
process.hrtime = delay(process.hrtime);

function delay(func) {
  return (...args) => {
    const delayNS = 1e6; // 1 ms delay in clock
    const start = realHrtime();
    let duration = realHrtime(start);
    while (duration[0] < 1 && duration[1] < delayNS) {
      duration = realHrtime(start);
    }
    return func(...args);
  };
}
```
</details>

And to test how well it works, the following is used:

<details><summary>Tester</summary>

```js
console.log("t1", t1);
const w = new Window();
console.log("timeOrigin", w.timeOrigin, new Date(w.timeOrigin));
setTimeout(() => {
  console.log(w.now());
}, 1000);
```
</details>

With fast clocks (the default), the following is shown:

```
t1 1514556060333.1968
timeOrigin 1514806602744.6213 2018-01-01T11:36:42.744Z
1003.210213
```

while with slower clocks (the delay code activated):

```
t1 undefined
timeOrigin 1514806783489 2018-01-01T11:39:43.489Z
1004
```

Zirro:
Rebased to use @TimothyGu's new [w3c-hr-time](https://github.com/jsdom/w3c-hr-time) package, which contains logic to achieve the most accurate timings currently possible with Node.js.

I had to figure out a second name for the `Performance` class from the dependency since there are now two of them in Window.js. Feel free to suggest other names if you think they'd be more appropriate.

lib/jsdom/living/hr-time/Performance-impl.js
```diff
@@ -0,0 +1,25 @@
...
+  now() {
+    return this._performance.now();
+  }
+
+  get timeOrigin() {
+    return this._performance.timeOrigin;
+  }
+
+  toJSON() {
+    return this._performance.toJSON();
```
domenic: Shouldn't this return the timeOrigin? That's what [Default] would imply. (One day we can implement [Default] in webidl2js.)

test/web-platform-tests/to-run.yaml
```diff
@@ -287,6 +287,15 @@ xml-serialization.xhtml: [fail, Unknown]
 
 ---
 
+DIR: hr-time
+
+idlharness.html: [fail, ReferenceError - WebIDL2 is not defined]
```
domenic: One day we should figure out what this is about. I guess it's probably script execution stuff :(

lib/jsdom/browser/Window.js
```diff
@@ -2,6 +2,7 @@
 
 const webIDLConversions = require("webidl-conversions");
 const { CSSStyleDeclaration } = require("cssstyle");
+const { Performance: PerformanceClass } = require("w3c-hr-time");
```
domenic: I'd call it maybe "RawPerformance" or "W3CHRTimePerformance". Then you can reuse this name (lowercased) for the variable/field name.

lib/jsdom/living/hr-time/Performance-impl.js
Zirro: Seems like Timothy looked into this: https://github.com/jsdom/w3c-hr-time/blob/master/lib/performance.js#L46

domenic: Hmm seems like either a spec bug or a Firefox bug, interesting.

lib/jsdom/browser/Window.js
Zirro: `RawPerformance` 👍

lib/jsdom/living/hr-time/Performance-impl.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
+
+const EventTargetImpl = require("../events/EventTarget-impl").implementation;
+
+class PerformanceImpl extends EventTargetImpl {
```
TimothyGu: My original plan was to make it so that it's possible to directly use the `Performance` class in w3c-hr-time as a webidl2js implementation, but seems like I forgot about `EventTaget` :/ This works well enough I guess.

test/web-platform-tests/to-run.yaml
Zirro: I had a quick look and the cause seems to be: https://github.com/w3c/web-platform-tests/issues/5608

The redirect for `/resources/WebIDLParser.js` to `/resources/webidl2/lib/webidl2.js` doesn't seem to occur with the way we launch the WPT server, as the test works when I use the actual path.

Maybe it can be resolved by modifying `wpt-config.json`?

domenic: I guess it wouldn't work anyway since these tests use `fetch()`, so not worth bothering with for now.

PR #2097, opened 2017-12-31
Set Document's default encoding to UTF-8

Zirro:
This is used for `new Document()`, where `document.characterSet` would previously be undefined. The spec says that "unless stated otherwise, a document's encoding is the utf-8 encoding".

The `dom/nodes/Document-constructor.html` test still fails one section since the `origin` property should inherit from the constructor's context, so I have added a dont-upstream test.


lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -205,6 +205,9 @@ class DocumentImpl extends NodeImpl {
     if (!privateData.options.parsingMode) {
       privateData.options.parsingMode = "xml";
     }
+    if (!privateData.options.encoding) {
+      privateData.options.encoding = "UTF-8";
```
TimothyGu: Does capitalization make a difference? If not I think `"utf-8"` is the more prevalent style.

test/web-platform-tests/to-run.yaml
```diff
@@ -202,9 +202,9 @@ CharacterData-appendChild.html: [fail, Unknown]
 Comment-constructor.html: [fail, Unknown]
 DOMImplementation-createDocument.html: [fail, Unknown]
 Document-URL.sub.html: [fail, Unknown]
-Document-characterSet-normalization.html: [fail, Unknown]
+Document-characterSet-normalization.html: [fail, Some encodings are not supported - see the whatwg-encoding module]
```
TimothyGu: Hmm, so is it a WPT bug that it's using encodings that are not in the Encoding Standard?

lib/jsdom/living/nodes/Document-impl.js
Zirro: Indeed, I also prefer `utf-8` but the test expects `UTF-8` for `document.characterSet` and this appears to be what browsers do as well.

The [spec](https://encoding.spec.whatwg.org/#names-and-labels) suggests that new protocols/formats should use the form `utf-8`, though.

test/web-platform-tests/to-run.yaml
Zirro: They are in the Encoding Standard, but `iconv-lite` which is used by `whatwg-encoding` [does not support](https://github.com/jsdom/whatwg-encoding#unsupported-encodings) all of them yet.

lib/jsdom/living/nodes/Document-impl.js
TimothyGu: Ah. The [spec for `characterSet`](https://dom.spec.whatwg.org/#dom-document-characterset) says it should "return context object’s encoding’s [name](https://encoding.spec.whatwg.org/#name)," which for UTF-8 is indeed UTF-8. `characterSet` definitely isn't by any means a new API, so UTF-8 it is.

PR #2108, opened 2018-01-10
virtualConsole.sendTo(c) clarified c

BebeSparkelSparkel:


domenic:
Thanks, but this code is not syntactically valid, so I don't think we can accept it.

PR #2111, opened 2018-01-12
Enable old custom resource loader in new API

sarvaje:
ref: #2050 

This PR just enables the old custom resource loader in the new API.

This PR doesn't intent to replace the need of a new design for the
resource loader, but can help to some people to update to the new API.


domenic:
Sorry, we'd rather not do this, as it will cause a transition period between new API and new-new API when we get the resource loader fixed.

benorgera:
When will this functionality be available again? The ability to run external scripts is much more powerful with control over resources (caching, only loading certain scripts...)

Issue #2128, opened 2018-01-26
SVG basic geometric shapes (such as SVGPathElement) are not implemented


marinho:
This is a follow up of https://github.com/jsdom/jsdom/issues/2126, as it wasn't properly formatted.

### Basic info:

- **Node.js version:** v8.9.0
- **jsdom version:** 11.6.1

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const options = {
};
const dom = new JSDOM(`
  <svg width="100%" height="100%" viewBox="0 0 400 400" xmlns="http://www.w3.org/2000/svg">

    <path d="M 100 100 L 300 100 L 200 300 z" fill="orange" stroke="black" stroke-width="3" />
    <rect x="10" y="10" width="100" height="100"/>
    <circle cx="100" cy="100" r="100"/>
    <ellipse cx="60" cy="60" rx="50" ry="25"/>
    <line x1="20" y1="100" x2="100" y2="20" stroke-width="2" stroke="black"/>
    <polyline points="0,40 40,40 40,80 80" fill="white" stroke="#D07735" stroke-width="6" />
    <polygon points="60,20 100,40 100,80 60,100 20,80 20,40"/>

  </svg>
`, options);

console.log(dom.window.document.firstChild.querySelector('path'));
```

Current output:

```
> SVGElement {}
```

How it should be:

```
> SVGPathElement {}
```

That's because the following SVG basic geometry shapes are not implemented:

- `SVGCircleElement`
- `SVGEllipseElement`
- `SVGLineElement`
- `SVGPathElement`
- `SVGPolygonElement`
- `SVGPolylineElement`
- `SVGRectElement`

### How does similar code behave in browsers?

They work as expected in SVG specs, by showing their respective graphic shapes.

I'm going to file a pull request as a follow up to this issue.

Undistraction:
Just ran into this. It isn't just shapes that are missing, but also others like `SVGAnimatedLength`.

felixfbecker:
I personally need `SVGAElement` (equivalent of `HTMLAnchorElement`) because I want to test code that needs to handle both `SVGAElement` and `HTMLAnchorElement`

ulfaslak:
**Bump** on this. None of the workarounds people suggest in this and related issues work for me. I'm writing test (using Jest) for some code that does SVG processing. Until common SVG geometries such as `SVGPathElement`, `SVGPolygonElement`, etc. are supported in JSDOM, I don't know how to do this.

[...]

PR #2129, opened 2018-01-26
Stub implementation of SVG basic geometric shapes

marinho:
This PR is aimed to solve https://github.com/jsdom/jsdom/issues/2128, mostly as a stub implemention, as it doesn't include full implementation for the given classes.

Instead, it provides the SVG geometric classes and their dependencies, including all attributes and methods as defined in specs, even though such methods have empty code.

This is strictly following the [official specs](https://www.w3.org/TR/SVG/), but as it doesn't provide much information about most of those classes, I also follow [Mozilla WebIDL files](https://dxr.mozilla.org/mozilla-beta/source/dom/webidl/), excluding Mozilla specific attributes.

I didn't enable **DIR: svg** entry in `tu-run.yaml`, as they don't pass, raising an error for missing **WebIDL2**, and they are unlike to pass anyway as my methods are empty.

I splitted the work into 2 commits, as I'm not sure if the attributes implementation based on Mozilla WebIDL is desirable. Please let me know if I should remove them.

It can be seen as a continuation of https://github.com/jsdom/jsdom/pull/2011


marinho:
It seems that errors happening on `referrer.any.html` and `security/001.html` aren't making Travis to fail, right?

So I guess it's caused by linting errors at the beginning of the output?

I'm going to take care of them.

TimothyGu:
Also just in general, we always want to follow the editor's draft of specifications, since that is what browser developers actually look at when implementing features. In other words, URLs that look like `https://www.w3.org/TR/` are pretty much never appropriate. `https://dev.w3.org/` and `https://w3c.github.io` however, are generally fine, except for [HTML](https://annevankesteren.nl/2016/01/film-at-11) and [some other specs maintained by WHATWG](https://wiki.whatwg.org/wiki/Fork_tracking).

marinho:
thanks for the review. I'm going to make corrections and push it again soon!

marinho:
@TimothyGu one question on empty interfaces missing implementation, what's the general guidance then?

Does it make sense to keep tagged element interfaces (i.e. `SVGPathElement`, `SVGRectElement`, etc.) with their attributes and comment their methods out?

Regarding to interfaces which aren't tagged elements (i.e. `SVGAnimatedLength`), should I replace them by some wildcard type, comment them out or implement them properly?

thanks

TimothyGu:
> Does it make sense to keep tagged element interfaces (i.e. `SVGPathElement`, `SVGRectElement`, etc.) with their attributes and comment their methods out?

If you don't want to implement the methods, then yes comment them out in the IDL files.

If the attributes work, then sure, keeping them is fine. Otherwise, comment them out too.

> Regarding to interfaces which aren't tagged elements (i.e. `SVGAnimatedLength`), should I replace them by some wildcard type, comment them out or implement them properly?

If you have the time and ability to implement them, then implement them properly.
 (`SVGAnimatedLength` is one such class where the things needed to be implemented are probably too plentiful to worth the effort.) If not, then remove them, and comment out the attributes in the IDL for Element classes that use them.

domenic:
In general I would prefer we not include stub implementations at all. Unless we implement a significant portion of a class, it's better to just leave it as SVGElement, as that indicates more clearly to consumers that it's not supported.

It would help if there was a prose explanation of this change in terms of what features it implements.

marinho:
@domenic, I was thinking that too after I thought on @TimothyGu comments.

My work plan is to get rid of `SVGAnimatedPathData`, `SVGPathSeg`, `SVGPathSegList` and related attributes and implement the rest as much as possible. I'm not sure what challenges I will face though (still couldn't work on it). I will probably have an update at the end of next week.

marinho:
@domenic and @TimothyGu 

I started some implementation, but I'd like to have your feedback before I go ahead. Please take a look at the latest 3 commits I made (I didn't touch on the 3 commits that were there before), considering the following concerns:

- I didn't update it to latest jsdom master yet
- I added 2 new dependencies (d3-polygon and svg-pathdata
- wrote unit tests (tuwpt) instead of enabling tests based of wpt (for the moment)
- Although path data is no longer exposed by Chrome SVG API, the SVG 2 draft state that all SVG shapes still resolve into paths, so that `getPathData` is present only for internal matter and I plan to hide it as a private method
- The SVG present in wpt tests is not easy to cover, as it's more focused on transitions, animations and other features. I'm going to investigate how to enable some of them

If you give a positive feedback, I'm going to implement the basic methods according to SVG2 draft for these shapes, but I would like to know if it's better push it as smaller steps or wait until I have all the shapes ready.

PS: By "shapes", I mean these elements:

- `SVGCircleElement`
- `SVGEllipsisElement`
- `SVGLineElement`
- `SVGPolygonElement`
- `SVGPolylineElement`
- `SVGPathElement`
- `SVGRectElement`

best,

lib/jsdom/living/register-elements.js
```diff
@@ -356,6 +356,10 @@ const mappings = {
       file: require("./generated/SVGGeometryElement.js"),
       tags: []
     },
+    SVGLength: {
```
domenic: This isn't an element so doesn't belong here.

test/web-platform-tests/to-upstream/svg/svgrectelement.html
```diff
@@ -0,0 +1,73 @@
...
+  assert_true(rect instanceof SVGElement);
+  assert_true(rect instanceof SVGGraphicsElement);
+  assert_true(rect instanceof SVGGeometryElement);
+  assert_true(rect instanceof SVGRectElement);
+  assert_false(rect instanceof HTMLElement);
+}, "Basic properties");
+
+test(() => {
+  const rect = document.querySelector("#aRect");
+  const pathData = rect.getPathData();
```
domenic: This isn't in any spec so shouldn't be tested in WPTs.

test/web-platform-tests/to-upstream/svg/svgpathelement.html
```diff
@@ -0,0 +1,66 @@
...
+  assert_true(path instanceof SVGElement);
+  assert_true(path instanceof SVGGraphicsElement);
+  assert_true(path instanceof SVGGeometryElement);
+  assert_true(path instanceof SVGPathElement);
+  assert_false(path instanceof HTMLElement);
+}, "Basic properties");
+
+test(() => {
+  const path = document.querySelector("#aPath");
+  const pathData = path.getPathData();
```
domenic: This isn't in any spec so shouldn't be tested in WPTs

seblucas:
Don't know if somebody is still working of this, but I'm also using latest d3.js and I'm also needing this ;)

domenic:
Closing this as it seems abandoned :(.

ulfaslak:
Such a shame this never went anywhere. @marinho's original problem was that when JSDOM does not have support for these geometric shapes, there is no clean way to write tests (with Jest) for code that uses them. Many people have this problem. In my case I (like @marinho) need a polyfill, so the fix is [a monkeypatch](https://github.com/facebook/jest/issues/5379#issuecomment-360044161).

Also, just because it struck me as a bystander: the attitude (especially @domenic's) towards OP in these interactions was hostile. Nobody wants to put time into a space like that.

Issue #2140, opened 2018-02-01
can't delete window.onpagehide


asos-luissilva:
### Basic info:

Since I migrated for the new version jsdom, looks like I can't delete window.onpagehide since is a property of his prototype.

- **Node.js version:** v8.9.4
- **jsdom version:** 11.6.2

### Minimal reproduction case

```js
const jsdom = require("jsdom");
const { JSDOM } = jsdom;
const { window } = new JSDOM(`<!DOCTYPE html><p>Hello world</p>`);

console.log(Object.getOwnPropertyDescriptor(window, "onpagehide")); // undefined
console.log("before", window.onpagehide); // null
console.log(delete window.onpagehide); // true
console.log("after", window.onpagehide); // null
```

### How does similar code behave in browsers?

```js
// Mac 10.11.6 (15G1421)
// in Chrome Version 63.0.3239.84 (Official Build) (64-bit)

console.log(Object.getOwnPropertyDescriptor(window, "onpagehide")); // {get: ƒ, set: ƒ, enumerable: true, configurable: true}
console.log("before", window.onpagehide); // null
console.log(delete window.onpagehide); // true
console.log("after", window.onpagehide); // undefined
```

domenic:
You omitted the section of the issue template that asks you how similar code behaves in browsers. Please run such tests there first and then we can reopen the issue.

asos-luissilva:
Hi @domenic I omitted it, since in my case I'm using jsdom, for testing purposes (jest). And for the test I'm doing I need, I need to be able to delete it so I can test, when a browser has onpagehide and when it doesn't has it.

But I updated the issue to include, how it behaves in browser

domenic:
Thanks for updating! In general we're not interested in supporting testing use cases that can't be run as tests in the browser, so it's good you did.

My best guess at this point is this is a Node.js bug with their vm module :(. It might be worth trying v9.5.0, although I don't remember there being any fixes to vm on the 9.x branch. Sigh. So the action for us jsdom folks is to create a minimal test case (not involving jsdom, just Node.js's vm module), check if it's still broken on v9.5.0, check open issues on Node.js to see if it's already reported, and if not, report it to Node.js. I'm afraid there's no easy workaround for your case :(

asos-luissilva:
I tested in Node.js v9.5.0 but it still doesn't.

I don't think is a problem with the Node but the current implementation of the jsdom window.

in node:
```javascript
// in jsdom
[...]

PR #2141, opened 2018-02-03
fix: allow to delete window[`on${event}`] property

Fetz:
Hi here is the possible fix for the issue jsdom/jsdom#2140

- the to-upstream test file window-properties.html that I created is basically a duplicated of:
w3c web-platform-tests [the-window-object/window-properties.html](https://github.com/w3c/web-platform-tests/blob/master/html/browsers/the-window-object/window-properties.html), but with the other breaking tests commented out.
- added some missing events like: "mousewheel", "show"
- added enumerable to events created with create-event-accessor.js
- for some reason, even if we are adding getter/setter to the {global|window}Events of the window, when the window is being used in the tests the getter/setter doesn't show in the test for some reason. I will probably will need help to fix it if is a blocker

Let me know if there is something I need to tweak or improve


lib/jsdom/living/nodes/GlobalEventHandlers-impl.js
```diff
@@ -22,14 +22,14 @@ const events = new Set([
...
-  "mouseup", "wheel",
+  "mouseup", "wheel", "mousewheel",
   "pause", "play",
   "playing", "progress",
   "ratechange", "reset",
   "resize", "scroll",
   "securitypolicyviolation",
   "seeked", "seeking",
-  "select", "sort", "stalled",
+  "select", "show", "sort", "stalled",
```
TimothyGu: These two additions don't seem to be in the [spec](https://html.spec.whatwg.org/multipage/webappapis.html#globaleventhandlers) anymore. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/GlobalEventHandlers.onmousewheel) for `mousewheel`.

Fetz:
The 3 tests that are failing, in my machine they are working. I suppose is because I'm using a different version of node (v8.5.0).

There are some linting warning in the tests, should I fix even if they will make it different from the source (like 'prefer arrow callback')

lib/jsdom/living/nodes/GlobalEventHandlers-impl.js
Fetz: but they are still on the test from https://github.com/w3c/web-platform-tests/blob/master/html/browsers/the-window-object/window-properties.html, should I ignore for our case and probably later do a contribution there to remove these 2 additions?

TimothyGu: Yeah that works.

Fetz:
Yes, I will rename the file later 👍 , regarding the tests I will see if there is something I can do (since looking at the build log looks like the tests that are failing are the ones that test the primitives values).

If is a problem with node v.6 I will remove or comment those tests (depending of what is the best for this project)

TimothyGu:
> There are some linting warning in the tests, should I fix even if they will make it different from the source (like 'prefer arrow callback')

Yes.

TimothyGu:
> If is a problem with node v.6 I will remove or comment those tests (depending of what is the best for this project)

The same failures [appear](https://travis-ci.org/jsdom/jsdom/jobs/336826288#L5679) with Node.js v9 as well so I'm not sure if it's just a version issue.

Zirro:
Great work, thanks for helping us debug the issue as well as fixing it!

> There are some linting warning in the tests, should I fix even if they will make it different from the source (like 'prefer arrow callback')
>> Yes

Tests which include `dont-upstream` in their name get [ignored](https://github.com/jsdom/jsdom/blob/master/.eslintignore#L30) by the linter now, so I think it'd be better to just rename the file without introducing differences to satisfy it.

Fetz:
the test failing looks like an issue with node, I created an issue in [nodejs project](https://github.com/nodejs/node/issues/18551). So I probably will comment those failing tests.

Regarding those failing tests is ok to be comment or is better to delete them?

TimothyGu:
@Fetz :disappointed: feel free to comment them out then. I'll take a look at them some time in the future as part of the Window overhaul.

Fetz:
The tests are now passing, I removed as well those 2 events handlers and renamed the file. Let me know if I need to do something else

test/web-platform-tests/to-upstream/html/browsers/the-window-object/window-properties-dont-upstream.html
```diff
@@ -0,0 +1,325 @@
...
+  "atob",
+
+  // WindowTimers
+  "setTimeout",
+  "clearTimeout",
+  "setInterval",
+  "clearInterval",
+
+  // HTML Editing APIs
+  // "getSelection", //doesn't work for some reason
```
Zirro: The reason that these methods/attributes - `getSelection`, `matchMedia`, `applicationCache`, `sessionStorage` and `localStorage` - don't work is that they have not yet been implemented. Would be good if you could change the comments to clarify that.

With that, LGTM.

PR #2142, opened 2018-02-03
Constraint Validation API implementation

kontomondo:
This adds constraints validation api support and enables (and passes) the related web platform tests.


lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -589,6 +590,96 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+  }
+
+  get validity() {
+    if (!this._validity) {
+        const value = () => convertStringToNumberByTypeMap.get(this.type)(this.value);
+        const supportsRange = () => this.type === "date" || this.type === "datetime"
+          || this.type === "datetime-local" || this.type === "month"
+          || this.type === "time" || this.type === "week" || this.type === "number"
+          || this.type === "range";
+        this._validity = ValidityState.createImpl({
```
kontomondo: as per the spec, ValidityState needs to be a live object

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -75,7 +75,7 @@ const convertStringToNumberByTypeMap = new Map([
       if (date === null) {
         return NaN;
       }
-      return getUTCMs(date.year, date.month - 1, date.day);
+      return getUTCMs(date.year, date.month, date.day);
```
kontomondo: getUTCMs already subtracts from the month, so in effect you get the wrong date

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -108,7 +108,7 @@ const convertStringToNumberByTypeMap = new Map([
       if (time === null) {
         return NaN;
       }
-      return ((time.hours * 60 + time.minutes) * 60 + time.seconds) * 1000;
+      return ((time.hour * 60 + time.minute) * 60 + time.second) * 1000;
```
kontomondo: the parser returns hour/minute/second, not hours/minutes/seconds

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -590,6 +590,143 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+              return email.split(",").some(oneEmail => !emailRegex.test(oneEmail));
+            }
+            return !emailRegex.test(email);
+          } else if (this.type === "url") {
+            const url = this.value;
+            const urlRegex = new RegExp("^(?!mailto:)(?:(?:http|https|ftp)://)(?:\\S+(?::\\S*)?@)?" +
+              "(?:(?:(?:[1-9]\\d?|1\\d\\d|2[01]\\d|22[0-3])(?:\\.(?:1?\\d{1,2}|2[0-4]\\d|25[0-5])){2}" +
+              "(?:\\.(?:[0-9]\\d?|1\\d\\d|2[0-4]\\d|25[0-4]))|(?:(?:[a-z\\u00a1-\\uffff0-9]+-?)*" +
+              "[a-z\\u00a1-\\uffff0-9]+)(?:\\.(?:[a-z\\u00a1-\\uffff0-9]+-?)*[a-z\\u00a1-\\uffff0-9]+)*" +
+              "(?:\\.(?:[a-z\\u00a1-\\uffff]{2,})))|localhost)(?::\\d{2,5})?(?:(/|\\?|#)[^\\s]*)?$", "i");
```
TimothyGu: We have a URL parser (`git grep whatwg-url lib`). No need for this regex.

lib/jsdom/living/validity/ValidityState-impl.js
```diff
@@ -0,0 +1,66 @@
...
+            return validationMethod();  
+        }
+    }
+
+    return false;  
+  }
+
+  constructor(state) {
+    this._state = state;
+  }
```
TimothyGu: 1. Can you move this to the top of the class?
2. All IDL constructors follow the format `constructor(args, privateData)`. See https://github.com/jsdom/webidl2js#writing-implementation-class-files
3. Would it be better to hold a reference to the element, while making sure all element classes implement a specific interface (e.g. `_sufferingFrom`)? This way, we don't have to create a state object with its own callback functions for every element.

lib/jsdom/living/validity/ValidityState-impl.js
```diff
@@ -0,0 +1,66 @@
...
+
+  get typeMismatch() {
+    return this.hasErrors("typeMismatch");
+  }
+
+  get valueMissing() {
+    return this.hasErrors("valueMissing");
+  }
+
+  hasErrors(method) {
```
TimothyGu: Methods private to IDL and not exposed should be prefixed with a single underscore.

lib/jsdom/living/nodes/HTMLButtonElement-impl.js
```diff
@@ -47,6 +49,60 @@ class HTMLButtonElementImpl extends HTMLElementImpl {
...
+    }
+    const { validity } = this;
+    if (!validity.valid) {
+      this._dispatch(Event.createImpl(["invalid", { cancelable: false } ]));
+    }
+    return validity.valid;
+  }
+
+  setCustomValidity(message) {
+    this._customValidityMessage = message;
```
TimothyGu: 1. This function appears in multiple places. Would it be possible to refactor this function (and possibly other related methods) into a dedicated mixin? Same for the implementations of `willValidate` and `customError` state methods in the `ValidityState` creation, which are the same for all elements. Only the impl method needs to be factored out, no need to divide the IDL file as well. Check out `git grep '^mixin('`.
2. Nit: `_customValidityErrorMessage` per [spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#custom-validity-error-message)
3. Make sure to initialize this property to the empty string in the constructor, per spec. If this method is factored out into a dedicated mixin, create a function called `_initFormControl` in which this property is initialized, and call `this._initFormControl()` in each element's constructor. This should also allow you to get rid of the `this._customValidityMessage != null` check in `customError`.

lib/jsdom/living/nodes/HTMLButtonElement-impl.js
```diff
@@ -47,6 +49,60 @@ class HTMLButtonElementImpl extends HTMLElementImpl {
         break;
     }
   }
+
+  get willValidate() {
+    return !isDisabled(this);
```
TimothyGu: This looks quite different from the [spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#candidate-for-constraint-validation).

lib/jsdom/living/validity/ValidityState-impl.js
```diff
@@ -0,0 +1,66 @@
+"use strict";
```
TimothyGu: You could name this file `lib/jsdom/living/form-control/ValidityState-impl.js` to make the directory a bit more generic.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -590,6 +591,153 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+          return false;
+        },
+        // Is a Boolean indicating the value is not in the required syntax (when type is email or url).
+        typeMismatch: () => {
+          if (this.value === "") {
+            return false;
+          }
+          if (this.type === "email") {
+            const email = this.value;
+            const emailRegex = new RegExp("[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)" +
```
Zirro: The specification [suggests](https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address) a regular expression which differs a bit from this one. Probably safer to go with that instead.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -590,6 +591,153 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+      const supportsRange = () => this.type === "date" || this.type === "datetime" ||
+          this.type === "datetime-local" || this.type === "month" ||
+          this.type === "time" || this.type === "week" || this.type === "number" ||
+          this.type === "range";
+      this._validity = ValidityState.createImpl({
+        willValidate: () => this.willValidate,
+        // Is a Boolean indicating the element's custom
+        // validity message has been set to a non-empty string by
+        // calling the element's setCustomValidity() method.
+        customError: () => this._customValidityMessage != null &&
```
Zirro: Some white space between each method might help with readability.

lib/jsdom/living/validity/ValidityState.webidl
```diff
@@ -0,0 +1,15 @@
...
+  readonly attribute boolean typeMismatch;
+  readonly attribute boolean patternMismatch;
+  readonly attribute boolean tooLong;
+  readonly attribute boolean tooShort;
+  readonly attribute boolean rangeUnderflow;
+  readonly attribute boolean rangeOverflow;
+  readonly attribute boolean stepMismatch;
+  readonly attribute boolean badInput;
+  readonly attribute boolean customError;
+
```
Zirro: We'd prefer to keep the IDL close to the way it looks in the spec to make updating it easier in the future. The IDL in the spec doesn't appear to include a blank line between the errors and the `valid` attribute here.

lib/jsdom/living/validity/ValidityState.webidl
```diff
@@ -0,0 +1,15 @@
...
+  readonly attribute boolean valid;
+};
```
Zirro: Missing trailing newline.

lib/jsdom/living/nodes/HTMLButtonElement-impl.js
```diff
@@ -47,6 +49,60 @@ class HTMLButtonElementImpl extends HTMLElementImpl {
...
+  }
+
+  get validationMessage() {
+    let message = "";
+    const { validity } = this;
+    if (!validity.valid) {
+      if (validity.customError) {
+        message = this._customValidityMessage;
+      } else {
+        message = "The input element does not satisfy its constraints";
```
Zirro: This message should presumably be different for each type of element.

lib/jsdom/living/helpers/dates-and-times.js
```diff
@@ -106,7 +106,7 @@ function serializeYearlessDate({ month, day }) {
   return `${monthStr}-${dayStr}`;
 }
 
-const timeRe = /^([0-9]{2}):([0-9]{2})(?::([0-9]{2}(?:\.[0-9]{1,3})?))?$/;
+const timeRe = /^([0-9]{2}):([0-9]{2})(?::([0-9]{2}(?:\.([0-9]{1,3}))?))?$/;
```
kontomondo: it's better to parse milliseconds directly in order to avoid loss of accuracy due to truncation and subtraction, eg. 1.001 - Math.trun(1.001) =/= 0.001, this lead to a bunch of failures with range overflow/underflows as the millis weren't being parsed correctly.

domenic:
Although I haven't had time to review this much myself, I'll say that my main way of reviewing is going to be comparing this to the spec and seeing how much of a direct translation it is, or is not. The closer it can be to the spec, the easier it will be for all of us to review and maintain.

kontomondo:
@TimothyGu @Zirro Thank you for the feedback, I'll get a few more individual tests going, figure out the needed spec for `fieldset` & `button` then get on refactoring the code with the feedback in mind.

kontomondo:
@domenic Part of the challenge in implementing a spec is of course unambiguously understanding the said spec, the more eyes on this the better, there are some things that I haven't dug into yet (button comes to mind as an example), the input element is hopefully close to spec, although certain things like maxLength/minLength (tooShort/tooLong) are a bit odd as most browsers just ignore the spec and basically always return false even though per spec it's supposed to do the actual check for certain types but only if the value is dirty, strangely enough the spec tests are all testing against a value of false which kind of defeats the tests' purpose but it makes sense given that it's what say Chrome & Firefox do. There are certain things that will need the reviewers insight to figure out, specifically if tooLong/toShort is to be implemented (which is easy enough) then is this._dirtyValue enough? In case of the the form is there a way to do a query that would give us all elements of a certain tag name but are not descendants of a fieldset, or should that operation be implemented, etc.

domenic:
In general we should implement according to the spec. If the tests don't match the spec, we can file bugs on the test suite, and/or the spec, but we shouldn't be forging our own path.

lib/jsdom/living/helpers/dates-and-times.js
```diff
@@ -149,7 +149,7 @@ function serializeTime({ hour, minute, second, millisecond }) {
 // https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#parse-a-local-date-and-time-string
 function parseLocalDateAndTimeString(str, normalized = false) {
   let separatorIdx = str.indexOf("T");
-  if (!normalized && separatorIdx < -1) {
```
kontomondo: An `indexOf` would never be lower than -1

lib/jsdom/living/helpers/dates-and-times.js
```diff
@@ -122,21 +122,22 @@ function parseTimeString(str) {
...
 
 // https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-time-string
 function isValidTimeString(str) {
   return parseTimeString(str) !== null;
 }
-function serializeTime({ hour, minute, second }) {
+function serializeTime({ hour, minute, second, millisecond }) {
   const hourStr = leftPad(`${hour}`, 2, "0");
   const minuteStr = leftPad(`${minute}`, 2, "0");
-  const milliseconds = Math.trunc(second * 1000);
```
kontomondo: This would  jump map the number of seconds onto milliseconds, which is wrong.

kontomondo:
Will close the PR for now, and will re-open in a week or two once I get to refactor the code.

kontomondo:
@TimothyGu @Zirro @domenic Did a refactor pass, this should be in a slightly better shape now.

Remaining:
- [ ] Full `button` spec implementation (mostly a stub at the moment)
- [ ] Full `select` spec implementation (mostly a stub at the moment)
- [ ] Change ValidityState constructor (I haven't looked yet into what args, privateData would mean for this case)
- [ ] Figure out why a test under constraint is failing (has to do with number of invalid events dispatched), may need more eyes on this to figure out why it wants two events to be fired at the same controls
- [ ] `willValidate` for input in button/reset status and `textarea`/`select`/`button` in submit mode with datalist ancestor 
- [ ] `tooLong`/`tooShort` implemetation?
- [ ] More spec url comments pinning to implementation 
- [ ] Further code cleanups

Apart from willValidate and the number of events fired assert all remaining tests under `forms/constraints` are passing (ditto for input-element validation related tests)

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -26,6 +36,23 @@ const selectAllowedTypes = new Set([
...
 
+const maxMinStepTypes = new Set(["date", "month", "week", "time", "datetime-local", "number", "range", "datetime"]);
+
+const allowedTypesForAttribute = {
+  readonly: new Set([
+    "text", "search", "url", "tel", "email", "password", "date",
+    "month", "week", "time", "number", "datetime-local",
+    // The below are NOT correct per spec, but that's what the '-willValidate' web platform test expects
+    // An issue needs to be filed against the tests
+    "color", "file", "submit"
```
kontomondo: Filed a PR for this: https://github.com/w3c/web-platform-tests/pull/9431

lib/jsdom/living/helpers/form-controls.js
```diff
@@ -58,6 +60,21 @@ exports.normalizeToCRLF = string => {
     .replace(/^\n/, "\r\n");
 };
 
+// https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address
+exports.isValidEmailAddress = (emailAddress, multiple = false) => {
+  const emailAddressRegexp = new RegExp("^[a-zA-Z0-9.!#$%&'*+\\/=?^_`{|}~-]+@[a-zA-Z0-9]" +
+    "(?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}" +
+    "[a-zA-Z0-9])?)*$", "i");
```
Zirro: The `ignore case`-flag is not in the spec (and seems like it might not make a difference based on how the regex is written).

test/web-platform-tests/to-run.yaml
```diff
@@ -498,19 +495,13 @@ button.html: [fail, Validation is not implemented]
 checkbox.html: [fail, Events weirdness]
 datetime-local.html: [fail, min/max constraints not implemented]
 datetime.html: [fail, step is not implemented]
-email.html: [fail, Validation is not implemented]
 hidden.html: [fail, list/step/validation/valueAsDate/valueAsNumber not implemented]
```
Zirro: If the validation-part of this test is passing now, you can remove "validation" from the description here.

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -22,6 +25,7 @@ const methods = new Set([
   "dialog"
 ]);
 
+
```
Zirro: I don't think we want the extra line here when it would be inconsistent with the other files.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -590,6 +632,209 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+    }
+    return 0;
+  }
+
+  _attributeApplies(attribute) {
+    return applicableTypesForAttribute[attribute].has(this.type);
+  }
+
+  get willValidate() {
+    // https://html.spec.whatwg.org/#hidden-state-(type=hidden)
```
Zirro: All URLs in comments should preferably go to the `/multipage` version of the spec. This applies to a few other files as well.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -590,6 +632,209 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+    // https://html.spec.whatwg.org/#reset-button-state-(type=reset)
+    // https://html.spec.whatwg.org/#button-state-(type=button)
+    const willNotValidateTypes = new Set(["hidden", "reset", "button"]);
+    // https://html.spec.whatwg.org/multipage/input.html#attr-input-readonly
+    const readOnly = this.hasAttribute("readonly");
+
+    // https:// html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-disabled
```
Zirro: Space between `https://` and the rest of the URL.

test/web-platform-tests/to-run.yaml
```diff
@@ -453,6 +453,9 @@ dirname-ltr.html: [timeout, Unknown]
 formAction_document_address.html: [fail, Unknown]
 formaction.html: [fail, Unknown]
 
+---
```
Zirro: Missing blank line here.

kontomondo:
Thank you, @Zirro , I've made the requested changes.

Zirro:
@kontomondo Sorry about the recent lack of responses. jsdom is a spare time project for everyone involved, and we've all been unusually busy lately.

That doesn't mean we're not excited for this pull request, though - it looks fantastic and we look forward to merging it once we've had time to give it a full review.

---

Something I noticed at an earlier point - the `html/semantics/forms/the-input-element/button.html` test is also passing now, so you can remove it from the list of failing tests.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -252,10 +275,18 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+        result.push(candidate);
       }
     }
+    return result;
   }
+
+  _removeOtherRadioCheckedness() {
+    this._otherRadioGroupElements.forEach(radioGroupElement => {
+      radioGroupElement._checkedness = false;
+    });
```
domenic: Style nit: use a for-of loop here instead of forEach.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -590,6 +632,209 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+    if (converter !== undefined) {
+      return converter(this.value);
+    }
+    return this.value;
+  }
+
+  // https://html.spec.whatwg.org/multipage/input.html#attr-input-step
+  get _step() {
+    let step = this._defaultStep;
+    if (this.hasAttribute("step") && this.getAttribute("step").toLowerCase() !== "any") {
```
domenic: Don't use toLowerCase here; instead use the ASCII case-insensitive version in strings.js

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -590,6 +632,209 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+            return false;
+          }
+          let regexp;
+          try {
+            regexp = new RegExp(this.getAttribute("pattern"), "u");
+          } catch (e) {
+            return false;
+          }
+          const multiple = this._attributeApplies("multiple") && this.hasAttribute("multiple");
+          return !regexpMatches({ regexp, testValue: this.value, split: multiple });
```
domenic: So, you have applied the concept of an elements "values" too widely here. Not every input element can have "values". Only email and file inputs can. So you shouldn't always split them. I think this should be special-cased for emails, not use a generic regexpMatches helper, as I noted previously.

test/web-platform-tests/to-run.yaml
```diff
@@ -466,14 +470,9 @@ reset-form.html: [fail, Unknown]
...
 button-activate.html: [fail, Unknown]
-button-checkvalidity.html: [fail, Validation is not implemented]
 button-events.html: [fail, Unknown]
 button-labels.html: [fail, labels attribute is not implemented]
-button-setcustomvalidity.html: [fail, Validation is not implemented]
-button-validation.html: [fail, Validation is not implemented]
-button-validationmessage.html: [fail, Validation is not implemented]
-button-validity.html: [fail, Validation is not implemented]
-button-willvalidate.html: [fail, Validation is not implemented]
+button-validation.html: [fail, Expects type 'menu' to not change to 'submit']
```
domenic: Sounds like this test is broken upstream; it'd be ideal to file a bug or send a PR to fix it. (Not a blocker for this PR.)

lib/jsdom/living/constraint-validation/DefaultConstraintValidation-impl.js
```diff
@@ -0,0 +1,60 @@
...
+
+  // https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-validationmessage
+  get validationMessage() {
+    const { validity } = this;
+    if (!this.willValidate || validity.valid) {
+      return "";
+    }
+    if (validity.customError) {
+      return this._customErrorMessage;
+    }
```
domenic: This is another area where, since you are using the public API, it is hard to tell if you are following the spec.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
kontomondo: I was under the impression that both `type="file"` and `type="email"` support the multiple attribute and hence "values". Due to that there's `this._attributeApplies("multiple")` above which resolves to:
```
const applicableTypesForAttribute = {
...
  multiple: new Set(["email", "file"]),
...
};
```

lib/jsdom/living/helpers/strings.js
```diff
@@ -99,3 +99,11 @@ exports.splitOnCommas = str => {
   }
   return tokens;
 };
+
+// https://html.spec.whatwg.org/#concept-fe-values
+exports.regexpMatches = ({ regexp, testValue, split = false }) => {
```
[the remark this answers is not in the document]
kontomondo: Done, regarding only email being the special case, maybe we're interpreting the spec differently but as far as I can tell for the pattern mismatch constraint the spec says:
>>> Constraint validation: If the element's value is not the empty string, and the element's multiple attribute is specified and applies to the input element, and the element has a compiled pattern regular expression but that regular expression does not match the entirety of each of the element's values, then the element is suffering from a pattern mismatch.

Both `email` and `file` have the attribute `multiple` apply to them, so I assume that we need to split for file as well

lib/jsdom/living/helpers/form-controls.js
```diff
@@ -58,6 +60,21 @@ exports.normalizeToCRLF = string => {
...
+    "(?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}" +
+    "[a-zA-Z0-9])?)*$");
+  // A valid e-mail address list is a set of comma-separated tokens, where each token is itself
+  // a valid e - mail address.To obtain the list of tokens from a valid e - mail address list,
+  // an implementation must split the string on commas.
+  return regexpMatches({ regexp: emailAddressRegexp, testValue: emailAddress, split: multiple });
+};
+
+exports.isValidAbsoluteUrl = url => {
+  return url.length <= 2083 && whatwgURL.parseURL(url) !== null;
```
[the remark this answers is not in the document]
kontomondo: done

lib/jsdom/living/nodes/HTMLInputElement-impl.js
kontomondo: done

lib/jsdom/living/nodes/HTMLInputElement-impl.js
kontomondo: done

test/web-platform-tests/to-run.yaml
kontomondo: It was a while ago, so I may not be recalling it well, I think it was jsdom at fault here as it's the behaviour that both Chrome and Firefox exhibited.

lib/jsdom/living/helpers/strings.js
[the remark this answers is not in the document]
kontomondo: Let me know if that changes things about needing `regExpMatchValues` or whether I should inline the two usages (`typeMismatch` for `email` validation and `patternMismatch` for` file` and `email`)

lib/jsdom/living/nodes/HTMLButtonElement-impl.js
```diff
@@ -47,8 +50,20 @@ class HTMLButtonElementImpl extends HTMLElementImpl {
         break;
     }
   }
+
+  get willValidate() {
```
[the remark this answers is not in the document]
kontomondo: Done

lib/jsdom/living/helpers/strings.js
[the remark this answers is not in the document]
domenic: So for files, selecting multiple files does not make `el.value` return a comma-separated list of the element's values. (Again, this is a difference between the public API---`value`---and the spec concept---"values".)

The public API is governed by https://html.spec.whatwg.org/multipage/input.html#dom-input-value-filename. As you can see, any commas that occur there are part of the filename, and not appropriate to split on.

So, that's why I think it should be inlined, and there only for emails.

lib/jsdom/living/nodes/HTMLButtonElement-impl.js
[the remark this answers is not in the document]
domenic: Looks great, thank you!

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -122,6 +125,20 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
     this.setAttribute("action", V);
   }
+
+  // https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#statically-validate-the-constraints
+  checkValidity() {
+    const controls = descendantsByHTMLLocalNames(this, submittableElements);
+    // Even though one negative result would be enough, as per spec the checkValidity code needs to run for all
+    // elements
+    const invalidControls = controls.filter(control => !control.checkValidity());
+    return invalidControls.length === 0;
```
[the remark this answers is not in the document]
kontomondo: Done.

> This does not fire any invalid events, or handle the result of firing them.
Since `.checkValidity()` was used, that one fires the invalid events
> Are there no tests for these missing steps? :(
There are tests for `invalid` events being fired, I don't think there are tests for form within a form however (so that form owner would of the descendant would not be the farther form).

lib/jsdom/living/constraint-validation/DefaultConstraintValidation-impl.js
kontomondo: To follow the spec would mean to put `this._customValidityErrorMessage = ''` in every constructor of elements that support validation (or alternatively have an `initialiseCustomValidityErorrMessage/init` method on the default constraint api mixin and call that from the constructor), let me know which way you prefer.

kontomondo:
@domenic Thanks! I think I've addressed most feedback, let me know if I missed anything, some outstanding things that I'd need your responses on are:
1) Clarification regarding your interpretation of the spec for `patternMismatch` constraint for input type `file` with `multiple=`true`, see my replies above.
2) Your preference for more closely following the spec for custom validity.

lib/jsdom/living/helpers/strings.js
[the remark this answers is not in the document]
kontomondo: done

lib/jsdom/living/nodes/HTMLInputElement-impl.js
kontomondo: updated to only do the split for emails

lib/jsdom/living/constraint-validation/DefaultConstraintValidation-impl.js
kontomondo: I went with this solution for now - `this._customValidityErrorMessage` set to '' in the constructor for elements that are not barred from constraint validation

kontomondo:
@domenic The CI is failing with 
```
1) web-platform-tests FileAPI reading-data-section/filereader_readAsBinaryString.html:
     Error: Timeout of 70000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
  
  2) web-platform-tests dom/events event-disabled-dynamic.html:
     Error: Timeout of 70000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
  
  3) web-platform-tests dom/nodes Element-classlist.html:
     Error: 120 errors in test:
```

Are these known issues? Not sure why my changes here would impact any of that.

Zirro:
@kontomondo Nice work! Provided Domenic is happy with the changes, this looks ready for merging now.

domenic:
Thanks so much @kontomondo, especially for your patience with our slow review process. Pulling this in to do a release now!

PR #2145, opened 2018-02-04
Extend fromUrl options

lucky-libora:
Also typescript definition file added


domenic:
Thanks for your interest, but we're not interested in either TypeScript nor in making fromURL more general purpose. For the latter, see https://github.com/jsdom/jsdom#fromurl:

> The initial request is not infinitely customizable to the same extent as is possible in a package like request; fromURL() is meant to be a convenience API for the majority of cases. If you need greater control over the initial request, you should perform it yourself, and then use the JSDOM constructor manually.

PR #2146, opened 2018-02-04
Use whatwg-mimetype and data-urls packages

domenic:
This improves our MIME type and data URL parsing to be per-spec.


domenic:
The prevalence of try/catch makes me wonder about the API design of `new MIMEType`. Maybe we want an additional static method, e.g. `MIMEType.parse()`, which is allowed to return null for unparseable MIME types. After all, often these strings come as user input, and so it's not "exceptional" for them to be unparseable. /cc @annevk.

annevk:
That sounds reasonable. Should we have that for `new URL()` as well or is that more exceptional?

domenic:
Somehow I haven't run across it as much for `new URL()`, but the same argument makes sense. I guess because my JS implementation already exposes the `parseURL` low-level primitive, instead of going through the full `URL` API. Yeah, if you were to implement web platform features using `new URL()`, you'd be doing a lot of try/catching.

Sebmaster:
Rebase to drop the isJavascript commits?

domenic:
Was planning to squash in general, but yeah.

Issue #2150, opened 2018-02-13
Shouldn't try to load image with removed src attribute


atsikov:
### Basic info:

- **Node.js version:** 9.5
- **jsdom version:** 11.6.2

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");
const dom = new JSDOM(
    "<!DOCTYPE html>",
    {
        resources: "usable",
    },
);

const grayPx = "data:image/gif;base64,R0lGODlhAQABAIAAAMLCwgAAACH5BAAAAAAALAAAAAABAAEAAAICRAEAOw==";
const image = new dom.window.Image();
const onLoad = () => {
    console.log("image loaded");
    image.removeEventListener("load", onLoad);
    image.removeAttribute("src");
}
image.addEventListener("load", onLoad);
image.src = grayPx;
```

This code causes an exception. JSDOM is trying to load image with null url.

```
image loaded, will try to remove src attribute
=====
Error: Uncaught [TypeError: Cannot read property 'replace' of null]
    at reportException (node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:63:24)
    at invokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:233:9)
    at HTMLImageElementImpl._dispatch (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:122:9)
    at HTMLImageElementImpl.dispatchEvent (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:83:17)
    at HTMLImageElementImpl.dispatchEvent (node_modules/jsdom/lib/jsdom/living/nodes/HTMLElement-impl.js:40:27)
    at node_modules/jsdom/lib/jsdom/browser/resource-loader.js:51:15
    at Object.check (node_modules/jsdom/lib/jsdom/living/nodes/Document-impl.js:91:11)
    at node_modules/jsdom/lib/jsdom/living/nodes/Document-impl.js:110:12
    at wrappedEnqueued (node_modules/jsdom/lib/jsdom/browser/resource-loader.js:255:16)
    at readDataUrl (node_modules/jsdom/lib/jsdom/browser/resource-loader.js:114:7)

TypeError: Cannot read property 'replace' of null
    at fetch (node_modules/jsdom/lib/jsdom/browser/resource-loader.js:139:6)
    at Object.exports.load (node_modules/jsdom/lib/jsdom/browser/resource-loader.js:270:11)
[...]

PR #2151, opened 2018-02-13
Fix HTMLImageElement removeAttribute for src

atsikov:
This change should fix exception raised while removing src attribute from HTMLImageElement (https://github.com/jsdom/jsdom/issues/2150)

### DIFF
--- a/test/web-platform-tests/to-upstream/html/semantics/embedded-content/the-img-element/remove-attribute-src.html
+++ b/test/web-platform-tests/to-upstream/html/semantics/embedded-content/the-img-element/remove-attribute-src.html
@@ -1,19 +1,18 @@
 <!DOCTYPE html>
 <meta charset="utf-8">
-<title>Removing src attribute from img</title>
+<title>Removing src attribute from img after it loads</title>
 <link rel="author" title="atsikov" href="mailto:alexey.tsikov@gmail.com">
-<link rel=help href="https://html.spec.whatwg.org/#the-img-element">
+<link rel="help" href="https://html.spec.whatwg.org/#the-img-element">
 
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 
-<body>
 <script>
 "use strict";
 
 async_test(t => {
   const image = new window.Image();
-  image.onload = () => {
+  const runTest = t.step_func(() => {
     image.onload = t.unreached_func("onload should not be called with removed src");
     image.onerror = t.unreached_func("onerror should not be called with removed src");
     image.removeAttribute("src");
@@ -23,19 +22,18 @@ async_test(t => {
     assert_equals(image.complete, false);
     assert_equals(image.src, "");
     assert_equals(image.currentSrc, "");
-  };
+
+    // Give it a second to make sure onload/onerror don't fire within that time frame.
+    t.step_timeout(() => {
+      t.done();
+    }, 1000);
+  });
 
   image.src = "data:image/gif;base64,R0lGODlhAQABAIAAAMLCwgAAACH5BAAAAAAALAAAAAABAAEAAAICRAEAOw==";
 
-  t.step_timeout(() => {
-    // JSDOM specific: in case image loading is not supported (no canvas api), run checks out of onload handler
-    image.removeAttribute("src");
-    assert_equals(image.src, "");
-    assert_equals(image.currentSrc, "");
-    // end of JSDOM specific
+  img.onload = runTest;
 
-    t.done();
-  }, 1000);
+  // JSDOM specific: when image loading is not supported (i.e. when node-canvas is not installed), run the test anyway.
+  t.step_timeout(runTest, 1000);
 }, "img is cleared when src attribute is removed");
 </script>
-</body>
--- a/test/web-platform-tests/to-upstream/html/semantics/embedded-content/the-img-element/remove-attribute-src.html
+++ b/test/web-platform-tests/to-upstream/html/semantics/embedded-content/the-img-element/remove-attribute-src.html
@@ -1,0 +1,34 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Disabled checkbox event handling</title>
+<link rel="author" title="atsikov" href="mailto:alexey.tsikov@gmail.com">
+<link rel=help href="https://html.spec.whatwg.org/#the-img-element">
+
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<body>
+<script>
+"use strict";
+
+async_test(t => {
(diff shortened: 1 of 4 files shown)


domenic:
Thanks, but we can't accept tests in the old legacy format. See CONTRIBUTING.md for how to write tests.

atsikov:
So I should add new test to `web-platform-tests/html/semantics/embedded-content/the-img-element/` and create PR there, right?

atsikov:
Nevermind, I think I have found proper test in WPT

atsikov:
Is there a way to re-run travis? Some random test failed with timeout.

test/web-platform-tests/to-upstream/html/semantics/embedded-content/the-img-element/remove-attribute-src.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Disabled checkbox event handling</title>
```
domenic: Needs a better title

test/web-platform-tests/to-upstream/html/semantics/embedded-content/the-img-element/remove-attribute-src.html
```diff
@@ -0,0 +1,43 @@
...
+  assert_equals(image.complete, false);
+  assert_equals(image.src, "");
+  assert_equals(image.currentSrc, "");
+  t.done();
+}
+
+async_test(t => {
+  const image = new window.Image();
+  image.onload = () => {
+    image.onload = () => {
```
domenic: You can make this a bit shorter: `image.onload = t.unreached_func("onload should not be called...");`.

test/web-platform-tests/to-upstream/html/semantics/embedded-content/the-img-element/remove-attribute-src.html
```diff
@@ -0,0 +1,43 @@
...
+      assert_true(false, "onload should not be called with removed src");
+    };
+    image.error = () => {
```
domenic: Should be `onerror`, not `error`. And you can use `unreached_func` again.

test/web-platform-tests/to-upstream/html/semantics/embedded-content/the-img-element/remove-attribute-src.html
```diff
@@ -0,0 +1,43 @@
...
+"use strict";
+
+function testRemoveSrc(image, t) {
+  image.removeAttribute("src");
+  assert_equals(image.width, 0);
+  assert_equals(image.height, 0);
+  assert_equals(image.complete, false);
+  assert_equals(image.src, "");
+  assert_equals(image.currentSrc, "");
+  t.done();
```
domenic: This will cause the test to finish, even if onload/onerror are going to be filed. I think you'll want to put this inside a `t.step_timeout()` call with a comment like "Wait for a bit to make sure onload/onerror are not fired".

test/web-platform-tests/to-upstream/html/semantics/embedded-content/the-img-element/remove-attribute-src.html
```diff
@@ -0,0 +1,43 @@
...
+    };
+    image.error = () => {
+      assert_true(false, "error should not be called with removed src");
+    };
+    testRemoveSrc(image, t);
+  };
+  image.src = "data:image/gif;base64,R0lGODlhAQABAIAAAMLCwgAAACH5BAAAAAAALAAAAAABAAEAAAICRAEAOw==";
+
+  t.step_timeout(() => {
+    // in case image loading is not supported, run checks out of onload handler
```
domenic: It'd be nice to put a note like "JSDOM specific:" here so that if/when we upstream these tests to the shared test suite used by real browsers, we remember to remove this code. This was a good idea for jsdom though.

test/web-platform-tests/to-upstream/html/semantics/embedded-content/the-img-element/remove-attribute-src.html
atsikov: `t.done()` was put here intentionally. Same checks as for timeout are performed for initial `onload` handler to make sure loaded image was cleared upon attribute removal. Then `t.done()` should be called for both cases.

atsikov: Actually I decided that since `t.step_timeout` will be removed in case of moving test to wpt repo, there's no need for a separate function. Code is a bit duplicated with timeout handler but then will be more straightforward without additional function.

test/web-platform-tests/to-upstream/html/semantics/embedded-content/the-img-element/remove-attribute-src.html
```diff
@@ -0,0 +1,43 @@
...
+    image.onerror = t.unreached_func("onerror should not be called with removed src");
+    image.removeAttribute("src");
+
+    assert_equals(image.width, 0);
+    assert_equals(image.height, 0);
+    assert_equals(image.complete, false);
+    assert_equals(image.src, "");
+    assert_equals(image.currentSrc, "");
+
+    t.done();
```
domenic: This is still not valid, because by calling the test done() here, you've made it impossible for image.onload/image.onerror to ever fire. (So the unreached_func assertions will never happen, even in a buggy implementation.)

In other words, I believe this test passes both with and without your change, making it a not-useful test.

atsikov: Without my changes `img.removeAttribute("src")` will raise an exception trying to load resource from `null` url (and that was the main reason for these changes), so test will fail even without any checks.

But you're right, since `onload` or `onerror` are async, it's better to leave `t.done()` in timeout only.

domenic: I don't believe tests in the WPT framework fail due to raised exceptions; they fail due to assertions getting hit. (But I might be wrong; I haven't checked recently.)

atsikov: I tried and it failed with exception (in case canvas was installed, otherwise it didn't try to load anything).

atsikov:
```
  1) web-platform-tests html/semantics/tabular-data processing-model-1/span-limits.html:
     Error: Timeout of 70000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:160:7)
```
This test keeps failing on Travis probably because of high load. It is really slow (took 45s on my machine).

Zirro:
Interestingly, we recently merged changes designed to handle that testcase faster (though still not fast relative to other tests, by any means).

Indeed, Travis seems to have been under particularly high load while those jobs were running given how much longer they took to complete compared to the other ones. They'll probably pass if we give it another try.

PR #2165, opened 2018-03-06
Implement requestAnimationFrame using performance.now

SimenB:
Shouldn't make a practical difference as `w3c-hr-time` uses `browser-process-hrtime` (thus no change in lockfile), but it simplifies things a tiny bit.


domenic:
Thanks!

PR #2170, opened 2018-03-12
Use latest Web Platform Tests

Zirro:
Including changes needed to pass some of the added/modified tests. More information can be found in the extended commit messages.


Zirro:
`FileAPI/historical.https.html` seems to have joined the group of tests which only fail in Node v6. Will fix tomorrow.

lib/jsdom/living/aborting/AbortSignal-impl.js
```diff
@@ -26,7 +26,14 @@ class AbortSignalImpl extends EventTargetImpl {
       algorithm();
     }
     this.abortAlgorithms.clear();
-    this._dispatch(Event.createImpl(["abort"]), false);
+
+    this._dispatch(Event.createImpl(
+      [
+        "abort",
+        { bubbles: false, cancelable: false }
```
domenic: We shouldn't need to explicitly set bubbles and cancelable, I think. But no big deal. I filed #2179 for future cleanup here.

Issue #1924, opened 2017-07-29
Can't load external js


dwqs:
### Basic info:

- **Node.js version:** v8.1.3
- **jsdom version:** v 11.1.0

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const options = {
    runScripts: "dangerously",
    resources: "usable"
};
const dom = new JSDOM(`
  <!DOCTYPE html><html><body><div id="test"></div><script src="http://cdn.bootcss.com/jquery/2.1.1/jquery.min.js"></script></body></html>
`, options);

console.log(dom.window.jQuery, dom.window.$)
```
The output in terminal is `undefined`. And if load external js from a relative path, will get error:

```
Error: Could not load script: "./vv.js"
    at /Users/pomy/git/vue-mobx/node_modules/jsdom/lib/jsdom/browser/resource-loader.js:44:23
    at Object.check (/Users/pomy/git/vue-mobx/node_modules/jsdom/lib/jsdom/living/nodes/Document-impl.js:91:11)
    at ReadStream.<anonymous> (/Users/pomy/git/vue-mobx/node_modules/jsdom/lib/jsdom/living/nodes/Document-impl.js:110:12)
    at ReadStream.wrappedEnqueued (/Users/pomy/git/vue-mobx/node_modules/jsdom/lib/jsdom/browser/resource-loader.js:255:16)
    at emitOne (events.js:115:13)
    at ReadStream.emit (events.js:210:7)
    at fs.js:1942:12
    at FSReqWrap.oncomplete (fs.js:135:15) { Error: ENOENT: no such file or directory, open './vv.js' errno: -2, code: 'ENOENT', syscall: 'open', path: './vv.js' }
```
For `vv.js`:
```
var a =1;
```
I've trawled the documentation and tried lots of variations but nothing works.

vsemozhetbyt:
I may be wrong, but it seems the case of https://github.com/tmpvar/jsdom#asynchronous-script-loading
Yor script may not be loaded and executed yet when you try to get `dom.window.jQuery`.

TimothyGu:
http://cdn.bootcss.com/jquery/ returns a 404 here.

dwqs:
Here is the link: [http://cdn.bootcss.com/jquery/2.1.1/jquery.min.js](http://cdn.bootcss.com/jquery/2.1.1/jquery.min.js) @TimothyGu

dwqs:
[...]

PR #2172, opened 2018-03-15
Document how to execute external scripts

futtetennista:
The only trace I could find is in [this issue](https://github.com/jsdom/jsdom/issues/1924). 

Ideally jsdom could log something to the console if the it sees a script element with an external script without the `resources: "usable"` option.


domenic:
Hmm, I can see how having to read two sections:

- https://github.com/jsdom/jsdom#executing-scripts
- https://github.com/jsdom/jsdom#loading-subresources

instead of one might be difficult. But I'd rather not introduce a third section. Maybe just add a single sentence at the bottom of "Executing scripts" that refers you to look at the "Loading subresources" section for information about loading external scripts?

futtetennista:
Oh I completely missed that second section! Yes the solution you proposed make sense to me.

PR #2174, opened 2018-03-22
fix: require canvas lib

ioslh:
`canvas` and `canvas-prebuilt` export different result
 for `canvas` situation,  which export an Object like this `{ Canvas, Image, ... }`, see [here](https://github.com/Automattic/node-canvas/blob/master/index.js#L51) for more details.

### DIFF
--- a/lib/jsdom/utils.js
+++ b/lib/jsdom/utils.js
@@ -183,7 +183,19 @@ exports.treeOrderSorter = function (a, b) {
 exports.Canvas = null;
 ["canvas", "canvas-prebuilt"].some(moduleName => {
   try {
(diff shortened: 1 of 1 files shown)


TimothyGu:
Seems like this is because you are using a v2.x version of node-canvas, which is [not yet supported](https://github.com/jsdom/jsdom/pull/1964#issuecomment-326359348). Please let me know if that is the case.

ioslh:
Bingo.

Maybe I have to use my own fork before node-canvas 2.x is out of alpha. Thank you.

ioslh:
Bingo.

Maybe I have to use my own fork before node-canvas 2.x is out of alpha. Thank you.

PR #2195, opened 2018-04-16
Fix XML-related issues in createDocument and createElement

domenic:
* createElement() was not correctly setting a null namespace for non-XHTML XML documents
* createDocument() was not correctly setting the content type based on the namespace

---

This is not ready for merge. I believe it depends on both #2192 and https://github.com/w3c/web-platform-tests/pull/10484 to pass all the relevant web platform tests.


domenic:
Blocked on https://github.com/whatwg/dom/issues/643

PR #2197, opened 2018-04-16
Roll Web Platform Tests

Zirro:
This includes updates to the configuration, accounting for changes upstream. I've also added support for Event's `srcElement` (an alias of `target`) and `returnValue`.

For the latter, the last test in this file is still failing: https://github.com/w3c/web-platform-tests/blob/master/dom/events/Event-returnValue.html#L55

The test seems incorrect to me as it relies on the `cancelable` flag being true by default for an event created with `new Event("foo")`, but I've left it enabled so that we can verify whether this is the case before merging.

### DIFF
--- a/lib/jsdom/living/events/Event-impl.js
+++ b/lib/jsdom/living/events/Event-impl.js
@@ -35,6 +35,20 @@ class EventImpl {
     this.timeStamp = Date.now();
   }
 
+  get srcElement() {
+    return this.target;
+  }
+
+  get returnValue() {
+    return !this._canceledFlag;
+  }
+
+  set returnValue(v) {
+    if (this.cancelable && v === false) {
+      this._canceledFlag = true;
+    }
+  }
+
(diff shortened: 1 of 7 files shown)


lib/jsdom/living/events/Event-impl.js
```diff
@@ -22,6 +22,7 @@ class EventImpl {
     }
 
     this.target = null;
+    this.srcElement = null;
```
domenic: Seems a bit better to just return `this.target`. That way it matches the spec and we don't have to worry about keeping them in sync.

TimothyGu:
Also CC @annevk and @miketaylr.

miketaylr:
Nice catch, thanks @TimothyGu.

Zirro:
The `Event.returnValue` test has been fixed upstream, and I've rebased this pull request on a new roll.

Issue #2177, opened 2018-03-26
'Error: Could not parse CSS stylesheet' since 11.6.x


aphofstede:
### Basic info:

- **Node.js version:** 8.9.1
- **jsdom version:** 11.6.2

I'm getting this error after upgrading from 11.5.1 to 11.6.2 on the sweetalert2 CSS, when mounting a Vue component during a vue-test-utils mocha-webpack test run. (which uses JSDOM) 
Let me know if I can supply more details; not sure how to debug this.

```
Error: Could not parse CSS stylesheet
    at exports.createStylesheet (node_modules/jsdom/lib/jsdom/living/helpers/stylesheets.js:35:21)
    at HTMLStyleElementImpl._updateAStyleBlock (node_modules/jsdom/lib/jsdom/living/nodes/HTMLStyleElement-impl.js:68:5)
    at HTMLStyleElementImpl._childTextContentChangeSteps (node_modules/jsdom/lib/jsdom/living/nodes/HTMLStyleElement-impl.js:37:12)
    at HTMLStyleElementImpl.insertBefore (node_modules/jsdom/lib/jsdom/living/nodes/Node-impl.js:225:14)
    at HTMLStyleElementImpl.insertBefore (node_modules/jsdom/lib/jsdom/living/nodes/Node-impl.js:202:14)
    at HTMLStyleElementImpl.appendChild (node_modules/jsdom/lib/jsdom/living/nodes/Node-impl.js:327:17)
    at HTMLToDOM._parseWithParse5 (node_modules/jsdom/lib/jsdom/browser/htmltodom.js:60:21)
    at HTMLToDOM._doParse (node_modules/jsdom/lib/jsdom/browser/htmltodom.js:47:42)
    at HTMLToDOM.appendToNode (node_modules/jsdom/lib/jsdom/browser/htmltodom.js:37:17)
    at setInnerHTML (node_modules/jsdom/lib/jsdom/living/nodes/Element-impl.js:40:25)
    at HTMLStyleElementImpl.set innerHTML [as innerHTML] (node_modules/jsdom/lib/jsdom/living/nodes/Element-impl.js:211:5)
    at HTMLStyleElement.set [as innerHTML] (node_modules/jsdom/lib/jsdom/living/generated/Element.js:874:29)
    at node_modules/sweetalert2/dist/sweetalert2.all.js:1978:1
    at Object.<anonymous> (node_modules/sweetalert2/dist/sweetalert2.all.js:1978:1)
[snip]
[...]

PR #2202, opened 2018-04-20
move from cssom to css-object-model for #2177

caub:


Zirro:
Sorry, @caub. I already have a [branch](https://github.com/Zirro/jsdom/tree/css-object-model) which integrates css-object-model with jsdom, but it is nowhere near ready to become a pull request yet.

PR #2203, opened 2018-04-21
Fixed a line in toBlob() where quality was always taken to be 0

challakoushik:
Just a single line change replacing 
`Math.min(0, Math.max(1, qualityArgument)) * 100 
`
with
`Math.max(0, Math.min(1, qualityArgument)) * 100`


domenic:
Thank you!

PR #2206, opened 2018-04-24
document.hidden follows visibilityState

### LINKED ISSUE #2205: [readme] Question about § Pretending to be a visual browser
Specifically:

> * Changing `document.hidden` to return `true` instead of `false`

Shouldn't it be the other way around, or am I missing something? A visible page would be `document.hidden == false`, I thought…

kapouer:
Per spec. Fix #2205

### DIFF
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -806,11 +806,7 @@ class DocumentImpl extends NodeImpl {
   }
 
   get hidden() {
-    if (this._defaultView && this._defaultView._pretendToBeVisual) {
-      return false;
-    }
-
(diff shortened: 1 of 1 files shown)


kapouer:
Note: i've just changed the === to !===

Zirro:
This does not fix #2205, which only concerns a mistake in the [README.md file](https://github.com/jsdom/jsdom/blame/master/README.md#L112).

I think that it's better to keep `hidden` and `visibilityState` separate in this case, to make it clear that both rely on the jsdom-specific `pretendToBeVisual` option rather than what the specification actually requires. The use of if-conditions also seem clearer to me here.

For the reasons above, I will close this pull request.

PR #2207, opened 2018-04-25
Fix old-api tests missing html property

Zirro:
Two of the old-api tests were missing their `html` property, apparently resulting in the HTML getting passed to `fs.readFile` as a filename through resource-loader instead.

This issue was revealed by changes in Node.js v10 - likely https://github.com/nodejs/node/commit/d8f73385e2 where the null byte included in one of the tests now causes a synchronous error to be thrown.

There could potentially be more tests with missing `html` properties. I only had a quick look through `env.js`.

### DIFF
--- a/test/old-api/env.js
+++ b/test/old-api/env.js
@@ -133,9 +133,9 @@ describe("jsdom/env", () => {
   });
 
   specify("string, full HTML document", { async: true }, t => {
-    env(
-      "<!DOCTYPE html><html><head><title>Hi</title></head><body>Hello</body></html>",
-      (err, window) => {
+    env({
+      html: "<!DOCTYPE html><html><head><title>Hi</title></head><body>Hello</body></html>",
+      done(err, window) {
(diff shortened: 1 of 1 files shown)


domenic:
I'm not sure changing the tests is the right fix here. env is/was supposed to be magic, and figure out whether you're passing HTML or a filename. We should probably preserve that contract even in Node 10, instead of requiring people to change their code in this way...

Zirro:
Aha, I see. The new API was introduced shortly after I took an interest in jsdom, so I never saw much of the old one. I concluded that this was the right fix after seeing that most tests which were passing HTML in that file also were using the `html` property but indeed, we shouldn't break the old API as long as we're keeping it around.

Looking at the logic in `old-api.js`, I suppose the proper solution is to catch potential errors and then proceed with the `if (err)`-path from the callback. I'll close this pull request since it's based on a faulty assumption.

PR #2215, opened 2018-05-04
Upgrade whatwg-url to version 6.4.1

Zirro:
It includes a fix to make the href setter propagate changes to `URL.searchParams`, allowing the `url/url-constructor.html` test to pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "webidl-conversions": "^4.0.2",
     "whatwg-encoding": "^1.0.3",
     "whatwg-mimetype": "^2.1.0",
-    "whatwg-url": "^6.4.0",
+    "whatwg-url": "^6.4.1",
     "ws": "^4.0.0",
     "xml-name-validator": "^3.0.0"
   },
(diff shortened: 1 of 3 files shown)


domenic:
This also needs a package.json update for people installing jsdom as a dependency, I think? Otherwise they will get whatever their local machine has cached that still fits our version range.

AviVahl:
Actually, people installing jsdom (current latest version) without an existing lock file already get whatwg-url v6.4.1. That's because you specify `^6.4.0`, which resolves to 6.4.1. That's good and that's how semver works (patch versions should not break api).

If you specify `^6.4.1` in `package.json`, you will also force any user resolving the next jsdom version to get whatwg-url 6.4.1 (even is they were locked to whatwg-url 6.4.0, as the dependency request changed and will be re-resolved).

The yarn.lock change in this PR just assures you yourself will get 6.4.1 while working on this project.

domenic:
I'm not sure that's true, e.g. if you use npm and your cache contains 6.4.0, and the package specifies `^6.4.0`, then npm will use 6.4.0, instead of hitting the network to download 6.4.1. Especially if you specify --prefer-offline.

AviVahl:
I wrote the above after testing it locally, heh.

If you specify -`-prefer-offline`, that's another case. One would argue this flag is way less common than people just using `yarn`/`npm i` (as-is) with lock-files.

Zirro:
Rebased to include `package.json` as well.

PR #2216, opened 2018-05-04
Remove test causing connections to several unrelated domains

Zirro:
I use a firewall for outbound connections, and have noticed that this test causes resources hosted at `redditstatic.com`, `ajax.googleapis.com` and `google-analytics.com` to be loaded. I'm not very keen on adding them to my permanent list of allowed domains, so they've remained an annoyance when I run the tests locally.

I'm reasonably sure that the issue being tested for here is sufficiently covered by newer tests at this point and that it's not suitable for porting to a WPT.

To my knowledge, the only external connection that remains after this change comes from a few tests which intentionally use `example.com`.


domenic:
This broke a browser-only test (which does not run on pull requests): https://travis-ci.org/jsdom/jsdom/jobs/376042102.

We should change that to refer to some other file. And also bump the priority on changing our browser testing infrastructure to use headless Chrome or similar.

Issue #2175, opened 2018-03-23
Is label.control unsupported?


kentcdodds:
### Basic info:

- **Node.js version:** v8
- **jsdom version:** 11.6.2

### Minimal reproduction case

```js
const {JSDOM} = require('jsdom');
const dom = new JSDOM('');
const {document} = dom.window;

var div = document.createElement('div');
div.innerHTML = '<label>hi<input /></label>';
console.log(div.firstChild.control); // undefined
```

### How does similar code behave in browsers?

In the browser it references the input. Copy/paste this into your browser's developer tools:

```javascript
var div = document.createElement('div');
div.innerHTML = '<label>hi<input /></label>';
console.log(div.firstChild.control); // references the input node
```

domenic:
It looks like this is indeed unimplemented. A pull request would be welcome!

FrecksterGIT:
#2217 brings support for HTMLLabelElement.control and *.labels where needed.
This can be closed now.

Zirro:
@FrecksterGIT Nice work. Thanks!

kentcdodds:
That's wonderful! Thanks @FrecksterGIT

PR #2217, opened 2018-05-04
Add support for HTMLLabelElement.control

FrecksterGIT:
Adds support for the HTMLLabelElement.control and *.labels properties. (jsdom/jsdom#2175)

Adds labels property support for:

- button
- input (if the type attribute is not in the Hidden state)
- meter
- output
- progress
- select
- textarea

Activated tests:
- button-labels.html
- input-labels.html
- labelable-elements.html

Added tests:
- label-attributes-dont-upstream.html (copy from wpt without 2 failing tests.)

### DIFF
--- a/lib/jsdom/living/helpers/form-controls.js
+++ b/lib/jsdom/living/helpers/form-controls.js
@@ -72,7 +72,6 @@ exports.isLabelable = node => {
 
   switch (node.tagName) {
     case "BUTTON":
-    case "KEYGEN":
     case "METER":
     case "OUTPUT":
     case "PROGRESS":
--- a/lib/jsdom/living/helpers/form-controls.js
+++ b/lib/jsdom/living/helpers/form-controls.js
@@ -18,6 +18,10 @@ const {
 } = require("./dates-and-times");
 const whatwgURL = require("whatwg-url");
 
+const NodeList = require("../generated/NodeList");
+const { domSymbolTree } = require("../helpers/internal-constants");
+const NODE_TYPE = require("../node-type");
+const { getRootNode } = require("../helpers/traversal");
 const submittableLocalNames = new Set(["button", "input", "keygen", "object", "select", "textarea"]);
 
 exports.isDisabled = formControl => {
@@ -58,6 +62,51 @@ exports.normalizeToCRLF = string => {
     .replace(/\r$/, "\r\n")
     .replace(/([^\r])\n/g, "$1\r\n")
     .replace(/^\n/, "\r\n");
+};
+
+exports.isLabelable = node => {
+  // labelable logic defined at: https://html.spec.whatwg.org/multipage/forms.html#category-label
+  if (node.nodeType !== NODE_TYPE.ELEMENT_NODE) {
+    return false;
+  }
+
+  switch (node.tagName) {
+    case "BUTTON":
+    case "KEYGEN":
+    case "METER":
+    case "OUTPUT":
+    case "PROGRESS":
+    case "SELECT":
+    case "TEXTAREA":
+      return true;
+
+    case "INPUT":
+      return node.type !== "hidden";
+  }
+
+  return false;
+};
+
+exports.getLabelsForLabelable = labelable => {
+  if (!exports.isLabelable(labelable)) {
+    return null;
+  }
+  if (!labelable._labels) {
+    const root = getRootNode(labelable);
+    labelable._labels = NodeList.create([], {
+      element: root,
+      query: () => {
+        const nodes = [];
+        for (const descendant of domSymbolTree.treeIterator(root)) {
+          if (descendant.control === labelable) {
+            nodes.push(descendant);
+          }
+        }
+        return nodes;
+      }
+    });
+  }
+  return labelable._labels;
 };
 
 // https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address
--- a/lib/jsdom/living/nodes/HTMLInputElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLInputElement-impl.js
@@ -4,6 +4,7 @@ const DOMException = require("domexception");
 const Event = require("../generated/Event");
 const FileList = require("../generated/FileList");
 const HTMLElementImpl = require("./HTMLElement-impl").implementation;
+const NodeList = require("../generated/NodeList");
 const idlUtils = require("../generated/utils");
 const DefaultConstraintValidationImpl =
   require("../constraint-validation/DefaultConstraintValidation-impl").implementation;
@@ -59,6 +60,7 @@ function allowVariableLengthSelection(type) {
 
 const valueAttributeDefaultMode = new Set(["hidden", "submit", "image", "reset", "button"]);
 const valueAttributeDefaultOnMode = new Set(["checkbox", "radio"]);
+
 function valueAttributeMode(type) {
   if (valueAttributeDefaultMode.has(type)) {
     return "default";
@@ -240,6 +242,7 @@ class HTMLInputElementImpl extends HTMLElementImpl {
 
     super._attrModified.apply(this, arguments);
   }
+
   _formReset() {
     const wrapper = idlUtils.wrapperForImpl(this);
     this._value = sanitizeValueByType(this, wrapper.defaultValue);
@@ -250,6 +253,7 @@ class HTMLInputElementImpl extends HTMLElementImpl {
       this._removeOtherRadioCheckedness();
     }
   }
+
   _changedFormOwner() {
     if (this._checkedness) {
       this._removeOtherRadioCheckedness();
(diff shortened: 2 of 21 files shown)


lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -48,6 +48,26 @@ function sendClickToAssociatedNode(node) {
 }
 
 class HTMLLabelElementImpl extends HTMLElementImpl {
+  get control() {
+    if (this.hasAttribute("for")) {
+      const node = this.ownerDocument.getElementById(this.getAttribute("for"));
+      if (node && isLabelable(node)) {
+        return node;
+      }
+      return undefined;
```
TimothyGu: This should probably be null.

FrecksterGIT: Fixed. Also changed the activation behaviour to make use of this. it didn't handle the _wrong-for-attribute-case_.

domenic:
Nice work! A lot of stuff about `control` seems to be covered by https://github.com/w3c/web-platform-tests/blob/master/html/semantics/forms/the-label-element/label-attributes.html. Can you enable that test and see how we do?

FrecksterGIT:
After working through the tests, I also changed the form to make use of the control.

Ignoring all tests using HTMLInputElement.labels the only one that is still failing:
```
Failed in "A label element not in a document can not label any element in the document.":
assert_not_equals: A label element not in a document should not label an element in a document. got disallowed value Element node <input id="test1"></input>
```
The problem here is, that the label has an ownerDocument, but isn't in the dom.

```javascript
    var label = document.createElement("label");
    label.htmlFor = "test1";
    assert_not_equals(label.control, document.getElementById("test1"),
                      "A label element not in a document should not label an element in a document.");
    document.body.appendChild(label);
```

Seems more like a generic problem with creating elements and adding/removing them from the dom.

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -48,27 +47,35 @@ function sendClickToAssociatedNode(node) {
 }
 
 class HTMLLabelElementImpl extends HTMLElementImpl {
-  _activationBehavior() {
+  get control() {
     if (this.hasAttribute("for")) {
       const node = this.ownerDocument.getElementById(this.getAttribute("for"));
-      if (node && isLabelable(node) && !isDisabled(node)) {
-        sendClickToAssociatedNode(node);
+      if (node && isLabelable(node)) {
```
TimothyGu: The spec [says](https://html.spec.whatwg.org/multipage/forms.html#attr-label-for):

> If the [[`for`](https://html.spec.whatwg.org/multipage/forms.html#attr-label-for)] attribute is specified and there is an element in the [tree](https://dom.spec.whatwg.org/#concept-tree) whose [ID](https://dom.spec.whatwg.org/#concept-id) is equal to the value of the [`for`](https://html.spec.whatwg.org/multipage/forms.html#attr-label-for) attribute, and the first such element in [tree order](https://dom.spec.whatwg.org/#concept-tree-order) is a [labelable element](https://html.spec.whatwg.org/multipage/forms.html#category-label), then that element is the [`label`](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) element's [labeled control](https://html.spec.whatwg.org/multipage/forms.html#labeled-control).

You need to check if the current element is in the same tree as node. A simple algorithm would be

```js
function hasCommonAncestor(a, b) {
  const ancestors = new Set(domSymbolTree.ancestorsIterator(a));
  for (const ancestor of domSymbolTree.ancestorsIterator(b)) {
    if (ancestors.has(ancestor)) {
      return true;
    }
  }
  return false;
}
```

This should fix the remaining test failure.

FrecksterGIT:
Added support for HTMLInputElement.labels.
All control-attribute related tests are passing now.

Only got 3 failing tests left in label-attributes.html.

TimothyGu:
@FrecksterGIT can you enable the passing `control` attribute tests in this pull request as well?

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -29,6 +28,11 @@ function isLabelable(node) {
   return false;
 }
 
+function hasCommonAncestor(a, b) {
+  const ancestors = new Set(domSymbolTree.ancestorsIterator(a));
+  return domSymbolTree.ancestorsToArray(b).some(ancestor => ancestors.has(ancestor));
```
TimothyGu: Can you use `ancestorsIterator` as I originally proposed? That would obviate the need to walk through the entire ancestor chain for `b` if not needed.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -316,6 +320,16 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
   }
 
+  get labels() {
+    const root = this.ownerDocument;
+    return NodeList.create([], {
+      element: root,
+      query: () => domSymbolTree.treeToArray(root, {
+        filter: node => node.tagName === "LABEL" && node.control === this
+      })
+    });
```
TimothyGu: Please always implement things according to the [spec](https://html.spec.whatwg.org/multipage/forms.html#dom-lfe-labels), which specifically says that the same object must be returned on separate invocations of `labels` getter, and that the `labels` getter should return null in case the type of this `input` element is Hidden.

FrecksterGIT:
```
html/semantics/forms/the-input-element
      √ input-labels.html
    html/semantics/forms/the-label-element
      1) label-attributes.html
```
Still two failing tests in label-attributes.html:

1. document.getElementById(...).attachShadow is not a function

> Web Components / Shadow Dom isn't supported by jsdom.

2. Failed in "A labelable element is moved to iframe.":
assert_equals: The number of labels should be 2 after the labelable element is moved to iframe. expected 2 but got 0

> not sure how to tackle this one.

I haven't figured out a way to disable single wpt-tests in one test-file. Guess it's not even possible?
And is there some documentation for debugging wpt? That took longer than it should have.

FrecksterGIT:
Refactored labels logic into helpers/form-controls.js and added labels support to all labelable elements i could find:

https://html.spec.whatwg.org/multipage/forms.html#category-label

- button
- input (if the type attribute is not in the Hidden state)
- meter
- output
- progress
- select
- textarea

```
  web-platform-tests html/semantics/forms/the-label-element labelable-elements.html:
     Error: Failed in "Check if the keygen element is not a labelable element": 
assert_not_equals: got disallowed value Element node <keygen id="testkeygen"></keygen>
```
Keygen is deprecated and not part of the living spec.

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -2,31 +2,16 @@
...
 const { domSymbolTree } = require("../helpers/internal-constants");
-const { isDisabled } = require("../helpers/form-controls");
+const NODE_TYPE = require("../node-type");
+const { isLabelable, isDisabled } = require("../helpers/form-controls");
 
-function isLabelable(node) {
-  // labelable logic defined at: https://html.spec.whatwg.org/multipage/forms.html#category-label
-  if (node.nodeType !== 1) {
-    return false;
+function getRootNode(node) {
```
TimothyGu: You can use the same function in traversal.js.

lib/jsdom/living/helpers/form-controls.js
```diff
@@ -60,6 +64,50 @@ exports.normalizeToCRLF = string => {
...
+  }
+
+  return false;
+};
+
+exports.getLabelsForLabelable = labelable => {
+  if (!exports.isLabelable(labelable)) {
+    return null;
+  }
+  if (!labelable._labels) {
```
TimothyGu: Can you declare `this._labels = undefined` in the constructors of all classes that use this function?

TimothyGu:
It also seems like there are some outstanding test failures: https://travis-ci.org/jsdom/jsdom/jobs/381198906

Some of the failures seem to have to do with `attachShadow` which we don't implement. In that case, you can copy the file from upstream into our to-upstream folder, commenting out the features we don't yet implement, and change the filename so that it has `dont-upstream` in the name. Like [this](https://github.com/jsdom/jsdom/blob/master/test/web-platform-tests/to-upstream/hr-time/timeOrigin-dont-upstream.html).

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
FrecksterGIT: I knew i missed something there.

lib/jsdom/living/helpers/form-controls.js
FrecksterGIT: Initialized `this._labels = null` everywhere.

PR #2218, opened 2018-05-05
Fix Content-Type for Documents created through frames and createDocument()

Zirro:
Documents created by frames and the `createDocument()`-method were not receiving the correct Content-Type. This pull request resolves this and enables several tests.

While I was fixing the latter, I noticed that the full namespaces were rather unwieldy to use in comparisons. For improved readability, I decided to create a helper with constants based on the [Namespaces](https://infra.spec.whatwg.org/#namespaces) section of the Infra Standard and replaced the strings where it made sense.


domenic:
Hmm this has a lot of overlap with my #2195 but is done better, so let's merge this and I'll work on rebasing that.

Zirro:
@domenic aha, I missed that your pull request addressed that issue too. I got started by looking at the `contentype_{format}.html` tests and continued with the one for `createDocument()` afterwards. Glad to hear you like it, though.

PR #2224, opened 2018-05-11
chore: clone last 5 commits

DanielRuf:


TimothyGu:
Is there an explicit reason for this? The cloning time [hardly](https://travis-ci.org/jsdom/jsdom/jobs/376526684#L407) looks like a performance bottleneck in testing, and Travis CI already only clones 50 commits.

DanielRuf:
YMMV, it depends on many things like the size of the diffs. So far I saw decreased clone times in most (big) projects with this setting.

TimothyGu:
Sounds like the cloning time was cut down by... 1 second. I think it'll be more fruitful if you could find a way to shallow clone the submodules, which do take a long time to clone.

domenic:
I don't think we need to add non-default stuff to the build process for this.

PR #2225, opened 2018-05-11
chore: add Node.js 10

DanielRuf:


DanielRuf:
`mocha-sugar-free` is not compatible so we need a replacement.

DanielRuf:
> The other comes from the Blob-slice.html test which appears to be affected by a change in the vm

Yes, seems so. At least the mentioned mocha package throws the error.
Ok, should I close this PR or wait until we got it fixed and then trigger a new build?

Zirro:
We could probably keep this open if we intend to merge it later, but I'm not sure that we want to test in all three versions at the same time due to CI limits. @Domenic?

DanielRuf:
Probably just 8 (LTS) and 10 (stable).

domenic:
Ended up doing this as part of https://github.com/jsdom/jsdom/commit/1c792e1a1bf9f2ff842ccef749b01f95b60bd364 where we're moving the minimum version to v8.

PR #2226, opened 2018-05-11
chore: use yarn cache

DanielRuf:


domenic:
It's not clear why fetching from the Travis CI servers is better than fetching from the Yarn servers. Some data would be helpful for justifying this change.

DanielRuf:
> It's not clear why fetching from the Travis CI servers is better than fetching from the Yarn servers. Some data would be helpful for justifying this change.

This is for use the build cache and improves the build times.
Travis and yarn know if there are changes to the dependencies by creating checksums on lockfiles and directories and either loads from cache (is faster) or downloads the new dependencies.

DanielRuf:
Because there is no persistent filesystem except the caches.

domenic:
My understanding is that the caches are not a persistent filesystem. They are stored in separate storage buckets and fetched each time onto the container running the CI job.

DanielRuf:
> My understanding is that the caches are not a persistent filesystem. They are stored in separate storage buckets and fetched each time onto the container running the CI job.

They are in the same network (often EC2 instances) which is much faster.

Wait a moment.

DanielRuf:
Hm, not yet sure why the results differ so much here (unlike in other projects).

Closing this PR for now as there might be other performance issues with CI.

PR #2229, opened 2018-05-14
Use NWSAPI

Zirro:
This replaces nwmatcher with [nwsapi](https://github.com/dperini/nwsapi) which resolves many known issues, supports several new selectors and is significantly faster.

Although a beta version of nwsapi is currently used, all tests should pass and this pull request is ready for review. A stable version of nwsapi is expected to be published later this week.

### DIFF
--- a/lib/jsdom/living/helpers/selectors.js
+++ b/lib/jsdom/living/helpers/selectors.js
@@ -1,28 +1,40 @@
 "use strict";
 
+const DOMException = require("domexception");
+const nwsapi = require("nwsapi");
 const idlUtils = require("../generated/utils");
-const nwmatcher = require("nwmatcher/src/nwmatcher-noqsa");
 
 exports.matchesDontThrow = (elImpl, selector) => {
   const document = elImpl._ownerDocument;
 
-  if (!document._nwmatcherDontThrow) {
-    document._nwmatcherDontThrow = nwmatcher({ document });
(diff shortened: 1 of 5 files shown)


lib/jsdom/living/helpers/selectors.js
```diff
@@ -1,28 +1,40 @@
...
 // stages of Document initialization, there is no documentElement present yet.
-exports.addNwmatcher = parentNode => {
+exports.addNwsapi = parentNode => {
   const document = parentNode._ownerDocument;
 
-  if (!document._nwmatcher) {
-    document._nwmatcher = nwmatcher({ document });
-    document._nwmatcher.configure({ UNIQUE_ID: false, LOGERRORS: false, SVG_LCASE: true });
+  if (!document._nwsapi) {
+    document._nwsapi = nwsapi({
```
Sebmaster: Do we have to put it on the document like this? Can we just store the reference in private state?

Zirro: This seems worth exploring separately, but with my time being rather limited right now I'd like to keep this pull request to just the changes needed for the switch.

Zirro:
The stable version of nwsapi is out, and this pull request has been updated to use it! I would like to give a big thanks to @dperini, who has worked long hours together with me to resolve all remaining issues and make it possible to use in jsdom.

With selector performance that not only rivals but often beats the native implementations in browsers, it's a very interesting project and something we're fortunate to be able to use here.

PR #2232, opened 2018-05-17
Minor tests script updates

caub:
minor changes on tests

### DIFF
--- a/test/web-platform-tests/start-wpt-server.js
+++ b/test/web-platform-tests/start-wpt-server.js
@@ -6,8 +6,9 @@ const dns = require("dns");
 const childProcess = require("child_process");
 const { inBrowserContext } = require("../util.js");
 const requestHead = require("request-promise-native").head;
-const dnsLookup = promisify(dns.lookup);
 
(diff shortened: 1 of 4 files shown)


test/web-platform-tests/start-wpt-server.js
```diff
@@ -69,6 +69,6 @@ function pollForServer(url) {
     })
     .catch(err => {
       console.log(`WPT server at ${url} is not up yet (${err.message}); trying again`);
-      return q.delay(500).then(() => pollForServer(url));
+      return new Promise(r => setTimeout(r, 500)).then(() => pollForServer(url));
```
TimothyGu: You could have a `promisify(setTimeout)` as well.

TimothyGu:
Seems like this broke Node.js v6.x due to a lack of `util.promisify`...

caub:
oh :(, let's temporarily close and reopen in 1 year when we can skip node v6 maybe

domenic:
I find most of these changes less clear than the original. Eliminating the q dependency would be nice if we could, but we can't, so yeah, let's close this.

PR #2234, opened 2018-05-18
Fix HTMLSourceElement.srcset getter

Zirro:
The getter for the `srcset` attribute on `<source>` elements would attempt to return the value of `cite` attributes instead. Thanks to @Treora for noticing and [letting us know](https://github.com/jsdom/jsdom/commit/d4d1cfea86d894ed665911dca0b39b626df56c0b#r29018531) about this.


domenic:
I suppose the tests are locked up in the giant reflection.html test; oh well.

PR #2235, opened 2018-05-19
Upgrade cssstyle >= 0.3.1 < 0.4.0

jsakas:
Greetings. I use jsdom a lot in testing applications with jest. We have run into issues when our application code cannot get style values that are not implemented in cssstyle. Like the css3 unit `rem`, for example.

```js

const { JSDOM } = jsdom;
const doc = new JSDOM(`
  <p style="width:10px" id="px">Hello world</p>
  <p style="width:10rem" id="rem">Hello world</p>
`).window.document;

assert.strictEqual(doc.querySelector('#px').style.width, '10px'); // ok

assert.strictEqual(doc.querySelector('#rem').style.width, '10rem');
// AssertionError [ERR_ASSERTION]: '' === '10rem'
```

I have taken ownership of the cssstyle github project and bumped it to 0.3.1 with a few PRs that had been open for awhile: https://github.com/jsakas/CSSStyleDeclaration/releases/tag/0.3.1

Ultimately the goal would be to upgrade cssstyle to use information gathered from csswg, and then release 1.0.0.

If the owners of jsdom are willing to continue upgrading this package we can have a better implementation of `CSSStyleDeclaration` for testing!

Thanks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "acorn-globals": "^4.1.0",
     "array-equal": "^1.0.0",
     "cssom": ">= 0.3.2 < 0.4.0",
-    "cssstyle": ">= 0.2.37 < 0.3.0",
+    "cssstyle": ">= 0.3.1 < 0.4.0",
     "data-urls": "^1.0.0",
     "domexception": "^1.0.0",
     "escodegen": "^1.9.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -917,9 +917,9 @@ cssom@0.3.x, "cssom@>= 0.3.2 < 0.4.0":
   version "0.3.2"
   resolved "https://registry.yarnpkg.com/cssom/-/cssom-0.3.2.tgz#b8036170c79f07a90ff2f16e22284027a243848b"
 
-"cssstyle@>= 0.2.37 < 0.3.0":
-  version "0.2.37"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "acorn-globals": "^4.1.0",
     "array-equal": "^1.0.0",
     "cssom": ">= 0.3.2 < 0.4.0",
-    "cssstyle": ">= 0.2.37 < 0.3.0",
+    "cssstyle": "^0.3.1",
```
TimothyGu: `^` doesn't really work with `0.` versions. `>= 0.3.1 < 0.4.0` should work just fine.

domenic: We should probably stick to exact versions actually until we have a 1.0.0 that's guaranteed to not make breaking changes...

jsakas:
@TimothyGu - amended, thanks!

package.json
jsakas: @domenic - That is what I had but @TimothyGu requested a change. I'm happy to pin it at 0.3.1 again.

There won't be any breaking API changes until I have time to do a re-write. And even then I think the API will probably be the same, I am just going to generate the properties in more scalable way.

TimothyGu: With @jsakas's guarantee, I think having a version range is fine.

domenic:
I'm not very comfortable with depending on a version range for a pre-1.0.0 module, but I guess that's what we were doing before, so we can go ahead and merge this.

I'd strongly suggest publishing a 1.0.0 as soon as possible, and not waiting to batch up changes, so that we can have a better guarantee.

PR #2237, opened 2018-05-21
Fix 'WebIDL2 is not defined'-error in tests

Zirro:
Figured I'd finally see if I could find the cause of this issue. Turns out that a path that was meant to be rewritten by the WPT server got overridden and used by the `/resources/`-condition here before that happened. The fix performs the same replacement as the [WPT server](https://github.com/w3c/web-platform-tests/blob/master/tools/serve/serve.py#L271).

This change alone hasn't made any tests pass since most of them also depend on `fetch()`, but I suppose it'll come in handy when we eventually do support it - assuming the rewrite is still in use at that point.

### DIFF
--- a/test/web-platform-tests/run-single-wpt.js
+++ b/test/web-platform-tests/run-single-wpt.js
@@ -61,7 +61,11 @@ function createJSDOM(urlPrefix, testPath) {
         } else if (resource.url.pathname.startsWith("/resources/")) {
           // When running to-upstream tests, the server doesn't have a /resources/ directory.
           // So, always go to the one in ./tests.
-          const filePath = path.resolve(__dirname, "tests" + resource.url.pathname);
+          // The path replacement accounts for a rewrite performed by the WPT server:
+          // https://github.com/w3c/web-platform-tests/blob/master/tools/serve/serve.py#L271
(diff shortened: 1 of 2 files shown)


test/web-platform-tests/run-single-wpt.js
```diff
@@ -61,7 +61,11 @@ function createJSDOM(urlPrefix, testPath) {
         } else if (resource.url.pathname.startsWith("/resources/")) {
           // When running to-upstream tests, the server doesn't have a /resources/ directory.
           // So, always go to the one in ./tests.
-          const filePath = path.resolve(__dirname, "tests" + resource.url.pathname);
+          // The path replacement accounts for a rewrite performed by the WPT server:
+          // https://github.com/w3c/web-platform-tests/blob/master/tools/serve/serve.py#L271
+          const filePath = path.resolve(__dirname, "tests" + resource.url.pathname)
+            .replace("/resources/WebIDLParser.js", "/resources/webidl2/lib/webidl2.js");
```
Sebmaster: Maybe we should check this with == instead to make it clearer this only a full match?

Zirro: I tested that approach first, but decided against it as I felt that the extra condition and reassignment made the code less readable and more complex than this version.

PR #2238, opened 2018-05-21
fix: pass vm scrip options through

SimenB:
In order to prevent scripts from potentially running forever, it'd be great to be able to pass a `timeout` to it.

Docs: https://nodejs.org/api/vm.html#vm_script_runincontext_contextifiedsandbox_options

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,7 +307,7 @@ console.log(dom.nodeLocation(imgEl));    // { startOffset: 13, endOffset: 32 }
 
 Note that this feature only works if you have set the `includeNodeLocations` option; node locations are off by default for performance reasons.
 
-### Running vm-created scripts with `runVMScript(script)`
+### Running vm-created scripts with `runVMScript(script[, options])`
 
 The built-in `vm` module of Node.js allows you to create `Script` instances, which can be compiled ahead of time and then run multiple times on a given "VM context". Behind the scenes, a jsdom `Window` is indeed a VM context. To get access to this ability, use the `runVMScript()` method:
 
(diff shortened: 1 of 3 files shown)


TimothyGu:
Can you update the corresponding documentation in README?

SimenB:
Done! 🙂

domenic:
Why not use the options argument to the `Script` constructor?

SimenB:
Oh, that's a very good point. That works for my use case then. Feel free to close if you don't want this 🙂

SimenB:
To my eyes it looks like `new Script` doesn't care about the timeout, and the docs are wrong:

https://github.com/nodejs/node/blob/9461f327f50c9885508392938053a62985d13259/lib/vm.js#L45-L52

SimenB:
Done! https://github.com/nodejs/node/issues/20982

SimenB:
This is apparently faulty docs, so the PR is back on the table, I think :)

README.md
```diff
@@ -282,6 +282,8 @@ dom.window.ran === 3;
 
 This is somewhat-advanced functionality, and we advise sticking to normal DOM APIs (such as `window.eval()` or `document.createElement("script")`) unless you have very specific needs.
 
+`runVMScript` also takes an `options` object as its second argument. See the [Node docs](https://nodejs.org/api/vm.html#vm_script_runincontext_contextifiedsandbox_options) for details.
```
domenic: Some nits:

- I've tried to stick with adding parens for function/method names, so `runVMScript()`.
- I've tried to stick to Node.js instead of Node.

Also maybe we should update the header to `runVMScript(script[, options])`? Dunno, could go either way.

SimenB:
@domenic sorry, I completely forgot about this! I've rebased, addressed your comments on the docs and added a test 🙂

Issue #1611, opened 2016-09-23
9.5 doesn't seem to allow querying into noscript


daniel-white:
I'm trying to use jsdom to munge with an html file to prepare it for a CDN. However, I'm not able to select into an element inside of a `noscript`. I have `ProcessExternalResources` set to `false`.

```
<noscript><img href="..." /></noscript>
```

```
var document = jsdom('path', {features: {ProcessExternalResources: false}});
var els = document.querySelectorAll('[href]');
els.length === 0;
```

Sebmaster:
From my testing in FF this does indeed seem like a bug.

Zirro:
For future reference, this looks to be due to the same reason as issue #1802 and can be resolved when https://github.com/inikulin/parse5/issues/192 has been merged.

Issue #1802, opened 2017-04-06
[DOMParser] Spec behavior for noscript tags


xtuc:
You can find the spec for DOMParser [here](https://w3c.github.io/DOM-Parsing/#the-domparser-interface)

The following example produce an incorrect output using JSDOM.

```js
console.log(
    (new window.DOMParser())
        .parseFromString('<noscript><div>foo</div><img /></noscript>', 'text/html')
        .body
        .innerHTML
);
```

#### Chrome

```html
<div>foo</div><img>
```

#### JSDom

Nothing.

If this is related to JSDom script configuration, I think (based on DOMParser spec) it should override the config.

I'm happy to help with a PR.

Edit:

In the HEAD:

```js
console.log(
    (new window.DOMParser())
        .parseFromString('<noscript><div>foo</div><img /></noscript>', 'text/html')
        .head // <---
        .innerHTML
);
```

#### JSDom

```html
<noscript><div>foo</div><img /></noscript>
```

#### Chrome

```html
<noscript></noscript>
```

inikulin:
`DOMParser` requires `scripting flag` to be disabled. However, currently [parse5](https://github.com/inikulin/parse5) (HTML parser used internally by jsdom) always acts as if we are in interactive user agent. Therefore, we need to implement `scripting flag` and related tree construction insertion modes in parse5 first.

domenic:
Of course, I think we are still behind by a major parse5 release or two, so getting that upgraded might be a prerequisite :(

inikulin:
@domenic It shouldn't be hard to implement. I guess we'll be able to lend it in `v4`. I've created an issue on parse5 side: https://github.com/inikulin/parse5/issues/192

xtuc:
Thanks for the fast answer.

`DOMParser` says that scripts are marked as executable. `scripting flag` configuration shouldn't alter this behavior.

inikulin:
> DOMParser says that scripts are marked as executable. scripting flag configuration shouldn't alter this behavior.

It says completely opposite: 
> The scripting flag must be set to "disabled".
NOTE
script elements get marked unexecutable and the contents of noscript get parsed as markup.

xtuc:
Yes absolutely, sorry. That's what I wanted to write.

inikulin:
@domenic 
> Of course, I think we are still behind by a major parse5 release or two, so getting that upgraded might be a prerequisite :(

[...]

PR #2242, opened 2018-05-23
Use parse5 v5.0.0

Zirro:
[Version 5.0.0](https://github.com/inikulin/parse5/blob/master/docs/version-history.md) of parse5 was just released and it implements the `scriptingEnabled` flag which may allow us to fix issues https://github.com/jsdom/jsdom/issues/1611 and https://github.com/jsdom/jsdom/issues/1802. This pull request only includes the changes necessary for the upgrade at the moment, though.

It also includes some breaking changes to the location info, but I'm happy to report that the only fix required because of the temporary monkey-patch was the replacement of underscores with dashes in a filename.

### DIFF
--- a/test/api/options-run-scripts.js
+++ b/test/api/options-run-scripts.js
@@ -46,7 +46,7 @@ describe("API: runScripts constructor option", () => {
         virtualConsole.on("jsdomError", err => {
           try {
             assert.strictEqual(err.type, "unhandled exception");
-            assert(err.detail.stack.includes("at about:blank:2"));
+            assert.isTrue(err.detail.stack.includes("at about:blank:2"));
             resolve();
           } catch (actualErr) {
             reject(actualErr);
--- a/lib/api.js
+++ b/lib/api.js
@@ -65,11 +65,11 @@ class JSDOM {
   }
 
   nodeLocation(node) {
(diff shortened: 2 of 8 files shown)


test/api/options-run-scripts.js
```diff
@@ -46,7 +46,7 @@ describe("API: runScripts constructor option", () => {
         virtualConsole.on("jsdomError", err => {
           try {
             assert.strictEqual(err.type, "unhandled exception");
-            assert(err.detail.stack.includes("at about:blank:2"));
+            assert.isTrue(err.detail.stack.includes("at about:blank:2"));
```
Zirro: I made this change since the error given from the failing `assert()` was a bit cryptic, and in order to be consistent with the rest of the tests.

domenic:
Awesome work! Can you update the readme example node location objects too?

Zirro:
It looks like the change doesn't affect the properties we currently use in the README examples, and including all of them would likely be too verbose.

This should probably be merged when its time for the next major release, even if we may not have implemented use of the `scriptingEnabled` yet.

piuccio:
I'd like to politely ask for an update on this PR. I looked into the milestones and a bunch of issues but couldn't find any information on what's holding a major release back.
Can anyone provide a rough estimate of when it'll be looked at again?

PR #2243, opened 2018-05-23
Enable various selector tests

### LINKED ISSUE #1832: Throws on Element.prototype.matches call with pseudo-element selector
### Basic info:

- **Node.js version:** 6.9.1
- **jsdom version:** 10.0.0 (also present in 9.x)

### Minimal reproduction case

```js
const dom = new JSDOM(`
  <p>text</p>
`);

dom.window.document.body.firstChild.matches('p:after');
```

Throws:
```
SyntaxError
    at HTMLParagraphElementImpl.<anonymous> (/Users/munter/git/assetgraph/node_modules/jsdom/lib/jsdom/living/nodes/Element-impl.js:514:11)
    at HTMLParagraphElementImpl.<anonymous> (/Users/munter/git/assetgraph/node_modules/jsdom/lib/jsdom/utils.js:113:45)
    at HTMLParagraphElement.matches (/Users/munter/git/assetgraph/node_modules/jsdom/lib/jsdom/living/generated/Element.js:278:29)
    at repl:1:37
    at sigintHandlersWrap (vm.js:22:35)
    at sigintHandlersWrap (vm.js:96:12)
    at ContextifyScript.Script.runInThisContext (vm.js:21:12)
    at REPLServer.defaultEval (repl.js:313:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
```

Returns `true` or `false`

### LINKED ISSUE #2204: SyntaxError thrown by querySelector when selector contains apostrophe
### Basic info:

- **Node.js version:** 9.11.0 
- **jsdom version:** 11.8.0 

### Minimal reproduction case

```js

const dom = new JSDOM(`<!DOCTYPE html><p>Hello <a href="#" onclick="window.open('http://example.com/')">world</a></p>`);

console.log(dom.window.document.querySelector(`a[onclick^="window.open('http://"]`).textContent);

console.log(dom.window.document.querySelector(`a[onclick^="window.open("]`).textContent)
// prints "world"

console.log(dom.window.document.querySelector(`a[onclick^="window.open('"]`).textContent)
// SyntaxError: Invalid or unexpected token
//     at DocumentImpl.<anonymous> (.../node_modules/jsdom/lib/jsdom/living/nodes/ParentNode-impl.js:70:11)
//     at DocumentImpl.querySelector (.../node_modules/jsdom/lib/jsdom/utils.js:123:45)
//     at Document.querySelector (.../node_modules/jsdom/lib/jsdom/living/generated/Document.js:689:45)
```

https://jsbin.com/nokegupeye/edit?js,console

domenic:
38b868b0ca2a127a55c8a59a0e1320e032074118 enabled the next-gen nwsapi selector engine, which enables us to pass a few more tests.

Closes #2204. Closes #1832.

---

This reveals a few failures in nwsapi, or perhaps our jsdom integration. It would be good to fix them and then merge this:

```
 1) web-platform-tests dom/nodes Element-matches.html:
     Error: Failed in "In-document Element.matches: Universal selector, matching all descendants of the specified refere
nce element (with refNode Element): *":
assert_false: The element #universal should not match the selector. expected false got true

(same for `Element-webkitMatchesSelector.html`)

 2) web-platform-tests dom/nodes ParentNode-querySelector-All.html:
     Error: 2 errors in test:

Failed in "Document.querySelectorAll: ID selector, matching multiple elements with duplicate id: #id-li-duplicate":
assert_equals: The method should return the expected number of matches. expected 4 but got 1
```

Additionally, there may be other tests we could enable; @Zirro and @dperini have been working on them.


dperini:
@domenic first of all thank you for reviewing this pull request.

The first error you mention in web-platform-tests dom/nodes Element-matches.html:

    In-document Element.matches: Universal selector, matching all descendants of the specified reference element (with refNode Element): *
 is present also in all browsers I could test: Firefox, Safari, Chrome and Opera so I thought this should be a problem related to the test itself, not a `jsdom` nor a `nwsapi`problem.

For the other 2 errors in web-platform-tests dom/nodes ParentNode-querySelector-All.html:

    Failed in "Document.querySelectorAll: ID selector, matching multiple elements with duplicate id: #id-li-duplicate"
they don't show up neither in  my own tests in browsers nor in `jsdom` test that I have run locally.
I have executed the tests by running:

     npm run test
from inside a fresh clone of the `jsdom` git. I get the following results:

      3723 passing (5m)
      775 pending
However tomorrow morning I will rerun all of these tests again and comment here on the results.

dperini:
I have found the difference I had in my local copy, the ParentNode-querySelector-All.html test was disabled. But it works and pass all the 1855 tests on browsers, so it might be an asynchronous quirk.

I will have a debug session with the help of @Zirro later today.

We should also consider that duplicated IDs is a high edge case (and a risky environment to say the least) but I am sure that having solved this in the browser will let us solve in `jsdom` too.

test/web-platform-tests/to-run.yaml
```diff
@@ -753,6 +752,26 @@ template-element/template-content-hierarcy.html: [fail, probably a combination o
...
+pseudo-classes/checked-type-change.html: [fail, CSS computed style computation]
+pseudo-classes/dir.html: [fail, nwsapi doesn't support dir]
+pseudo-classes/dir01.html: [fail, nwsapi doesn't support dir]
+pseudo-classes/focus-autofocus.html: [fail, Unknown]
+pseudo-classes/focus.html: [fail, Unknown]
+pseudo-classes/indeterminate-radio.html: [fail, CSS computed style computation]
+pseudo-classes/indeterminate-type-change.html: [fail, CSS computed style computation]
+pseudo-classes/inrange-outofrange-type-change.html: [fail, CSS computed style computation]
+pseudo-classes/inrange-outofrange.html: [fail, Unknown]
+pseudo-classes/link.html: [fail, Unknown]
```
Zirro: This test also [fails the same way](https://wpt.fyi/results/html/semantics/selectors/pseudo-classes/link.html) in browsers. It appears to have relied on elements in `<head>` staying the same despite being shared across tests, and a mass-change inevitably broke it at one point.

domenic: Nice catch, want to file a bug on web-platform-tests/wpt?

Zirro: Done: https://github.com/web-platform-tests/wpt/issues/11231

dperini: @domenic @Zirro the other error noticed above in Element-matches.html:

    In-document Element.matches: Universal selector, matching all descendants of the specified reference element (with refNode Element): *

seems to be due to a missing functionality in qS(), qSA() and matches() methods related to the current inability to pass a `refNode` as extra argument that was considered at some point in the past, see here: https://www.w3.org/TR/selectors-api2/#interface-definitions

Current QS API do not include such functionality (passing more params) that I am aware of.

The only current way I know to pass a `refNode` is through the use of the `:scope` pseudo-class so the only fix for current browser would be changing the selector of that test:

 from '*'
 to ':scope *'

I haven't tested it completely yet but:

    document.querySelector('#universal').matches('*'); // return true
    document.querySelector('#universal').matches(':scope *') // return false

Note that this is commented in the file `selectors.js` from Line 415 to Line 445.
More testing and checking are needed, maybe also asking to the test author..

domenic: Yep, this is a test bug. Being fixed in https://github.com/web-platform-tests/wpt/pull/11229

domenic:
Currently (with nwsapi v2.0.7) getting failures where things should throw, but do not.  See Element-matches.html and ParentNode-querySelector-All.html.

dperini:
@domenic / @Zirro  could you please check Line#46 of the `Element-matches.js` file ?

Currently it says to only execute the `invalid` tests on the Element interface:

    if (root.nodeType === root.ELEMENT_NODE) {

but maybe the WPT tests expect that the `invalid` tests be executed also on both DOCUMENT_FRAGMENT_NODE and DOCUMENT_NODE.

I tested by either removing the conditional completely or by changing it to:

    if (root.nodeType !== root.DOCUMENT_NODE) {

and in both cases that seems to fix the failures you were getting from running the WPT tests.
My knowledge is a bit limited on these tests and their inner working so I need your help confirming.
Thank you in advance for the help.

domenic:
Hmm, I don't quite understand. That line ensures the tests only run for element node roots, right? So why would running it for _more_ node types cause _less_ failures in the tests?

Note that browsers all pass the tests (or mostly do): https://wpt.fyi/results/dom/nodes/Element-matches.html So it seems unlikely it's a test bug. Could still be a jsdom bug, of course... maybe @Zirro's #2312 would help? I should probably merge that before doing any more tests.

dperini:
@domenic ... as I said I am still trying to guess how these test are organized and learn.

However my debugging sessions revealed that this was the most probable point of failure and that coincides with what you said in the comment above about "getting failures where things should throw, but do not".

It might be that those tests are expected to fail in `runInvalidSelectorTestMatches`, so if that conditional line does skip those tests no exceptions are thrown for tests on the other node types and those may be considered failed tests.

These are the lines from "Element-matches-init.js" that are invoking the mentioned function:

    runInvalidSelectorTestMatches(method, "Document", doc, invalidSelectors);
    runInvalidSelectorTestMatches(method, "Detached Element", detached, invalidSelectors);
    runInvalidSelectorTestMatches(method, "Fragment", fragment, invalidSelectors);
    runInvalidSelectorTestMatches(method, "In-document Element", element, invalidSelectors);

the conditional line I pointed out above prevents the execution of the tests present in "selectors.js" for two types of nodes (11 - DOCUMENT_FRAGMENT_NODE and 9 - DOCUMENT_NODE) the reason might be these nodes do not have the `matches()` method implemented.

There is definitely something wrong in my reasoning, since, as you pointed out, most browsers are already passing all these tests but I have run out of ideas about these failures. I will try to run these same tests in the browsers tomorrow and I will let you know the outcome.

Issue #2223, opened 2018-05-09
JSDom can't set and retrieve CSS styles


OlofT:
### Basic info:

- **Node.js version:** v10
- **jsdom version:** v11.10

### Minimal reproduction case

As the title says, I can't set css styles, and get them later. It's sad since it means you can't use JSDom to inline your CSS. 

At least its easy to reproduce:

```js
const { JSDOM } = require("jsdom");

const options = {
url: "http://localhost/",
		referrer: "http://aggressive.se",
		contentType: "text/html",
		userAgent: "AggressiveDev",
		includeNodeLocations: false,
		resources: "usable",		
		pretendToBeVisual: true,
		runScripts: "dangerously",
};
const dom = new JSDOM(`<body></body>`, options);

window = dom.window
document = window.document;
var styleSheet = document.createElement("style");
styleSheet.setAttribute('type', 'text/css');
var head = document.getElementsByTagName("head")[0];
head.appendChild(styleSheet);
styleSheet = document.styleSheets[document.styleSheets.length - 1];

styleSheet.insertRule(".workingClass {  dontAddThis }", 0);
var rule = styleSheet.cssRules[0];
var style = rule.style;
style.borderColor = "green";

console.log("rule.cssText gets selector but no css");
console.log(rule.cssText);
console.log("should be:\n.workingClass {border-color: green;}");

console.log("but using setProperty works (almost)");
style.setProperty("borderColor", "green")
console.log(rule.cssText);
console.log("cssText is wrong (still js keys)");

```

### How does similar code behave in browsers?
(can be copy/pasted into browser after removing "window = dom.window") or jsbin:
https://jsbin.com/sumadul/edit?js,console

jcubic:
I have a similar problem, `insertRule` can't be modified:

```typescript
const $style = document.createElement('style');
document.head.appendChild($style);
const sheet = $style.sheet!;

const style = {
   color: 'red'
};

const index = sheet.cssRules.length;
sheet.insertRule(`body { }`, index);
const rule = sheet.cssRules[index];
if (rule instanceof CSSStyleRule) {
    Object.assign(rule.style, style);
}
console.log(rule.cssText);
// prints body {}
// should be body { color: red; }
```

This is open for almost 6 years. Will you accept a PR if I can implement the missing feature of updating `rule.cssText`?

jcubic:
I was investigating and it seems that the problem is that when the CSS property is not present in the rule when inserting you can't change it afterwards:

```typescript
[...]

PR #2250, opened 2018-05-27
Fix for css-tags to get its attributes, now you can see @media or similar for each css (a must when inlining).

OlofT:
Fixes step 1 of the problem that you cannot inline CSS. Now, you can at least see media-queries and other attributes for the CSS-elements.

Issue: #2223


OlofT:
Now there are tests for the css bugs I found so far. The first test succeeds since it is fixed by this commit.

There are no mention of "lint" in Contributing.md

domenic:
See contributing.md; that is not an acceptable kind of test. It needs to be a web platform test

PR #2257, opened 2018-06-05
Run failing tests to confirm their status

Zirro:
A bit of work remains to appropriately handle the `execution-timing` tests on Node.js v6 where they take ten seconds each before they timeout, compared to later versions where they fail quickly.


Zirro:
This is ready for review now.

test/web-platform-tests/run-single-wpt.js
```diff
@@ -118,9 +124,16 @@ function createJSDOM(urlPrefix, testPath) {
           errors.push(...doneErrors);
           errors.push(...unhandledExceptions);
 
-          if (errors.length === 1) {
+          if (errors.length === 0 && expectFail === true) {
```
Zirro: Just as I request a review, I realise that I ought to simplify these to `expectFail` and `!expectFail`. Will do tomorrow.

PR #2267, opened 2018-06-18
Add the `storageQuota` option

Zirro:
Fixes the problem noticed by @fstoerkle in https://github.com/jsdom/jsdom/commit/3afbc0feb7b2b4ac4c84530ea6af90e1f79a88ee#r29395675. Thanks!

It's probably not appropriate to write a WPT for this given that the quota merely is a recommendation.

### DIFF
--- a/test/api/options.js
+++ b/test/api/options.js
@@ -289,4 +289,74 @@ describe("API: constructor options", () => {
       });
     });
   });
+
+  describe("storageQuota", () => {
+    describe("not set", () => {
+      it("should be 5000000 bytes by default", () => {
+        const { localStorage, sessionStorage } = (new JSDOM(``, { url: "https://example.com" })).window;
+        const dataWithinQuota = "0".repeat(4000000);
+
+        localStorage.setItem("foo", dataWithinQuota);
+        sessionStorage.setItem("bar", dataWithinQuota);
+
+        assert.strictEqual(localStorage.foo, dataWithinQuota);
+        assert.strictEqual(sessionStorage.bar, dataWithinQuota);
+
+        const dataExceedingQuota = "0".repeat(6000000);
+
+        assert.throws(() => localStorage.setItem("foo", dataExceedingQuota));
+        assert.throws(() => sessionStorage.setItem("bar", dataExceedingQuota));
+      });
+    });
+
+    describe("set to 10000 bytes", () => {
+      it("should only allow setting data within the custom quota", () => {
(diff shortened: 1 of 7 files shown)


domenic:
Would it be feasible to write a non-WPT test?

Zirro:
Yes, that should be possible. I can add an option for changing the storage quota as well, to make it a justifiable addition to the API tests.

Zirro:
I've added a `storageQuota` option, along with documentation and API tests.

TimothyGu:
Seems like this tests only the size of a single entry, while I think the spirit of "size per origin" is the entire storage area combined… In later versions of Node.js one can use [`v8.serialize()`](https://nodejs.org/api/v8.html#v8_v8_serialize_value) to find out how big an object actually is.

Zirro:
This test is only intended to verify that the option gets applied by the API - the important part is checking that it throws if the storage exceeds the limit. More exhaustive tests are located among the `webstorage` WPTs.

TimothyGu:
Oh sorry, by "this" I meant the `Storage` code itself.

Zirro:
I don't think that's the case. The total size is calculated by adding together each key and value here: https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/webstorage/Storage-impl.js#L63

This test should verify that it works: https://github.com/web-platform-tests/wpt/blob/master/webstorage/storage_local_setitem_quotaexceedederr.html

TimothyGu:
@Zirro Oh, seems like I misread it. Sorry about the noise

test/api/options.js
```diff
@@ -289,4 +289,66 @@ describe("API: constructor options", () => {
...
+        })).window;
+        const dataWithinQuota = "0".repeat(5);
+
+        localStorage.setItem("foo", dataWithinQuota);
+        sessionStorage.setItem("bar", dataWithinQuota);
+
+        assert.strictEqual(localStorage.foo, dataWithinQuota);
+        assert.strictEqual(sessionStorage.bar, dataWithinQuota);
+
+        const dataExceedingQuota = "0".repeat(15000);
```
TimothyGu: It would be interesting to test something like 9995 here, since even though the string itself won't exceed the quota, the sum of all strings will.

Zirro: Thanks for the suggestion! I'm not sure if this is what you were thinking of, but it revealed an issue where replacing the value of an existing key would include the size of both entries in the count. I've fixed this and added the test now.

PR #2279, opened 2018-06-27
New resource loader [work in progress]

sarvaje:
Hi,

I'm working on the new Resource Loader. I would like to know if something as simple as what I have now is what you have in mind, if not, any comment is more than welcome :)

There is also a couple of questions in the code.


sarvaje:
I have also added a draft for the stylesheet.js to see if that is what we need, or if I missing something.

Zirro:
Hey, thanks for working on this! We're excited about the prospect of replacing the old resource loader.

@domenic has the clearest idea of what structure we're looking for (and what to avoid) in the new resource loader so getting his feedback would be ideal, though he's currently very busy with work. The same seems to be true for many of us maintainers right now, so responses may come a bit slow for a while.

In the meantime I just want to say that this looks like a good start. This code has much better separation of responsibilities, and it's great to hear that work is underway on the tests too.

sarvaje:
Thanks @Zirro.

Update:
I've been taking a look to the old tests for the resource loader and IMO the are not bad, I just add a few more tests to cover other scenarios (404 error, 503 error, etc.)

Now I'm going to start working on figure out why there are so many tests failing :( (any help will be more than welcome :smile:)

sarvaje:
I have found out that some of the errors are because:
- how I check if a script needs to be run or not (fix in my pc)
- cookies are not sent in the request

- cookies are not saved when we fetch something with cookies
- custom resource loader for the old api is broken and some tests fail.

Zirro:
> custom resource loader for the old api is broken and some tests fail.

The new Resource Loader is actually the last piece we require to move completely to the new API. When it's complete, we would like to remove the old API. That means that it's alright if this branch only works with the new API, though the old tests which require `old-api.js` would need to be moved to the new API as well. Glad to see the new tests you added, as well.

test/old-api/env.js
```diff
@@ -216,51 +216,53 @@ describe("jsdom/env", () => {
...
-        SkipExternalResources: false
-      },
-      done(err, window) {
-        assert.ifError(err);
-        assert.strictEqual(window.frames.bar.resourceLoaderWasOverridden, true);
-        t.done();
-      }
-    });
-  });
+  // this tests doesn't work with the new resource loader
```
domenic: Just delete, no need to comment out :)

test/to-port-to-wpts/htmlimageelement.js
```diff
@@ -100,18 +100,19 @@ describe("htmlimageelement", { skipIfBrowser: true }, () => {
     const src = fs.readFileSync(path.resolve(__dirname, "files/image.txt"), { encoding: "utf-8" }).trim();
     image.onload = () => {
       assert.ok(true, "onload should be triggered when loading from data URL.");
+      assert.strictEqual(image.width, 168, "after setting data URL, width should be 168");
```
domenic: Good catch

lib/api.js
```diff
@@ -314,6 +315,10 @@ function transformOptions(options, encoding) {
     transformed.windowOptions.pretendToBeVisual = Boolean(options.pretendToBeVisual);
   }
 
+  if (options.ResourceLoader) {
+    transformed.windowOptions.ResourceLoader = options.ResourceLoader;
+  }
```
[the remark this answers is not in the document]
sarvaje: > The option should be lowercase resourceLoader, and accept an instance of ResourceLoader, instead of accepting a constructor. This is similar to the virtualConsole or cookieJar options.

Ok. I wasn't what was the best, if the Constructor or the instance.

> Ideally a resource loader should be able to handle multiple documents/windows. This may involve creating a new internal class, something like PerDocumentResourceLoader, which gets spawned from the ResourceLoader (using a "private" method). But users of jsdom should not worry about this, if possible.

Sorry, I don't understand this 100% hehehe.

Do you mean something like this?
`class PerDocumentResourceLoader {
   constructor(resourceLoader, document) {
     this.document = document;
     this.resourceLoader = resourceLoader;
   }
   
   fetch(url, options) {
     // do some job with options
     return this.resourceLoader(url, newOptions);
   }  
}`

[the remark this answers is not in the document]
domenic: Something like that, yeah. The idea being that fetch() on PDRL can use Document-specific stuff like tracking Window shutdown.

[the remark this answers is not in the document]
domenic: An alternate design, maybe simpler for now, is to just add document to fetch()s options. Each call site then needs to pass it in, but that isn't too bad.

lib/jsdom/browser/new-resource-loader.js
```diff
@@ -0,0 +1,99 @@
...
+
+    if (!url) {
+      return Promise.resolve(null);
+    }
+
+    if (url.protocol === "data:") {
+      return Promise.resolve(parseDataURL(urlString).body);
+    }
+
+    if (!url.hostname) {
```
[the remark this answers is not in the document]
TimothyGu: See https://github.com/nodejs/node/blob/602da6492f278c1345f7f2d82014d9248254476b/lib/internal/url.js#L1342-L1403 for an already-made URL-to-file path routine. I've been meaning to publish that as a separate module for quite a while but still haven't gotten a chance to.

lib/api.js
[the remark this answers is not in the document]
sarvaje: I will try the `PerDocumentResourceLoader` if I see that things get complicated, I can go back and use pass document as an option.

But I prefer the PerDocumentResourceLoader, because if someone create a custom resource loader and they don't use `super.fetch` then they will lose the Document-specific stuff

lib/jsdom/living/helpers/stylesheets.js
```diff
@@ -57,18 +57,73 @@ function addStylesheet(sheet, elementImpl) {
...
+  const request = resourceLoader.fetch(urlString, {
+    element: elementImpl,
+    jar: wrapCookieJarForRequest(document._cookieJar),
+    proxy: document._proxy,
+    strictSSL: document._strictSSL
+  });
+
+  document._requestManager.add(request);
+
+  function onStylesheetLoad(err, data) {
```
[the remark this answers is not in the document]
sarvaje: Ok, I think I need to change the ResourceQueue to support promises.

lib/jsdom/living/nodes/HTMLFrameElement-impl.js
```diff
@@ -1,14 +1,91 @@
...
+  const { _resourceLoader: resourceLoader } = document._defaultView;
+  const documentImpl = document.implementation;
+
+  if (!documentImpl._hasFeature("FetchExternalResources", frame.tagName.toLowerCase())) {
+    return;
+  }
+
+  const request = resourceLoader.fetch(serializedURL, {
+    element: frame,
+    jar: wrapCookieJarForRequest(document._cookieJar),
```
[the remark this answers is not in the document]
sarvaje: For what I saw fixing some tests, I need those things (cookie jar, proxy, strictSSL) from the document, at least cookie jar. Maybe proxy and strictSSL can be configure in `PerDocumentResourceLoader`

lib/jsdom/browser/new-resource-loader.js
```diff
@@ -0,0 +1,99 @@
...
+
+    return promise;
+  }
+
+  /**
+   * Fetch a resource.
+   * @param {string} urlString - Absolute url (data URLs are supported).
+   * @param {object} options - Options to configure fetch.
+   */
+  fetch(urlString, options = {}) {
```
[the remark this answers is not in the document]
sarvaje: ok.

lib/jsdom/browser/new-resource-loader.js
```diff
@@ -0,0 +1,99 @@
...
+const { URL } = require("whatwg-url");
+const parseDataURL = require("data-urls");
+const request = require("request-promise-native");
+const { version: packageVersion } = require("../../../package.json");
+const DEFAULT_USER_AGENT = `Mozilla/5.0 (${process.platform}) AppleWebKit/537.36 (KHTML, like Gecko) ` +
+                           `jsdom/${packageVersion}`;
+
+class ResourceLoader {
+  constructor() {
+    this._defaultOptions = {
```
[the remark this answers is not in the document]
sarvaje: > Did you see the requestOptions design in #2050 ?

Yep, I just add some other default options that I though It was use everywhere, but I like the idea.

sarvaje:
I have just push an implementation of `PerDocumentResourceLoader`.
I still need to fix a lot of tests and feedback, but it would be nice to know if the `PerDocumentResourceLoader` looks good or not :)

Thanks!

domenic:
Sorry I haven't had time to look at PDRL yet, will try to get to that soon. Great question to flag though about the sync-ness.

Looking through the tests, it does seem that would break a lot of the simpler use cases people use jsdom for. I don't quite understand why though. That is, why does changing the readyState later mean you can't use compareDocumentPosition immediately anymore?

In other words, I think it's fine if the load event fires async, and the readyState changes async. But I don't think we should break the ability to start manipulating the DOM synchronously.

sarvaje:
> That is, why does changing the readyState later mean you can't use compareDocumentPosition immediately anymore?
>
>In other words, I think it's fine if the load event fires async, and the readyState changes async. But I don't think we should break the ability to start manipulating the DOM synchronously.

I agree.

I think it is related with the method `write` in document-impl. Anyway, let me take a look to what is different.

sarvaje:
@domenic I have taken a look, and the problem is that `write` depends on the `readyStatus` so, when the status is loading it modifies the html using this line https://github.com/jsdom/jsdom/blob/d6f8a97b5fb7709d6ad0215c1ae95fd4cab58489/lib/jsdom/living/nodes/Document-impl.js#L440

but when `readyStatus !== 'loading'` it use this line
https://github.com/jsdom/jsdom/blob/d6f8a97b5fb7709d6ad0215c1ae95fd4cab58489/lib/jsdom/living/nodes/Document-impl.js#L447

in the case of the doctype, the first method doesn't update it, but the second does, I don't know exactly why.

I think you can modify the DOM even if the `readyStatus` is loading, the problem is when you use `write` to change it.

domenic:
Got it, that makes sense! I guess the right fix then is to just change the load() test util to use `new JSDOM("x")` instead of `document.write("x")`.

sarvaje:
@domenic I think this is ready to review.

There is a tests failing, but I'm not sure why or how to solve it.
The differences I have notice is that in the previous version, the events for the storage are never dispatched because when we call to `setItem` the window is already closed.
In the new version, when we call to `setItem` the window element is not closed yet, so the events are dispatched, and that produce the error.

domenic:
I spent a decent amount of time debugging that test. It's very interesting, and I guess you must be right that before the failure was hidden. We should file a new issue about it, and disable it for this PR, and fix it later. From what I can tell the way that we're scheduling the storage events does not play well with the microtasks used by the EventWatcher abstraction used in that test: we fire both storage events before we ever reach the second `w.wait_for()` call.

Given that, I'll work on doing a review pass on the code, since the only test it doesn't pass is independently broken.

lib/api.js
```diff
@@ -314,6 +320,10 @@ function transformOptions(options, encoding) {
     transformed.windowOptions.pretendToBeVisual = Boolean(options.pretendToBeVisual);
   }
 
+  if (options.ResourceLoader) {
```
domenic: Why is there also a capital-R `ResourceLoader` option?

lib/api.js
```diff
@@ -241,6 +242,7 @@ function transformOptions(options, encoding) {
       runScripts: undefined,
       encoding,
       pretendToBeVisual: false,
+      resourceLoader: undefined,
```
domenic: This should be in the "defaults filled in later" section a couple lines down.

lib/jsdom/browser/Window.js
```diff
@@ -76,6 +77,16 @@ function Window(options) {
 
   ///// PRIVATE DATA PROPERTIES
 
+  if (options.resourceLoader) {
+    if (!(options.resourceLoader instanceof ResourceLoader)) {
+      throw new RangeError("options.ResourceLoader must implement ResourceLoader");
+    }
+
+    this._resourceLoader = options.resourceLoader;
+  } else {
```
domenic: This should never happen, right? Since you already did the defaulting in api.js. Note that the VirtualConsole pattern has a comment saying it is not necessary in the new API :).

lib/jsdom/browser/new-resource-loader.js
```diff
@@ -0,0 +1,131 @@
+"use strict";
+
+const fs = require("fs");
+const { parseURL } = require("whatwg-url");
+const { fromURLRecord } = require("data-urls");
```
domenic: Maybe rename this while importing so it's clearer, e.g. `const dataURLFromRecord = require("data-urls").fromURLRecord`.

lib/jsdom/browser/new-resource-loader.js
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * Returns the options to make the request
+   * @param {object} options - Custom options for the request.
+   */
+  _getRequestOptions(options = {}) {
+    const rqOptions = options.requestOptions;
+    const requestOptions = {
+      agent: rqOptions.agent,
+      agentOptions: rqOptions.agentOptions ?
+        Object.assign({}, this._defaultOptions.agentOptions, rqOptions.agentOptions) :
```
domenic: You can just always do the `Object.assign`; it'll work the same way. Here and below.

lib/jsdom/browser/new-resource-loader.js
```diff
@@ -0,0 +1,131 @@
...
+        this._defaultOptions.agentOptions,
+      agentClass: rqOptions.agentClass,
+      encoding: null,
+      gzip: true,
+      jar: rqOptions.jar,
+      pool: rqOptions.pool ? Object.assign({}, this._defaultOptions.pool, rqOptions.pool) : this._defaultOptions.pool,
+      // If strictSSL is not defined, then strictSSL will be true.
+      strictSSL: rqOptions.strictSSL !== false,
+      headers: {
+        "User-Agent": rqOptions.userAgent || DEFAULT_USER_AGENT,
```
domenic: Hmm, should we put DEFAULT_USER_AGENT in the this._defaultOptions too? Not quite sure why it's different, but there could definitely be a reason I'm missing.

lib/jsdom/browser/new-resource-loader.js
```diff
@@ -0,0 +1,131 @@
...
+   * @param {object} options - Options to configure fetch.
+   */
+  fetch(urlString, options = {}) {
+    if (!urlString) {
+      return Promise.reject(new Error("Invalid url string"));
+    }
+
+    let url = parseURL(urlString);
+
+    // Question: Can we assume this?
```
domenic: This seems wrong. Why wouldn't the URL string be a `file://` URL?

I'd rather this section omit the empty-string check, and just always reject if parsing fails.

lib/jsdom/browser/new-resource-loader.js
```diff
@@ -0,0 +1,131 @@
...
+    if (!url) {
+      url = { scheme: "file" };
+    }
+
+    if (url.scheme === "data") {
+      return Promise.resolve(fromURLRecord(url).body);
+    }
+
+    if (!url.cannotBeABaseURL && url.scheme !== "file") {
```
domenic: Hmm, I don't think this is the right way to do this. Instead, check if the scheme is http or https.

lib/jsdom/browser/new-resource-loader.js
```diff
@@ -0,0 +1,131 @@
...
+      const requestOptions = this._getRequestOptions(options);
+
+      return request(urlString, requestOptions);
+    }
+
+    const filePath = urlString
```
domenic: So here, check first that the scheme is file. If it is not, we should have the final fallback be reject.

So the logic is:

- Parse
- Parsing fails? Reject
- Is data? Process
- Is http or https? Process
- Is file? Process
- Reject, invalid scheme

domenic: Add a TODO here to come up with a better url -> filename algorithm. @TimothyGu was mentioning Node core has a good one we should adopt. But, it does not need to block this pull request; this is great as-is.

lib/jsdom/browser/new-resource-loader.js
```diff
@@ -0,0 +1,131 @@
+"use strict";
```
domenic: This file is really nice and clear and readable; great work.

lib/jsdom/living/node.js
```diff
@@ -55,6 +55,11 @@ module.exports.clone = function (node, document, cloneChildren) {
       break;
   }
 
+  // https://html.spec.whatwg.org/multipage/scripting.html#script-processing-model
+  if (node._localName === "script") {
```
domenic: Can you add a [cloningSteps] to HTMLScriptElement-impl instead?

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -56,44 +55,187 @@ const ProgressEvent = require("../generated/ProgressEvent");
...
+const { ResourceLoader } = require("../../browser/new-resource-loader");
+const { wrapCookieJarForRequest } = require("../xhr-utils");
+
 function clearChildNodes(node) {
   for (let child = domSymbolTree.firstChild(node); child; child = domSymbolTree.firstChild(node)) {
     node.removeChild(child);
   }
 }
 
+class AsyncResourceQueue {
```
domenic: Can we move this to its own file?

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -56,44 +55,187 @@ const ProgressEvent = require("../generated/ProgressEvent");
...
+        if (q.count() === 0) {
+          q._notify();
+        }
+
+        return null;
+      });
+  }
+}
+
 class ResourceQueue {
```
domenic: Can we move this to its own file too?

And can we make sure the files have good explanations of each class's purpose? It's a bit confusing trying to figure out why there are both ResourceQueue and AsyncResourceQueue. Maybe they need better names?

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -140,6 +282,128 @@ class RequestManager {
   }
 }
 
+class PerDocumentResourceLoader {
```
domenic: Again, new class, in its own file :)

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -140,6 +282,128 @@ class RequestManager {
...
+  constructor(document) {
+    this._document = document;
+    this._defaultEncoding = document._encoding;
+    this._resourceLoader = document._defaultView ? document._defaultView._resourceLoader : new ResourceLoader();
```
domenic: Hmm, for documents without a _defaultView, we should generally not be loading resources, I am pretty sure. Does this happen in the tests or something?

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -140,6 +282,128 @@ class RequestManager {
...
+    this._requestManager = document._requestManager;
+    this._queue = document._queue;
+    this._deferQueue = document._deferQueue;
+    this._asyncQueue = document._asyncQueue;
+
+    // TODO: Remove when _hasFeature is no longer necessary
```
domenic: If you want you can remove all the _hasFeature stuff yourself in this PR :)

lib/jsdom/living/helpers/stylesheets.js
```diff
@@ -2,7 +2,6 @@
 const cssom = require("cssom");
```
domenic: This file is also very clean. It looks like the classes in Document-impl.js worked out really well.

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -56,44 +55,187 @@ const ProgressEvent = require("../generated/ProgressEvent");
 const TouchEvent = require("../generated/TouchEvent");
 const UIEvent = require("../generated/UIEvent");
 
+const { ResourceLoader } = require("../../browser/new-resource-loader");
+const { wrapCookieJarForRequest } = require("../xhr-utils");
```
domenic: Seeing how wrapCookieJarForRequest is used, I think it does not belong in xhr-utils. Maybe its own file inside living/helpers/?

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -140,6 +282,128 @@ class RequestManager {
...
+    const ev = this._document.createEvent("HTMLEvents");
+
+    const onError = error => {
+      this._requestManager.remove(request);
+
+      if (options.onError) {
+        options.onError(error);
+      }
+
+      // TODO: Create events in the modern way https://github.com/jsdom/jsdom/pull/2279#discussion_r199969734
```
domenic: Just flagging this to fix before we merge.

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -140,6 +282,128 @@ class RequestManager {
...
+    };
+
+    const onLoad = data => {
+      this._requestManager.remove(request);
+
+      this._addCookies(url, request.response ? request.response.headers : {});
+
+      if (options.onLoad) {
+        try {
+          const result = options.onLoad(data);
```
domenic: It's a bit sad that the callers have to use onLoad callbacks instead of promises, but this is a good compromise, given that you have to run code after they do.

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -140,6 +282,128 @@ class RequestManager {
...
+
+          if (result && result.then) {
```
domenic: Instead of this pattern, do `return Promise.resolve(result).then(() => { ... })`. Then remove the sync code branch (except the catch is still necessary I guess).

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -140,6 +282,128 @@ class RequestManager {
...
+        }
+      }
+
+      ev.initEvent("load", false, false);
+      element.dispatchEvent(ev);
+
+      return Promise.resolve();
+    };
+
+    if (element.nodeName.toLowerCase() === "script" && element.getAttribute("async") !== null) {
```
domenic: Use element.localName so you don't need to toLowerCase().

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -521,17 +788,79 @@ class DocumentImpl extends NodeImpl {
...
+      this.dispatchEvent(ev2);
+
+      return;
+    }
     this._queue.resume();
 
-    // Set the readyState to 'complete' once all resources are loaded.
-    // As a side-effect the document's load-event will be dispatched.
-    resourceLoader.enqueue(this, null, function () {
+    const dummyPromise = new Promise(resolve => {
```
domenic: Just do `const dummyPromise = Promise.resolve()`

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -521,17 +788,79 @@ class DocumentImpl extends NodeImpl {
...
       this.readyState = "complete";
       const ev = this.createEvent("HTMLEvents");
-      ev.initEvent("DOMContentLoaded", false, false);
+
+      ev.initEvent("load", false, false);
+
       this.dispatchEvent(ev);
-    })(null, true);
+
+      return Promise.resolve();
```
domenic: This will not be necessary if you do the above changes to avoid `.then` testing.

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -589,7 +918,7 @@ class DocumentImpl extends NodeImpl {
   get body() {
     const { documentElement } = this;
     if (!documentElement || documentElement._localName !== "html" ||
-        documentElement._namespaceURI !== HTML_NS) {
+      documentElement._namespaceURI !== HTML_NS) {
```
domenic: All these reformatting changes are unfortunate :(. If you could undo them that would be really really nice.

lib/jsdom/living/nodes/HTMLFrameElement-impl.js
```diff
@@ -1,16 +1,92 @@
...
+  if (attaching) {
+    const ev = document.createEvent("HTMLEvents");
+
+    ev.initEvent("load", false, false);
+    frame.dispatchEvent(ev);
+
+    return;
+  }
+
+  const dummyPromise = new Promise(resolve => {
```
domenic: Use `Promise.resolve()`. Or, if this pattern is common, maybe that should be the default value for the first argument to queue.push?

lib/jsdom/living/nodes/HTMLFrameElement-impl.js
```diff
@@ -1,16 +1,92 @@
...
+    frame.dispatchEvent(ev);
+
+    return Promise.resolve();
+  }
+
+  document._queue.push(dummyPromise, onLoad);
+}
+
+function fetchFrame(serializedURL, frame, document, contentDoc) {
+  const { _resourceLoader: resourceLoader } = document;
```
domenic: Avoid this syntax. Instead just do `const resourceLoader = document._resourceLoader`. Clearer and similar amount of typing. I think I saw this earlier too.

lib/old-api.js
```diff
@@ -397,7 +397,12 @@ function processHTML(config) {
   } else if (window.document.readyState === "complete") {
     scriptComplete();
   } else {
-    window.addEventListener("load", scriptComplete);
+    function onWindowLoad() {
+      window.removeEventListener("load", onWindowLoad);
+      scriptComplete();
+    }
+
+    window.addEventListener("load", onWindowLoad);
```
domenic: You could do `window.addEventListener("load", scriptComplete, { once: true })`, I think.

test/api/methods.js
```diff
@@ -183,13 +183,17 @@ describe("API: JSDOM class's methods", () => {
         assert.strictEqual(dom.window.top, newTop);
       });
 
-      it("should reconfigure the window.top property (tested from the inside)", () => {
+      it("should reconfigure the window.top property (tested from the inside)", { async: true }, t => {
```
domenic: Hmm. So the change is that here we're no longer executing inline script synchronously? That seems bad. Although I know we were not perfect at it, browsers do execute inline scripts sync, so preserving that seems better if we can...

test/api/resources.js
```diff
@@ -238,26 +239,223 @@ describe("API: resource loading configuration", { skipIfBrowser: true }, () => {
...
+      setUpLoadingAsserts(element);
+      element.src = url;
+      dom.window.document.body.appendChild(element);
+
+      return assertLoaded(element).then(() => {
+        assert.strictEqual(
+          dom.window.frames[0].document.body.textContent, "Hello",
+          "The frame must have been downloaded"
+        );
+        assert.isTrue(dom.window._resourceLoader.called, "The custom resource should be called");
```
domenic: Don't probe underscore-prefixed properties, even in tests. Instead, you can use a closure variable at the top of this test, and just set `called = true` instead of `this.called = true`.

test/api/resources.js
```diff
@@ -238,26 +239,223 @@ describe("API: resource loading configuration", { skipIfBrowser: true }, () => {
...
+      });
+    });
+
+    describe("resources returns 404", () => {
+      it("should fire an error event donwloading images if and only if canvas is installed", { slow: 500 }, () => { // eslint-disable-line max-len
```
domenic: Don't disable ESLint. Split the line into two. You can see other tests that do this.

Mispelled "downloading"

test/old-api/env.js
```diff
@@ -647,63 +601,65 @@ describe("jsdom/env", () => {
...
-              FetchExternalResources: ["script"],
-              ProcessExternalResources: ["script"],
-              SkipExternalResources: false
-            }
-          });
-        });
-      }
-    );
+    // TODO: With the current implementation, this is not possible.
+    // QUESTION: Is this important to have in the new resource loader?
```
domenic: It's possible by overriding fetch, right?

test/web-platform-tests/run-single-wpt.js
```diff
@@ -1,11 +1,13 @@
 "use strict";
 /* eslint-disable no-console */
-const fs = require("fs");
 const path = require("path");
+const { URL } = require("url");
 const { specify } = require("mocha-sugar-free");
 const { inBrowserContext, nodeResolverPromise } = require("../util.js");
 const jsdom = require("../../lib/old-api.js");
+const { ResourceLoader } = require("../../lib/jsdom/browser/new-resource-loader");
```
domenic: Can we convert this file to use new API now?

lib/jsdom/living/nodes/Document-impl.js
sarvaje: I think I did this before testing, I can try to remove the default resource loader.

lib/jsdom/living/nodes/Document-impl.js
sarvaje: Sure, I'm glad you ask for that hehehe.

lib/jsdom/living/nodes/Document-impl.js
sarvaje: I will move all the classes to a new file

lib/jsdom/living/node.js
sarvaje: ok!

lib/jsdom/browser/new-resource-loader.js
sarvaje: Thanks! :)

lib/jsdom/browser/new-resource-loader.js
sarvaje: I will take a look to all of this, because when I change to the `parseURL` I had a lot of problems with tests. I'm going to try all the changes related with this and try to explain the problems I have.

sarvaje:
> The other meta-comment is that all the reformatting changes make it hard to review. I imagine your editor might be doing that for you, but I'd ask you to turn off that setting and undo those changes if at all possible.

Sorry about that, I tried to avoid them, but it seems that I couldn't do it hehehe. I will review everything to remove the formatting changes.

>My major worry at this point is making all script execution asynchronous, as seen from some of the test changes. If I'm understanding correctly, that's a change away from browser compatibility, and one that will probably be disruptive to consumers. So if there's any way to avoid that, that'd be ideal. Hopefully it's just a matter of tweaks to HTMLScriptElement-impl for the inline script case.

>So the change is that here we're no longer executing inline script synchronously? That seems bad. Although I know we were not perfect at it, browsers do execute inline scripts sync, so preserving that seems better if we can...

I'm going to try to explain the problem, the problem I have in the tests is not that the script not run synchronously (I add it to the queue as soon I get it) , the thing is that the queue is using promises, so even if the script is running as soon as I read it, because of the promises, jsdom is finishing the parse before the script finish the execution.

If I remember correctly, these problems appear when we change the "callback" queue to "promises" queue.

I don't know if I explain myself well hehehe

lib/api.js
sarvaje: this is an error

sarvaje:
I also need to fix the conflict hehehe

lib/jsdom/browser/Window.js
sarvaje: You mean the `else` part right? At least we still need the type checking.

lib/jsdom/browser/new-resource-loader.js
sarvaje: Do you prefer that way, or `const { fromURLRecord: dataURLFromRecord } = require("data-urls");` is ok?

lib/jsdom/browser/new-resource-loader.js
sarvaje: Ok, that was because there are a couple of tests that use the `script` property in the old API. and they don't use `file://`

I will fail and modify those tests

lib/jsdom/living/nodes/Document-impl.js
sarvaje: Yep, sorry about that, I tried to be careful about them, but I missed them.

lib/jsdom/living/nodes/HTMLFrameElement-impl.js
sarvaje: IIRC I was doing that because of an eslint rule, but now it works. I will change them.

lib/jsdom/living/nodes/Document-impl.js
sarvaje: I know, I didn't find a better way to do that :(

lib/jsdom/living/nodes/Document-impl.js
sarvaje: This would be easier if we could use async/await hehehe.

sarvaje: But it is true, I don't need the if :)

lib/jsdom/living/nodes/Document-impl.js
sarvaje: This is used in the queue, not in the load you comment above, but I can apply something similar to the queue :)

test/api/methods.js
sarvaje: I'm not sure if I'm talking to soon, but looks like now this is not necessary.
My theory is that the changes in `dummyPromise` for `Promise.resolve()` has helped with this.

test/old-api/env.js
sarvaje: It wasn't, but I have done some changes to make it possible.

lib/jsdom/living/nodes/Document-impl.js
sarvaje: If I remove that, I need to remove the old api too, right?

sarvaje:
@domenic I think this is ready for another review :)

I'm working on `test/web-platform-tests/run-single-wpt.js` to use the new API, and for now it looks good, but I think I need to modify some code to use the right properties in the request like `strictSSL`, etc.

I will finish it on Monday.

About the scripts, looks like now it is working as expected, but if there is something else, please, let me know.

lib/jsdom/living/per-document-resource-loader.js
```diff
@@ -0,0 +1,119 @@
...
+          })
+          .catch(err => {
+            return onError(err);
+          });
+      } catch (err) {
+        return onError(err);
+      }
+    };
+
+    if (element.localName === "script" && element.getAttribute("async") !== null) {
```
Zirro: Adding support for `async` scripts is great, but `HTMLScriptElement` also needs to be updated to handle the attribute. You can do this by replacing the `async` attribute line [here](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/nodes/HTMLScriptElement.webidl#L7) with:

`[CEReactions, Reflect] attribute boolean async;`

...and run `yarn convert-idl` to process the WebIDL file.

We'll also need test coverage since this is a new feature. Some of the `async` tests in https://github.com/jsdom/jsdom/blob/master/test/web-platform-tests/to-run.yaml#L589 should be passing with these changes, so you can enable them by removing them from the list.

sarvaje:
> Maybe we should continue to not expose the async property,

But do you want to keep the current code to manage the async scripts? Like kind of working but not officially supported?

Tomorrow I will take a look to change the file `test/web-platform-tests/run-single-wpt.js` to use the new API and I will review all the tests that are now using the onload to see if we can remove them.

There is something else you want me to change/review?

Thanks @Zirro and @domenic for the reviews and the commits :)

domenic:
Yeah, kind of working but not officially supported was my thinking for async. Welcome to other thoughts though.

Nothing else besides the sync/onload issue from my end!

Zirro:
I noticed that many of the new files are placed directly inside the `living` directory. I think it would help to keep things organised by having a specific location for them - perhaps a new `network` or `resources` directory inside `jsdom/lib/jsdom/living/helpers`?

domenic:
+1 on a resources directory inside helpers.

domenic:
Actually, maybe inside `browser/` or just inside `jsdom/?` Hrm, that whole directory structure is still a mess... anyway, as long as its its own directory...

sarvaje:
> I noticed that many of the new files are placed directly inside the living directory. I think it would help to keep things organised by having a specific location for them - perhaps a new network or resources directory inside jsdom/lib/jsdom/living/helpers?

Good idea!.

Here is a summary of the changes I have done today:
* Moved the new files to a `resources` folder inside `browser`.
* Changed `web-platform-tests` to use the new API instead of the old one. I had to do some changes in the new API to pass the tests.
* Removed all the unnecessary `onload` from the tests.
* Fixed most of the code format changes.

domenic:
OK! I spent much of the weekend working on finishing up all the revisions I wanted to do on top of this. I'd love your review, @sarvaje, as well as other collaborators if they have time.

I tried to divide things up into commits that are reasonably reviewable (e.g. you can skip the "port tests" commit). Perhaps the most important thing to review is the new README; the code is less important.

We may need to make some (breaking?) changes to the resource loader after it lands to make sure it meets everyone's use cases. For example in the course of this work I noticed that XHRs don't go through the resource loader, which is annoying, and we should work toward fixing it eventually. But hopefully this lays a solid groundwork.

(Oh, I just realized, some tests are probably going to fail on browsers, and perhaps with canvas installed. I have to test in browsers and disable tests as appropriate before merging.)

Some particular thoughts I had:

- What is the right naming for the option? I left it as `resources` for now. `resourceLoader` is a bit weird because `resourceLoader: "usable"` doesn't make as much sense... Suggestions welcome.
- I had always envisioned that HTML elements (etc.) would call the resource loader for all possible resources, and the resource loader would decide to not issue requests for unusable resources. But, in the end, the architecture is that HTML elements decide not to call the resource loader unless they can use the resulting resource. Probably this is OK?
- Maybe we can avoid passing `options` to `resourceLoader.fetch()` directly, so that subclassers don't have to remember to pass them back up to `super.fetch()`...
- All this centralization makes it much easier to see our request dependency in action, and potentially replace it in the future. But that will have to wait until XHR goes through the resource loader as well...

sarvaje:
Nice work in the PR @domenic.

Everything looks good to me, just a couple of comments.

> Maybe we can avoid passing options to resourceLoader.fetch() directly, so that subclassers don't have to remember to pass them back up to super.fetch()…

I like the idea, Setting the option when you create the object `new ResourceLoader(options)` should be enough. 

>What is the right naming for the option? I left it as resources for now. resourceLoader is a bit weird because resourceLoader: "usable" doesn't make as much sense...

I'm talking with @molant and, if `usable` is the only "string" options, what do you think of change it to a Boolean? so the `resources` options could be `false` `true` or a custom resource Loader.

domenic:
Wasn't able to figure out a reasonable way to not require the pass-through options.

Going to stick with "usable" instead of true/false for now; it seems nice and I don't see a big reason to cause folks to switch.

Updated the docs and inlined contextifyWindow per review.

Locally at least, `yarn test-browser` passes, somewhat surprisingly.

Going to merge this once CI is happy!!

PR #2282, opened 2018-07-01
Move to w3c-xmlserializer

Sebmaster:
This finally gives us proper serialization for XML documents and the
XMLSerializer interface.


domenic:
@Sebmaster mind taking a look at the test failures? Probably all just bad tests...

lib/api.js
```diff
@@ -61,7 +61,7 @@ class JSDOM {
   }
 
   serialize() {
-    return domToHtml([idlUtils.implForWrapper(this[window]._document)]);
+    return fragmentSerialization(idlUtils.implForWrapper(this[window]._document), false);
```
domenic: Can we use an options object instead? https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

Sebmaster:
Fixed feedback and rebased on master. Greeeen 🌈

papandreou:
Really looking forward to this, great work!

Is it intentional that `jsdom.serialize(document)` does not preserve the XML declaration?

```js
const JSDOM = require('jsdom').JSDOM;
const text = `<?xml version="1.0" encoding="utf-8"?><foo></foo>`;
const jsdom = new JSDOM(text, {
  contentType: 'application/xhtml+xml'
});
const document = jsdom.window.document;
console.log(jsdom.serialize(document)); // <foo/>
```

I can easily work around that, but just wanted to check ;)

domenic:
Hmm, there is no spec for serialize(), so we get to make up the rules. I tend to think it should be included?

I'll try to do that on my next jsdom weekend (and then merge this), unless anyone has counterobjections.

papandreou:
Thanks for the reply! For my case the most handy thing would be if the XML declaration were included if and only if it was in the string that was originally parsed, similar to how `<!DOCTYPE ...>` behaves for HTML:

```js
const JSDOM = require('jsdom').JSDOM;
function roundTrip(text, contentType = 'text/html') {
  const jsdom = new JSDOM(text, { contentType });
  console.log(jsdom.serialize(jsdom.window.document));
}

roundTrip(`<html><head></head><body></body></html>`);
//         <html><head></head><body></body></html>

roundTrip(`<!DOCTYPE html><html><head></head><body></body></html>`);
//         <!DOCTYPE html><html><head></head><body></body></html>

roundTrip('<foo/>', 'application/xml');
//         <foo/>

roundTrip('<?xml version="1.0"?><foo/>', 'application/xml');
//         <foo/>
```

Would be great if the last one logged `<?xml version="1.0"?><foo/>`.

domenic:
After some investigation it seems that the XML declaration is properly treated as non-existent in the DOM; it's basically a sort of invisible comment/header that identifies the file, and not part of the document object model. Thus, any serializations of the DOM must not include it.

As such, this is ready to merge as-is.

PR #2283, opened 2018-07-02
Add and expose @trust/webcrypto as window.crypto

### LINKED ISSUE #1612: window.crypto is undefined
i'm trying to run tests using jsdom and mocha but I got this error:

```
ReferenceError: crypto is not defined
```

Is crypto really defined in jsdom?

koresar:
Hello.

Noticed that `self.crypto` is supported by all major browsers now, but missing in jsdom.

There is this excellent project - https://github.com/anvilresearch/webcrypto
> W3C Web Cryptography API for Node.js

The `require('@trust/webcrypto')` object is compatible with `window.crypto` ([MDN](http://mdn.io/window.crypto))

Although, only parts of algorithms are supported. See their [table](https://github.com/anvilresearch/webcrypto#supported-algorithms). Probably, the jsdom README should warn about it.

If you believe this is an acceptable solution then I'd be glad to continue working on this PR.

Closes #1612

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -30,6 +30,7 @@ const reportException = require("../living/helpers/runtime-script-errors");
 const { matchesDontThrow } = require("../living/helpers/selectors");
 const SessionHistory = require("../living/window/SessionHistory");
 const { contextifyWindow } = require("./documentfeatures.js");
+const webcrypto = require("@trust/webcrypto");
 
 const GlobalEventHandlersImpl = require("../living/nodes/GlobalEventHandlers-impl").implementation;
 const WindowEventHandlersImpl = require("../living/nodes/WindowEventHandlers-impl").implementation;
@@ -248,6 +249,9 @@ function Window(options) {
     },
     get screen() {
       return screen;
+    },
+    get crypto() {
+      return webcrypto;
     },
     get localStorage() {
       if (this._document.origin === "null") {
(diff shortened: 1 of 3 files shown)


Zirro:
It's hard to say if this package is suitable for use in jsdom at this stage. At a glance it seems pretty limited in the algorithms and methods it supports, though I don't know how that compares to the support in current browsers.

It would be very interesting if it was able to pass a number of tests from the [WebCryptoAPI](https://github.com/web-platform-tests/wpt/tree/master/WebCryptoAPI) web platform tests (see  [Contributing.md](https://github.com/jsdom/jsdom/blob/master/Contributing.md#web-platform-feature-tests) for details) but the APIs would likely need to be exposed through [webidl2js](https://github.com/jsdom/webidl2js) for this happen.

koresar:
Okay. That's looks like a longer path than I expected. Let me see what I can do here.

Btw, is there a place in the internets where I can ask questions regarding platform tests and the webidl?

Zirro:
Since they're not very common outside the development of browsers and browser-like projects, it's probably easiest if you ask here. The Contributing.md file should hopefully provide most information you'll need, though.

domenic:
Closing this for now; happy to reopen if we get a spec-compliant version.

givethemheller:
This deserves to be re-opened. Having a partial solution, is orders of magnitude better than having no solution. Implementing this could drive changes on the library. 4 years later - maybe its already up to the standard for usage?

PR #2303, opened 2018-07-26
Change storageQuota to operate on code units, not bytes

domenic:
Test runs in browsers on CI revealed that the buffer shim used in browsers is very slow at calculating byte length. Even in Node, it takes an unnecessarily-long amount of time. Instead, we can just set the quota to be derived from the string length, since that's the native format of the data. This is much less expensive to compute.

Please review especially for any stray references to "bytes" I might have missed.


lib/jsdom/living/webstorage/Storage-impl.js
```diff
@@ -56,15 +56,15 @@ class StorageImpl {
...
     for (const [curKey, curValue] of this._items) {
       // If the key already exists, skip it as it will be set to the new value instead
       if (key !== curKey) {
-        itemsConcat += curKey + curValue;
+        itemsTotalLength += curKey.length + curValue.length;
       }
     }
-    if (Buffer.byteLength(itemsConcat) > this._quota) {
+    if (itemsTotalLength > this._quota) {
       throw new DOMException(`The ${this._quota} byte storage quota has been exceeded.`, "QuotaExceededError");
```
Zirro: This DOMException references bytes as well.

domenic: Great catch, thank you!!

PR #2311, opened 2018-07-28
HTMLElement.prototype.draggable

### LINKED ISSUE #2268: `HTMLElement.prototype.draggable` descriptor is missing
### Basic info:

- **Node.js version:** v10.1.0
- **jsdom version:** 11.11.0

### Minimal reproduction case

```js
const jsdom = require("./");
const { JSDOM } = jsdom;
const dom = new JSDOM(`<body></body>`);

const HTMLElement = dom.window.HTMLElement;
const draggable = Object.getOwnPropertyDescriptor(HTMLElement.prototype, 'draggable');

console.log(draggable); // yields undefined
```

Documentation:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable

```js
Object.getOwnPropertyDescriptor(HTMLElement.prototype, 'draggable');
```

It yields `{get: ƒ, set: ƒ, enumerable: true, configurable: true}` in all major browsers.

pmdartus:
This PR adds support for the [draggable](https://html.spec.whatwg.org/multipage/dnd.html#the-draggable-attribute) global HTML attribute.

I also have been investigating adding support for the rest of the [Drag and Drop](https://html.spec.whatwg.org/multipage/dnd.html#dnd) APIs, but I am not sure it would make sense to add support for those in the context of jsdom.

Fix #2268

### DIFF
--- a/lib/jsdom/living/nodes/HTMLAnchorElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLAnchorElement-impl.js
@@ -19,7 +19,14 @@ class HTMLAnchorElementImpl extends HTMLElementImpl {
 
   get draggable() {
     const draggableValue = this.getAttribute("draggable");
-    return draggableValue !== "false";
+
+    if (draggableValue === "true") {
+      return true;
+    } else if (draggableValue === "false") {
+      return false;
+    }
+
+    return this.hasAttribute("href");
   }
 
   set draggable(value) {
--- a/lib/jsdom/living/nodes/HTMLAnchorElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLAnchorElement-impl.js
@@ -16,22 +16,6 @@ class HTMLAnchorElementImpl extends HTMLElementImpl {
   set text(v) {
     this.textContent = v;
   }
-
-  get draggable() {
-    const draggableValue = this.getAttribute("draggable");
-
-    if (draggableValue === "true") {
-      return true;
-    } else if (draggableValue === "false") {
-      return false;
-    }
-
-    return this.hasAttribute("href");
-  }
-
-  set draggable(value) {
-    super.draggable = value;
-  }
 }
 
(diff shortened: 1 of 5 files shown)


diervo:
@domenic The first of many :)

lib/jsdom/living/nodes/HTMLAnchorElement-impl.js
```diff
@@ -16,6 +16,22 @@ class HTMLAnchorElementImpl extends HTMLElementImpl {
   set text(v) {
     this.textContent = v;
   }
+
+  get draggable() {
```
domenic: This shouldn't be here; it's an attribute for HTMLElement, not for HTMLAnchorElement. So only implement it on HTMLElement.

lib/jsdom/living/nodes/HTMLElement-impl.js
```diff
@@ -75,6 +75,22 @@ class HTMLElementImpl extends ElementImpl {
     this._clickInProgress = false;
   }
 
+  get draggable() {
```
domenic: Implement all the logic here. Test if `this._localName === "img"` or `this._localName === "a"`.

lib/jsdom/living/nodes/HTMLElement-impl.js
```diff
@@ -75,6 +75,22 @@ class HTMLElementImpl extends ElementImpl {
...
+    const attributeValue = this.getAttribute("draggable");
+    return attributeValue === "true";
+  }
+  set draggable(value) {
```
domenic: Just do `this.setAttribute("draggable", String(value))`.

pmdartus:
Thanks for the review @domenic, I updated PR based on your feedbacks.

domenic:
Thanks so much for this, release coming soon! Also I wanted to say I appreciate a lot that you immediately did the right thing with the web platform tests framework; many first-time contributors miss that. Looking forward to more!

PR #2322, opened 2018-08-05
Wrap new console methods

Zirro:
This wraps three console methods which were added in Node.js v8 (or earlier, in one case). I also moved the `wrapConsoleMethod`-method to a more reasonable location inside `Window.js` since it had drifted away from the rest of the console code over time.

[`countReset()`](https://nodejs.org/api/console.html#console_console_countreset_label) was added in Node.js v8.3.0 and [to the spec in May 2018](https://github.com/whatwg/console/commit/40ec790b96f50e9bf78a47985556d803d3d9712a).

[`dir()`](https://nodejs.org/api/console.html#console_console_dir_obj_options) has existed in Node.js since v0.1.101 but was added [to the spec in 2017](https://github.com/whatwg/console/commit/e5d4557d66a260ccc02c1feb72754e64a20cc1f1).

[`dirxml()`](https://nodejs.org/api/console.html#console_console_dirxml_data) was added in Node.js v8.0.0 and to the spec in 2017 together with `dir()`. In the future we could consider applying XML-formatting to the content since we have the tooling available.


domenic:
This doesn't seem like a breaking change to me? Unsure how to note it in the changelog.

PR #2323, opened 2018-08-05
Upgrade dependencies

Zirro:
This upgrades most dependencies to their latest version, with exceptions for some major ones.

This is a breaking change since [`ws`](https://github.com/websockets/ws/releases) drops Node.js v4 support in version 6.0.0.

nwsapi v2.0.8 resolves an issue making two tests pass. To avoid conflicts, this can be resolved in https://github.com/jsdom/jsdom/pull/2243 by removing them from `to-run.yaml`.


domenic:
Adding a couple more easy upgrades, and letting this take precedence over #2243.

PR #2332, opened 2018-08-15
Implement stub for Element.scrollIntoView()

### LINKED ISSUE #1695: scrollIntoView is not a function
I'm running a test using a library (`grommet`) that uses `scrollIntoView` on an `a` tag. `jsdom` is throwing errors when hitting that line of code. I was able to get it to work by adding `window.HTMLElement.prototype.scrollIntoView = function() {};` in my testHelper file to setup the tests, but I thought I'd report it to see if it's something you want to fix.

ovidiubute:
Fixes #1695

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -413,6 +413,9 @@ class ElementImpl extends NodeImpl {
     const matcher = addNwsapi(this);
     return matcher.closest(selectors, idlUtils.wrapperForImpl(this));
   }
+
+  scrollIntoView() {
+  }
 }
 
 mixin(ElementImpl.prototype, NonDocumentTypeChildNode.prototype);
--- a/lib/jsdom/living/nodes/Element.webidl
+++ b/lib/jsdom/living/nodes/Element.webidl
@@ -71,7 +71,7 @@ dictionary ScrollIntoViewOptions : ScrollOptions {
 partial interface Element {
   DOMRectList getClientRects();
   [NewObject] DOMRect getBoundingClientRect();
-//  void scrollIntoView(optional (boolean or ScrollIntoViewOptions) arg);
+  void scrollIntoView(optional (boolean or ScrollIntoViewOptions) arg);
 //  void scroll(optional ScrollToOptions options);
 //  void scroll(unrestricted double x, unrestricted double y);
 //  void scrollTo(optional ScrollToOptions options);


ovidiubute:
This PR adds a stub for `Element.scrollIntoView`. It is a noop but is very useful when writing unit-tests in case you need to spy on this particular method.

domenic:
Thanks, but I'm not sure we should do this kind of no-op thing. For example, I don't like implementing things that we fail the tests for. It needs a bit more thought and discussion.

In the meantime, have you tried doing this in your unit tests? `Element.prototype.scrollIntoView = () => {}` works fine in your code too; it doesn't have to built in to jsdom.

ovidiubute:
I wasn't aware tests failed because of this, where can I see that? 

> In the meantime, have you tried doing this in your unit tests? Element.prototype.scrollIntoView = () => {} works fine in your code too; it doesn't have to built in to jsdom.

This works just fine, but I just think JSDOM should implement as many DOM APIs as possible. I imagine lots of developers run into this issue while unit-testing and they probably work around it, it would be nice if they didn't have to.

There are other examples of noop APIs in JSDOM such as [this](https://github.com/jsdom/jsdom/blob/df3abb4cfc95d058588b1949c04588b9881c8ca6/lib/jsdom/living/nodes/Element-impl.js#L334), and since the signature of scrollIntoView returns void, I thought it would be the same thing.

domenic:
> I wasn't aware tests failed because of this, where can I see that?

The issue is you didn't add any tests. And if you did, per CONTRIBUTING.md, enable the appropriate web platform tests, they would fail, since it doesn't scroll things.

> This works just fine, but I just think JSDOM should implement as many DOM APIs as possible. 

I agree, but I think we have different definitions of "implement" :). You're right we have some no-op APIs. It's a complicated subject and I'm not sure whether we've made the right tradeoffs. Basically, we haven't figured out a coherent story for what we're doing with these sorts of layout-related things, and so every time we discuss adding one of these APIs, we end up unsure what to do. That's a bit unfair to you, and I'm sorry about that... I'll try to give this some more thought.

Zirro:
For scroll methods, we have a precedent in the `notImplementedMethod()` used on `window`: https://github.com/jsdom/jsdom/blob/master/lib/jsdom/browser/Window.js#L614

I think it'd make sense for us to behave consistently in all scroll methods, at least.

ovidiubute:
@Zirro does it make sense to change this PR to use the `notImplementedMethod` function?

Zirro:
I think that'd make more sense than an empty method, yes, but whether it gets merged in the end is up to Domenic. That said, if we add `scrollIntoView()` we ought to have `scroll()`, `scrollBy()` and `scrollTo()` with the same behaviour as well.

G-Rath:
Any update on this? 

If all that's required is adding 

```
ElementImpl.prototype.scroll = notImplementedMethod("element.scroll");
ElementImpl.prototype.scrollBy = notImplementedMethod("element.scrollBy");
ElementImpl.prototype.scrollTo = notImplementedMethod("element.scrollTo");
ElementImpl.prototype.scrollIntoView = notImplementedMethod("element.scrollIntoView");
```

Then I'd be happy to take a whack at it.

I'm wanting these methods implemented asap (if at all possible) so that I can spy on them in some `jest` tests. I'm using TypeScript, which makes it a bit harder to correct this at my end.

I'm not at all familiar with jsdom (outside of how to use it in `jest`, and it's overall aim/use case), so I can't really make a strong proposal on how solve this (& #1422), but personally I'd think it might be enough (or a good start) to just update the `scrollX` values appropriately (i.e `scrollTop` if `options.top`, `scrollLeft` if `options.left`, etc), while emitting any expected scroll events, and throwing `DOMException`s if the `behaviour` is not `auto` or `smooth`.

I'd also argue that perfect is the enemy of good: it might be better to charge ahead with an implementation like I've suggested (which I think is essentially a browser that "scrolls" at light speed), and then adjusting if turns out that's not good enough?

Like I said, I'm not that familiar with jsdom so feel free to disregard my comments, but I'd love to get this resolved asap, and currently there seems to be no other proposals on the table 😃

domenic:
We need to tackle this problem comprehensively. Let's roll discussions into https://github.com/jsdom/jsdom/issues/2751.

PR #2334, opened 2018-08-18
Update dependencies

domenic:
This notably includes a change to the origin of file: URLs in whatwg-url v7, which is a breaking change for jsdom users.


yarn.lock
```diff
@@ -3,16 +3,12 @@
...
+  version "1.3.4"
+  resolved "https://registry.yarnpkg.com/JSONStream/-/JSONStream-1.3.4.tgz#615bb2adb0cd34c8f4c447b5f6512fa1d8f16a2e"
   dependencies:
     jsonparse "^1.2.0"
     through ">=2.2.7 <3"
 
-abab@^1.0.4:
-  version "1.0.4"
-  resolved "https://registry.yarnpkg.com/abab/-/abab-1.0.4.tgz#5faad9c2c07f60dd76770f71cf025b62a63cfd4e"
-
```
TimothyGu: Yay for only one copy of `abab`!

PR #2335, opened 2018-08-19
Roll Web Platform Tests

Zirro:
Also disables a flaky test and removes the `needs-await` reason.


test/web-platform-tests/to-run.yaml
```diff
@@ -195,7 +196,9 @@ xmldomparser.html: [fail, The status returned is "loading", instead of "complete
 
 DIR: hr-time
 
-idlharness.html: [fail, Depends on fetch]
+idlharness.any.html: [fail, Depends on fetch]
+idlharness.any.serviceworker.html: [timeout, Needs Service Worker implementation]
```
domenic: We should automatically skip/omit these like we do .worker.html

test/web-platform-tests/to-run.yaml
```diff
@@ -413,6 +417,12 @@ DIR: html/editing/focus
 composed.window.html: [fail, Not implemented]
 focus-01.html: [timeout, scrollIntoView() not implemented]
 focus-02.html: [timeout, scrollIntoView() not implemented]
+inert/inert-does-not-match-disabled-selector.html: [fail, Not implemented]
```
domenic: Interesting that these got merged before the corresponding spec change... they should have ".tentative" in their name, in theory. Ah well.

PR #2339, opened 2018-08-21
Add html element to `options` for custom resource loader

### LINKED ISSUE #2338: Element is not passed to a custom resource loader.
### Basic info:

- **Node.js version:** 8.11.3
- **jsdom version:** v12

### Minimal reproduction case

```js
const ResourceLoader = require('jsdom').ResourceLoader;

export class CustomResourceLoader extends ResourceLoader {
    fetch(url, options) {
        console.log(options.element);
    }
}
```

sarvaje:
Fix #2338


domenic:
Thanks, this seems reasonable, but can you add tests and documentation?

sarvaje:
@domenic I have added a test and some documentation. Let me know if I need to add something else.

domenic:
@sarvaje I made some tweaks so that the element you get is a wrapper, not an impl. Please take a look and tell me if they look good. If so we can merge this,

sarvaje:
@domenic the changes look good to me :) Thanks!

Issue #2213, opened 2018-05-01
When parsing XML, JSDOM should produce CDATASection and Comment nodes even when their content is the empty string


lddubeau:
### Basic info:

- **Node.js version:** 8.11.1
- **jsdom version:** 11.10.0

### Minimal reproduction case

```js

const { JSDOM, VirtualConsole } = require("jsdom");

const vc = new VirtualConsole();
vc.sendTo(console);

const { window } = new JSDOM("", { virtualConsole: vc });

const { DOMParser, Node } = window;

function assert(msg, actual, expected) {
  if (actual !== expected) {
    console.log(`${msg}: got ${actual} but expected ${expected}`);
  }
  else {
    console.log(`${msg}: passed`);
  }
}

function assertCDATA(node, value) {
  try {
    assert("CDATA node", node.nodeType, Node.CDATA_SECTION_NODE);
    assert("CDATA node content", node.textContent, value);
  }
  catch (ex) {
    console.log(`Failed with: ${ex}`);
  }
}

function assertComment(node, value) {
  try {
    assert("comment node", node.nodeType, Node.COMMENT_NODE);
    assert("comment node value", node.textContent, value);
  }
  catch (ex) {
    console.log(`Failed with: ${ex}`);
  }
}

const parser = new DOMParser();
const doc = parser.parseFromString(
  "<doc><![CDATA[]]><!----></doc>",
  "text/xml");
const de = doc.documentElement;

function check() {
  assert("content", de.textContent, "");
  assert("length", de.childNodes.length, 2);
  assertCDATA(de.childNodes[0], "");
  assertComment(de.childNodes[1], "");
}

check();
```

`check` fails because the CDATA and comment nodes are missing.

### How does similar code behave in browsers?

In Chrome and Firefox `check` does not fail. In those browsers, the result of parsing the XML shown above includes a CDATA section that has the text "" and a comment with the text "", whereas in JSDOM the nodes are just not created.

The XML standard does not prescribe a behavior. It specifies what a CDATA section is, and what a comment is but does not specify whether such constructs should be seen by client applications if the data they contain is the empty string. It is up to the parser to decide whether it shows them to the client code or not. I did not see anything in the DOM or HTML living standards that mandates one behavior or another.

Although I've not found a specification that mandates creating the nodes, I have two arguments:

1. Chrome and Firefox create a CDATA node that has for data an empty string and the same for Comment.

[...]

PR #2340, opened 2018-08-21
Move from sax to saxes for XML parsing

lddubeau:
## Motivation

Sax does not appear to be actively developed anymore, which means that request for fixes or features go unanswered. The [PR filed](https://github.com/isaacs/sax-js/pull/231) to fix the issue reported in #2213 has gone unanswered for over 3 months.  See also [this issue](https://github.com/isaacs/sax-js/issues/238) asking whether the project is still alive.

I recently forked saxes from sax for my own XML parsing purposes, and fixed a lot of issues with how sax parses XML. 

## Benefits for JSDOM

* Better conformance with the XML specification. ([XML 1.0 fifth edition](https://www.w3.org/TR/2008/REC-xml-20081126/) and [XML Namespaces 1.0 third edition](http://www.w3.org/TR/2009/REC-xml-names-20091208/), to be specific.) Known limitation regarding conformance: saxes does not do well-formedness checks *inside* the DTD that may be included in a document. It is rare nowadays to rely on DTDs when doing XML processing in browser environments. Adding code to check that the DTD is well-formed would require *a lot of work* for very little benefit.

* Parsing behavior more closely aligned with those browsers that people normally use (like Chrome and Firefox.) The issue with empty comments and empty CDATA sections is one example: JSDOM+sax would not create nodes for `<!---->` and `<![CDATA[]]>` because sax did not report them. JSDOM+saxes produces nodes for both, which is consistent with other browsers.

## Breaking Changes

Prior to this PR, JSDOM accepted malformed XML. A prime example is the empty string. An empty string is not a well-formed XML document. Prior to this PR you could do `new DOMParser().parseFromString("", "text/xml")` and get an unpopulated document. After this PR, this is not possible anymore: you get an error. (This is consistent with how Chrome and Firefox behave.) So users may have been able to run JSDOM with malformed XML documents in the past. They won't be able to do this anymore.

## Other Considerations

I've noticed that prior to this PR, JSDOM was unable to properly handle `<script><![CDATA[ ... ]]></script>`. It so happens that a lot of the tests in web-platform-test suite that are in XML files have to make use of CDATA sections in `script` elements, so that `<` and `&` don't need escaping. Prior to this PR, upon encountering such `script` construct, JSDOM would build an `HTMLScriptElement` containing literally `<![CDATA[...`, which obviously is not valid JavaScript and this would cause the test to immediately fail. This made quite a few tests that are part of the WPT fail. This PR fixes the immediate problem of handling CDATA in `script`. There were a few tests that immediately became runnable after the parsing changes were made. This was reported by the test runner and I've turned on these tests (edited `to-run.yaml`). However, I've not reviewed the rest of the tests that were failing due to CDATA in `script`. There may be some tests in the lot that are failing for some easily fixable reason.


lib/jsdom/browser/htmltodom.js
```diff
@@ -67,90 +67,89 @@ module.exports = class HTMLToDOM {
...
-          localName = prefix;
-          prefix = null;
-        }
+      for (const key of Object.keys(tag.attributes)) {
+        const attr = tag.attributes[key];
 
-        if (prefix === "") {
+        let { prefix, local } = attr;
+        const { uri, value } = attr;
+        if (prefix === "xmlns" && local === "") {
```
domenic: Any interest in changing saxes to fix this bug?

lddubeau: I don't know that I'd call it a "bug". Though I did call it a "peculiarity" in the comment. (I reworded an existing comment but "peculiarity" is my choice of words :smiley:  .) Looking at the commit history of sax, it seems to me it was a conscious choice to do it this way. There's some logic to it, and there are use-case scenarios where the current behavior is quite handy. Then again, as a "peculiarity", it violates the principle of least surprise, and it is not concordant with the XML Infoset. (Not that the Infoset drives how saxes should do things, but again: principle of least surprise.) So I guess the balance weighs in favor of removing that peculiar behavior.

lddubeau: I've released a new version of saxes and amended this PR.

Zirro:
Great to see the many failures due to CDATA-in-script fixed! Unfortunately, the previously quick fail in `/dom/nodes/ParentNode-querySelector-All-xht.xht` seems to have turned into a one minute timeout instead and its failure reason should be changed to reflect this.

lddubeau:
@Zirro I've amended the last commit to include the change you pointed out.

Two questions come to my mind though:

1. Before marking `ParentNode-querySelector-All-xht.xht` as a timeout, I reran it and it *passed*, after something like 60 seconds. In every single test suite I'm responsible for, a timeout is a failure. Is there some special situation here that requires handling timeouts the way they are handled?

2. How can I detect cases like `ParentNode-querySelector-All-xht.xht`. Can I run the suite with a flag that would immediately point these out? Or do I have to look at each and every test in the test report and look for those that may happen to be timing out?

Zirro:
Those are good questions. In this case the passing status actually means that the test failed as expected. We recently started running tests that are known to fail quickly in order to see if that's still the case after any given change. Considering the large amount of WPTs and their complexity, it's actually been pretty helpful in finding unexpected passing tests in various directories.

While we may be a bit limited by Mocha in the ways we can express "expected failures" during the run, we could certainly look into some ways of making it clearer that the tests have indeed still failed.

As for timeouts, we separate them from other failures by giving them the `timeout` status since they would take too long to run every time. The process could do with some refinement though - much of it can likely be automated later on.

> Or do I have to look at each and every test in the test report and look for those that may happen to be timing out?

It depends on the cause of the timeout. Some kinds are caught by `testharness.js` and result in an error message regarding a timeout at the end of the run, while others are skipped by the test framework after too much time has passed. These can usually be found by looking for their run time printed in red next to their names in the test log, though it may still take some work to separate them from tests which actually do pass but do so very slowly.

domenic:
> In this case the passing status actually means that the test failed as expected.

We should change the "test name" in the harness output to something like "[EXPECTED FAIL] .../filename/...

PR #2347, opened 2018-08-23
Add support for Shadow DOM

### LINKED ISSUE #2343: Support for Shadow DOM
I have been working over the last few weeks on the implementation of Shadow DOM API in JSDOM. You can find the current implementation here: https://github.com/jsdom/jsdom/compare/master...pmdartus:shadow-root

There are still a handful of tests failing, but most of it is there. I am currently looking for guidance/help from the JSDOM core team and community to make this real. 🎉 

There are multiple outstanding issues with this branch:

* With the introduction of Shadow DOM comes `HTMLSlotElement` and `ShadowRoot`. Since those objects are used in multiple places in the codes, it creates circular dependencies. In order to avoid this issue, I lazy require those when needed, however, I am looking for a better pattern now.
* I have spent a lot of time on the [dispatch event](https://dom.spec.whatwg.org/#concept-event-dispatch). With dispatching the events during the CAPTURE_PHASE. I can speak in more details if someone is interested to debug that with me.
* Performance! The addition of slots and slotable has a non-negligible performance impact on the DOM update. I solved some of the low-hanging fruits (without those wpt takes more than 2 hours to run), but I still feel that the test are slower to run.
* And much more ...

pmdartus:
This PR adds support for Shadow DOM APIs.

Fix #2343


lib/jsdom/living/helpers/shadow-dom.js
```diff
@@ -0,0 +1,252 @@
...
+  "h6",
+  "header",
+  "main",
+  "nav",
+  "p",
+  "section",
+  "span"
+]);
+
+// TODO: Evaluate if it should be somewhere else.
```
pmdartus: Importing `../generated/Node` or `../generated/ShadowRoot` would result in a circular dependency. Can we find a good way to avoid this?

lib/jsdom/living/nodes/Node-impl.js
```diff
@@ -294,23 +312,52 @@ class NodeImpl extends EventTargetImpl {
...
+          break;
+        }
+      }
+    }
+
+    if (hasSlotDescendant) {
+      assignSlotableForTree(getRoot(this));
+
+      // TODO: if the node is removed from is parent at 9., the root will always be itself ?!
+      assignSlotableForTree(getRoot(oldChildImpl));
```
pmdartus: In the [remove node algorithm](https://dom.spec.whatwg.org/#concept-node-remove), the root of the node at this point ([12.2](https://dom.spec.whatwg.org/#ref-for-concept-tree-root%E2%91%A1%E2%93%AA)) is always the node itself since it has already been detached.
Am I missing something here?

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -82,63 +87,134 @@ class EventTargetImpl {
...
+        if (isSlotable(parent) && parent._assignedSlot) {
+          slotable = parent;
+        }
+
+        relatedTarget = retarget(eventImpl.relatedTarget, parent);
+
+        // TODO: Understand why checking for Window is needed? Window is not a Node...
+        if (
+          (isNode(parent) && isShadowInclusiveAncestor(getRoot(targetImpl), parent)) ||
+          idlUtils.wrapperForImpl(parent).constructor.name === "Window"
```
pmdartus: In order to properly dispatch event on `Window`, we also need to check if the *parent* is `Window` since `Window` is not a `Node`. This is not consistent with the [dispatch event](https://dom.spec.whatwg.org/#ref-for-boundary-point-node%E2%91%A0) algorithm

lib/jsdom/living/events/Event-impl.js
```diff
@@ -78,6 +80,57 @@ class EventImpl {
     }
   }
 
+  // https://dom.spec.whatwg.org/#dom-event-composedpath
+  composedPath() {
```
pmdartus: The last failing tests are related to `composedPath` are still failing:
* [Event Path with a slot in a closed Shadow Root.](https://github.com/web-platform-tests/wpt/blob/master/shadow-dom/event-composed-path.html#L198-L219)
* [Event Path with slots in nested ShadowRoots: closed > closed.](https://github.com/web-platform-tests/wpt/blob/master/shadow-dom/event-composed-path.html#L251-L280)

I don't understand how the [composedPath algorithm](https://dom.spec.whatwg.org/#dom-event-composedpath) would work in the case of slot content in closed shadow trees.

----

Let's take for example: [Event Path with a slot in a closed Shadow Root.](https://github.com/web-platform-tests/wpt/blob/master/shadow-dom/event-composed-path.html#L198-L219)

```
test9
└── host1
    ├──#shadowRoot (closed)
    │   └──slot
    └── target
```

For an event dispatched on target, the following event path would get produced:

| item       | target | rootOfClosedTree | slotInSlosedTree |
|------------|--------|------------------|------------------|
| target     | target | false            | false            |
| slot       |        | false            | true             |
| shadowRoot |        | true             | false            |
| host1      |        | false            | false            |
| test9      |        | false            | false            |

Now if we consider target to the context object of composedPath the result would be: `[target, slot, shadowRoot, host1, test9]` while it should be `[target, host1, test9]`

| item       | hiddenSubtreeLevel | hasSeenCurrentTarget | reversedComposedPath                   |
|------------|--------------------|----------------------|----------------------------------------|
| target     | 0                  | true                 | test9, host1, shadowRoot, slot, target |
| slot       | 0                  | false                | test9, host1, shadowRoot, slot         |
| shadowRoot | 0                  | false                | test9, host1, shadowRoot               |
| host1      | 0                  | false                | test9, host1                           |
| test9      | 0                  | false                | test9                                  |

lib/jsdom/living/events/EventTarget-impl.js
domenic: I assume tests fail if you follow the spec exactly? In that case, maybe file a spec bug?

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -147,41 +223,63 @@ module.exports = {
...
 };
 
-function invokeEventListeners(listeners, target, eventImpl) {
-  const wrapper = idlUtils.wrapperForImpl(target);
-  const document = target._ownerDocument || (wrapper && (wrapper._document || wrapper._ownerDocument));
-  // Will be falsy for windows that have closed
-  if (!document) {
-    return;
+// https://dom.spec.whatwg.org/#concept-event-listener-invoke
+function invokeEventListeners(tuple, eventImpl /* , legacyOutputDidListenersThrowFlag */) {
```
domenic: Why is the flag commented out, here and below? We should leave a comment explaining.

lib/jsdom/living/helpers/shadow-dom.js
domenic: You could do nodeImpl instanceof NodeImpl, if you import Node-impl.js. That's only slightly better though and probably also has circular dependency issues. This is probably fine.

lib/jsdom/living/helpers/shadow-dom.js
```diff
@@ -0,0 +1,252 @@
...
+function isNode(nodeImpl) {
+  // Use an approximation by checking the presence of nodeType instead of instead of using the isImpl from
+  // "../generated/Node" to avoid introduction of circular dependencies.
+  return Boolean(nodeImpl && nodeImpl.nodeType);
```
domenic: Maybe check `"nodeType" in nodeImpl` instead. This only works because nodeType is never 0, but I didn't know that until I checked the spec.

lib/jsdom/living/helpers/shadow-dom.js
```diff
@@ -0,0 +1,252 @@
...
+    ) {
+      return a;
+    }
+
+    a = getRoot(a).host;
+  }
+}
+
+// https://dom.spec.whatwg.org/#get-the-parent
+function getParent(eventTarget, event) {
```
domenic: I would rather implement this as private methods on ShadowRoot, Document, and Node, that are called polymorphically. That would be closer to the spec.

lib/jsdom/living/helpers/shadow-dom.js
```diff
@@ -0,0 +1,252 @@
...
+    if (child.localName === "slot" && child.name === slotable._slotableName) {
+      return child;
+    }
+  }
+
+  return null;
+}
+
+module.exports = {
+  VALID_HOST_ELEMENT_NAME,
```
domenic: Can we export a testing function, instead of a set?

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -367,10 +368,13 @@ class DocumentImpl extends NodeImpl {
     this.write(...arguments, "\n");
   }
 
-  // This is implemented separately for Document (which has a _ids cache) and DocumentFragment (which does not).
-  getElementById(id) {
-    // Return the first element with this ID.
-    return this._ids[id] && this._ids[id].length > 0 ? this._ids[id][0] : null;
+  getElementById(elementId) {
```
domenic: I am worried about the performance degradation here of removing the cache. Have you done any tests?

lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -91,39 +53,33 @@ class ElementImpl extends NodeImpl {
...
-
     // update classList
     if (name === "class" && this._classList !== undefined) {
       this._classList.attrModified();
     }
+
+    this._attrModifiedSlotableMixin(name, value, oldValue);
+  }
+
+  get slot() {
```
domenic: Can we just use [Reflect]? If not, we should add a comment explaining why.

lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -359,6 +315,51 @@ class ElementImpl extends NodeImpl {
     return 0;
   }
 
+  // https://dom.spec.whatwg.org/#dom-element-attachshadow
+  attachShadow(init) {
+    if (this.namespaceURI !== HTML_NS) {
+      throw new DOMException(
+        "This element does not support attachShadow.",
```
domenic: Let's explain a bit more in the error messages, so that the two cases are distinct.

lib/jsdom/living/nodes/HTMLOutputElement-impl.js
```diff
@@ -19,6 +20,10 @@ class HTMLOutputElementImpl extends HTMLElementImpl {
   get labels() {
     return getLabelsForLabelable(this);
   }
+
+  get form() {
+    return closest(this, "form");
```
domenic: This does not fully implement https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner . Does implementing it in this incomplete way help pass other tests? (E.g. shadow DOM tests.) If so then this is OK, but add a TODO comment explaining that it's incomplete.

If it's not necessary for shadow DOM tests, then let's remove this change until we can do a complete implementation.

lib/jsdom/living/nodes/HTMLSlotElement-impl.js
```diff
@@ -0,0 +1,53 @@
...
+const { getRoot, assignSlotableForTree, findFlattenedSlotables } = require("../helpers/shadow-dom");
+
+class HTMLSlotElementImpl extends HTMLElementImpl {
+  constructor(args, privateData) {
+    super(args, privateData);
+    this._assignedNodes = [];
+  }
+
+  // https://html.spec.whatwg.org/#dom-slot-name
+  get name() {
```
domenic: Use [Reflect]

lib/jsdom/living/nodes/HTMLSlotElement-impl.js
```diff
@@ -0,0 +1,53 @@
...
+      if ((!value && oldValue === "") || (value === "" && !oldValue)) {
+        return;
+      }
+
+      assignSlotableForTree(getRoot(this));
+    }
+  }
+
+  // https://html.spec.whatwg.org/#dom-slot-assignednodes
+  assignedNodes(options) {
```
domenic: These two method implementations are quite nice.

lib/jsdom/living/nodes/HTMLSlotElement-impl.js
```diff
@@ -0,0 +1,53 @@
...
+  set name(value) {
+    this.setAttribute("name", value);
+  }
+
+  _attrModified(name, value, oldValue) {
+    super._attrModified(name, value, oldValue);
+
+    // https://dom.spec.whatwg.org/#slot-name
+    if (name === "name" && value !== oldValue) {
+      if ((!value && oldValue === "") || (value === "" && !oldValue)) {
```
domenic: Follow the spec more exactly here, with a series of `if` statements that do early returns. Collapsing them all into a big if statement makes it harder for me to understand the correspondence.

lib/jsdom/living/nodes/Node-impl.js
```diff
@@ -86,13 +90,9 @@ class NodeImpl extends EventTargetImpl {
...
   }
 
-  getRootNode() {
-    // ignore option for composed, because of no Shadow DOM support
-    let root;
-    for (const ancestor of domSymbolTree.ancestorsIterator(this)) {
-      root = ancestor;
-    }
-    return root;
+  getRootNode(options = {}) {
```
domenic: The defaulting should be taken care of by the Web IDL layer. So this should just be `getRootNode(options)` and the next line should be unnecessary.

lib/jsdom/living/nodes/Node-impl.js
```diff
@@ -167,6 +167,8 @@ class NodeImpl extends EventTargetImpl {
       newChildImpl._ownerDocument = this._ownerDocument;
     }
 
+    // https://dom.spec.whatwg.org/#concept-node-ensure-pre-insertion-validity
```
domenic: Not necessary in this pull request, but it would be really nice to factor out these things into helper functions, like they are in the spec, instead of inlining them into the methods of the Node class. This has been a long-standing problem.

lib/jsdom/living/nodes/Node-impl.js
domenic: Not sure! Please file a spec bug.

lib/jsdom/living/nodes/ShadowRoot-impl.js
```diff
@@ -0,0 +1,50 @@
+"use strict";
+
+const { domToHtml } = require("../../browser/domtohtml");
+
+const { HTML_NS } = require("../helpers/namespaces");
+const { domSymbolTree } = require("../helpers/internal-constants");
+const { setInnerHTML } = require("../helpers/html");
+
+const DocumentFragment = require("./DocumentFragment-impl").implementation;
```
domenic: DocumentFragmentImpl

lib/jsdom/living/nodes/ShadowRoot-impl.js
```diff
@@ -0,0 +1,50 @@
...
+
+  set innerHTML(html) {
+    if (html === null) {
+      html = "";
+    }
+
+    setInnerHTML(this._ownerDocument, this, html);
+  }
+
+  // ShadowRoot doesn't have a namespace URI per say. However the namespaceURI is used by
```
domenic: Great comment.

lib/jsdom/living/nodes/Slotable-impl.js
```diff
@@ -0,0 +1,35 @@
...
+// https://dom.spec.whatwg.org/#mixin-slotable
+// https://dom.spec.whatwg.org/#light-tree-slotables
+class SlotableMixinImpl {
+  _initSlotableMixin() {
+    this._slotableName = "";
+  }
+
+  _attrModifiedSlotableMixin(name, value, oldValue) {
+    if (name === "slot") {
+      if ((!value && oldValue === "") || (value === "" && !oldValue)) {
```
domenic: As before, don't collapse down the if statements; make the spec correspondence exact.

lib/jsdom/living/helpers/shadow-dom.js
pmdartus: Same issue when importing `Node-impl.js` and `ShadowRoot-impl.js`.

lib/jsdom/living/nodes/Node-impl.js
pmdartus: https://github.com/whatwg/dom/issues/691

lib/jsdom/living/helpers/shadow-dom.js
pmdartus: I went down this route originally. However, adding `getParent` private method requires the invoker to invoke it defensively.

In the case where you want to climb the parent chain, invoking getParent on `Document` return the global object but invoking getParent on the global object  throws since the global object doesn't have the private method. So you need to guard the getParent invocation with an if statement. The `getParent` method doesn't suffer from the same issue.

What about an intermediary approach where `getParent` is a function invoking the private method if present, otherwise returns `null`. This approach would be closer to the spec without requiring the invoker to be defensive.

```js
funtion getParent(nodeImpl, eventImpl) {
	return "_getParent" in nodeImpl ? nodImpl.getParent(eventImpl) : null;
}

class DocumentImpl {
	// ...

	_getParent(eventImpl) {
		// The getParent logic on document
	}
}
```

lib/jsdom/living/helpers/custom-elements.js
```diff
@@ -0,0 +1,25 @@
...
+  "missing-glyph"
+]);
+
+// https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name
+function isValidCustomElementName(name) {
+  if (RESTRICTED_CUSTOM_ELEMENT_NAME.has(name)) {
+    return false;
+  }
+
+  return name.match(/^[a-z][-.0-9_a-z]*-[-.0-9_a-z]*$/);
```
Zirro: [`RegExp.prototype.test()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) is likely a better option here, since [`String.prototype.match()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match) returns an array for matches.

lib/jsdom/living/nodes/Element-impl.js
pmdartus: I don't feel extremely inspired, let me know what you think about those error messages.

lib/jsdom/living/nodes/Document-impl.js
pmdartus: I changed the implementation to use the cached ids and find the first one that is not enclosed in a shadow tree.

lib/jsdom/living/helpers/shadow-dom.js
pmdartus: I tried to move the `_getParent` to the window object, however, I ran in issues since the window impl is not like the standard ones. I don't know yet where do the issues come from.

pmdartus:
@domenic This PR is ready for another round of review. I am still facing some issues to move the `getParent` on the window object. If you still want to move it to the appropriate objects I will probably need someone help.

pmdartus:
The PR is now rebased. I update the comment to reflect the spec bugs status. I would love if you can get a stab at the polymorphic call to `getParent`.

caridy:
Thank you so much @domenic and @Zirro for supporting @pmdartus' amazing effort to get this in jsdom.

PR #2358, opened 2018-09-05
Implement logic in play and pause media element methods

georgesbiaux:
This PR aims to implement a minimal logic for the `play` and `pause` methods by setting the `paused` property.


domenic:
Thanks, but this doesn't follow the spec for those methods, so it's best to leave them as not-implemented.

Issue #1365, opened 2016-01-25
Add optional support for image loading through Canvas.Image


lehni:
Now that there is support for the Canvas API through node-canvas, I think it makes sense to also support image loading through the `Canvas.Image` class from the same module.

I am willing to work on this, but would need a few pointers, as I am new to the JSDom internals.

I have looked at the `HTMLCanvasElementImpl` class and understand roughly how to patch this into `HTMLImageElementImpl`. The main problem I am currently facing is that I'm not sure how i am supposed to load the binary image files. Should I use `resourceLoader` for this, or ist that only for text files? 

This relates to #189, but since this is about an actual implementation through node-canvas, I thought it might be better to open a separate issue.

Sebmaster:
> Should I use resourceLoader for this, or ist that only for text files?

Yes, resourceLoader is for all resources used for a specific Window. You should fine the resourceLoader in `this.ownerDocument._resourceLoader` (iirc). It takes care of automatically canceling requests when `window.close` is called and firing an error event when the request fails (which might actually not be applicable for every case, but I haven't found one yet, so we've left it in there).

lehni:
Ok, but how to load the binary data with it? So far I have tried:

    resourceLoader.download(url, null, document._cookieJar, null, (error, data) => {...}

As well as:

    resourceLoader.load(this, url, data => {...}

In both cases, data is a string.

Sebmaster:
The actual request is done with the [request](https://github.com/request/request#requestoptions-callback) library, which "specifies" encoding: utf-8 by default, you should be able to provide an option object with that set to null to get a Buffer.

lehni:
Ah, great! This works, I get binary data now:

    resourceLoader.download(value, { encoding: null }, document._cookieJar, null, (error, data) => {...}

lehni:
I have a feeling I should be using `resourceLoader.load` instead, as that appears to be handling the closed windows and all, but there is no way to pass on encoding options to that, unfortunately (it only receives these arguments: `(element, url, callback)`).

lehni:
Another reason for using `resourceLoader.load()` is that only this function handles loading of local files... But there really is no way to pass on the encoding option here. File loading is always done with utf8 encoding also, and the data that it is reading into is a string, not a buffer. It looks like this would require quite a few changes to `resourceLoader`, unfortunately.

domenic:
Please feel free to fix resource loader :)

lehni:
Great, I shall try : )

[...]

PR #2361, opened 2018-09-12
canvas/canvas-prebuilt: show error when module is available but fails to load

joscha:
In some environments, neither `canvas`, nor `canvas-prebuilt` will be installed - as it is an optional dependency, we will just ignore their absence, however when one or both of them are installed, there is a chance that loading them fails, as they are depending on native bindings.

For example when building a docker image that runs `sid`, `libpng12-0` is not available, which means that loading (requiring) canvas-prebuilt fails, due to:

> `error while loading shared libraries: libpng12.so.0: cannot open shared object file: No such file or directory`

It took me quite a while to figure out why canvas-prebuilt worked outside the docker container, but not inside, even though both systems were linux.
The way the try/catch is currently set up means that loading errors are swallowed. The change in this PR makes loading changes at least visible in a similar manner to how they would show up when just doing `require('canvas-prebuilt')` anywhere in the bootstrap code.

references https://github.com/jsdom/jsdom/pull/1754, https://github.com/jsdom/jsdom/issues/1365, https://github.com/tcoopman/image-webpack-loader/issues/95

### DIFF
--- a/lib/jsdom/utils.js
+++ b/lib/jsdom/utils.js
@@ -169,6 +169,13 @@ exports.Canvas = null;
       exports.Canvas = null;
     }
   } catch (e) {
+    try {
+      require.resolve(moduleName);
+      // the module is installed, but loading it failed
+      console.error("Failed loading " + moduleName, e);
+    } catch(_) {
+      // module is not installed
+    }
     exports.Canvas = null;
   }
   return exports.Canvas !== null;
--- a/lib/jsdom/utils.js
+++ b/lib/jsdom/utils.js
@@ -163,20 +163,15 @@ exports.treeOrderSorter = function (a, b) {
 exports.Canvas = null;
 ["canvas", "canvas-prebuilt"].some(moduleName => {
   try {
-    exports.Canvas = require(moduleName);
-    if (typeof exports.Canvas !== "function") {
-      // In browserify, the require will succeed but return an empty object
-      exports.Canvas = null;
-    }
+    require.resolve(moduleName);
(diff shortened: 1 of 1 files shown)


lib/jsdom/utils.js
```diff
@@ -169,6 +169,13 @@ exports.Canvas = null;
       exports.Canvas = null;
     }
   } catch (e) {
+    try {
+      require.resolve(moduleName);
+      // the module is installed, but loading it failed
+      console.error("Failed loading " + moduleName, e);
```
joscha: Alternatively, we could potentially also rethrow here, which would make it even more obvious, however that would be a backwards incompatible change. Let me know if you prefer to rethrow.

domenic:
Very interesting, thanks! Let's flip this around though: first let's do require.resolve. If that fails (how does it fail? return null? Docs aren't clear) then set exports.Canvas to null. Otherwise set exports.Canvas to `require(moduleName)`, without any try/catch. Then errors will bubble naturally.

WDYT?

joscha:
That sounds good, I can make the change?

joscha:
> how does it fail? return null? Docs aren't clear

it errors with Node 10:

1__node__node_

I'd assume that behaviour hasn't changed.

joscha:
changes in 6763097 @domenic - unsure if I like it more, but happy with either as long as there is some feedback.

domenic:
Thanks. I like this a lot better since the error directly comes from Node.js.

Issue #2368, opened 2018-09-26
`button` associated with a form using `form` attribute doesn't trigger form `submit` event


jamietre:
### Basic info:

- **Node.js version:** 8.12.0
- **jsdom version:** 12.0.0

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const dom = new JSDOM(`
<form id="foo"></form><button id="foo-button" form="foo"></button>
`, options);

const document = dom.window.document;
document.getElementById('foo').addEventListener('submit', (e) => {
    console.log('submit event fired');
});
document.getElementById('foo-button').click();
// event listener callback is never invoked
```

Repro repo: https://github.com/jamietre/jsdom-form-attribute/

### How does similar code behave in browsers?

https://codepen.io/jamietre/pen/BOggdj

Triggering a click event on a `button` associated to a form using the `form` attribute should behave the same as if it were associated by nature of being a child, e.g. cause a `submit` event on the form.

Relevant spec: https://html.spec.whatwg.org/#association-of-controls-and-forms

PR #2369, opened 2018-09-26
Allow binding to forms using 'form' attribute

jamietre:
Resolves https://github.com/jsdom/jsdom/issues/2368
Spec: https://html.spec.whatwg.org/#association-of-controls-and-forms

This change permits using the `form` attribute of form elements to associate them to a form without them being children.

This is my first pull request to jsdom so apologies in advance if I haven't done something correctly; did my best to follow the contribution guidelines for testing.  I wasn't sure whether I should add a new file or extend and existing one to test this since one existed that already covered related functionality. If it's not expected to merge new work into existing files in web-platform-tests lmk and I'll make a new file for this.


lib/jsdom/living/helpers/form-controls.js
```diff
@@ -214,3 +215,15 @@ exports.sanitizeValueByType = (input, val) => {
 
   return val;
 };
+
+exports.associatedForm = formControl => {
```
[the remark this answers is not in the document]
jamietre: Yes - as far as I can tell from reading the spec, It doesn't seem necessary to actually follow that imperative implementation for initializing/assigning the form owner based on certain events rather than just calculating the form owner on demand. I'll add a comment to this effect. While it's possible that there is some nuance I don't understand I think this implementation covers the vast majority of use cases.

jamietre:
Most of the `form-control-infrastructure` tests pass when `output` is added as a listed element and a few details of the implementation revised to match spec edge cases. The ones that won't pass are related to invalid html and "parser inserted flag"; since the form element isn't actually in the dom for these cases it would likely be a lot of work to fully support that part of the spec. I think this implementation works for all valid HTML.

lib/jsdom/living/helpers/form-controls.js
```diff
@@ -20,7 +20,10 @@ const whatwgURL = require("whatwg-url");
 
 const NodeList = require("../generated/NodeList");
 const { domSymbolTree } = require("../helpers/internal-constants");
+const { closest } = require("../helpers/traversal");
```
Zirro: It's just a small detail, but I noticed that the path on this line (and the older one above) goes out of the `helpers` directory and back in again.

jamietre: Oh, sure does. I think I just copied the pattern from the line above ;) I'll clean it up.

domenic:
It looks like the spec's super-complicated processing is necessary in order to pass some of those tests. I'll update the comments here to reflect that, and then merge this good-enough-for-now implementation.

PR #2373, opened 2018-10-03
Upgrade saxes from 3.1.2 to 3.1.3 to fix deprecation warning

dogan-acuris:
Running saxes version 3.1.2 displays a deprecation warning related to its dependency 'xmlchars'.

Patch version 3.1.3 fixes that deprecation warning.


domenic:
Thanks. Note that new installations of jsdom should not get this deprecation warning though, as `^3.1.2` will get you 3.1.3 automatically.

PR #2376, opened 2018-10-05
Lint scripts in HTML and JS files during same run

Zirro:
I occasionally see pull requests where linting errors from `to-upstream` don't appear on the initial CI run, as the second linting command only runs if there's no error from the first.

Using `;` will ensure that errors in both locations show up right away.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,7 @@
 node_modules
 
 benchmark/browser-bundle.js
+benchmark/browser-runner.html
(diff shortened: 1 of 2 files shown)


Zirro:
Ah, that's right. I focused on output for humans and forgot about the error codes.

However, as an alternative I found that (with a few changes to `.eslintignore`) it seems feasible to combine the two commands into one, even getting a small boost to performance.

PR #2377, opened 2018-10-06
Implement toggleAttribute()

Zirro:
https://dom.spec.whatwg.org/#dom-element-toggleattribute


lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -278,6 +278,32 @@ class ElementImpl extends NodeImpl {
     attributes.removeAttributeByNameNS(this, namespace, localName);
   }
 
+  toggleAttribute(name, force) {
```
domenic: Nit: s/name/qualifiedName/g?

Zirro: Yeah, I went with `name` to be consistent with the other methods in this file but I suppose we can change those as well later on.

PR #2379, opened 2018-10-06
potential runaway promise fix

fgarcia:
I am getting a warning from Bluebird which seems to match a runaway promise in the code above

http://bluebirdjs.com/docs/warning-explanations.html#warning-a-promise-was-created-in-a-handler-but-was-not-returned-from-it

Adding a return statement fixes the warning and "seems" logic.

Not sure if one should do the same for the catch statement

If the sensible thing is not to chain the return promise of item.check() please consider adding a "return null;" statement for Bluebird users. This silences the warning too and confirms the intention to not chain promises

### DIFF
--- a/lib/jsdom/browser/resources/resource-queue.js
+++ b/lib/jsdom/browser/resources/resource-queue.js
@@ -116,7 +116,7 @@ module.exports = class ResourceQueue {
       .then(data => {
         item.fired = 1;
         item.data = data;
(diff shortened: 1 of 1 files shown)


domenic:
Can you file a new issue using the issue template that explains what problem this fixes? Let's close the PR so we can have a discussion based around a concrete example.

Issue #1949, opened 2017-08-20
XHR is not sending empty headers correctly


domenic:
I'm having trouble figuring out whether this problem is in:

- jsdom
- request
- Node.js
- The tests themselves

But it's probably not a problem in the tests themselves, since Chrome can pass them. (Firefox fails similarly to us, however.)

The tests in question are the web platform tests

- XMLHttpRequest/setrequestheader-allow-empty-value.htm
- XMLHttpRequest/setrequestheader-allow-whitespace-in-value.htm

(which recently got stricter and more interesting)

PR #2380, opened 2018-10-06
Roll Web Platform Tests

Zirro:
In addition to the regular roll, I've been generous with the timeout status for slow (8+ seconds, mostly) failing tests. I think it's reasonable as most of them are unlikely to start passing without an effort targeted at them specifically.

This roll also seems to fix https://github.com/jsdom/jsdom/issues/1949 🎉, as the `xhr/setrequestheader-allow-empty-value.htm` and `xhr/setrequestheader-allow-whitespace-in-value.htm` started passing. The tests themselves haven't changed, so I suppose this comes down to changes in the WPT tooling.


domenic:
Can we add a new status, instead of using timeout for things that don't actually timeout? We'd treat it the same, but it would be useful when working on the tests later.

domenic:
Also, I'm confused by the sudden build failures on the to-upstream tests...

Zirro:
> Can we add a new status...

Sure. `fail-slow`?

> Also, I'm confused by the sudden build failures on the to-upstream tests...

I guess the `assert_exists()` method was removed from testharness. I'll look into replacing it.

Zirro:
I've resolved the issues.

PR #2388, opened 2018-10-09
Issues/2387 form reset input activation

### LINKED ISSUE #2387: Form reset functionality has been implemented without Reset event from spec
### Basic info:

- **Node.js version:** v8+
- **jsdom version:** v12.2.0

### Expected behavior
A reset button should dispatch a [Reset Event](https://developer.mozilla.org/en-US/docs/Web/Events/reset), on which default can be prevented, and custom logic can be initiated.

### Current Behavior:
JSDOM attempts to reset all child elements by looping over child elements and calling `el._formReset();`

### Minimal reproduction case

```js

const dom = new JSDOM(`
  <form id="my-form">
    <input name="test"/>
    <button type="reset" id="reset-button">Reset</button>
  </form>
`, {});

dom.window.document.getElementById('my-form').addEventListener('reset', function(e) {
  // This never gets called.
  e.preventDefault();
  console.log('resetting!', e);
});

dom.window.getElementById('reset-button').click(); 
```

https://jsbin.com/ciqacat/7/edit?html,js,console,output
(Link to a jsbin or similar suggested.)

epfremmer:
I think that this should resolve #2387, and get us closer to meeting the requirements in `html/semantics/forms/resetting-a-form/` specs.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLFormElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLFormElement-impl.js
@@ -79,7 +79,7 @@ class HTMLFormElementImpl extends HTMLElementImpl {
       { isTrusted: true }
     );
 
-    if (!this.dispatchEvent(submitEvent)) {
+    if (!this._dispatch(submitEvent)) {
       return;
     }
 
@@ -103,7 +103,7 @@ class HTMLFormElementImpl extends HTMLElementImpl {
       { isTrusted: true }
     );
 
-    if (!this.dispatchEvent(resetEvent)) {
+    if (!this._dispatch(resetEvent)) {
       return;
     }
 
--- a/test/web-platform-tests/to-run.yaml
+++ b/test/web-platform-tests/to-run.yaml
@@ -472,8 +472,6 @@ form_owner_and_table_3.html: [fail, complex form owner semantics not implemented
 
 DIR: html/semantics/forms/resetting-a-form
 
-reset-event.html: [fail, Expects event.isTrusted to equal true]
-
 ---
 
 DIR: html/semantics/forms/the-button-element


lib/jsdom/living/nodes/HTMLOutputElement-impl.js
```diff
@@ -23,6 +24,14 @@ class HTMLOutputElementImpl extends HTMLElementImpl {
...
     return formOwner(this);
   }
+
+  get defaultValue() {
+    return childTextContent(this);
+  }
+
+  set defaultValue(val) {
+    this.textContent = val;
+  }
```
Zirro: This code is missing some steps from the [spec](https://html.spec.whatwg.org/multipage/form-elements.html#dom-output-defaultvalue), and I don't think it makes sense to implement `defaultValue` without having `value` at the same time. It'd be great if you'd like to do that too, but otherwise you could add a dont-upstream version ([example](https://github.com/jsdom/jsdom/blob/master/test/web-platform-tests/to-upstream/dom/nodes/Element-closest-dont-upstream.html)) of the `reset-form.html` test with the `<output>`-specific tests disabled for now.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -232,6 +232,11 @@ class HTMLInputElementImpl extends HTMLElementImpl {
       if (form) {
         form._doSubmit();
       }
+    } else if (this.type === "reset") {
+      const { form } = this;
+      if (form) {
```
Zirro: Can you deduplicate the form-related lines with the ones in the submit condition code above, making it similar to the code in HTMLButtonElement?

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -77,6 +77,17 @@ class HTMLFormElementImpl extends HTMLElementImpl {
     notImplemented("HTMLFormElement.prototype.submit", this._ownerDocument._defaultView);
   }
 
+  _doReset() {
+    if (!isConnected(this)) {
+      return;
+    }
+    const ev = this._ownerDocument.createEvent("HTMLEvents");
+    ev.initEvent("reset", true, true);
```
Zirro: `createEvent`/`initEvent` are older methods for creating events which we don't use in newer code. [This approach](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/nodes/HTMLInputElement-impl.js#L222) is preferred instead.

lib/jsdom/living/nodes/HTMLOutputElement-impl.js
epfremmer: Added `value` attribute to output element, and fixed up the implementation to satisfy `value`, `defaultValue`, and form reset steps mentioned in that spec. I also added the `type` type attribute for posterity, and enabled some of the output element web-platform-tests in `html/semantics/forms/the-output-element`. Hopefully this is helpful =)

lib/jsdom/living/nodes/HTMLInputElement-impl.js
epfremmer: This should be fixed. It is formatted a little different than HTMLButtonElement because of the additional "checkbox"/"radio" condition, but at least no longer initializing the same constant in 2 different places.

lib/jsdom/living/nodes/HTMLFormElement-impl.js
epfremmer: I updated these to use the current method of creating/initializing events. I hope I am not overstepping, but I took the initiative to move the dispatch of both submit/reset events into the reset/submit methods so that `html/semantics/forms/resetting-a-form/reset-form-event-realm.html` tests could also be re-enabled.

test/web-platform-tests/to-run.yaml
```diff
@@ -472,9 +472,7 @@ form_owner_and_table_3.html: [fail, complex form owner semantics not implemented
 
 DIR: html/semantics/forms/resetting-a-form
 
-reset-event.html: [fail, Unknown]
-reset-form-event-realm.html: [fail, Unknown]
-reset-form.html: [fail, Unknown]
+reset-event.html: [fail, Expects event.isTrusted to equal true]
```
Zirro: This can be fixed by using the internal `_dispatch()` method which allows setting `isTrusted`, instead of `dispatchEvent()`.

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -66,18 +66,47 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
+      [
+        "submit",
+        { bubbles: true, cancelable: true }
+      ],
+      { isTrusted: true }
+    );
+
+    if (!this.dispatchEvent(submitEvent)) {
+      return;
+    }
```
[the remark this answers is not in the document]
epfremmer: I was thinking about the same thing. I moved event dispatching to `submit()/reset()` because having that in `_doSubmit / _doReset()` results in events not being dispatched when reset or submit is called directly.

I think that the method names may be a little confusing since `_doSubmit`  is being called from `HTMLInputElement` (https://github.com/jsdom/jsdom/blob/d5339dfb882ff2c03d7f5fa05cb842abc30a4efc/lib/jsdom/living/nodes/HTMLButtonElement-impl.js#L21). I decided to leave it that way because I wasn't sure about the the `isConnected(this)` check in those private methods and didn't want to cause any regression.

lib/jsdom/living/nodes/HTMLOutputElement-impl.js
```diff
@@ -10,19 +10,56 @@ class HTMLOutputElementImpl extends HTMLElementImpl {
   constructor(args, privateData) {
     super(args, privateData);
     this._labels = null;
+    this._defaultValue = null;
```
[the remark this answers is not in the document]
epfremmer: This should be fixed

lib/jsdom/living/nodes/HTMLOutputElement-impl.js
```diff
@@ -10,19 +10,56 @@ class HTMLOutputElementImpl extends HTMLElementImpl {
...
     return formOwner(this);
   }
+
+  get value() {
+    return this.textContent;
+  }
+
+  set value(val) {
+    this._valueMode = "value";
+    this._defaultValue = this.textContent;
```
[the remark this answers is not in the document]
epfremmer: I was storing the `defaultValue` to be able to satisfy from the spec.

> The reset algorithm for output elements is to set the element's value mode flag to default and then to set the element's textContent IDL attribute to the value of the element's default value (thus replacing the element's child nodes).

After playing around with the output element it behaved (in chrome) as the `defaultValue` being set to the last `textContent` when `value` is set. 
```
const output = document.createElement('output');
output.textContent = '123';
output.value; // "123"
output.defaultValue; // "123"

output.value = 'abc';
output.value; // "abc"
output.defaultValue; // "123"
``

Looking at this again today I am realizing that this would not work correctly when the value is set a second time because the default value will be overridden. It should probably change to this:

```
  set value(val) {
    if (this._valueMode === "default") {
      this._defaultValue = this.textContent;
    }

    this._valueMode = "value";
    this.textContent = val;
  }
```

I think in either case we need to store the defaultValue when putting the output element into "value" mode in order to properly reset its value on form reset don't we?

test/web-platform-tests/to-run.yaml
epfremmer: Oh awesome! I did not realize that was the problem. Thank you! This has been fixed as well.

domenic:
> I'd like to ask @domenic about this - do we have an appropriate method corresponding to the spec instructions here? _descendantAdded()/descendantRemoved() or childTextContentChangeSteps() would cover some of it, but "in any way" is very broad. Are we better off approximating this behaviour after all?

As far as I can tell, the spec should be using "child text content changed steps", as that's all that could cause the textContent value to change. I will file a spec bug.

domenic:
Oh, I see, the problem is that per spec it does descendants too. But I wonder if that's true in browsers. Very interesting.

domenic:
OK, I did some investigation. My suggested course of action is here: https://github.com/whatwg/html/issues/4163#issuecomment-435924443

PR #2389, opened 2018-10-11
Add support for srcdoc

### LINKED ISSUE #1792: iframe improvements
Would be great to have `srcdoc` for iframes.

It also seems that `iframe`'s `onload` fires synchronously from the time it is defined rather than inserted into the document.

carpasse:
Fixes #1792


domenic:
Thanks for working on this, and sorry for the long delay. Part of it was, I intimidated by srcdoc.

I don't think this follows the spec very exactly, but the existing code doesn't either, for all this iframe stuff. Unfortunately the test coverage for srcdoc in web platform tests seems pretty bad. But, you've made all the ones that do exist, pass!

So, with that all in mind, I am going to merge this PR as-is (with a minor style tweak). It probably doesn't make things much worse, and definitely helps at least some folks who have been wanting srcdoc support. If we find that aspects of it are imperfect, we can tweak them over time. Or maybe one day we'll totally rewrite our iframe implementation.

carpasse:
Thanks :-)

domenic:
Unfortunately this looks like it caused some errors in our test suite, which were not caught because they didn't actually fail the tests: https://travis-ci.org/jsdom/jsdom/jobs/462645344#L1054

Because of this I will revert for now, and look in to how we can make our tests properly fail when this sort of thing happens. Sorry about that :-/.

beiyehou:
@domenic Hello,any progress on this? I got this error since last week, finally reached to this post. The lastest version of jsdom seems still not support iframe srcdoc property.

PR #2395, opened 2018-10-17
Add Node pre-insert validation

pmdartus:
Implements node mutation algorithm.
Update references to mutation algorithm in the rest of the code base.

Fixes #1349 #2364

PR #2398, opened 2018-10-19
MutationObserver implementation

### LINKED ISSUE #639: add support for MutationObserver
https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver

pmdartus:
Fixes #639.
Depends on #2395, This PR borrows some of the code from the pre-insert removal refactor.


lib/jsdom/living/helpers/mutation-observers.js
```diff
@@ -0,0 +1,191 @@
...
+    oldValue,
+    [],
+    [],
+    null,
+    null
+  );
+}
+
+// https://dom.spec.whatwg.org/#mutation-observer-compound-microtask-queued-flag
+// TODO: Add this to the browsing context
```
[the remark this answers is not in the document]
domenic: Oof. That'd be a big ask.

We have something for storage... Each window has a `_currentOriginData`. That's not quite the same, as you can be in the same origin, but not in the same UORSOBCs, if you are two independent windows with no parent/child or opener relationship.

Let's see how far we can get with this as a global. It might mess things up, but we'll see...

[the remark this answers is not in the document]
domenic: Alternately we could put this on windows, although that won't work for elements created in documents without a browsing context (i.e. for which `._ownerDocument._defaultView` is undefined).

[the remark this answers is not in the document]
pmdartus: I guess I am fine with globals for now. It's something we can iterate over in the future.

test/web-platform-tests/to-run.yaml
```diff
@@ -800,8 +803,7 @@ leaktests/html-collection.html: [fail, Document.all is not implemented]
 leaktests/window-frames.html: [fail, Window.name is not implemeneted]
 scroll-to-the-fragment-in-shadow-tree.html: [fail, Requires a layout engine]
 slotchange-customelements.html: [fail, CustomElement.define is not implemented]
-slotchange-event.html: [timeout, MutationObserver is not implemented]
-slotchange.html: [timeout, MutationObserver is not implemented]
+slotchange-event.html: [timeout, Compound microtask scheduling issue]
```
pmdartus: This is an interesting aspect of this PR. Is that since we are using a microtask to simulate the behavior of a compound microtask we run into some scheduling issues between `MutationObserver` and `slotchange` events.

I don't see how to get around this issue without having to fully reimplement an event-loop in the context of jsdom.

domenic: Can you say a bit more? I was vaguely under the impression that we could implement compound microtasks by posting a single microtask that does several things.

test/web-platform-tests/to-run.yaml
```diff
@@ -135,7 +135,10 @@ Element-insertAdjacentElement.html: [fail, Not implemented]
 Element-insertAdjacentText.html: [fail, Not implemented]
 Element-matches.html: [fail, Whitespace in pseudo-class parameter]
 Element-webkitMatchesSelector.html: [fail, Whitespace in pseudo-class parameter]
-MutationObserver-**: [fail, Not implemented]
+MutationObserver-characterData.html: [fail, Range is not implemented]
+MutationObserver-childList.html: [fail, Range is not implemented]
+MutationObserver-document.html: [fail, Compound microtask scheduling issue]
+MutationObserver-inner-outer.html: [fail, innerHTML and outerHTML are not spec compliant]
```
domenic: In what way is this still true?

pmdartus: The `innerHTML` and `outerHTML` are directly manipulating the context object, using [`appendChild`](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/browser/htmltodom.js#L58-L65), [`removeChild`](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/helpers/html.js#L6-L8). If a mutation observer is attached to the tree, those mutations will enqueue new records.

As far as I understand [Element.innerHTML](https://w3c.github.io/DOM-Parsing/#dfn-innerhtml), the parser should produce a document fragment and execute the replace all algorithm on the context object. Same goes for `Element.outerHTML`.

It will require a large amount of refactor to the `htmltodom.js`, and I think it would deserve another PR. Here is an ugly draft of what I have in mind: https://github.com/pmdartus/jsdom/compare/impl-mutation-observer...pmdartus:inner-outer-html-parsing. The code `htmltodom.js` is duplicated, but at least `Element.innerHTML` and `Element.outerHTML` are conform to the spec. Those changes make `MutationObserver-inner-outer.html` pass.

I would be happy to wait for the parsing aglo to be fixed before merging this PR.

test/web-platform-tests/to-run.yaml
pmdartus: I was wrong for the microtask here. 

If you look at the test: `slotchange event must fire at the end of current microtask after mutation observers are invoked`:
1. Adds 2 mutation observers (one on `id` attribute and another one on `title`) and a slot with a `slotchange` event handler.
2. Update the `id` attribute and the append a child in the light dom of the host. We then enqueue a microtask to notify the listeners. At this point the `id` attribute mutation observer contains 1 mutation record, and the signal slot list contains the slot.
3. In the next microtask, the notifiy observer algorithm is executed:
	1. The mutation observer list is copied, it contains the 2 mutation observers, and it iterates over the 2 mutation observers
	2. The callback for the `id` mutation observer is executed and in it's callback it updates title attribute and append a new element the light dom of the host. By updating the `title` attribute a new mutation record is enqueued but this time for the `title` mutation observer.
	3. The callback fro the `title` mutation observer is invoked because it has a mutation record. And the test throws here since it expects that the `slotchange` event is invoked first.

I don't understand, how this test can pass according to the specification. [Chrome](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/dom/mutation_observer.cc?sq=package:chromium&g=0&l=349-350) and [Webkit](https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/dom/MutationObserver.cpp#L273) don't run into the same issue since they keep track of an active mutation observer list instead of iterating over all the mutation observers.

The test pass if we introduce a similar logic in `notifyMutationObservers` by only iterating over the mutations observers with mutation records.

```js
const notifyList = mutationObserverList.filter(mo => {
    return mo._recordQueue.length;
});
```

domenic: Very interesting! Sounds like it's worth opening a spec issue? And probably introducing our own similar logic for now, so that we behave similar to browsers?

test/web-platform-tests/to-run.yaml
domenic: Ah OK, makes sense! I'm happy to land fixes for innerHTML/outerHTML as a follow-up; it seems like a clean separation. Although I hope you have time to contribute them as well even after we land MutationObserver, as conformant innerHTML/outerHTML would be another great improvement :).

pmdartus: Yes, will do!

ijsnow:
Very excited for this to land!

lib/jsdom/living/mutation-observer/MutationObserver-impl.js
```diff
@@ -0,0 +1,88 @@
...
+        return registeredObserver.observer !== this;
+      });
+    }
+
+    this._recordQueue = [];
+  }
+
+  // https://dom.spec.whatwg.org/#dom-mutationobserver-takerecords
+  takeRecords() {
+    // TODO: Understand why wrapper is needed here.
```
domenic: This is needed because we don't unwrap collections. See https://github.com/jsdom/webidl2js/pull/108

lib/jsdom/living/mutation-observer/MutationObserver-impl.js
```diff
@@ -0,0 +1,88 @@
...
+class MutationObserverImpl {
+  // https://dom.spec.whatwg.org/#dom-mutationobserver-mutationobserver
+  constructor(args) {
+    const [callback] = args;
+
+    this._callback = callback;
+    this._nodeList = [];
+    this._recordQueue = [];
+
+    mutationObserverList.push(this);
```
domenic: The fact that this never gets deleted is a bit worrying. A program that adds lots of MOs will seem to "leak" memory, even if it loses all references to them. I filed https://github.com/whatwg/dom/issues/720 to check.

I don't think there's any solution for this except either (a) waiting for weak references to be added to JS; (b) implementing a maximum number of them, like we did with concurrentNodeIterators (no longer a publicly tune-able option).

I'm not sure what we should do exactly, but probably we should at least add a note to the README. I'm asking people on Twitter what they think... https://mobile.twitter.com/domenic/status/1069393241131401217

lib/jsdom/living/mutation-observer/MutationRecord-impl.js
```diff
@@ -0,0 +1,35 @@
+"use strict";
+
+const NodeList = require("../generated/NodeList");
+
+// https://dom.spec.whatwg.org/#mutationrecord
+class MutationRecordImpl {
```
domenic: The existence of this class (instead of using a dictionary) makes me angry. Ah well.

lib/jsdom/living/nodes/Node-impl.js
```diff
@@ -677,6 +689,10 @@ class NodeImpl extends EventTargetImpl {
         this._childTextContentChangeSteps();
       }
 
+      if (this.tagName === "SLOT" && this._assignedNodes.length === 0 && isShadowRoot(getRoot(this))) {
```
domenic: Also check localName and namespace, not tagName. I'll add a helper.

domenic:
Alternately, I wonder if there's a version of this that doesn't leak memory. https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/dom/mutation_observer.cc doesn't seem to use any weak pointers. It might be connected to https://github.com/whatwg/dom/issues/713, because I notice Blink has a collection of "Active mutation observers" that gets cleared every time "notify mutation observers" is called.

There is a use of a weak pointer though at https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/dom/mutation_observer.h?l=61&rcl=c057542b56106084df770c534fd76994b9025f9f , so maybe it's all the same thing.

lib/jsdom/living/mutation-observer/MutationObserver-impl.js
pmdartus: I get around this issue by avoiding having to keep a list of all the MOs. At the end of the day, the list is only used to keep the order of the mutation observer.

Instead, now each new MO get assigned a unique incrementing id, and we keep a set of MO that needs to be invoked. We can then invoke the MO in the order based on their id. You can find the details in: https://github.com/jsdom/jsdom/pull/2398/commits/02340384f5d680b4f4e99084d009ab014058b3d3

Issue #1890, opened 2017-06-19
Implement Element.insertAdjacentElement


Treora:
From the [whatwg dom spec](https://dom.spec.whatwg.org/#dom-element-insertadjacentelement). See also on [mdn](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentElement).

valishah:
Hi @Treora : I see it's not been taken care. Is anyone working on it currently?

Treora:
Not that I know of; but I am not involved in this project, just complaining about its imperfections. ;)

valishah:
Oh, Sorry then @Treora. Even I'm looking for the fix of `insertAdjusentElement`. Thinking to make the changes in the code base and open a PR so people will use it.

Treora:
Cool; for reference, you may want to look at (and perhaps share code with) the implementation of [insertAdjacentHTML](https://github.com/jsdom/jsdom/blob/ae4ab59f44918245c8427ec585ecd69a104f7d3b/lib/jsdom/living/nodes/Element-impl.js#L434) (and its discussion in #1355)

valishah:
Sure @Treora: Thanks for the information.

PR #2399, opened 2018-10-21
Implement `insertAdjacentElement()` and `insertAdjacentText()`

Zirro:
This builds on the code in https://github.com/jsdom/jsdom/pull/2395, and implements the `insertAdjacentElement()` and `insertAdjacentText()` methods on `Element`.

It also includes a fix to use ASCII lowercase for the `position` argument in `insertAdjacentHTML()`.

Fixes https://github.com/jsdom/jsdom/issues/1890.

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -451,7 +451,7 @@ class ElementImpl extends NodeImpl {
 
   // https://w3c.github.io/DOM-Parsing/#dom-element-insertadjacenthtml
   insertAdjacentHTML(position, text) {
-    position = position.toLowerCase();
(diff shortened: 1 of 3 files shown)


domenic:
I'd ideally like to add a test for the ASCII lowercase business before merging, since that's a help to the ecosystem. But otherwise things look great.

Zirro:
@domenic It looks like there already is one, but my understanding is that it would only reveal the issue on a system with a Turkish locale:
https://github.com/web-platform-tests/wpt/blob/master/domparsing/insert_adjacent_html.html#L54

domenic:
I think `toLowercase()` is locale-insensitive, so even then that test would not fail. I guess this might not be testable, in which case I'll just go ahead and merge :).

PR #2406, opened 2018-10-28
Fix and run tests on Node.js v11

Zirro:
On Node.js v11, one test began to fail due to a changed error message. Under the assumption that testing for the specific kind of error will be helpful in the future, I've replaced it with a regex that matches both versions.

In addition, one test which had a timing issue - `event_no_duplicates.html` - started passing. Quite likely due to the changes in https://github.com/nodejs/node/pull/22842 which brings Node.js timers closer to browser behaviour. 🎉

I'm not sure if we want to start testing in v11 yet, so I've not changed `.travis.yml`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: node_js
 node_js:
-  - "8"
-  - "10"
+  - 8
+  - 10
+  - stable
 sudo: false
 
 git:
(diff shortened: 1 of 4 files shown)


domenic:
We should test on Node v11 as well (maybe just on Node stable?), since we want to protect against regressions of this sort.

Zirro:
Okay, I've updated the Travis configuration. Since https://github.com/jsdom/jsdom/pull/1964 made one build slot available, we can now have tests covering both LTS versions as well as latest stable.

Zirro:
This is currently failing on a timeout that I'm only able to reproduce by running API tests and WPT tests in sequence.

It seems likely that the cause could be https://github.com/nodejs/node/pull/23870 which has been fixed in the just-released v11.0.1. I'll restart the build in a couple of hours and see if Travis will use that version.

PR #2407, opened 2018-10-31
Disable two recently added <link> tests

Zirro:
Resolves a conflict where the roll in https://github.com/jsdom/jsdom/pull/2380 added two failing <link> tests that didn't get disabled in https://github.com/jsdom/jsdom/pull/2405 since they were merged at the same time.


domenic:
Thank you!!

Issue #2410, opened 2018-10-31
Ignore error if canvas can't be loaded?


molant:
Since https://github.com/jsdom/jsdom/pull/2361 got merged it's easier to know when `canvas` fails to load because of binary issues. This is great because, I've run into this a few times and spent sometime until I find out what was going on.
Nevertheless, there are some scenarios under which a developer might still want to continue the execution even if `canvas` failed to load:

[webhint](https://github.com/webhintio/hint) has a VS Code extension that uses the workspace version, similar to what the ESLint extension does so a user can run it from the CLI or the editor. The problem we have at the moment is that VS Code uses node 8.9.3 which could be different than the one the developer has (we are already at node 11) and thus making jsdom fail to load under VS Code but not the CLI.

I have a couple ideas on how this could be solved:
* Ignore the error, set `exports.Canvas = null;`  and print a `console.error`.
* Add a new parameter to jsdom to change how canvas is loaded and allow the error
* Something else?

wizebin:
In the meantime if someone does want debug their tests but can't through vscode they can always use the `--debug-brk` option from their cli and use the `chrome://inspect` url

Also in vscode you can specify the node runtime with the `runtimeVersion` command
```javascript
{
  "type": "node",
  "request": "launch",
  "name": "Jest Test Everything",
  "program": "${workspaceRoot}/node_modules/jest/bin/jest.js",
  "runtimeVersion": "11.0.0",
  "internalConsoleOptions": "openOnSessionStart"
}
```

PR #2411, opened 2018-11-01
Don't attempt to load canvas-prebuilt

### LINKED ISSUE #2409: jsdom shouldn't try to load canvas-prebuilt
With the release of `canvas` v2.0, `canvas-prebuilt` has been deprecated as `canvas` tries to download the binaries directly. On jsdom 13 support for canvas < 2 was dropped and there will be no more releases of `canvas-prebuilt` on npm. 

Right now jsdom's code searches for both modules if one fails in [`utils.js`](https://github.com/jsdom/jsdom/blob/7e37b5623cc8850d47ec921e965ef9358d3cced2/lib/jsdom/utils.js#L163):

```js
exports.Canvas = null;
["canvas", "canvas-prebuilt"].some(moduleName => {
  try {
    require.resolve(moduleName);
  } catch (e) {
    // ${moduleName} is not installed
    return false;
  }
  exports.Canvas = require(moduleName);
  if (typeof exports.Canvas.createCanvas !== "function") {
    // In browserify, the require will succeed but return an empty object
    exports.Canvas = null;
  }
  return exports.Canvas !== null;
});
```
This could probably be something like (haven't tested it yet):

```js
exports.Canvas = null;

try {
  exports.Canvas = require(moduleName);
  if (typeof exports.Canvas.createCanvas !== "function") {
    // In browserify, the require will succeed but return an empty object
    exports.Canvas = null;
  }
} catch (e) {
  // canvas is not installed
}
```

This is a very minor thing. Happy to do a PR if you want me to.

/CC @zbjornson as he did the work to support canvas 2 in jsdom in https://github.com/jsdom/jsdom/pull/1964

zbjornson:
Fixes #2409

### DIFF
--- a/lib/jsdom/utils.js
+++ b/lib/jsdom/utils.js
@@ -161,17 +161,17 @@ exports.treeOrderSorter = function (a, b) {
 /* eslint-disable global-require */
 
 exports.Canvas = null;
-["canvas", "canvas-prebuilt"].some(moduleName => {
-  try {
-    require.resolve(moduleName);
-  } catch (e) {
-    // ${moduleName} is not installed
-    return false;
-  }
-  exports.Canvas = require(moduleName);
-  if (typeof exports.Canvas.createCanvas !== "function") {
+let canvasInstalled = false;
+try {
+  require.resolve("canvas");
+  canvasInstalled = true;
+} catch (e) {
+  // canvas is not installed
+}
+if (canvasInstalled) {
+  const Canvas = require("canvas");
+  if (typeof Canvas.createCanvas === "function") {
     // In browserify, the require will succeed but return an empty object
-    exports.Canvas = null;
+    exports.Canvas = Canvas;
   }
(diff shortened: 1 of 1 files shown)


domenic:
Thanks, but this deletes a lot of important logic, besides just removing the canvas-prebuilt loop iteration. We should keep that logic.

zbjornson:
How so? `exports.Canvas` becomes `null` or the desired object under the same sets of circumstances as far as I can tell...

zbjornson:
Ah I see, https://github.com/jsdom/jsdom/commit/84987d03dd45f5acf56c1a633e41f07fd077a868#diff-b4d2b12fce9c87d002a9a282c86134ae...

molant:
Which are related to #2410 too.


________________________________

zbjornson:
@domenic I restored the original logic. (Feel free to edit the commit if it's faster/easier than commenting.)

PR #2412, opened 2018-11-01
XML parsing: ingore text outside the root element

lddubeau:
Saxes generates text events for text appearing outside the root element. However, DOM Documents created from XML must not have text nodes outside their root element.

This change allows the wpt ProcessingInstruction-literal-1.xhtml to pass.

The problem was originally reported at
https://github.com/lddubeau/saxes/issues/3


lddubeau:
Of course, I noticed the typo in the commit message just after submitting the PR. :cry: It is fixed now, not only here but in the commit.

PR #2418, opened 2018-11-08
Fixes #1579 JSDOM throws error for nonexistent file uri

### LINKED ISSUE #1579: jsdom throws when an XHR is made for a nonexistent file:// uri
When using jsdom with the file: configuration, any relative URLs will throw if they do not exist.

The EventEmitter created in xhr-utils has no error event handler and so EventEmitter rethrows the error.  

Shouldn't jsdom attempt to continue in this case?  A nonexistent file does not necessarily indicate a failure of the page to load, and this makes for a very fragile experience.

pascalbayer:
Handle nonexistent file uris similar to urls instead of throwing an exception that is not caught anywhere and causes test frameworks to fail.

Details can be found in #1579

### DIFF
--- a/lib/jsdom/living/xhr-utils.js
+++ b/lib/jsdom/living/xhr-utils.js
@@ -160,30 +160,21 @@ function createClient(xhr) {
 
     const client = new EventEmitter();
 
-    let readableStream;
+    const readableStream = fs.createReadStream(filePath, { encoding: null });
 
-    if (fs.existsSync(filePath)) {
-      readableStream = fs.createReadStream(filePath, { encoding: null });
+    readableStream.on("data", chunk => {
+      response.emit("data", chunk);
+      client.emit("data", chunk);
+    });
 
-      readableStream.on("data", chunk => {
-        response.emit("data", chunk);
-        client.emit("data", chunk);
-      });
+    readableStream.on("end", () => {
+      response.emit("end");
+      client.emit("end");
+    });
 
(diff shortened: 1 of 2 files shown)


pascalbayer:
Sample code that caused an uncaught exception before:

```
const xhttp = new XMLHttpRequest();
xhttp.onreadystatechange = () => {};
xhttp.open('GET', 'file:///somefile.png', true);
xhttp.send();
```

Now works similar to a 404 file not found exception for normal URLs.

domenic:
Heya, thanks for taking this on. However, we should not do an exists check, for the race condition reasons [mentioned in the docs](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback). Also, we definitely should not do sync I/O in jsdom.

Instead, we should catch any ENOENT errors from the stream.

I suggest writing some tests first illustrating the problem. They will need to be API tests, instead of web platform tests, since it's about file URLs. But then you can work on a fix that does the right thing.

pascalbayer:
@domenic You're totally right and that makes the fix even easier. I'll create some tests and update the PR.

pascalbayer:
@domenic Let me know if that is what you were looking for.

domenic:
I apologize for my long absence here. I've confirmed that your test is good, and your fix fixes the test. I've also expanded the test coverage with some other interesting things I've learned in this area around XHR and file URLs.

Issue #2350, opened 2018-08-27
Performance regression in version >= 11.6 (5x slower than 11.5)


weijarz:
### Basic info:

- **Node.js version:** 
v8.11.4

- **jsdom version:** 
11.6.x - 12.0.x

### Minimal reproduction case

Checkout this repo:
https://github.com/weijarz/jsdom-perf-bug

```
[jsdom-slow]$ npm i jsdom@11.5
[jsdom-slow]$ node test.js 
parse-0: 896.238ms
parse-1: 641.438ms
parse-2: 539.048ms

[jsdom-slow]$ npm i jsdom@12
[jsdom-slow]$ node test.js 
parse-0: 3739.243ms
parse-1: 3626.432ms
parse-2: 3859.575ms
```

```js
const fs = require('fs')
const jsdom = require('jsdom')

const h = fs.readFileSync(__dirname + '/bug.html', 'utf8')
for (let i = 0; i < 3; i++) {
  console.time(`parse-${i}`)
  let opts = {
    url: 'http://example.com',
    runScripts: 'outside-only'
  }
  dom = new jsdom.JSDOM(h, opts)
  console.timeEnd(`parse-${i}`)
}
```

Zirro:
The document provided in this case includes an enormous inline stylesheet. When it's removed, the remaining markup parses as fast as expected. The regression is probably a consequence of the changes in https://github.com/jsdom/jsdom/pull/2098, or the fixes that followed.

Some early profiling suggests that the slowdown comes from the [`replaceData()`](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/nodes/CharacterData-impl.js#L54)-method in `CharacterData-impl`. It's not that it's slow per se, as it usually finishes in much less than a millisecond. However, in this test case it ends up being executed 81276 times in total, which adds up over time.

Issue #2416, opened 2018-11-05
JSDOM spends a lot of time while parsing <script> tag content, even though JS is disabled


mrtcode:
### Basic info:

- **Node.js version:** 8.10.0
- **jsdom version:** 13.0.0

### Minimal reproduction case:

[An example of an HTML file](https://gist.githubusercontent.com/mrtcode/29ce92eff745561c2baf12be9f629ef7/raw/539ea8e3a9724bd46146aad2304321bd9594b02b/example.html) that has only a `<script>` tag and takes more than half a minute to process it. It's enough to just run JSDOM without any options.
```js
const { JSDOM } = require("jsdom");
const dom = new JSDOM(html);
```

The same is with `<style>` tags. I guess that JSDOM treats `<script>` and `<style>` tags just like regular tags and tries to parse HTML tags inside even though they are special tags and have JS or CSS.

More web pages that have this problem:
https://data.worldbank.org/indicator/EG.USE.COMM.FO.ZS?end=2015&start=1995&view=chart
https://www.newyorker.com/magazine/2011/10/31/foreign-campaigns
https://www.wsj.com/articles/SB10001424052970203471004577144672783559392
https://theintercept.com/

Zirro:
Thanks! This is likely the same issue as https://github.com/jsdom/jsdom/issues/2416, but we can keep this open until that's been verified.

I don't think the issue is that we're parsing the content of `<script>` and `<style>` as HTML, but rather that we're updating the `data` property of the text inside these elements far too frequently during parsing. It can probably be made a lot more efficient while still following the standard.

domenic:
Fixed by https://github.com/jsdom/jsdom/pull/2419, will do a new release soon!

PR #2419, opened 2018-11-11
Improve the performance of parsing text

Zirro:
While testing various solutions for the performance regression described in https://github.com/jsdom/jsdom/issues/2350 and https://github.com/jsdom/jsdom/issues/2416, I came across this optimisation.

As it's parsing, parse5 feeds us a new chunk of text when it encounters a whitespace character in the current text node, which we insert into the existing text through CharacterData's `replaceData()` method with the help of `String.prototype.substring()`.

It turns out that despite being very similar in purpose, using `String.prototype.slice()` repeatedly can be significantly faster than `String.prototype.substring()` in V8 under certain conditions. The difference really adds up in the test cases supplied in https://github.com/jsdom/jsdom/issues/2350 which goes from ~3000ms to ~500ms and https://github.com/jsdom/jsdom/issues/2416 which goes from ~43000ms to ~230ms on my computer.

After the first use, their initial times are very close at an average of 0.004ms. However, the time it takes to use `substring()` again seems to increase slightly with the length of the string, eventually leading to an average of several milliseconds. The timing of `slice()` in the same place on the other hand has even improved at this point and averages 0.001ms.

The impact of this change is particularly noticeable for large inline `<style>` or `<script>` elements, since these are more likely to contain a single uninterrupted text node than other elements.

This resolves https://github.com/jsdom/jsdom/issues/2350 and https://github.com/jsdom/jsdom/issues/2416.

### DIFF
--- a/benchmark/html/index.js
+++ b/benchmark/html/index.js
@@ -1,0 +1,5 @@
+"use strict";
+
+module.exports = {
+  parsing: require("./parsing")
+};
--- a/benchmark/html/parsing.js
+++ b/benchmark/html/parsing.js
@@ -1,0 +1,12 @@
+"use strict";
+const suite = require("../document-suite");
+
+// Measures the time spent parsing and creating a large text node
+// https://github.com/jsdom/jsdom/pull/2419
+
+exports.text = suite(document => {
+  document.body.innerHTML = `
+  <p>
+    ${"Some methods have unexpected implications for performance. ".repeat(5000)}
+  </p>`;
+});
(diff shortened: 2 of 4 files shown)


Zirro:
I added a benchmark which can be used to tell if we regress or improve further on this in the future.

domenic:
Commented at https://bugs.chromium.org/p/v8/issues/detail?id=6730#c4 and will tweet for some extra attention. https://bugs.chromium.org/p/v8/issues/list?can=2&q=slice+substring&colspec=ID+Type+Status+Priority+Owner+Summary+HW+OS+Component+Stars&x=priority&y=owner&cells=ids has some other vaguely-interesting issues.

mathiasbynens:
The fix for this just landed in V8: https://chromium-review.googlesource.com/c/v8/v8/+/1365282

PR #2426, opened 2018-11-18
Add adoption step and fix document owner for HTMLTemplateElement

### LINKED ISSUE #2041: images download automatically on a template
### Basic info:

- **Node.js version:** 8.9
- **jsdom version:** 11.3

### Minimal reproduction case

The same code with a `package.json` is in https://github.com/molant/template-tests

```js
// Server configuration

const templateWithImage = `<template>
<img src="/image.svg" />
</template>`;

const svg = `svg version="1.1"
baseProfile="full"
width="300" height="200"
xmlns="http://www.w3.org/2000/svg">
<rect width="100%" height="100%" fill="red" />
<circle cx="150" cy="100" r="80" fill="green" />
<text x="150" y="125" font-size="60" text-anchor="middle" fill="white">SVG</text>
</svg>`;

const express = require('express');
const app = express();

app.get('/', function (req, res) {
    console.log('Server: HTML requested');
    res.send(templateWithImage);
});

app.get('/image.svg', function (req, res) {
    console.log('Server: image requested');
    res.send(svg);
});

// jsdom API tests

function newApi() {
    const { JSDOM } = jsdom;
    console.log('---------------');
    console.log('Testing new API\n');

    JSDOM.fromURL('http://localhost:3000', {
        retendToBeVisual: true
    });
}

function oldApi() {
    const oldApi = require('jsdom/lib/old-api.js');
    console.log('---------------');
    console.log('Testing old API\n');

    oldApi.env('http://localhost:3000', {
        resourceLoader: function (resource, callback) {
            console.log(`jsdom resource loader: ${resource.url.href}`);
            return resource.defaultFetch(callback);
        },
        features: {
            FetchExternalResources: ["img"],
            SkipExternalResources: false
        },
            newApi();
        }
    });
}

app.listen(3000, oldApi);
```

The current output is:

```
---------------
Testing old API

Server: HTML requested
jsdom resource loader: http://localhost:3000/image.svg
Server: image requested
---------------
Testing new API

Server: HTML requested
```

The problem is that `old-api` automatically downloads the image in the `template` tag. The new API behaves as expected.

Any browser with `template` support behaves as expected (no image downloaded).

pmdartus:
* Implement proper template content owner document semantics
* Add missing adoption hooks for the `template` element.

Fix #2041

### DIFF
--- a/lib/jsdom/browser/parse5-adapter-parsing.js
+++ b/lib/jsdom/browser/parse5-adapter-parsing.js
@@ -1,4 +1,5 @@
 "use strict";
+
 const DocumentType = require("../living/generated/DocumentType");
 const DocumentFragment = require("../living/generated/DocumentFragment");
 const Text = require("../living/generated/Text");
@@ -7,9 +8,46 @@ const attributes = require("../living/attributes");
 const nodeTypes = require("../living/node-type");
 const serializationAdapter = require("../living/domparsing/parse5-adapter-serialization");
 
+const OpenElementStack = require("parse5/lib/parser/open-element-stack");
+
+const OpenElementStackOriginalPop = OpenElementStack.prototype.pop;
+const OpenElementStackOriginalPush = OpenElementStack.prototype.push;
+
 module.exports = class JSDOMParse5Adapter {
   constructor(documentImpl) {
     this._documentImpl = documentImpl;
+
+    // Since the createElement hook doesn't provide the parent element, we keep track of this using _currentElement:
+    // https://github.com/inikulin/parse5/issues/285
+    this._currentElement = undefined;
+
+    // Horrible monkey-patch to implement https://github.com/inikulin/parse5/issues/237
+    const adapter = this;
+    OpenElementStack.prototype.push = function (...args) {
+      OpenElementStackOriginalPush.apply(this, args);
+      adapter._currentElement = this.current;
+
+      const after = this.items[this.stackTop];
+      if (after._pushedOnStackOfOpenElements) {
+        after._pushedOnStackOfOpenElements();
+      }
+    };
+    OpenElementStack.prototype.pop = function (...args) {
+      const before = this.items[this.stackTop];
+
+      OpenElementStackOriginalPop.apply(this, args);
+      adapter._currentElement = this.current;
+
+      if (before._poppedOffStackOfOpenElements) {
+        before._poppedOffStackOfOpenElements();
+      }
+    };
+  }
+
+  _ownerDocument() {
+    // The _currentElement is undefined when parsing elements at the root of the document. In this case we would
+    // fallback to the global _documentImpl.
+    return this._currentElement ? this._currentElement._ownerDocument : this._documentImpl;
   }
 
   createDocument() {
@@ -21,11 +59,13 @@ module.exports = class JSDOMParse5Adapter {
   }
 
   createDocumentFragment() {
-    return DocumentFragment.createImpl([], { ownerDocument: this._documentImpl });
+    return DocumentFragment.createImpl([], { ownerDocument: this._currentElement._ownerDocument });
   }
 
   createElement(localName, namespace, attrs) {
-    const element = this._documentImpl._createElementWithCorrectElementInterface(localName, namespace);
+    const ownerDocument = this._ownerDocument();
+
+    const element = ownerDocument._createElementWithCorrectElementInterface(localName, namespace);
     element._namespaceURI = namespace;
     this.adoptAttributes(element, attrs);
 
@@ -37,7 +77,8 @@ module.exports = class JSDOMParse5Adapter {
   }
 
   createCommentNode(data) {
-    return Comment.createImpl([], { data, ownerDocument: this._documentImpl });
+    const ownerDocument = this._ownerDocument();
+    return Comment.createImpl([], { data, ownerDocument });
   }
 
   appendChild(parentNode, newNode) {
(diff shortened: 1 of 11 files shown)


pmdartus:
The failing test exhibit an interesting issue. The test tries to access the `ownerDocument.defaultView` property from the fragment created via `JSDOM.fragment` API. This PR currently break the behavior since the owner document of the fragment doesn't have a browsing context anymore.

I was originally thinking about updating the `_ownerDocument` of the document fragment with the `sharedFragmentDocument` and adopting all the nodes before returning it to avoid introducing a breaking change. 
On the other, as the documentation describes the `fragment` API to be used _for the very simplest of cases_ I also like the idea of not having access to the browsing context.

Thoughts?

lib/jsdom/browser/parse5-adapter-parsing.js
```diff
@@ -49,6 +90,14 @@ module.exports = class JSDOMParse5Adapter {
   }
 
   setTemplateContent(templateElement, contentFragment) {
+    // Internally during the construction of the HTMLTemplateElement, a DocumentFragment is also created. At this point
+    // we have 2 fragments the one created in the HTMLTemplateElement constructor and the one created by
+    // createDocumentFragment. We now have to reconciliate the one created internally with one created by the tree
+    // adapter.
```
domenic: It'd be good to expand this comment to explain why we create one in the HTMLTemplateElement constructor. Would it be accurate to change it to something like "...a DocumentFragment is also created. (This is important for cases like `document.createElement("template")`, where `setTemplateContent` never runs.)

Also for "the one created by createDocumentFragment", do you mean the one created by parse5?

I think it's important to make this comment as explicit as possible as it's kind of a weird situation.

lib/jsdom/living/helpers/shadow-dom.js
```diff
@@ -210,6 +210,30 @@ function findSlot(slotable, openFlag) {
   return null;
 }
 
+// https://dom.spec.whatwg.org/#concept-shadow-including-descendant
+function* shadowDomIncludingInclusiveDescendantIterator(node) {
```
domenic: Can we name these "shadowIncludingInclusiveDescendantsIterator" / "shadowIncludingDescendantsIterator"? That seems to fit the spec better.

lib/jsdom/living/nodes/HTMLTemplateElement-impl.js
```diff
@@ -2,13 +2,54 @@
...
+      }
+
+      doc = doc[ASSOCIATED_INERT_TEMPLATE_DOCUMENT];
+    }
+
+    return doc;
+  }
+
+  // https://html.spec.whatwg.org/multipage/scripting.html#template-adopting-steps
+  _adopted() {
```
domenic: Can we call this _adoptingSteps?

lib/jsdom/living/nodes/HTMLTemplateElement-impl.js
```diff
@@ -2,13 +2,54 @@
...
+const DocumentFragment = require("../generated/DocumentFragment");
+
 const { cloningSteps, domSymbolTree } = require("../helpers/internal-constants");
 const { clone } = require("../node");
 
+/* Symbol assigned to a document to mark it as inert */
+const INERT_TEMPLATE_DOCUMENT = Symbol("inhert-template-document");
+
+/* Symbol assigned to a document to store the associated inert template document. */
+const ASSOCIATED_INERT_TEMPLATE_DOCUMENT = Symbol("associated-inert-template-document");
```
domenic: Why use symbols? I'd rather just do something like `doc._isInertTemplateDocument` and `doc._associatedInertTemplateDocument`.

Also don't use multiline comments for singleline comments.

pmdartus: Since we are attaching properties on the document that are not defined in the constructor, my original thought was to create `Symbol` to prevent property clashing. Will make the change.

lib/jsdom/browser/parse5-adapter-parsing.js
pmdartus: > Also for "the one created by createDocumentFragment", do you mean the one created by parse5?

Yes. Will update the comment.

test/api/fragment.js
```diff
@@ -18,11 +18,10 @@ describe("API: JSDOM.fragment()", () => {
...
   });
 
-  it("should return fragments with shared Windows each time", () => {
-    const frag1 = JSDOM.fragment(``);
-    const frag2 = JSDOM.fragment(``);
+  it("should return a fragment with no associated browsing context", () => {
+    const frag = JSDOM.fragment(``);
 
-    assert.strictEqual(frag1.ownerDocument.defaultView, frag2.ownerDocument.defaultView);
+    assert.isNull(frag.ownerDocument.defaultView);
```
pmdartus: >  E.g. XHR shouldn't work, if we've implemented it right.

@domenic I don't know exactly how you would like to test XHR since the globals are not accessible via the `defaultView`.

test/api/fragment.js
```diff
@@ -45,7 +44,7 @@ describe("API: JSDOM.fragment()", () => {
     assert.strictEqual(frag.firstChild.textContent, "Hi");
   });
 
-  it("should respect ignore any options passed in", () => {
+  it("should ignore any options passed in", () => {
```
pmdartus: > Also while you're there if you could fix the test title from "should respect ignore any options passed in" to "should respect the relevant options passed in", that'd be ideal.

Changed the name to ignore instead of respect since this API doesn't allow any options.

test/api/fragment.js
domenic: Oh, I actually looked at the diff, and I see that of course, if `defaultView` is null, then you wouldn't be able to get an XHR.

pmdartus: > The defaultView IDL attribute of the Document interface, on getting, must return this Document's browsing context's WindowProxy object, if this Document has an associated browsing context, or null otherwise.

https://html.spec.whatwg.org/#dom-document-defaultview

pmdartus: Based on this, it think it worth adding this change for the next major version, since it's a real breaking change. 

I did a search on github to see the usage of the `JSDOM.fragment` API, non of the code that I have seen try to access the `defaultView` out of the fragment. Better be safe than sorry.

PR #2428, opened 2018-11-21
Removes pn dependency

dresende:
Since `pn` has deprecation warning and `jsdom` uses to little from it to really need the module, we're better off just removing it.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -1,6 +1,6 @@
 "use strict";
 const path = require("path");
-const fs = require("pn/fs");
+const fs = require("fs");
 const vm = require("vm");
 const toughCookie = require("tough-cookie");
 const sniffHTMLEncoding = require("html-encoding-sniffer");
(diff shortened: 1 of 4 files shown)


dresende:
I have no idea why the 2 failed tests, they just end with:

```sh
The command "yarn test --retries 1" exited with 0.

Done. Your build exited with 1.
```

dresende:
Nevermind, linting.

domenic:
I can't find any deprecation warning; could you explain where you saw one?

dresende:
screenshot 2018-12-03 at 10 37 59

dresende:
I was having these warning lines using `jsdom` and found that it was because of `pn`. Opened a pull request there but then looked at `jsdom` and saw that it uses too little of this module so I'm proposing not to use this dependency at all.

domenic:
Ah, well, we're not using `require("pn/timers").enroll`, so this isn't applicable to us. In fact, I believe `require("timers").enroll` is what's deprecated, and using it via pn's wrapper or not is unrelated.

PR #2436, opened 2018-11-27
Support SVGTitleElement, fix document.title and <title>'s text property

Zirro:
This adds `SVGTitleElement` and support for using it in `document.title` according to the spec at https://html.spec.whatwg.org/multipage/dom.html#document.title.

As part of this I've renamed and added an optional namespace argument for the traversal helpers to make matching elements outside the HTML namespace easier.

I've also fixed the getter of `<title>`s `text` property to correctly use the child text content instead of innerHTML: https://html.spec.whatwg.org/multipage/semantics.html#dom-title-text

This currently includes the changes from https://github.com/jsdom/jsdom/pull/2434 since one of the new tests, `document.title-not-in-html-svg.html`, is fixed through this.

### DIFF
--- a/lib/jsdom/living/helpers/traversal.js
+++ b/lib/jsdom/living/helpers/traversal.js
@@ -4,9 +4,9 @@ const { HTML_NS } = require("./namespaces");
 
 // All these operate on and return impls, not wrappers!
 
-exports.closest = (e, localName) => {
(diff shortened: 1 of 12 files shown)


domenic:
Nice work, and especially nice factoring into separate logical commits.

Issue #2437, opened 2018-11-27
Inline style of the element is not returned anymore with `.style` 


minas90:
### Basic info:

- **Node.js version:** 10.13.0
- **jsdom version:** 13.0.0

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");
const dom = new JSDOM('<a style="display:none;color:red;" href="https://google.com">google.com</a>');

console.log(document.querySelector('a').style)
```

Hi there,
It's no longer possible to get style of the element with `.style`, as a workaround `el.getAttribute('style')` works properly, but `.style` breaks our code in many places.

domenic:
How does this code work in browsers? Can you produce an example that tests something, instead of just logging something?

minas90:
Hi @domenic,
In browsers it returns CSSStyleDeclaration.

For example here is a basic usage:
```
let a = document.querySelector('a');
if (document.querySelector('a').style.display == 'none') {
    a.style.display = 'block'
}
```
It's checking if it's a hidden tag via display property and makes it visible by changing it's value to `block `

minas90:
It works properly on version 12.2.0, it's broken only on 13.0.0

minas90:
Yes, the log will not hit, downgrading to 12.2.0 fixes that.
I'm also not sure what changed, but it broke our code in many places, we are removing hidden elements by checking if `a.style.display === 'none'`

Zirro:
This is due to the `_attrModified` hook added in https://github.com/jsdom/jsdom/commit/064527e2cbc003debe0bc826f893519f1ceb64bc, where I missed the need to include `super._attrModified`. I'll create a fix for this later today.

PR #2438, opened 2018-11-27
Fix style property on <a> and <area> elements

Zirro:
Fixes a regression introduced through https://github.com/jsdom/jsdom/commit/064527e2cbc003debe0bc826f893519f1ceb64bc where the `style` property on `<a>` and `<area>` elements would not get updated correctly. Given how specific the issue is, I'm not sure if adding a test would make sense here.

Resolves https://github.com/jsdom/jsdom/issues/2437.


domenic:
This problem of forgetting to call `super.___` is annoying. The best solution I can think of is to write a custom lint rule. I'll open a new issue about that...

PR #2441, opened 2018-11-30
Run child text content change steps when inserting CDATASection node

Zirro:
While the specification only references Text nodes [in this step](https://dom.spec.whatwg.org/#ref-for-concept-node-text-change-ext), it seems like the inclusion of CDATASection nodes is implied unless the term [exclusive Text node](https://dom.spec.whatwg.org/#exclusive-text-node) is used.

If that's the case there are several places in the code where we should be matching against both Text nodes and CDATASection nodes instead of just the former, though only this location seems to be covered by our current tests. I'll create an issue for the remaining cases if this is correct.

### DIFF
--- a/lib/jsdom/living/nodes/Node-impl.js
+++ b/lib/jsdom/living/nodes/Node-impl.js
@@ -673,7 +673,8 @@ class NodeImpl extends EventTargetImpl {
 
       this._modified();
 
-      if (node.nodeType === NODE_TYPE.TEXT_NODE) {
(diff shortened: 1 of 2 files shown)


domenic:
While I think this is correct (and the tests bear it out), I think the spec's use of "`Text` node", instead of e.g. "text node" or (better) "inclusive text node", is confusing. I will file a spec bug for that, and perhaps you can chime in with all the relevant locations.

PR #2443, opened 2018-12-03
Update webidl2js

domenic:
This should only impact making [Symbol.unscopables] more accurate for Node; see https://github.com/jsdom/webidl2js/pull/114.

---

Locally when I run this I get failures on `web-platform-tests dom/nodes DOMImplementation-createHTMLDocument.html`, which is strange. Hopefully just a quirk of my local configuration and not actually a result of the upgrade... Let's see what CI says.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "st": "^1.2.2",
     "watchify": "^3.11.0",
     "wd": "^1.11.1",
-    "webidl2js": "^9.1.2"
+    "webidl2js": "^9.2.0"
   },
   "browser": {
     "canvas": false,
--- a/test/web-platform-tests/to-run.yaml
+++ b/test/web-platform-tests/to-run.yaml
@@ -143,7 +143,6 @@ Node-isEqualNode-xhtml.xhtml: [timeout, Unknown]
 ParentNode-querySelector-All-xht.xht: [timeout, Unknown]
 ProcessingInstruction-escapes-1.xhtml: [fail, Unknown]
 Text-constructor.html: [fail, Unknown]
-remove-unscopable.html: [fail, Unknown]
 
 ---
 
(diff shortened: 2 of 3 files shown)


domenic:
Uh oh, now I am worried that webidl2js has made a breaking change somehow. Maybe via the webidl2 dependency.

domenic:
I guess I'll unpublish webidl2js until someone can get to the bottom of this!?! It does seem to be because of the webidl2 upgrade; reverting webidl2js but not webidl2 still has the failing test.

domenic:
@saschanaz if you have any way to tell what changed in webidl2 between versions 9.1.0 and 10.3.1 that could have affected how createHTMLDocument gets processed, we'd love your help. (Indeed, if you have time to update webidl2js to the latest webidl2 release, that would also be amazing. It would probably require jsdom changes to stop using old mixin syntax.)

domenic:
After a tragedy of errors webidl2js v9.1.2 reverts us to the same state as v9.0.1, so consider this on hold. At least we haven't broken any consumers though.

saschanaz:
>if you have any way to tell what changed in webidl2 between versions 9.1.0 and 10.3.1

The only breaking change should be the removal of old iterator syntax support, per the changelog: https://github.com/w3c/webidl2.js/blob/develop/CHANGELOG.md#v1000-2017-12-20

domenic:
Yeah, I checked the changelog, but that doesn't appear to be the case. Something changed around the handling of

```
  [NewObject] Document createHTMLDocument(optional DOMString title);
```

it seems, or maybe in the handling of

```
  attribute [TreatNullAs=EmptyString] DOMString data;
```

saschanaz:
Hmm. [I tried diffing via runkit](https://runkit.com/embed/j0vs98akwuik) but couldn't find any meaningful change. The only changes are some added fields.

saschanaz:
Tested on my own branch and found that the last known good version is 10.3.0.

saschanaz:
The only difference between 10.3.0 and 10.3.1 is this commit, not sure how this can break the test. https://github.com/w3c/webidl2.js/commit/96deeb348054c562d817a8f5068ad914c83d450a

saschanaz:
Okay, I got the problem. https://github.com/w3c/webidl2.js/commit/96deeb348054c562d817a8f5068ad914c83d450a incorrectly caused an empty `.extAttrs` incorrectly be shared across types and this caused unexpected sharing of `[TreatNullAs]`. This is fixed in later versions, but we should workaround this for old versions.

Issue #2367, opened 2018-09-22
Click event on label child does not toggle label checkbox checked value


eddyerburgh:
### Basic info:

- **Node.js version:** 8.12.0
- **jsdom version:** 12.0.0

I've read through the WhatWG spec, and I think this is true:

Dispatching a MoustEvent of type `click` on a child element of a label should cause the first labeled control to toggle checkedness. (If clicking a label normally causes this behavior)

This is implemented inconsistently between browsers. Chrome 69, Safari 11.1.2, IE 11 toggle the inputs checked value, Firerfox 62 does not.
 
 From reading the specs, this is what I understand:
 
- `click()` dispatches a MouseEvent with type `click` on the target element (div)
- div does not have activation behavior, so activationTarget is null
- parent node (label) has activation behavior, activationTarget is set to parent (label)
- Run activation behavior of activationTarget

### Minimal reproduction case

```js
const { JSDOM } = require('jsdom');

var dom = new JSDOM('\
  <!DOCTYPE html>\
  <html>\
    <body>\
      <label><div /><input type="checkbox"></label>\
    </body>\
  </html>\
');

var div = dom.window.document.querySelector('div');
var input = dom.window.document.querySelector('input');
var label = dom.window.document.querySelector('label')

label.click()

console.log('expecting true, actual', input.checked); // actual = true

div.click();

console.log('expecting false, actual', input.checked); // actual = true
```

### How does similar code behave in browsers?

Input checkedness is toggled in most browsers (https://jsfiddle.net/4x0tjLms/5/).

Input checkedness is not toggled in Firerfox 62, but Firefox *does not* toggle checkedness when `click` is called on a label with labeled control. JSDOM *does* toggle checkedness when `click` is called on label with a labeled control.

vegarringdal:
I have another case, can use same fiddle as above

```js
const event = new window.CustomEvent('click');
input.dispatchEvent(event);
```

chrome does not toggle checked value, while jsdom does
I dont not knows who is right, but this have tricked me a few times :-)

Zirro:
Thanks for the reports! These issues are fixed by #2450.

PR #2450, opened 2018-12-07
Run activation behavior hooks from the event dispatch algorithm 

### LINKED ISSUE #2449: clicking span inside a submit button does not submit the form
### Basic info:

- **Node.js version:** >=8
- **jsdom version:** 13

### Minimal reproduction case

[Repl.it](https://repl.it/@kentcdodds/jsdom-submit-bug)

```js
const {JSDOM} = require('jsdom')


const {window, window: {MouseEvent, document}} = new JSDOM(`
<!doctype html>
<html>
  <body>
    <form>
      <button type="submit">
        <span>submit</span>
      </button>
    </form>
  </body>
</html>
`)

const form = document.querySelector('form')
const submit = document.querySelector('span')

let calls = 0

form.addEventListener('submit', event => {
  event.preventDefault()
  console.log('submitted')
  calls++
})

const clickEvent = new MouseEvent('click', {bubbles: true, cancelable: true, button: 0})
submit.dispatchEvent(clickEvent)

console.assert(calls === 1, 'the form was not submitted')
```

[jsbin](https://jsbin.com/recekor/edit?html,js,console)

In the browser, firing a click event on the span within a submit button will call the form's submit handler. In jsdom it does not.

Note that if I select the button instead of the span and fire the click event on the button then the form is submitted just fine. It's just when I select something within the button.

Real world situation: https://github.com/kentcdodds/react-testing-library/issues/234

Thanks!

Zirro:
This adds steps regarding activation behavior from the [Dispatch an event](https://dom.spec.whatwg.org/#concept-event-dispatch) algorithm, replacing older code where the activation didn't bubble correctly.
I've also renamed the existing hooks to match the names used in the specification.

This fixes #2449 and #2367.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLElement-impl.js
@@ -18,28 +18,6 @@ class HTMLElementImpl extends ElementImpl {
     this._clickInProgress = false;
   }
 
-  // Add default event behavior (click link to navigate, click button to submit
-  // form, etc). We start by wrapping dispatchEvent so we can forward events to
-  // the element's default functions (only events that did not incur
-  // preventDefault).
-  dispatchEvent(event) {
-    if (event.type === "click") {
-      callEventBehaviorHook(event, "_preClickActivationSteps", this);
-    }
-
-    const outcome = super.dispatchEvent(event);
-
-    if (event.type === "click") {
-      if (event.defaultPrevented) {
-        callEventBehaviorHook(event, "_canceledActivationSteps");
-      } else {
-        callEventBehaviorHook(event, "_activationBehavior");
-      }
-    }
-
-    return outcome;
-  }
-
   click() {
     // https://html.spec.whatwg.org/multipage/interaction.html#dom-click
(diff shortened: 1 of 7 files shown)


lib/jsdom/living/nodes/HTMLElement-impl.js
```diff
@@ -18,25 +18,8 @@ class HTMLElementImpl extends ElementImpl {
     this._clickInProgress = false;
   }
 
-  // Add default event behavior (click link to navigate, click button to submit
-  // form, etc). We start by wrapping dispatchEvent so we can forward events to
-  // the element's default functions (only events that did not incur
-  // preventDefault).
   dispatchEvent(event) {
```
TimothyGu: Can we remove this method now that it purely forwards?

Zirro: Yes, thanks for the suggestion.

Zirro:
Updated to add `_hasActivationBehavior` instead of directly checking for the existence of the method, as it occurred to me that it'll need to be conditional for `<summary>` which uses `HTMLElement`.

domenic:
Nice work. Let's wait to merge this until a WPT roll though, just out of principle :).

PR #2451, opened 2018-12-09
Remove unused feature methods

Zirro:
These have not been in use since https://github.com/jsdom/jsdom/pull/2279.


domenic:
Oh wow, so funny that I didn't delete these, when I was looking forward to removing them so much.

PR #2454, opened 2018-12-11
Bug fixed, charset is undefined when call FileReader.readAsDataURL for type text blob.

ytetsuro:
Bug fixed, charset is undefined when call FileReader.readAsDataURL for type text blob.

I write code following.

```javascript
const blob = new jsdom.window.Blob(['Hello'], {
   type: 'text/plain;charset=utf-8',
});

const fileReader = new jsdom.window.FileReader();
fileReader.addEventListener('load', () => {
   // false, result is 'data:text/plain;charset=undefined,Hello'
   console.log(fileReader.result === 'data:text/plain;charset=utf-8,Hello'); 
});

fileReader.readAsDataURL(blob);
```

I think result is setting blob type charset.
But, result charset is undefined.

### DIFF
--- a/lib/jsdom/living/file-api/FileReader-impl.js
+++ b/lib/jsdom/living/file-api/FileReader-impl.js
@@ -103,15 +103,9 @@ class FileReaderImpl extends EventTargetImpl {
             break;
           }
           case "dataURL": {
-            // Spec seems very unclear here; see https://github.com/whatwg/fetch/issues/665#issuecomment-362930079.
-            let dataUrl = "data:";
-            const contentType = MIMEType.parse(file.type);
-            if (contentType) {
-              dataUrl += contentType.toString();
-            }
-            dataUrl += ";base64,";
-            dataUrl += data.toString("base64");
-            this.result = dataUrl;
+            // Spec seems very unclear here; see https://github.com/w3c/FileAPI/issues/104.
+            const contentType = MIMEType.parse(file.type) || "application/octet-stream";
+            this.result = `data:${contentType};base64,${data.toString("base64")}`;
             break;
           }
           case "text": {
--- a/lib/jsdom/living/file-api/FileReader-impl.js
+++ b/lib/jsdom/living/file-api/FileReader-impl.js
@@ -2,7 +2,6 @@
 
 const whatwgEncoding = require("whatwg-encoding");
 const MIMEType = require("whatwg-mimetype");
-const querystring = require("querystring");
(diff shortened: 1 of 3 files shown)


domenic:
Thanks, however, when I run your test in a browser per https://jsbin.com/wotiyijete/edit?html,console,output , I get the `console.log` outputting `false`, because in browsers `result` is always `undefined`. So, I don't think this is the right fix, since it does not match browsers.

ytetsuro:
> I don't think this is the right fix, since it does not match browsers.

Sorry.
The sample was wrong.

`result` was a property of FileReader, not an argument.

domenic:
Thanks, but that still gives `false` when I run it in my browser.

ytetsuro:
> Thanks, but that still gives false when I run it in my browser.

Humm..

I want write test code.
May i make test code in following directory?

https://github.com/jsdom/jsdom/tree/master/test/api

domenic:
No; since this is not an API, but instead a web platform feature, you need to use https://github.com/jsdom/jsdom/tree/master/test/web-platform-tests/to-upstream

ytetsuro:
@domenic 

Thanks.

ytetsuro:
I writed test.
My code is mistaken.
I rewrite code.
Please review.

I expected result is following.

```javascript
const blob = new jsdom.window.Blob(['Hello'], {
   type: 'text/plain;charset=utf-8',
});

const fileReader = new jsdom.window.FileReader();
fileReader.addEventListener('load', () => {
   // false, result is 'data:text/plain;charset=undefined,Hello'
   console.log(fileReader.result === 'data:text/plain;charset=utf-8;base64,SGVsbG8='); 
});

fileReader.readAsDataURL(blob);
```

ytetsuro:
@domenic

I'm sorry to bother.
Could you review?

domenic:
Hi @ytetsuro, sorry for the long delays and thank you for your patience. This looks good. I pushed an extra commit with some cleanups, which also enables some of the existing tests. That revealed that the fallback for unparseable MIME types should probably be `application/octet-stream`; at least, that matches browsers better, and the web platform tests. (See https://wpt.fyi/results/FileAPI/reading-data-section/filereader_readAsDataURL.html?label=master&label=experimental&aligned&q=readasdata.)

I'll merge this after the CI runs, and hopefully there will be a release soon.

ytetsuro:
@domenic Thanks merged!

PR #2458, opened 2018-12-17
Document how runScripts installs globals

### LINKED ISSUE #2446: window.Date is undefined
### Basic info:

- **Node.js version:** > 9
- **jsdom version:** 13

### Minimal reproduction case

```js
const { JSDOM } = jsdom;

const dom = new JSDOM(`<!DOCTYPE html><p>Hello world</p>`);

console.log(!!dom.window.Date); // logs false
```

https://repl.it/repls/WellinformedSmartFiles

I'd expect `Date` to exist in `dom.window`. We use `jsdom` for testing and in our environment setup, we have:
```js
global.window = dom.window
```
That makes tools like [this one](https://github.com/TheBrainFamily/wait-for-expect/blob/master/src/index.ts#L13) fail.

The workaround is easy:
```js
dom.window.Date = Date;
```

But perhaps `jsdom` should include that by default.

domenic:
Closes #2446.

/cc @pladaria and @Xeppy. Would this documentation have helped you avoid the issue you ran into?

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ const { document } = (new JSDOM(`...`)).window;
 
 Full documentation on everything you can do with the `JSDOM` class is below, in the section "`JSDOM` Object API".
 
+_Important note: in the default configuration, JavaScript globals like `window.Date` or `window.Map` will not exist. Read the "Executing scripts" section below for more._
+
 ## Customizing jsdom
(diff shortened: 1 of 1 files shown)


pladaria:
Sure! Thanks!

Issue #2179, opened 2018-04-01
Consider adding helpers for firing/creating/etc. events


domenic:
In particular, per https://dom.spec.whatwg.org/#constructing-events and https://dom.spec.whatwg.org/#concept-event-fire.

This would help us avoid forgetting to set isTrusted, for example.

PR #2463, opened 2018-12-21
Add helpers for creating and firing events

Zirro:
This resulted in a pretty large clean-up of the old event code. Fixes https://github.com/jsdom/jsdom/issues/2179.

### DIFF
--- a/lib/jsdom/living/helpers/events.js
+++ b/lib/jsdom/living/helpers/events.js
@@ -1,0 +1,23 @@
+"use strict";
+
+const Event = require("../generated/Event");
+const { tryImplForWrapper } = require("../generated/utils");
+
+function createAnEvent(e, eventInterface = Event, attributes = {}) {
+  return eventInterface.createImpl(
+    [e, attributes],
+    { isTrusted: attributes.isTrusted !== false }
+  );
+}
+
+function fireAnEvent(e, target, eventInterface, attributes, legacyTargetOverrideFlag) {
+  const event = createAnEvent(e, eventInterface, attributes);
+
+  // tryImplForWrapper() is currently required due to use in Window.js and xmlhttprequest.js
+  return tryImplForWrapper(target)._dispatch(event, legacyTargetOverrideFlag);
+}
+
+module.exports = {
+  createAnEvent,
+  fireAnEvent
+};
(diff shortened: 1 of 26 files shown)


lib/jsdom/living/helpers/events.js
```diff
@@ -0,0 +1,22 @@
...
+  return eventConstructor.createImpl(
+    [e, attributes],
+    { isTrusted: attributes.isTrusted !== false }
+  );
+}
+
+function fireAnEvent(e, target, eventConstructor = Event, attributes = {}, legacyTargetOverrideFlag) {
+  const event = createAnEvent(e, eventConstructor, attributes);
+
+  return tryImplForWrapper(target)._dispatch(event, legacyTargetOverrideFlag);
```
Sebmaster: Do we really need to implForWrapper this? Seems like this should only be called for impls anyways?

Zirro: It seems to be required for when the method is used in `Window.js` and `xmlhttprequest.js`, without additional changes to those files at least.

lib/jsdom/living/helpers/events.js
```diff
@@ -0,0 +1,22 @@
+"use strict";
+
+const Event = require("../generated/Event");
+const { tryImplForWrapper } = require("../generated/utils");
+
+function createAnEvent(e, eventConstructor = Event, attributes = {}) {
```
domenic: It's not really eventConstructor, right? It's more like eventModule?

lib/jsdom/living/helpers/events.js
domenic: Let's add a comment to explain that it's for those cases; eventually we will be able to clean lots of such tryImplForWrapper() cases up.

lib/jsdom/living/nodes/HTMLElement-impl.js
```diff
@@ -46,21 +47,17 @@ class HTMLElementImpl extends ElementImpl {
...
-        cancelable: true,
-        composed: true,
-        view: this.ownerDocument.defaultView
-      }
-    ], {});
-
     // Run synthetic click activation steps. According to the spec,
     // this should not be calling dispatchEvent, but it matches browser behavior.
     // See: https://www.w3.org/Bugs/Public/show_bug.cgi?id=12230
     // See also: https://github.com/whatwg/html/issues/805
```
domenic: The spec has since been changed to handle this, so this comment is no longer accurate. Probably best to do any changes here in a followup though.

(For example, this seems buggy in that disabled elements will have _clickInProgress forever! We should write a test to show that bug and then fix it. Also the comment in click()'s header says we have no implementation of isTrusted, heh.)

lib/jsdom/living/nodes/HTMLMediaElement-impl.js
```diff
@@ -49,9 +50,7 @@ class HTMLMediaElementImpl extends HTMLElementImpl {
   }
 
   _dispatchRateChange() {
```
domenic: It may be simpler to just get rid of these functions now.

PR #2464, opened 2018-12-21
Add support for links to javascript: urls and fragments

Zirro:
A work in progress. The code related to choosing a browsing context and `noreferrer`/`noopener` may be removed as it does not have an effect on the resulting window or document at this point.


domenic:
Should we merge this? Seems not so bad :)

Zirro:
I'm a bit concerned about our test coverage here, but otherwise merging it now is okay with me.

PR #2469, opened 2018-12-24
Use UTF-8 as the default encoding for XML documents

Zirro:
XML has separate rules for detecting the encoding of documents where it is not explicitly declared through markup or headers, which tend to result in UTF-8 rather than windows-1252 as for HTML. Further work is most likely needed to be accurate here, but this should bring us closer to what browsers do and lets us enable two tests for documents using non-ASCII characters.


lib/api.js
```diff
@@ -227,8 +228,6 @@ function transformOptions(options, encoding) {
   };
 
   if (options.contentType !== undefined) {
```
domenic: I think we can just remove this guard now, and ignore options.contentType for the purposes of this function. I'll try it.

PR #2486, opened 2019-01-11
Implement navigator.plugins and navigator.mimeTypes

Zirro:
These are largely no-ops since we don't actually support any plugins.

There are tests for these in https://github.com/web-platform-tests/wpt/blob/master/html/webappapis/system-state-and-capabilities/the-navigator-object/navigator-pluginarray.html.

### DIFF
--- a/lib/jsdom/living/navigator/MimeType-impl.js
+++ b/lib/jsdom/living/navigator/MimeType-impl.js
@@ -1,0 +1,3 @@
+"use strict";
+
+exports.implementation = class MimeType {};
--- a/lib/jsdom/living/navigator/PluginArray-impl.js
+++ b/lib/jsdom/living/navigator/PluginArray-impl.js
@@ -1,0 +1,27 @@
+"use strict";
+
+const idlUtils = require("../generated/utils");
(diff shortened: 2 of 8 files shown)


lib/jsdom/living/navigator/PluginArray-impl.js
```diff
@@ -0,0 +1,27 @@
...
+  item() {
+    return null;
+  }
+
+  namedItem() {
+    return null;
+  }
+
+  get [idlUtils.supportedPropertyIndices]() {
+    return new Set();
```
TimothyGu: I believe you could return an empty array (which might be slightly cheaper), but the difference is insignificant anyway.

lib/jsdom/living/navigator/MimeType-impl.js
```diff
@@ -0,0 +1,3 @@
+"use strict";
+
+exports.implementation = class MimeType {};
```
domenic: The reason this works is because it's not actually possible to get an instance of one of these, I guess? Fascinating.

PR #2488, opened 2019-01-11
Request's timeout

### LINKED ISSUE #596: request.setTimeout
Hi!

It will be great if you will implement some timeout support for jsdom requests.
Thank you for package!

dmitriyGarden:
Resolve https://github.com/jsdom/jsdom/issues/596


domenic:
Closing as the resource loader has been redesigned. I think a timeout should be pretty easy with the interceptors these days, but we can leave #596 open in case it isn't.

PR #2490, opened 2019-01-16
Roll Web Platform Tests

Zirro:


domenic:
The hosts listing is getting a bit ridiculous. Let me try some things and see if they work.

domenic:
So I think I made the .travis.yml better, and then I updated it a bit, but now we're failing on one of the new expected-fail tests, due to some interesting problem in our cleanup code...

Zirro:
I noticed that the failure only occurs with the most recent commits applied, and eventually tracked it down these lines: https://github.com/jsdom/jsdom/commit/75a921eab52c239b2468e08be9f547f46c7f86bd#diff-f9316c180d590a33d166424a6b98059fL11

The error disappears if the changes in that file are reverted. It doesn't necessarily mean that those changes are to blame of course, but might instead be a side-effect of incorrect document-manipulation elsewhere (such as using `<template>` internally where we shouldn't).

If so, it seems like https://github.com/jsdom/jsdom/pull/2522 might provide the solution. I've run out of time and can't test further right now, but figured I'd mention what I've found in case someone else wants to take a look.

PR #2502, opened 2019-01-29
Fix bug when appending initially empty fragment to dom node

tjenkinson:
I discovered a bug, which is triggered by the test I added (f13cb62)

The next commit fixes it. I am not sure if it is the right fix. The real problem might be a bit deeper, but I am not very familiar with jsdom.

The test output without the fix is

```
1) API: JSDOM.fragment() should be possible to append to a dom node:

      AssertionError: expected '<div>undefined<div>test</div></div>' to equal '<div><div>test</div></div>'
      + expected - actual

      -<div>undefined<div>test</div></div>
      +<div><div>test</div></div>

      at it (test/api/fragment.js:71:12)
      at $mochaNoSugar$runUserTestFunc (node_modules/mocha-sugar-free/lib/mocha-sugar-free.js:64:35)
      at Context.$mochaNoSugar (node_modules/mocha-sugar-free/lib/mocha-sugar-free.js:267:32)
```

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -100,7 +100,7 @@ class JSDOM {
     }
   }
 
-  static fragment(string) {
+  static fragment(string = "") {
     if (!sharedFragmentDocument) {
       sharedFragmentDocument = (new JSDOM()).window.document;
     }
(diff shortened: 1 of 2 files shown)


domenic:
The general idea of the fragment() function was that you should always pass an argument to it. But, making it optional seems reasonable. We can include this as a breaking change in the next major version.

Issue #2499, opened 2019-01-26
Calling input.blur() fails if input was focused


Rich-Harris:
### Basic info:

- **Node.js version:** 11.6.0
- **jsdom version:** 13.2.0

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const { window } = new JSDOM();
const { document } = window;

const input = document.createElement('input');
input.focus();
input.blur();
```

This results in a "Cannot read property '_defaultView' of undefined" error:

```
[project]/node_modules/jsdom/lib/jsdom/living/helpers/focusing.js:71
    view: target._ownerDocument._defaultView,
                                ^

TypeError: Cannot read property '_defaultView' of undefined
    at Object.exports.fireFocusEventWithTargetAdjustment ([project]/node_modules/jsdom/lib/jsdom/living/helpers/focusing.js:71:33)
    at HTMLInputElementImpl.blur ([project]/node_modules/jsdom/lib/jsdom/living/nodes/HTMLAndSVGElementShared-impl.js:52:14)
    at HTMLInputElement.blur ([project]/node_modules/jsdom/lib/jsdom/living/generated/HTMLElement.js:38:23)
    at Object.<anonymous> ([project]/index.js:8:7)
    at Module._compile (internal/modules/cjs/loader.js:721:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:732:10)
    at Module.load (internal/modules/cjs/loader.js:620:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:560:12)
    at Function.Module._load (internal/modules/cjs/loader.js:552:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:774:12)
```

Rich-Harris:
(Should have added: this works in v12, just not in v13.)

danielhusar:
+1 I see the same issue

Zirro:
Thanks for the reports! This will be fixed by https://github.com/jsdom/jsdom/pull/2503.

Rich-Harris:
Excellent, thank you

asturur:
see the same issue.

diegohaz:
Is it possible to release a patch on `v13` for this fix?

domenic:
No, development has moved on to v14.

asturur:
breaking changes for v14 are relatively small, update to 14 should be easy.

diegohaz:
[...]

PR #2503, opened 2019-02-01
Fix element.blur() on a focused element

Zirro:
This regressed in https://github.com/jsdom/jsdom/pull/2463 as it changed the location where the event's `view` property was set.

Resolves https://github.com/jsdom/jsdom/issues/2499.


asturur:
I would really love to see this merged.
This bug is currently killing my UTs and stop my migration away from jsdom 9.
I would work with jsdom 12, but i need version 13 for node-canvas 2.

asturur:
Well looks like i can use 13.1 meantime. Does not have this regression.

PR #2521, opened 2019-02-23
Add linting rules for super hooks invocation

### LINKED ISSUE #2442: Write a lint rule for remembering to call super._attrModified() etc.
There are some methods in jsdom's internals, which subclasses can override. But if those subclasses forget to call the super version of those methods, they break key jsdom functionality.

See https://github.com/jsdom/jsdom/pull/2438 for an example of restoring the missing `super._attrModified()`, which broke the `style` property on those elements.

The best solution I can think of for this kind of scenario is to write a custom lint rule. We'd feet it a list of method names (e.g. `_attrModified()`, `_modified()`, etc.) and it would require that somewhere in the method body we call `super._thatMethodName()`.

Does something like this already exist? Does anyone want to write it?

pmdartus:
Fix #2442

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,6 +7,9 @@
   "parserOptions": {
     "ecmaVersion": 6
   },
+  "plugins": [
+    "jsdom-internal"
+  ],
   "rules": {
     // Possible errors
     "for-direction": "error",
@@ -276,6 +279,9 @@
     "sort-imports": "off",
     "symbol-description": "error",
     "template-curly-spacing": ["error", "never"],
-    "yield-star-spacing": ["error", "after"]
+    "yield-star-spacing": ["error", "after"],
+
+    // Custom rules
+    "jsdom-internal/hook-super-invocation": "error"
   }
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -282,6 +282,13 @@
     "yield-star-spacing": ["error", "after"],
 
     // Custom rules
-    "jsdom-internal/hook-super-invocation": "error"
+    "jsdom-internal/hook-super-invocation": ["error",
+      { "ancestor": "NodeImpl", "hook": "_attach" },
+      { "ancestor": "NodeImpl", "hook": "_detach" },
+      { "ancestor": "NodeImpl", "hook": "_descendantAdded" },
+      { "ancestor": "NodeImpl", "hook": "_descendantRemoved" },
+      { "ancestor": "NodeImpl", "hook": "_childTextContentChangeSteps" },
+      { "ancestor": "ElementImpl", "hook": "_attrModified" }
+    ]
   }
 }
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -99,7 +99,6 @@ class ElementImpl extends NodeImpl {
     }
   }
 
-  /* eslint-disable-next-line jsdom-internal/hook-super-invocation */
   _attrModified(name, value, oldValue) {
     this._modified();
     namedPropertiesWindow.elementAttributeModified(this, name, value, oldValue);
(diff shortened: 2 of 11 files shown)


lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -99,6 +99,7 @@ class ElementImpl extends NodeImpl {
     }
   }
 
+  /* eslint-disable-next-line jsdom-internal/hook-super-invocation */
```
pmdartus: Disable eslint rule on top hooks that don't require super invocation.

lib/jsdom/living/nodes/HTMLTextAreaElement-impl.js
```diff
@@ -40,6 +40,8 @@ class HTMLTextAreaElementImpl extends HTMLElementImpl {
   }
 
   _childTextContentChangeSteps() {
+    super._childTextContentChangeSteps();
```
pmdartus: The new custom rule found one missing super hook call. 🎉

package.json
```diff
@@ -55,6 +55,7 @@
     "chai": "^4.2.0",
     "eslint": "^4.19.1",
     "eslint-find-rules": "^3.3.1",
+    "eslint-plugin-jsdom-internal": "link:./scripts/eslint-plugin",
```
pmdartus: Eslint doesn't allow to reference in the eslint config local customs rules. Custom rules can be injected by either passing the [`--ruledir` flag](https://eslint.org/docs/user-guide/command-line-interface#--rulesdir), or by defining an eslint plugin package. The main advantage of using a local package over passing a flag to the CLI is that, eslint integration in the code editor will work out of the box without any configuration. This approach is borrowed from the react repo: [`package.json`](https://github.com/facebook/react/blob/master/package.json#L55)

Using [yarn `link:` dependency](https://github.com/yarnpkg/rfcs/blob/8015fd14f572bc444c35f9625eaef0d7d3bb3bf9/implemented/0000-link-dependency-type.md), to create a symbolic link instead of a copy of the local files in the `node_modules` directory.

scripts/eslint-plugin/rules/hook-super-invocation.js
```diff
@@ -0,0 +1,81 @@
+"use strict";
+
+const JSDOM_HOOK = new Set([
+  "_attach",
+  "_detach",
+  "_attrModified",
+  "_descendantAdded",
+  "_descendantRemoved",
+  "_childTextContentChangeSteps"
+]);
```
Zirro: It might be more consistent with the other rules to define these method names as [options](https://eslint.org/docs/developer-guide/working-with-rules#contextoptions) in the ESLint config file rather than hardcoding them here.

pmdartus:
Since jsdom is the only package interested in consuming this rule I think it makes sense to keep the rule definition as close as possible to the code for now. This would also allow to quickly add new rules in the future as we see the need arise.

I am also fine porting the rule in another repo, if you feel it's better on the long term. It's your call here.

lib/jsdom/living/nodes/Element-impl.js
domenic: It seems like it'd be better if you stated the class names of the top-hierarchy classes for each hook in the config file; then these wouldn't be necessary.

scripts/eslint-plugin/rules/hook-super-invocation.js
domenic: I agree it would be nicer if we could configure linting-related things in one place. (I.e., if/when we add a future hook.) If it's not too much work that'd be preferable, but if it is, then feel free to leave them here.

.eslintrc.json
```diff
@@ -276,6 +279,16 @@
     "sort-imports": "off",
     "symbol-description": "error",
     "template-curly-spacing": ["error", "never"],
-    "yield-star-spacing": ["error", "after"]
+    "yield-star-spacing": ["error", "after"],
+
+    // Custom rules
+    "jsdom-internal/hook-super-invocation": ["error",
```
pmdartus: @domenic @Zirro The hooks configuration is now part of the `.eslintrc`. It contains the hooks name along with the ancestor class defining the hook. This way no need to disable the rule for the base class anymore. Let me know what you think.

PR #2522, opened 2019-02-26
Make innerHMTL and outerHTML parsing spec compliant

pmdartus:
This pull request fixes the issue discussed here: https://github.com/jsdom/jsdom/pull/2398#discussion_r230072618. The `innerHTML` and `outerHTML`are currently directly manipulating the context object, using [`appendChild`](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/browser/htmltodom.js#L58-L65), [`removeChild`](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/helpers/html.js#L6-L8) instead of creating a fragment and inserting it in the DOM at the end of the parsing.

**Changes:**
* Create a generic parser module exposing 2 methods: `parseFragment` and `parseDocument`
*  Migrate all the internal usages of `htmltodom` module to use the new parser module
* Adjust implementation from `Element.innerHTML`, `Element.outerHTML`, `Element. insertAdjacentHTML`, `ShadowRoot.innerHTML` to closer match to the spec
* Update `saxes` to `3.1.9`


lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -150,49 +151,41 @@ class ElementImpl extends NodeImpl {
...
     return fragmentSerialization(this, { requireWellFormed: true });
   }
+  set innerHTML(markup) {
+    const fragment = parseFragment(markup, this);
 
-  set innerHTML(html) {
-    if (html === null) {
-      html = "";
+    let contextObject = this;
+    if (this.tagName === "TEMPLATE" && this.namespaceURI === HTML_NS) {
```
domenic:
```suggestion
    if (this.localName === "template" && this.namespaceURI === HTML_NS) {
```

lib/jsdom/living/nodes/HTMLFrameElement-impl.js
```diff
@@ -54,7 +55,13 @@ function fetchFrame(serializedURL, frame, document, contentDoc) {
     contentDoc._encoding = encoding;
 
     const html = whatwgEncoding.decode(data, contentDoc._encoding);
-    contentDoc._htmlToDom.appendToDocument(html, contentDoc);
+
+    try {
+      parseDocument(html, contentDoc);
+    } catch (error) {
+      // Ignore parsing errors when XML is malformed
```
domenic: I think we should fire a jsdomError, at least. The browser does a yellow screen of death; I'm not sure a "console error" is quite drastic enough, but it's a good start, probably.

test/web-platform-tests/to-run.yaml
```diff
@@ -152,11 +150,11 @@ NodeIterator-removal.html: [fail, Unknown]
...
 
 DOMParser-parseFromString-html.html: [fail, needs to get "the active document's URL" which is not possible with one DOMParser shared across all windows]
-DOMParser-parseFromString-xml-doctype.html: [fail, Unknown]
+DOMParser-parseFromString-xml-doctype.html: [fail, saxes doesn't support  DTD validation]
 DOMParser-parseFromString-xml.html: [fail, needs to get "the active document's URL" which is not possible with one DOMParser shared across all windows]
-createContextualFragment.html: [fail, Unknown]
+createContextualFragment.html: [fail, Range is not implemented]
 innerhtml-05.xhtml: [fail, Unknown]
-insert_adjacent_html-xhtml.xhtml: [fail, Unknown]
+insert_adjacent_html-xhtml.xhtml: [fail, XML fragment parsing doesn't enforce well-formedness checks]
```
domenic: That seems like a saxes bug; can you file it and link to it here?

test/web-platform-tests/to-run.yaml
```diff
@@ -845,7 +843,7 @@ DocumentOrShadowRoot-prototype-elementFromPoint.html: [fail, offsetTop not imple
 Element-interface-attachShadow-custom-element.html: [fail, CustomElement.define is not implemented]
 MouseEvent-prototype-offsetX-offsetY.html: [fail, offsetTop not implemented]
 Range-prototype-insertNode.html: [fail, Range is not implemented]
-ShadowRoot-interface.html: [fail, ShadowRoot.innerHTML is still at the stage of draft]
+ShadowRoot-interface.html: [fail, ShadowRoot.styleSheet and ShadowRoot.activeElement are not yet implemented]
```
domenic:
```suggestion
ShadowRoot-interface.html: [fail, shadowRoot.styleSheet and shadowRoot.activeElement are not yet implemented]
```

(Preexisting problem, but good to fix...)

lib/jsdom/browser/parser/xml.js
```diff
@@ -0,0 +1,189 @@
...
+}
+
+function parseFragment(markup, contextElement) {
+  const ownerDocument = contextElement._ownerDocument;
+  const fragment = DocumentFragment.createImpl([], { ownerDocument });
+
+  const parser = createParser(fragment, ownerDocument, {
+    xmlns: true,
+    fragment: true,
+    resolvePrefix(prefix) {
```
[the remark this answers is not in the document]
pmdartus: Based on the saxes documentation, the [resolvePrefix](https://github.com/lddubeau/saxes#parsing-xml-fragments) callback is only used for fragments when the parser is not able to resolve a namespace by itself. I guess, when parsing a document, the parser should have the knowledge of all the namespace and associated prefixes.

test/web-platform-tests/to-run.yaml
pmdartus: This issue has been fixed with saxes version [`3.1.6`](https://github.com/lddubeau/saxes/blob/master/CHANGELOG.md#316-2019-01-17). I took the opportunity to the dependency to the latest version.

lib/jsdom/living/nodes/HTMLFrameElement-impl.js
pmdartus: I updated the code to make it throw a jsdomError (https://github.com/jsdom/jsdom/pull/2522/files#diff-de5536e24fb5f7c11bd2a04cbecc0700R64).

After reading the spec, this line caught my attention (https://html.spec.whatwg.org/#read-xml):

> Error messages from the parse process (e.g., XML namespace well-formedness errors) may be reported inline by mutating the Document.

I am wondering if it makes more sense when the XML parser is parsing a document instead of throwing `DOMException` it should instead report the error directly in the document. This approach would remove the need to add a try/catch here and would probably stick closer to the spec. Any thoughts about this?

Zirro: Yes, replacing malformed documents with one that contains a `<parsererror>` element seems to be the approach taken by browsers. We already do so for documents created through DOMParser: https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/domparsing/DOMParser-impl.js#L20

benjamingr:
Hey, currently running into a nasty bug in jsdom that this fixes - any ETA on this since it looks pretty ready?

lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -141,6 +141,7 @@ class ElementImpl extends NodeImpl {
     return this._attributes;
   }
 
+  // https://w3c.github.io/DOM-Parsing/#dom-element-outerhtml
   get outerHTML() {
```
domenic: Worth noting that we should use the XML serializer here one day.

domenic: Nevermind, we already do!

lib/jsdom/living/nodes/HTMLFrameElement-impl.js
```diff
@@ -54,7 +57,26 @@ function fetchFrame(serializedURL, frame, document, contentDoc) {
...
+        error instanceof DOMException &&
+        error.code === DOMException.SYNTAX_ERR &&
+        contentDoc._parsingMode === "xml"
+      ) {
+        // As defined (https://html.spec.whatwg.org/#read-xml) parsing error in XML document may be reported inline by
+        // mutating the document.
+        const element = document.createElementNS("http://www.mozilla.org/newlayout/xml/parsererror.xml", "parsererror");
+        element.textContent = error.message;
+
+        document.body.appendChild(element);
```
Zirro: The spec does not provide much detail regarding this, but browsers seem to differ in their handling of parsing errors. Chrome and Safari seem to append an error to the existing document, while Firefox appears to throw it away and return a new document with only a `<parsererror>` element ([example](http://malform.no/testing/html5/bom/ascii.us-ascii.xhtml)). Perhaps we should follow Firefox here to simplify our error handling.

domenic: I like that idea, let me try it.

domenic:
Grr, I'm getting mysterious test failures in `Document-createElement-namespace` with the new parser error behavior, which I think might point to a deeper problem, so I want to avoid merging until we figure it out. Help appreciated, although I'll probably keep picking at it a bit...

domenic:
@pmdartus 	would you mind rebasing on top of master? It looks like the problem is https://github.com/jsdom/jsdom/commit/75a921eab52c239b2468e08be9f547f46c7f86bd . Please feel free to squash as necessary if that makes it easier.

pmdartus:
@domenic Rebased on top of master on pushed.

PR #2532, opened 2019-03-16
Implement PageTransitionEvent interface and pageshow event

Zirro:
Also sets the correct target for the load event - the document rather than window - through the legacy target override flag per https://html.spec.whatwg.org/multipage/parsing.html#the-end.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -7,6 +7,7 @@ const notImplemented = require("./not-implemented");
 const { define, mixin } = require("../utils");
 const Element = require("../living/generated/Element");
 const EventTarget = require("../living/generated/EventTarget");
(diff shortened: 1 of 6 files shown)


lib/jsdom/browser/Window.js
```diff
@@ -632,10 +633,15 @@ function Window(options) {
...
 
     if (window.document.readyState === "complete") {
-      fireAnEvent("load", window);
+      fireAnEvent("load", window, undefined, {}, window.document);
     } else {
       window.document.addEventListener("load", () => {
-        fireAnEvent("load", window);
+        fireAnEvent("load", window, undefined, {}, window.document);
+
+        if (!idlUtils.implForWrapper(window._document)._pageShowingFlag) {
```
domenic: I am surprised `_document` is not an impl... /me shakes fist at `Window`.

PR #2536, opened 2019-03-19
Implement changes to event path iteration

Zirro:
This is an implementation of the changes made to the event path iteration in https://github.com/whatwg/dom/pull/686, though it's currently causing many existing tests to fail. I've not yet been able to determine whether this is due to an issue with the spec or a bug in this implementation.

### DIFF
--- a/test/to-port-to-wpts/level2/events.js
+++ b/test/to-port-to-wpts/level2/events.js
@@ -377,7 +377,7 @@ describe("level2/events", () => {
       this.plist.addEventListener("foo",  this.monitor.handleEvent, true);
       this.plist.addEventListener("foo",  this.monitor.handleEvent, false);
       this.plist.dispatchEvent(this.event);
-      assert.equal(this.monitor.atEvents.length, 4, 'should be at 4 events');
+      assert.equal(this.monitor.atEvents.length, 2, 'should be at 2 events'); // Changed from 4 to 2 after https://github.com/whatwg/dom/commit/98564fc5284439d2555f545fa04288e230a37a03
       assert.equal(this.monitor.bubbledEvents.length, 0, 'should have no bubbled events');
       assert.equal(this.monitor.capturedEvents.length, 0, 'should have no captured events');
     });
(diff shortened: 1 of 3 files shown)


domenic:
@pmdartus I noticed you opened https://github.com/whatwg/dom/issues/742. Do you think that could be what's causing all the test failures? Even if not, you might be best-positioned to help @Zirro debug the issues here.

pmdartus:
I will try to debug soon. I am suspecting another issue with the spec if there are so many failures unrelated to `composed`.

pmdartus:
With the change proposed https://github.com/whatwg/dom/issues/742#issuecomment-476558030, tests start passing again. In the current state of the spec, the `load` event handler is never invoked causing the test harness to timeout.

That being said there is still a test failure (`dom/events/Event-dispatch-handlers-changed.html`) that I am still investigating.

annevk:
That was a mistake in that test, PR here: https://github.com/web-platform-tests/wpt/pull/16307.

pmdartus:
Thanks for the info @annevk.

domenic:
@Zirro, up for updating this with https://github.com/whatwg/dom/pull/750, and probably rolling the web platform tests too?

Zirro:
I've updated the code to include the changes from https://github.com/whatwg/dom/pull/750.

test/to-port-to-wpts/level2/events.js
```diff
@@ -377,7 +377,7 @@ describe("level2/events", () => {
       this.plist.addEventListener("foo",  this.monitor.handleEvent, true);
       this.plist.addEventListener("foo",  this.monitor.handleEvent, false);
       this.plist.dispatchEvent(this.event);
-      assert.equal(this.monitor.atEvents.length, 4, 'should be at 4 events');
+      assert.equal(this.monitor.atEvents.length, 2, 'should be at 2 events'); // Changed from 4 to 2 after https://github.com/whatwg/dom/commit/98564fc5284439d2555f545fa04288e230a37a03
```
Zirro: @domenic I'm not familiar enough with these older tests, but it might be worth verifying that this change isn't an indication of a problem with the spec change.

domenic: Filed https://github.com/whatwg/dom/issues/759

Issue #2495, opened 2019-01-21
JSDOM accepts a buffer but doesn't allow to set a content type encoding and defaults to windows-1252


mrtcode:
### Basic info:

- **Node.js version: 8.10
- **jsdom version: 13.1

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");
const dom = new JSDOM(buffer, {
  contentType: res.headers['content-type'], // Doesn't use the charset from content-type
  encoding: 'UTF-8' // Doesn't use the passed encoding
});
```

It should either automatically extract charset from `options.contentType`, or it should allow to pass `options.encoding`.

Internally it's using `transportLayerEncodingLabelHiddenOption` to pass the encoding, therefore `fromURL` method works correctly.

https://github.com/jsdom/jsdom/blob/master/lib/api.js#L33

domenic:
This is intentional. If you pass a buffer, we act like the browser does, and do encoding sniffing on it. To set the encoding, use meta charset inside the HTML that is encoded in that buffer.

mrtcode:
Well, it's good that JSDOM is guessing encoding from the content, but isn't the browser also using the encoding from the `content-type` header? Doesn't it has a higher priority?

For example let's take [this](https://www.ynet.co.il/articles/0,7340,L-5037054,00.html) website as an example. If I pass it as a buffer, it will be incorrectly decoded. But if I pass the URL to `fromURL` method, JSDOM is internally using the encoding from the `content-type` [header](https://github.com/jsdom/jsdom/blob/master/lib/api.js#L135), and decodes succesfully. So why JSDOM is able to pass the encoding with internal `transportLayerEncodingLabelHiddenOption`, but the client script doesn`t have a way to pass the correct encoding from header?

This limitation makes it impossible to correctly decode some websites, and the only way is to move the content encoding guessing logic to the client script and pass JSDOM the already decoded string instead of the buffer. Which doesn't sound right. So I really think the smartest way to solve this would be to make JSDOM to watch into `content-type` header encoding.

domenic:
Well, the `contentType` option is not the same as a content-type header. You can't supply headers with a `Buffer`.

But, maybe we should treat it similarly, instead of only treating it as an XML vs. HTML mode switch. Hmm.

mrtcode:
[...]

PR #2543, opened 2019-03-26
Recognize the .xht file extension

Zirro:
The [`.xht`](https://en.wikipedia.org/wiki/XHTML) file extension is a shorter form of `.xhtml` which for completeness we ought to match alongside `.xhtml` and `.xml` in the API. This could potentially be seen as a breaking change that we should save for the next major release (though it's certainly not worth making a major release over this change alone).


domenic:
We should do https://github.com/jsdom/jsdom/issues/2495 at the same time, and do a major release soon.

PR #2546, opened 2019-03-28
Fix <option>'s label and value properties

Zirro:
For reasons that are not clear to me, this [test](https://github.com/web-platform-tests/wpt/blob/master/html/semantics/forms/the-option-element/option-label-value.js) which passes in all browsers requires that the `label` and `value` attributes don't belong to a namespace. There doesn't appear to be any special instructions regarding this in the [specification](https://html.spec.whatwg.org/multipage/form-elements.html#dom-option-label).

Is this merely missing from the spec, or is it something we're actually supposed to check for all attributes?

### DIFF
--- a/lib/jsdom/living/nodes/HTMLOptionElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLOptionElement-impl.js
@@ -1,4 +1,5 @@
 "use strict";
+
 const HTMLElementImpl = require("./HTMLElement-impl").implementation;
 const { stripAndCollapseASCIIWhitespace } = require("../helpers/strings");
 const { domSymbolTree } = require("../helpers/internal-constants");
@@ -14,6 +15,7 @@ class HTMLOptionElementImpl extends HTMLElementImpl {
     this._selectedness = false;
     this._dirtyness = false;
   }
+
   _removeOtherSelectedness() {
     // Remove the selectedness flag from all other options in this select
     const select = this._selectNode;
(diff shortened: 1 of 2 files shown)


domenic:
This is generally something we're supposed to check for all attributes. See https://html.spec.whatwg.org/multipage/infrastructure.html#xml paragraph 2.

Zirro:
Hmm, I don't think we're checking this anywhere else currently when we're handling attributes. I wonder whether just mass-replacing `getAttribute(...)` with `getAttributeNS(null, ...)` or introducing helpers would be the better approach if we wanted to do this everywhere.

domenic:
Yeah, I'm unsure myself. Maybe helpers, perhaps even as a method on Element-impl. I used to have a dream of using `attributes.js` as the helper-location, but in practice that's not so convenient for people to use. And the HTML spec very rarely calls the appropriate operations such as https://dom.spec.whatwg.org/#concept-element-attributes-get-by-name .

I'm curious what browser code does to make the default-path the correct one. (Assuming they get it right.)

Zirro:
ESLint's [`no-restricted-properties`](https://eslint.org/docs/rules/no-restricted-properties) rule allows us to forbid certain property names and provide a custom error message ("getAttribute is restricted from being used. To access attributes inside jsdom you'll want getAttributeNS() with null as namespace instead").

If we restrict it to `lib/`, I imagine it could be quite helpful. What do you think?

domenic:
I like that idea quite a lot.

domenic:
Do you want to update this to use setAttributeNS, and then land https://github.com/jsdom/jsdom/pull/2547 ? Or land that first (with fixed text expectations), then this as a further cleanup?

Zirro:
You can merge this now, and I'll rebase and bump webidl2js in #2547 tomorrow.

PR #2547, opened 2019-03-30
Verify that attributes don't have namespaces

Zirro:
Following the discussion in https://github.com/jsdom/jsdom/pull/2546.

This includes linting rules to prevent `getAttribute()`, `hasAttribute()`, `removeAttribute()` and `toggleAttribute()` from being used in favour of their NS-alternatives.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -290,5 +290,21 @@
       { "ancestor": "NodeImpl", "hook": "_childTextContentChangeSteps" },
       { "ancestor": "ElementImpl", "hook": "_attrModified" }
     ]
-  }
+  },
+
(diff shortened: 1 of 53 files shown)


domenic:
It'd be good to include a bump of webidl2js to v9.2.1 in this as well.

Zirro:
I've rebased this now, but it seems like the new version of webidl2js hasn't been published.

domenic:
I wish there were a way to configure ESLint smarter so we didn't have the disable-eslint comments. I tried using `!`-prefixed rules in the `"files"` entry, but they did not work. (And Element-impl.js is a bit of a special case.) Oh well.

PR #2548, opened 2019-04-01
Add custom element support

### LINKED ISSUE #1030: Support for WebComponents API
We are trying to use jsdom for our unit tests for React core (http://facebook.github.io/react/).

Unfortunately, the web components spec is not natively supported by jsdom, and the webcomponents.js polyfill does not run on jsdom.  This issue requests the addition of WebComponent support (custom elements, shadow dom, html imports, etc).

pmdartus:
This PR introduces native support for custom elements in jsdom. 

While the core of the implementation is in place, the PR is currently at the draft stage to flush out some design decisions mainly around `[HTMLConstructor]` and `[CEReactions]`.

Fixes #1030


lib/jsdom/living/helpers/html-element-constructor.js
```diff
@@ -0,0 +1,162 @@
...
+  "HTMLTemplateElement",
+  "HTMLTextAreaElement",
+  "HTMLTimeElement",
+  "HTMLTitleElement",
+  "HTMLTrackElement",
+  "HTMLUListElement",
+  "HTMLVideoElement",
+];
+
+function patchHTMLConstructor(window) {
```
pmdartus: This implementation of `[HTMLConstructor]` patches the interface constructor and keep in place the interface prototype. This approach avoids paying the performance overhead of creating a brand new interface per window. Is there something that can be done at webidl2js level to avoid hardcoding a list of interfaces to patch?

lib/jsdom/living/helpers/ce-reactions.js
```diff
@@ -0,0 +1,680 @@
...
+  ],
+  CSSStyleDeclaration: [
+    "cssText",
+    "setProperty",
+    "removeProperty",
+    "cssFloat"
+  ]
+};
+
+function wrapWithCEReactions(original) {
```
pmdartus: Since webidl2js currently doesn't process the `[CEReactions]` annotation, for the purpose of this PR I used 2 different approaches:

**Interface patching:** Before exporting the interfaces, each property that requires `[CEReactions]` is patched. The approach requires to keep a mapping of interfaces and properties to patch which is not ideal. This approach also doesn't work for interfaces using named and index setters (eg. `DOMStringMap`), because there is not known properties to patch in advance on the interface.

**Impl patching:** Instead of patching the generated interface, in this approach the Impl class prototype is directly patched. It allows solving the issue described above for named and index setters: https://github.com/jsdom/jsdom/pull/2548/files#diff-2be81b507625c47c1dab5327ca4ace8aR65. If we were to use this approach everywhere it will require each impl to have list of all the properties that are required to be patched (vs. a central place).

A final approach would be to delegate this webidl2js, but I didn't spend enough time thinking about how to implement this.

Thoughts?

domenic: I would implement this in webidl2js, similar to [Reflect], with the following design: it outputs a method body of roughly:

```js
// do all argument conversions etc.

impl._ceReactionsPreSteps();

try {
  ... original stuff, including calling impl[methodName](...args);
} finally {
  impl._ceReactionsPostSteps();
}
```

Then, in jsdom, we would implement `_ceReactionsPreSteps()` and `_ceReactionsPostSteps()` in Node-impl.js.

domenic: The better architecture would be to add plugins to webidl2js, to let jsdom write a plugin that did this CEReactions stuff. Maybe you would want to give that a try, but it sounds hard to me. The _ceReactionsPre/PostSteps insertion is my attempt at something simple but still slightly decoupled.

I guess another approach would be to add an option to webidl2js's generate, e.g.

```js
transformer.generate(outputDir, {
  processCEReactions: normalMethodBody => expandedMethodBody
});
```

That's pretty simple, and even more decoupled.

(A "plugin" architecture is probably like that, but more general, so that you can intercept any extended attribute, applied to any IDL construct, not just CEReactions applied to operations.)

lib/jsdom/living/helpers/html-element-constructor.js
domenic: This is the hard one. See below for my thoughts on webidl2js plugins and interfacing. I think a `processHTMLConstructor` option might be the best bet here, so that webidl2js calls you every time it sees HTMLConstructor, and asks you to provide the text it wants to generate.

Then, you can reuse the architecture of https://github.com/jsdom/webidl2js#webidl2jsfactory so that it generates constructors per-Window in the same way. But, right now we just manually know that there are 4 interfaces which need to be set up in Window's constructor. (See e.g. the `createAbortController` call.) It'd be best to generalize this, so that webidl2js spits out a file we can require(), which exports a function we can use like `installPerWindowConstructors(window)`. That would then take care of both [Webidl2jsFactory] and [HTMLConstructor].

test/web-platform-tests/to-run.yaml
```diff
@@ -81,6 +81,41 @@ window-screen-width.html: [fail, Test not applicable - no output device]
 
 ---
 
+DIR: custom-elements
+
+CustomElementRegistry.html: [fail, Promise identity discontinuity,
+                                   webidl2js doesn't deal well with tests using Proxies to verify properties access]
```
domenic: Can you say more about these failures? webidl2js strives to work well in these ways, so fixing it should not be too hard.

test/web-platform-tests/to-run.yaml
```diff
@@ -81,6 +81,41 @@ window-screen-width.html: [fail, Test not applicable - no output device]
...
+Document-createElement.html: [fail, :defined is not defined and throws,
+                                    TODO - createElement should report error instead of throwing]
+HTMLElement-constructor.html: [fail, webidl2js doesn't deal well with tests using Proxies to verify properties access]
+attribute-changed-callback.html: [fail, attributeChangedCallback doesn't work with CSSStyleDeclaration]
+custom-element-reaction-queue.html: [fail, Document.write implementation is not spec compliant]
```
domenic: I didn't see an implementation of the flag in our document.write implementation. Even if we can't get the tests fully passing, maybe it's worth adding? Not sure.

test/web-platform-tests/to-run.yaml
```diff
@@ -81,6 +81,41 @@ window-screen-width.html: [fail, Test not applicable - no output device]
...
+reactions/Document.html: [fail,
+  Document.execCommand is not implemented, https://github.com/jsdom/jsdom/issues/1539
+  Document.write implementation is not spec compliant]
+reactions/ElementContentEditable.html: [fail, contentEditable is not implemented]
+reactions/HTMLButtonElement.html: [fail, HTMLButtonElement doesn't implement formAction formEnctype and formMethod]
+reactions/HTMLElement.html: [fail, translate and spellcheck attributes are not implemented on HTMLElement]
+reactions/Range.html: [fail, Range is not implemented, https://github.com/jsdom/jsdom/issues/317]
+reactions/Selection.html: [fail, Selection is not implemented, https://github.com/jsdom/jsdom/issues/937]
+throw-on-dynamic-markup-insertion-counter-construct.html: [timeout, Document.write implementation is not spec compliant]
+throw-on-dynamic-markup-insertion-counter-reactions.html: [timeout, Document.write implementation is not spec compliant]
```
domenic: I didn't see an implementation of the flag in our document.write implementation. Even if we can't get the tests fully passing, maybe it's worth adding? Not sure.

lib/jsdom/browser/parser/html.js
```diff
@@ -158,10 +173,16 @@ class JSDOMParse5Adapter {
 Object.assign(JSDOMParse5Adapter.prototype, serializationAdapter);
 
 function parseFragment(markup, contextElement) {
```
domenic: These changes look independently valuable. Would it make sense to split them out? Hopefully there are some targeted tests they fix, separate from custom elements, although we might have to write them...

alexgwolff:
So, this PR will be merged?

domenic:
As you can see there are unaddressed review comments.

marioandrest:
@pmdartus @domenic 
thank you so much for the addition of customElements to jsdom
this PR has enabled our test stack to use jsdom
would it be possible to rebase your branch to a jsdom 15 release?
(_this would remove the need to polyfill MutationObserver_)
also, what is the status of moving this PR forward?
thanks!

tibysko:
Hi all,
I understand that there is some issues left to be solved before this PR can be merged. However, is it possible to say roughly when / if is going to be merged? 

Thanks!

domenic:
No, sorry. Everyone involved is doing this in their free time, not on a schedule.

package.json
```diff
@@ -82,7 +83,7 @@
     "st": "^1.2.2",
     "watchify": "^3.11.1",
     "wd": "^1.11.2",
-    "webidl2js": "^9.2.2"
+    "webidl2js": "https://github.com/pmdartus/webidl2js.git#pmdartus/add-processor"
```
pmdartus: **⚠️Use the released version of webidl2js before merging ⚠️** 
Using the WIP version to run the test suite on the CI.

pmdartus:
Finally, the PR is in pretty good shape.

However, I just realized that the `[HTMLConstructor]` processor introduces a major regression with the constructor check. The following snippet will start failing:

```js
const elm = document.createElement('div');
elm.constructor === HTMLDivElement 
// Expected: true, Actual: false
```

I don't see any way to get around this without creating a brand new constructor and a brand new prototype chain per realm (https://github.com/jsdom/webidl2js/issues/133: part. 1 + part. 2). I would be happy to take a stab at this it's a blocker.

gusnaughton:
Thank you to all working on this, it is very helpful!

TimothyGu:
> However, I just realized that the `[HTMLConstructor]` processor introduces a major regression with the constructor check.

I would really prefer this not to be the case on the default setting. For the purpose of getting this merged in a reasonable timeline, however, I'd be happy with an “experimental flag” that enables custom elements at the cost of such regression, and iterate on the design of webidl2js.

pmdartus:
Sound good to me, I will hide the custom elements behind a flag.

pmdartus:
Let's put on hold this PR, until the constructor reform lands.
Discussion: https://github.com/jsdom/webidl2js/issues/133#issuecomment-543798213

thernstig:
@pmdartus Seems you fixed https://github.com/jsdom/webidl2js/issues/133 (in PR https://github.com/jsdom/webidl2js/pull/140) which is awesome! Will this PR here be continued? :) Just thankful you are doing all this.

pmdartus:
@thernstig Thanks for the kind words. I can confirm you that I am still working on this, I am curerntly working on integrating the latest changes in this branch.

thernstig:
@pmdartus Eager and all, did you manage to continue on this or did life come in the way? :)

ExE-Boss:
The&nbsp;`create‑element.js`&nbsp;helper should&nbsp;be&nbsp;moved in&nbsp;its&nbsp;own&nbsp;commit, like&nbsp;<https://github.com/jsdom/jsdom/pull/2764>.

lib/jsdom/living/webstorage/Storage-impl.js
```diff
@@ -20,6 +20,10 @@ class StorageImpl {
   }
 
   _dispatchStorageEvent(key, oldValue, newValue) {
+    // TODO: clean that up
+    // eslint-disable-next-line
+    const StorageEvent = require("../generated/StorageEvent");
```
pmdartus: The new evaluation order introduced by this PR, exhibit the circular dependency between `Storage` and `StorageEvent`. Since this method is not invoked often, it should be fine paying the cost of importing the interface every time the `_dispatchStorageEvent` is invoked.

test/web-platform-tests/to-run.yaml
pmdartus: The test wraps the custom element constructor into a Proxy and checks all the property access on it (`constructor`, `prototype`, `attributeChangedCallback`, ...). Because webidl2js uses a Symbol to like the wrapper to the implementation, the Proxy catches the Symbol property access making the test fail.

lib/jsdom/browser/Window.js
```diff
@@ -28,15 +29,15 @@ const { fireAnEvent } = require("../living/helpers/events");
...
 const jsGlobals = require("./js-globals.json");
 
 const GlobalEventHandlersImpl = require("../living/nodes/GlobalEventHandlers-impl").implementation;
 const WindowEventHandlersImpl = require("../living/nodes/WindowEventHandlers-impl").implementation;
 
-// NB: the require() must be after assigning `module.exports` because this require() is circular
-// TODO: this above note might not even be true anymore... figure out the cycle and document it, or clean up.
-module.exports = Window;
-const { installInterfaces } = require("../living/interfaces");
+exports.createWindow = function (options) {
```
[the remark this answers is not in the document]
ExE-Boss: You&nbsp;might&nbsp;want to&nbsp;add this&nbsp;comment from&nbsp;<https://github.com/jsdom/jsdom/pull/2771>:
```suggestion
// We can't re-assign `module.exports`, as there's the following cycle:
// /lib/jsdom/browser/Window
// -> /lib/jsdom/living/interfaces
// -> /lib/jsdom/living/generated/HTMLFrameElement
// -> /lib/jsdom/living/nodes/HTMLFrameElement-impl
// -> /lib/jsdom/browser/Window
exports.createWindow = function (options) {
```

test/web-platform-tests/to-run.yaml
ExE-Boss: **WebIDL2JS**&nbsp;should&nbsp;use a&nbsp;**WeakMap** to&nbsp;map&nbsp;wrappers to&nbsp;implementations.

pmdartus: I don't it is the case: https://github.com/jsdom/webidl2js/blob/master/lib/output/utils.js#L34-L36

ExE-Boss: I’ve&nbsp;created&nbsp;<https://github.com/jsdom/webidl2js/pull/155> to&nbsp;update&nbsp;**WebIDL2JS** to&nbsp;use&nbsp;**WeakMap**s for&nbsp;the&nbsp;wrapper&nbsp;→&nbsp;impl&nbsp;mapping.

test/web-platform-tests/to-run.yaml
```diff
@@ -920,7 +961,7 @@ DIR: shadow-dom
 
 Document-prototype-currentScript.html: [timeout, Test not up to date next with updating wpt it should work]
 DocumentOrShadowRoot-prototype-elementFromPoint.html: [fail, offsetTop not implemented]
-Element-interface-attachShadow-custom-element.html: [fail, CustomElement.define is not implemented]
+Element-interface-attachShadow-custom-element.html: [needs-node12, Need static class fields support]
```
ExE-Boss: This&nbsp;test seems&nbsp;to&nbsp;pass in&nbsp;**Node&nbsp;10**.

lib/jsdom/living/webstorage/Storage-impl.js
```diff
@@ -1,10 +1,12 @@
 "use strict";
 
 const DOMException = require("domexception/webidl2js-wrapper");
-const StorageEvent = require("../generated/StorageEvent");
-const idlUtils = require("../generated/utils");
+
 const { fireAnEvent } = require("../helpers/events");
 
+const idlUtils = require("../generated/utils");
+const StorageEvent = require("../generated/StorageEvent");
```
ExE-Boss: Is&nbsp;it&nbsp;necessary to&nbsp;move&nbsp;this after&nbsp;`fireAnEvent` and&nbsp;`idlUtils`?

pmdartus: That's an old change that I haven't reverted properly.

package.json
```diff
@@ -81,7 +82,7 @@
     "st": "^2.0.0",
     "watchify": "^3.11.1",
     "wd": "^1.11.2",
-    "webidl2js": "^13.0.0"
+    "webidl2js": "pmdartus/webidl2js#pmdartus/add-processor"
```
ExE-Boss: Don’t&nbsp;forget to&nbsp;use the&nbsp;production **WebIDL2JS**&nbsp;release before&nbsp;merging.

lib/jsdom/browser/Window.js
```diff
@@ -576,6 +577,8 @@ function Window(options) {
     return window._document.getSelection();
   };
 
+  this.customElements = CustomElementRegistry.create(window);
```
domenic: This should be a getter; remember we don't have webidl2js generated code to handle that for us here.

lib/jsdom/browser/parser/html.js
```diff
@@ -11,15 +13,19 @@ const attributes = require("../../living/attributes");
 const nodeTypes = require("../../living/node-type");
 
 const serializationAdapter = require("../../living/domparsing/parse5-adapter-serialization");
+const {
+  CUSTOM_ELEMENT_REACTIONS_STACK, invokeCEReactions, lookupCEDefinition
```
domenic: I don't think CUSTOM_ELEMENTS_REACTIONS_STACK should be capitalized in this way; it's just an object.

lib/jsdom/living/attributes.js
```diff
@@ -28,21 +30,40 @@ exports.changeAttribute = function (element, attribute, value) {
   // https://dom.spec.whatwg.org/#concept-element-attributes-change
 
   const { _localName, _namespace, _value } = attribute;
+
   queueAttributeMutationRecord(element, _localName, _namespace, _value);
 
-  const oldValue = attribute._value;
+  if (element._ceState === CUSTOM_ELEMENT_STATE.CUSTOM) {
```
domenic: I'd rather just use strings directly like the spec does, instead of strings after a `.`

lib/jsdom/living/custom-elements/CustomElementRegistry-impl.js
```diff
@@ -0,0 +1,253 @@
...
+      construct() {
+        return {};
+      }
+    });
+
+    // eslint-disable-next-line no-new
+    new P();
+
+    return true;
+  } catch (err) {
```
domenic:
```suggestion
  } catch {
```

lib/jsdom/living/custom-elements/CustomElementRegistry-impl.js
```diff
@@ -0,0 +1,253 @@
...
+    this._whenDefinedPromiseMap = Object.create(null);
+
+    this._globalObject = globalObject;
+  }
+
+  // https://html.spec.whatwg.org/#dom-customelementregistry-define
+  define(name, ctor, options) {
+    const { _globalObject } = this;
+
+    if (typeof ctor !== "function" || !isConstructor(ctor)) {
```
domenic:
```suggestion
    if (!isConstructor(ctor)) {
```

(or, move that check into isConstructor)

lib/jsdom/living/helpers/custom-elements.js
```diff
@@ -11,17 +70,205 @@ const RESTRICTED_CUSTOM_ELEMENT_NAME = new Set([
...
+            upgradeElement(reaction.definition, element);
+            break;
+
+          case "callback":
+            reaction.callback.apply(wrapperForImpl(element), reaction.args);
+            break;
+        }
+      }
+    } catch (error) {
+      reportException(element._ownerDocument._defaultView, error);
```
domenic:
```suggestion
      reportException(element._globalObject, error);
```

lib/jsdom/living/helpers/html-constructor.js
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable global-require */
```
domenic: Why?

lib/jsdom/living/interfaces.js
```diff
@@ -1,186 +1,187 @@
...
-  require("./generated/ShadowRoot"),
-
-  require("./generated/MutationObserver"),
-  require("./generated/MutationRecord"),
-
-  require("./generated/Headers"),
-  require("./generated/AbortController"),
-  require("./generated/AbortSignal")
-];
+const generatedInterfaces = {
```
domenic: You could probably leave this as-is and then do a single pass over the array to build a hash table from `arrayMember.interface.name` or something like that... seems a bit nicer for maintenance.

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -196,6 +195,9 @@ class DocumentImpl extends NodeImpl {
     // to which the browsing context's session history was most recently traversed. When a Document is created,
     // it initially has no latest entry.
     this._latestEntry = null;
+
+    // https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#throw-on-dynamic-markup-insertion-counter
+    this._throwOwnDynamicMarkupInsertionCounter = 0;
```
domenic: Own -> On

test/web-platform-tests/to-run.yaml
```diff
@@ -103,6 +103,47 @@ window-screen-width.html: [fail, Test not applicable - no output device]
...
+form-associated/form-associated-callback.html: [fail, Not implemented]
+form-associated/form-disabled-callback.html: [timeout, Not implemented]
+form-associated/form-reset-callback.html: [fail, Not implemented]
+htmlconstructor/newtarget.html: [fail, Currently impossible to get the active function associated realm]
+microtasks-and-constructors.html: [fail, Usage of external scripts doesn't block HTML parsing, https://github.com/jsdom/jsdom/issues/2413]
+parser/parser-constructs-custom-element-synchronously.html: [fail, Usage of external scripts doesn't block HTML parsing, https://github.com/jsdom/jsdom/issues/2413]
+parser/parser-fallsback-to-unknown-element.html: [fail, Usage of external scripts doesn't block HTML parsing, https://github.com/jsdom/jsdom/issues/2413]
+parser/parser-sets-attributes-and-children.html: [fail, Usage of external scripts doesn't block HTML parsing, https://github.com/jsdom/jsdom/issues/2413]
+parser/parser-uses-constructed-element.html: [fail, Usage of external scripts doesn't block HTML parsing, https://github.com/jsdom/jsdom/issues/2413]
+parser/serializing-html-fragments.html: [fail, parse5 doesn't support is attribute for serialization]
```
domenic: Can we file an issue for this?

test/web-platform-tests/to-run.yaml
```diff
@@ -103,6 +103,47 @@ window-screen-width.html: [fail, Test not applicable - no output device]
...
 
+DIR: custom-elements
+
+CustomElementRegistry.html: [fail, webidl2js doesn't deal well with tests using Proxies to verify properties access]
+Document-createElement.html: [fail, :defined is not defined and throws]
+HTMLElement-attachInternals.html: [fail, Not implemented]
+HTMLElement-constructor.html: [fail, webidl2js doesn't deal well with tests using Proxies to verify properties access]
+adopted-callback.html: [fail, "Failing test due to https://github.com/whatwg/dom/issues/813"]
+attribute-changed-callback.html: [fail, attributeChangedCallback doesn't work with CSSStyleDeclaration]
+custom-element-reaction-queue.html: [fail, Document.write implementation is not spec compliant]
```
domenic: Here and elsewhere refer to methods as `document.write()`, not `Document.write`. There is no static `write` property on `Document`.

lib/jsdom/living/helpers/html-constructor.js
pmdartus: Good catch, forgot to remove this in a previous refactor.

lib/jsdom/living/interfaces.js
pmdartus: The `interface` is not exposed anymore on the generated interfaces since the constructor and prototype reform. And there is, unfortunately, no other way today to retrieve the interface name.

test/web-platform-tests/to-run.yaml
pmdartus: I realized as I was writing the issue for this, that adding a new parse5 tree adapter method is probably overkill for handing the `is` attribute serialization. The parse5 API is already complex, I don't think adding more complexity into the mix is the right option. For this reason, I added some special logic to handle the `is` attribute in the `getAttrList` handler to add the `is` attribute if necessary there.

lib/jsdom/living/helpers/create-element.js
```diff
@@ -0,0 +1,306 @@
...
+function getSVGInterface(name) {
+  const specDefinedInterface = TAG_INTERFACE_LOOKUP[SVG_NS][name];
+  if (specDefinedInterface !== undefined) {
+    return interfaces.getInterfaceWrapper(specDefinedInterface);
+  }
+
+  return interfaces.getInterfaceWrapper("SVGElement");
+}
+
+/** Returns the list of valid tag names that can bo associated with a element given its namespace and name. */
```
domenic:
```suggestion
// Returns the list of valid tag names that can be associated with a element given its namespace and name.
```

lib/jsdom/living/helpers/create-element.js
```diff
@@ -0,0 +1,306 @@
...
+    if (namespace === HTML_NS && (isValidCustomElementName(localName) || isValue !== null)) {
+      result._ceState = "undefined";
+    }
+  }
+
+  return result;
+}
+
+// https://dom.spec.whatwg.org/#internal-createelementns-steps
+function createElementNS(document, namespace, qualifiedName, options) {
```
domenic: Could we call this `internalCreateElementNSSteps`? Right now it seems too much like a peer of `createElement` where in the spec "create an element" is a lot more fundamental.

lib/jsdom/living/helpers/custom-elements.js
```diff
@@ -1,5 +1,56 @@
...
+    const { _stack } = this;
+    return _stack[_stack.length - 1];
+  }
+
+  isEmpty() {
+    return this._stack.length === 0;
+  }
+}
+
+const customElementReactionsStack = new CEReactionsStack();
```
domenic: Maybe add a comment that in the spec this is per agent but we don't deal with that for now in jsdom.

(E.g. in theory separate cross-origin `Window`s created by separate `JSDOM` instances could have separate stacks. But, we would need to implement the whole agent architecture. Which is kind of questionable given that we don't run our `Window`s in their own separate threads, which is what agents are meant to represent.)

lib/jsdom/living/interfaces.js
domenic: Good point. Can we add a comment explaining that this needs to stay as-is in order to make bundlers not freak out? Right now it seems like a really tempting refactoring target.

lib/jsdom/living/interfaces.js
```diff
@@ -192,21 +193,23 @@ function install(window, name, interfaceConstructor) {
...
   // These need to be cleaned up...
   style.addToCore(window);
   xpath(window);
 
   // This one is OK but needs migration to webidl2js eventually.
   nodeFilter(window);
 };
+
+/** Returns an interface webidl2js wrapper given its an interface name. */
+exports.getInterfaceWrapper = function (name) {
```
domenic:
```suggestion
exports.getInterfaceWrapper = name => {
```

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -325,6 +301,13 @@ class DocumentImpl extends NodeImpl {
       ]);
     }
 
+    if (this._throwOnDynamicMarkupInsertionCounter > 0) {
+      throw DOMException.create(this._globalObject, [
+        "Cannot use document.write while an element upgrades",
```
```suggestion
        "Cannot use document.write while a custom element upgrades",
```

test/web-platform-tests/to-run.yaml
```diff
@@ -103,6 +103,46 @@ window-screen-width.html: [fail, Test not applicable - no output device]
 
 ---
 
+DIR: custom-elements
+
+CustomElementRegistry.html: [fail, webidl2js doesn't deal well with tests using Proxies to verify properties access]
+Document-createElement.html: [fail, :defined is not defined and throws]
+HTMLElement-attachInternals.html: [fail, Not implemented]
+HTMLElement-constructor.html: [fail, webidl2js doesn't deal well with tests using Proxies to verify properties access]
+adopted-callback.html: [fail, "Failing test due to https://github.com/whatwg/dom/issues/813"]
```
domenic: It seems like there's been some progress on the spec issue here; is this still right?

test/web-platform-tests/to-run.yaml
```diff
@@ -103,6 +103,46 @@ window-screen-width.html: [fail, Test not applicable - no output device]
...
+attribute-changed-callback.html: [fail, attributeChangedCallback doesn't work with CSSStyleDeclaration]
```
domenic: Is this fixable? We already do some fun stuff to synchronize the style="" attribute and CSSStyleDeclaration, if I recall.

If it's fixable but more work than you want to tackle right now, then filing a follow-up issue with a description of the problem/solution space, and pointing to it, would work well.

test/web-platform-tests/to-run.yaml
```diff
@@ -103,6 +103,46 @@ window-screen-width.html: [fail, Test not applicable - no output device]
...
+custom-element-reaction-queue.html: [fail, document.write() implementation is not spec compliant]
+custom-element-registry/per-global.html: [fail, iframe location related issue]
+form-associated/ElementInternals-NotSupportedError.html: [fail, Not implemented]
```
domenic: Let's just use `form-associated/**`?

domenic:
Hmm, well, I guess we still have https://github.com/jsdom/webidl2js/pull/159, so I won't do a release until we get that settled...

pmdartus:
I rebase the PR and upgraded the version of webidl2js. However, I realized that the Proxy performance improvement broke (https://github.com/jsdom/webidl2js/commit/7c8037ffd4afcff01fd4ac48423675aa64aa1cea) broke [CEReactions] and [HTMLConstructor] processor (https://github.com/jsdom/webidl2js/commit/fda810e8d36fe64612bba533228b1b4e91f1a569) only after upgrading webidl2js 😢. 

In short, the [CEReactions] changes rely on the fact that the `globalObject` is available in all the wrapper methods via the `install` closure. Since the Proxy traps object is created outside the closure, the `globalObject` is not accessible anymore.

I don't see yet how we can preserve the performance optimization by hoisting the trap object out of the closure and at the same time be able to retrieve the associated `globalObject` that is needed for the [CEReactions]. @ExE-Boss what are your thoughts on this?

ExE-Boss:
You can use:
```js
const globalObject = target[impl]._globalObject;
```

The&nbsp;`_globalObject` property is&nbsp;available on&nbsp;`EventTargetImpl`&nbsp;objects:

https://github.com/jsdom/jsdom/blob/1ebb1fcd3823882e2cee36afcd1b78217b629ee4/lib/jsdom/living/events/EventTarget-impl.js#L21-L25

test/web-platform-tests/to-run.yaml
pmdartus: We have [multiple](https://github.com/jsdom/jsdom/blob/master/test/web-platform-tests/to-run.yaml#L161) [tests](https://github.com/jsdom/jsdom/blob/master/test/web-platform-tests/to-run.yaml#L822) that have been disabled because of this issue. I would be happy to work on this in a different PR.

test/web-platform-tests/to-run.yaml
pmdartus: It's fixable, however, it would require some work on the `cssstyle` package: https://github.com/jsdom/cssstyle/issues/113

ExE-Boss:
**WebIDL2JS v14** should’ve included <https://github.com/jsdom/webidl2js/pull/168>.

bellindj:
@pmdartus thanks so much for your work on this!! Congrats on merging!

Issue #1723, opened 2017-01-31
Fix option.value/text/label


domenic:
I'm filing this as a good bug for someone who wants to contribute to jsdom, as it's a pretty straightforward instance of updating jsdom to match the spec.

Currently, our implementations of the `value`, `text`, and `label` properties of `<option>` elements are incorrect. They can be seen [in HTMLOptionElement-impl.js](https://github.com/tmpvar/jsdom/blob/30a7772ed92b31b039d87f2d21d136b1d1901732/lib/jsdom/living/nodes/HTMLOptionElement-impl.js), with comments on the wrong parts.

We have tests for these already in place, but commented out: after https://github.com/tmpvar/jsdom/pull/1722 lands, there will be four lines in [web-platform-tests/index.js](https://github.com/tmpvar/jsdom/blob/master/test/web-platform-tests/index.js) for "html/semantics/forms/the-option-element/" that are commented out. The goal is to uncomment those lines, and then get `npm run test-wpt` to pass.

You may find it easier to run the tests if you comment out all the other tests in that file besides those four during development, as the rest of the tests in that file take a while to run.

To make the tests pass, you need to implement the spec:

- [For `text`](https://html.spec.whatwg.org/multipage/forms.html#dom-option-text)
- [For `value`](https://html.spec.whatwg.org/multipage/forms.html#dom-option-value)
- [For `label`](https://html.spec.whatwg.org/multipage/forms.html#dom-option-label)

Note how fixing `text` makes the others much easier to fix, as they delegate to it.

The hardest part of implementing `text` will be doing appropriate DOM node crawling. You're going to want to use domSymbolTree.treeIterator. Ctrl+F the codebase to find similar usages. Don't forget CDATA sections (which are a type of Text node). The code in Node-impl.js's textContent implementation would be a good starting point, although the script element business will be tricky. You may find the `closest` utility, in helpers/traversal, helpful.

Hope this sounds fun for someone!

rajpatel507:
can I take this one? I am new to jsdom so give some time to refer some code.

domenic:
For sure, feel free!

ev1stensberg:
@rajpatel507 If you need a hand, ping me 😄

rajpatel507:
@ev1stensberg  I got the local environment running, have changed `web-platform-tests/index.js ` so that only 4 test will execute, now i am confuse how to change `HTMLOptionElement-impl.js` so that all error can be disappear.

[...]

PR #2550, opened 2019-04-04
Fix <option>'s text getter

Zirro:
It now correctly excludes descendants of HTML and SVG <script> elements.

EDIT: I've realised that this needs a fix and new test. I'll close it in the meantime.


Zirro:
This together with https://github.com/jsdom/jsdom/commit/9a382982d8ab1f498b4e789fd16d494b1f37ebbb resolves https://github.com/jsdom/jsdom/issues/1723.

domenic:
@Zirro I lost track of what happened here; why didn't we end up landing this?

PR #2558, opened 2019-04-20
Fix HTML serialization breaking after setting certain properties

### LINKED ISSUE #2554: JSDOM does not typecheck attributes
### Basic info:

- **Node.js version:** 10.15
- **jsdom version:** 14.0.0

### Minimal reproduction case
```

var jsdom = require('jsdom')

var jsdom = new jsdom.JSDOM(
  `<div>
      <progress
        class="StudyPlan-load-progress svelte-g1ijzk"
        value="0"
        max="100"
      />
    </div>`,
  {
    url: 'http://localhost',
    pretendToBeVisual: true,
  }
)
var window = jsdom.window

window.document.querySelector('.StudyPlan-load-progress').value = 0

console.log(window.document.body.innerHTML)
```

Output:

```
$ node /tmp/testcase.js
0
/tmp/node_modules/parse5/lib/serializer/index.js:168
    str = str.replace(AMP_REGEX, '&amp;').replace(NBSP_REGEX, '&nbsp;');
              ^

TypeError: str.replace is not a function
    at Function.Serializer.escapeString (/tmp/node_modules/parse5/lib/serializer/index.js:168:15)
    at Serializer._serializeAttributes (/tmp/node_modules/parse5/lib/serializer/index.js:103:38)
    at Serializer._serializeElement (/tmp/node_modules/parse5/lib/serializer/index.js:67:14)
    at Serializer._serializeChildNodes (/tmp/node_modules/parse5/lib/serializer/index.js:50:26)
    at Serializer._serializeElement (/tmp/node_modules/parse5/lib/serializer/index.js:93:18)
    at Serializer._serializeChildNodes (/tmp/node_modules/parse5/lib/serializer/index.js:50:26)
    at Serializer.serialize (/tmp/node_modules/parse5/lib/serializer/index.js:36:14)
    at Object.exports.serialize (/tmp/node_modules/parse5/lib/index.js:28:23)
    at htmlSerialization (/tmp/node_modules/jsdom/lib/jsdom/living/domparsing/serialization.js:20:17)
    at module.exports.fragmentSerialization (/tmp/node_modules/jsdom/lib/jsdom/living/domparsing/serialization.js:27:12)

```

The browser will return different types based how you call it: 

Selection_005

Is parse5 using the wrong api when accessing the value?

domenic:
Various properties that reflect HTML attributes, e.g. <progress>'s value and max properties, would cause HTML serialization (e.g. via innerHTML) to break after they were set, due to a missing conversion causing jsdom to store the attribute's value as a non-string. This fixes the problem in one central location for robustness, even if it may be more per-spec to fix it all the entry points.

Fixes #2554.

---

I'm curious on collaborators thoughts on this approach. As the commit message notes, it's a tradeoff.

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -244,6 +244,10 @@ class ElementImpl extends NodeImpl {
   setAttributeNS(namespace, name, value) {
     const extracted = validateNames.validateAndExtract(namespace, name);
 
+    // Because of widespread use of this method internally, e.g. to manually implement attribute/content reflection, we
+    // centralize the conversion to a string here, so that all call sites don't have to do it.
+    value = `${value}`;
+
     attributes.setAttributeValue(this, extracted.localName, value, extracted.prefix, extracted.namespace);
   }
 
--- a/test/web-platform-tests/to-upstream/html/semantics/forms/the-progress-element/serialization-regression-test.html
+++ b/test/web-platform-tests/to-upstream/html/semantics/forms/the-progress-element/serialization-regression-test.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<title>Regression test for serialization bug with progress.value</title>
+<!-- https://github.com/jsdom/jsdom/issues/2554 -->
+
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<div><progress></progress></div>
+
+<script>
+"use strict";
+
+test(() => {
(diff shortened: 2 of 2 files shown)


Sebmaster:
> I'm curious on collaborators thoughts on this approach. As the commit message notes, it's a tradeoff.

Seems reasonable, just ensuring type safety here, really.

Somewhat related because I just thought about this: Do we have any opinions on typings generally? Like jsdocing internally maybe?

domenic:
I dunno, I've historically not been a big fan, but if people think it'd make a difference, maybe we could experiment...

PR #2560, opened 2019-04-20
Allow the contentType option to override sniffed encoding

### LINKED ISSUE #2495: JSDOM accepts a buffer but doesn't allow to set a content type encoding and defaults to windows-1252

domenic:
Fixes #2495, by making the contentType option behave more like the Content-Type header. This also simplifies the code, by unifying the code paths used by the JSDOM constructor and the JSOM.fromURL() function, now that they behave similarly.


ms502040:
not working in 16.4.0

PR #2565, opened 2019-04-21
Implement Headers from the fetch standard

Sebmaster:


Sebmaster:
Blocked on jsdom/webidl2js#121.

lib/jsdom/living/fetch/Headers-impl.js
```diff
@@ -0,0 +1,231 @@
...
+      return null;
+    }
+    return values.map(h => h.value).join(", ");
+  }
+
+  _removePrivilegedNoCORSHeaders() {
+    this.headersList.delete("range");
+  }
+
+  _append(name, value) {
```
domenic: The difference between _append() and append() is confusing. I think what is happening here is that _append is https://fetch.spec.whatwg.org/#concept-header-list-append for header-list, and append() is for the Headers class.

Instead, could we separate out the headers list class into its own file, with its own methods (e.g. append, delete, set, combine, sort and combine)?

lib/jsdom/living/fetch/reservedHeaders.js
```diff
@@ -0,0 +1,42 @@
+"use strict";
```
domenic: Filename should be `reserved-headers`.

Although, why are these called "reserved headers", instead of e.g. "header classes" or "header types" or similar?

lib/jsdom/living/fetch/Headers-impl.js
```diff
@@ -0,0 +1,231 @@
...
+  }
+}
+
+function assertValue(value) {
+  if (value.match(/[\0\r\n]/)) {
+    throw new TypeError("value is invalid");
+  }
+}
+
+function isForbiddenHeader(name) {
```
domenic: It seems like we have two ways of testing a header: certain Sets, and certain functions. Maybe it would be better if the reserved-headers/header-types/whatever file just exported functions, leaving the sets as an encapsulated implementation detail? Then this file could focus mostly on the Headers class itself, not on header types or on header list stuff.

lib/jsdom/living/fetch/Headers-impl.js
```diff
@@ -0,0 +1,231 @@
...
+        }
+        break;
+    }
+
+    this._append(name, value);
+    this._removePrivilegedNoCORSHeaders();
+  }
+
+  set(name, value) {
+    value = value.trim();
```
domenic: trim() will remove Unicode whitespace... I guess it's OK though since we're behind a ByteString guard, interesting.

lib/jsdom/living/fetch/Headers-impl.js
```diff
@@ -0,0 +1,231 @@
...
+    assertName(name);
+    assertValue(value);
+
+    switch (this.guard) {
+      case "immutable":
```
domenic: Braces around case blocks? I thought we had an ESLint rule enforcing that...

lib/jsdom/living/fetch/Headers-impl.js
domenic: I see use of toLowerCase() as well. Perhaps a comment at the top of the file explaining that it's OK to use normally-Unicode string operations because the bindings layer guarantees that the strings are ByteStrings.

lib/jsdom/living/fetch/Headers-impl.js
```diff
@@ -0,0 +1,231 @@
...
+        throw new TypeError("Headers is immutable");
+      case "request":
+        if (isForbiddenHeader(name)) {
+          return;
+        }
+        break;
+      case "request-no-cors": {
+        if (
+          !NO_CORS_SAFELISTED_REQUEST.has(name.toLowerCase()) &&
```
domenic: Yeah, making these functions instead of sets would be much nicer, so you didn't have to remember to lowercase every time.

lib/jsdom/living/fetch/Headers-impl.js
Sebmaster: Variable declarations inside require it.

lib/jsdom/living/fetch/reservedHeaders.js
Sebmaster: Naming things is hard.

PR #2566, opened 2019-04-21
Use frozen lockfile to keep yarn lockfile in sync

Sebmaster:


Sebmaster:
Just had unrelated changes in my lockfile when I installed modules, which was kinda confusing.

domenic:
But this is only on CI, right?

Sebmaster:
> But this is only on CI, right?

Yeah, but it'll prevent committing package.json upgrades without corresponding lockfile updates.

PR #2569, opened 2019-04-22
Roll Web Platform Tests

Zirro:
A few changes to the code were made as the `/` prefix has been removed from the paths in the manifest.


test/web-platform-tests/to-run.yaml
```diff
@@ -28,7 +28,7 @@ basic.htm: [fail, Unknown]
 client-hint-request-headers-2.tentative.htm: [fail, Client Hints not implemented]
 client-hint-request-headers.htm: [fail-slow, Client Hints not implemented]
 cors-safelisted-request-header.any.html: [fail, Depends on fetch]
-credentials-flag.htm: [fail, Unknown]
+credentials-flag.htm: [needs-node12, Unknown]
```
domenic: Weird.

test/web-platform-tests/to-run.yaml
```diff
@@ -126,7 +126,7 @@ DIR: dom/nodes
 Comment-constructor.html: [fail, Unknown]
 DOMImplementation-createDocument.html: [fail, Unknown]
 Document-URL.html: [fail, Unknown]
-Document-characterSet-normalization.html: [fail, Some encodings are not supported - see the whatwg-encoding module]
+Document-characterSet-normalization.html: [timeout, Some encodings are not supported - see the whatwg-encoding module]
```
domenic: Why this change?

test/web-platform-tests/to-run.yaml
Zirro: Yep, presumably a fix from the new HTTP parser.

test/web-platform-tests/to-run.yaml
Zirro: With v12, I saw this test take various times between 10000ms to 60000ms until it failed.

Zirro: Interestingly it only seems to occur when all tests up to that point have been run. If just that test or only the tests from that directory run in isolation, I don't see the same long delay.

Zirro:
Looks like there needs to be a new release of canvas with Node.js v12 support: https://github.com/Automattic/node-canvas/issues/1410

I'll disable testing of the canvas dependency in stable Node for now.

Zirro:
Rolled again to include the tests for https://github.com/jsdom/jsdom/pull/2571.

I also ended up disabling `response-method.htm` as `flaky`, since I saw it passing a couple of times even though it still fails most of the time in Node.js v12.

PR #2570, opened 2019-04-24
Fix MutationObserver _nodeList memory leak

kschzt:
The MutationObserver kept a `_nodeList` of nodes added to it via `observe()`. When the node was removed from the DOM however, the MO didn't remove its reference to it, causing a memory leak.

From what I can tell, the `_nodeList` isn't actually _really_ needed for anything.

Also, the for-loop in notifying MO's called each MO as many times as it had nodes in the `_nodeList`, when calling each MO once is enough.

### DIFF
--- a/lib/jsdom/living/helpers/mutation-observers.js
+++ b/lib/jsdom/living/helpers/mutation-observers.js
@@ -148,23 +148,17 @@ function notifyMutationObservers() {
     const records = [...mo._recordQueue];
     mo._recordQueue = [];
 
-    for (const node of mo._nodeList) {
-      node._registeredObserverList = node._registeredObserverList.filter(registeredObserver => {
-        return registeredObserver.source !== mo;
-      });
+    if (records.length) {
+      try {
+        mo._callback(
+          records.map(idlUtils.wrapperForImpl),
+          idlUtils.wrapperForImpl(mo)
+        );
+      } catch (e) {
+        const { target } = records[0];
+        const window = target._ownerDocument._defaultView;
 
-      if (records.length) {
-        try {
-          mo._callback(
-            records.map(idlUtils.wrapperForImpl),
-            idlUtils.wrapperForImpl(mo)
-          );
-        } catch (e) {
-          const { target } = records[0];
-          const window = target._ownerDocument._defaultView;
-
-          reportException(window, e);
-        }
+        reportException(window, e);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


kschzt:
Doesn't look like the tests failed due to this change?

pmdartus:
Thanks for reporting the issue @kschzt.

The [node list](https://dom.spec.whatwg.org/#mutationobserver-node-list) is used to manage the [transient registered observer](https://dom.spec.whatwg.org/#transient-registered-observer). Looking back at the PR introducing Mutation Observer, I missed the step in the [remove](https://dom.spec.whatwg.org/#ref-for-transient-registered-observer) adding a new transient observer.

When investigating the leak, I stumbled upon the [garbage collection](https://dom.spec.whatwg.org/#garbage-collection) section. I am confused about the following sentences:

> Nodes have a strong reference to registered observers in their registered observer list.
> Registered observers in a node’s registered observer list have a weak reference to the node.

There is a reference to the registered observers using a weak reference to the node, but the registered observers don't reference nodes directly.

kschzt:
@pmdartus I'm not exactly sure what the implications of your message are – do you think the approach in the PR is correct? Without it, MO retains a strong reference to the node.

domenic:
Ping @pmdartus, I'm counting on you for further review here :)

TimothyGu:
I spent some time looking at this, and I think by "to the node" in

> Registered observers in a node’s registered observer list have a weak reference to the node.

the spec really meant the node in the node list of the registered observer, with the assumption being that GC'd nodes would disappear automatically from the node list.

Since we don't yet have `WeakRef`s or enumerable `WeakSet`s in the language, this looks a bit tough to resolve. However, the Chromium implementation does something curious: rather than making the node list a weak container, it clears the node list (they call it transient registration nodes) each time after iterating through it.

We might be able to experiment with this approach: instead of not having a `_nodeList`, clear it often.

domenic:
**Posted by GPT-6 Astra Medium**

Closing as superseded by the weak-reference implementation in [8ae47bda](https://github.com/jsdom/jsdom/commit/8ae47bda).

The existing [“does not retain observed nodes through MutationObserver instances” test](https://github.com/jsdom/jsdom/blob/556b11fc3cc4676a22e34fa45dfa09cbd95adfb1/test/api/from-outside.js#L85) keeps the observer alive while checking that its observed node can be garbage-collected, using [this fixture](https://github.com/jsdom/jsdom/blob/556b11fc3cc4676a22e34fa45dfa09cbd95adfb1/test/api/fixtures/mutation-observer-with-gc.js). It passes on current main and runs unconditionally in CI through `test:api`. Observer bookkeeping is retained without keeping the observed nodes alive.

PR #2594, opened 2019-05-24
added _attributeModified for style in SVGElementImpl

kbruneel:
Hi, 

I bumped into this small bug when trying to parse an SVG with inline styles using DOMParser. I'm far from an expert on jsdom, but I managed to do a quick fix.

Let me know if you need any more info.

Cheers,
Karel

### DIFF
--- a/lib/jsdom/living/nodes/ElementCSSInlineStyle-impl.js
+++ b/lib/jsdom/living/nodes/ElementCSSInlineStyle-impl.js
@@ -3,6 +3,7 @@ const cssstyle = require("cssstyle");
 
 class ElementCSSInlineStyle {
   _initElementCSSInlineStyle() {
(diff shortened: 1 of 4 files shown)


TimothyGu:
I know you got the piece of code from https://github.com/jsdom/jsdom/blob/93944f64c8a7580dc16ce1252299efae7fa414cb/lib/jsdom/living/nodes/HTMLElement-impl.js#L119-L129, but I think the other part about `_globalEventChanged` would also be correct. I've amended this change to account for that, and also added a test. See if this still works for you.

kbruneel:
@TimothyGu Yes it's very possible that the other part is also needed, but I was hoping somebody, like you, would come along that knows what he is doing :) Thanks!

Issue #2603, opened 2019-05-29
Jsdom with canvas does not support yarn pnp


etiennedupont:
### Basic info:

- **Node.js version:** 10.15.3 
- **jsdom version:** 15.1.1
- **canvas version** 2.5.0
- **yarn pnp enabled**

### Minimal reproduction case
Install the project with yarn pnp enabled

```js
const { JSDOM } = require("jsdom");

const options = {
  pretendToBeVisual: true,
  storageQuota: 1e9,
  resources: "usable",
  runScripts: "dangerously"
};
const window = new JSDOM(`
   <!DOCTYPE html><body id="main"><p >Hello world</p></body>
`, options).window;
const document = window.document;
const navigator = window.navigator;


const main = async () => {
  console.log("before loading image");
  const imageLoaded = await new Promise((resolve, reject) => {
    console.log("beginning code promise");
    var image = document.createElementNS("http://www.w3.org/1999/xhtml", "img");
    function onImageLoad() {
      image.removeEventListener("load", onImageLoad, false);
      image.removeEventListener("error", onImageError, false);
      console.log("IMAGE LOADED");
      resolve(image);
    }

    function onImageError(event) {
      image.removeEventListener("load", onImageLoad, false);
      image.removeEventListener("error", onImageError, false);
      console.log("IMAGE LOADING FAILED");
      reject(event);
    }
    image.addEventListener("load", onImageLoad, false);
    image.addEventListener("error", onImageError, false);
    image.src = "foo.png";
    console.log("end code promise");
  });
  console.log("image loaded", imageLoaded);
};

main()
```
The image fails to load without launching any of error or load events, as if canvas was not installed. 
Indeed, with the code in https://github.com/jsdom/jsdom/blob/master/lib/jsdom/utils.js#L166 we see that with pnp, jsdom will fail finding the package which will result in boolean `canvasInstalled` being false all the time.

### How does similar code behave in browsers?

canvas should be detected.
 
(Link to a jsbin or similar suggested.)

domenic:
Sounds like a good bug to file on yarn pnp, if it doesn't work the same way as yarn.

etiennedupont:
This short modification fixes the problem, it's not a big deal...
https://github.com/jsdom/jsdom/pull/2604

@arcanis what is your opinion on this case?

domenic:
#2604 is definitely not acceptable; we're not adding yarnpnp-specific code to jsdom.

arcanis:
[...]

PR #2604, opened 2019-05-29
Update utils.js to comply with yarn pnp

etiennedupont:
following issue https://github.com/jsdom/jsdom/issues/2603


domenic:
We'd rather not make changes to jsdom to support tools that don't work with standard Node.js module system practices; instead, those tools need to get better.

PR #2630, opened 2019-07-10
Include the url of the XML document in error messages

papandreou:
Previously it would say eg. `undefined:1:17: malformed comment.`
Fixed by passing in the url as the `fileName` when creating the parser.


papandreou:
As far as my own use case is concerned, I don't really need this if https://github.com/lddubeau/saxes/pull/21 gets merged. Still, I'll leave it on the table since it still feels like a good idea.

papandreou:
> They may be failing on master for some reason, or they may be failing because of this test change---most of the test failures do seem XML related.

They were breaking because of this change. Apparently `ownerDocument.location` isn't available when parsing `XMLHttpRequest` response bodies. Added a guard against that.

PR #2631, opened 2019-07-11
Allow FormData constructor to handle input elements without a value

### LINKED ISSUE #2523: Creating FormData from form fails on input with no value set
### Basic info:

- **Node.js version:** v10.11.0
- **jsdom version:** v13.2.0

### Minimal reproduction case

```js
import { JSDOM } from 'jsdom'

const jsdom = new JSDOM( `<!doctype html>` )
const { document, FormData } = jsdom.window

const form = document.createElement( 'form' )
const input = document.createElement( 'input' )

input.type = 'text'
input.name = 'foo'

form.appendChild( input )

const formData = new FormData( form )
```

The formData object is created and entries contains [ "foo", "" ]

In jsdom, when building the formDataset, for an input element with no value set, when _getValue() is called, it returns null:

https://github.com/jsdom/jsdom/blob/16d3913eea0360c7757e75f7e266c3873c85b7dd/lib/jsdom/living/xhr/FormData-impl.js#L149

This is then passed to normalizeToCRLF, which expects a string:

https://github.com/jsdom/jsdom/blob/16d3913eea0360c7757e75f7e266c3873c85b7dd/lib/jsdom/living/xhr/FormData-impl.js#L163

```
xxx\node_modules\jsdom\lib\jsdom\living\helpers\form-controls.js:63
  return string.replace(/\r([^\n])/g, "\r\n$1")
                ^

TypeError: Cannot read property 'replace' of null
    at exports.normalizeToCRLF.string (xxx\node_modules\jsdom\lib\jsdom\living\helpers\form-controls.js:63:17)
```

brendo:
Fixes #2523

This PR enables the `FormData` constructor to handle input elements that are defined like:

```html
<input type="text" name="look-at-me" />
```

I couldn't quite see anything in the [Living Standard](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-form-data-set) that describes what should happen when the value is not defined, but the browser behaviour for this (Chrome/Firefox at least) is to return an empty string.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLInputElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLInputElement-impl.js
@@ -180,7 +180,7 @@ class HTMLInputElementImpl extends HTMLElementImpl {
   }
 
   _getValue() {
-    return this._value;
+    return this._value !== null ? this._value : "";
   }
 
   _legacyPreActivationBehavior() {
@@ -340,7 +340,7 @@ class HTMLInputElementImpl extends HTMLElementImpl {
     switch (valueAttributeMode(this.type)) {
       // https://html.spec.whatwg.org/multipage/input.html#dom-input-value-value
       case "value":
-        return this._value !== null ? this._value : "";
+        return this._getValue();
       // https://html.spec.whatwg.org/multipage/input.html#dom-input-value-default
       case "default": {
         const attr = this.getAttributeNS(null, "value");
--- a/lib/jsdom/living/nodes/HTMLInputElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLInputElement-impl.js
@@ -158,7 +158,7 @@ class HTMLInputElementImpl extends HTMLElementImpl {
 
     this._selectionStart = this._selectionEnd = 0;
     this._selectionDirection = "none";
-    this._value = null;
+    this._value = "";
     this._dirtyValue = false;
     this._checkedness = false;
     this._dirtyCheckedness = false;
@@ -179,8 +179,10 @@ class HTMLInputElementImpl extends HTMLElementImpl {
     return convertStringToNumberByTypeMap.get(this.type);
   }
 
+  // For <input>, https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-value
+  // is a simple value that is gotten and set, not computed.
   _getValue() {
-    return this._value !== null ? this._value : "";
+    return this._value;
   }
 
   _legacyPreActivationBehavior() {
@@ -364,11 +366,7 @@ class HTMLInputElementImpl extends HTMLElementImpl {
       // https://html.spec.whatwg.org/multipage/input.html#dom-input-value-value
       case "value": {
         const oldValue = this._value;
-        if (val === null) {
-          this._value = null;
-        } else {
-          this._value = sanitizeValueByType(this, String(val));
-        }
+        this._value = sanitizeValueByType(this, val);
         this._dirtyValue = true;
 
         if (oldValue !== this._value) {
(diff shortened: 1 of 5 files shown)


domenic:
Nice catch. From what I can tell of the standard, we should revise `_getValue()` so that it always returns a string. Could you push a commit that takes that approach instead?

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -340,7 +340,7 @@ class HTMLInputElementImpl extends HTMLElementImpl {
     switch (valueAttributeMode(this.type)) {
       // https://html.spec.whatwg.org/multipage/input.html#dom-input-value-value
       case "value":
-        return this._value !== null ? this._value : "";
+        return this._getValue();
```
brendo: @domenic Interestingly it appears that the accessor for the input already handles this case and [has test coverage](https://github.com/jsdom/jsdom/blob/master/test/to-port-to-wpts/htmlinputelement.js#L10-L12). Therefore in order to maintain the existing API I've pushed the logic back to `_getValue()` and have called that method from here. 

That said, I do wonder if `_getValue()` should be removed and the `FormData` logic changed to be:

```js
formDataSet.push({ name, value: field.value, type });
``` 

I am not familiar with this project so am likely missing a lot of context but a project search showed that `_getValue` was implemented in:

- [`HTMLButtonElement`](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/nodes/HTMLButtonElement-impl.js#L31-L34)
- [`HTMLTextareaElement`](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/nodes/HTMLTextAreaElement-impl.js#L37-L40) (which oddly _also_ has a getter for [`value()`](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/nodes/HTMLTextAreaElement-impl.js#L66-L68) which calls [`_getAPIValue `](https://github.com/jsdom/jsdom/blob/ecb22a27627ace4e3d5db09cabf566ec606c36ad/lib/jsdom/living/nodes/HTMLTextAreaElement-impl.js#L33-L35)
- [`HTMLInputElement`](https://github.com/jsdom/jsdom/blob/ecb22a27627ace4e3d5db09cabf566ec606c36ad/lib/jsdom/living/nodes/HTMLInputElement-impl.js#L182-L184) (this file).

It appears to only called by the [`FormData`](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/xhr/FormData-impl.js#L149) logic and I have a suspicion the cases it is handling are already handled by `value` accessor.

1. While `HTMLButtonElement` defines `getValue()`, the loop in `FormData` [will bail out if the element is a button](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/xhr/FormData-impl.js#L108-L110).
1. The logic for [handling checkboxes/radio](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/xhr/FormData-impl.js#L137-L139) is already handled in this switch block with the `default` and `default/on` cases.
1. The `HTMLTextAreaElement` implementation of `_getValue()` already calls `normalizeToCRLF`. This is likely the reason for the [`entry.type !== "textarea"`](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/xhr/FormData-impl.js#L162) guard later in this method.

What do you think?

brendo: @domenic Is there another way I could approach this or are you happy with the changes in ad1415d?

domenic: Hmm, I think this is confusing, indeed. Let me try playing around with something more like the spec...

PR #2633, opened 2019-07-13
fix postMessage bug; add documentImpl cross Origin 

lljxx1:
- add `documentImpl` property, for check the network is in idle (requestManager.size)
- skip cross origin check, always use `preflight` (modern browser will send `Origin` header for cross origin check)
- fix `postMessage` error, fire event throw fatal error when impl not ready

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -44,6 +44,11 @@ class JSDOM {
     documentImpl.close();
   }
 
+  get documentImpl() {
+    // for access requestManager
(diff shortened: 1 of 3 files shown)


domenic:
This diff includes unacceptable whitespace changes (lint will fail), no tests, and no description of the bug it's fixing. Please start with a test case based on the issue template.

lljxx1:
alright.
I am trying to use jsdom for server-side rendering.

PR #2657, opened 2019-09-06
Add origin support to MessageEvent

wkh237:
Add `MessageEvent.origin` to the implementation of `PostMessage`. 

The idea is, when postMessage invoked, create an `Error` object and parse its `stack`.
Since the stack trace is in this format

```
function_name (filename:line_no:column_no)
```

If the function is called from other context, the `filename` would be a string starts with its `host` (which is the origin we look for).

I know this is a bit tricky, but I think it might be a go.

Besides the tests, according to the README files, I know the tests should be composed in wpt format. 

However, it looks like most test cases about `PostMessage` is located in `test/to-port-to-wpts/post-message.js` but anywhere else, I just put my test case there for now.

Please let me know if there's any thoughts about this PR, thanks 😃


domenic:
Heya, we're trying to not introduce new test files in the "to port" folder, so please do write or find an existing web platform test for this functionality. Thanks!

wkh237:
Okay, thanks for heads up! Will move tests to wpt folder !

test/web-platform-tests/to-run.yaml
```diff
@@ -910,6 +910,106 @@ urlencoded-parser.any.html: [fail, Depends on fetch]
 
 ---
 
+DIR: webmessaging
```
wkh237: Add files in `webmessaging` into `to-run.yaml` and excluded failed cases.

test/web-platform-tests/to-upstream/webmessaging/event.data.sub.htm
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
```
wkh237: This test cases are based on :
- https://github.com/web-platform-tests/wpt/blob/master/webmessaging/event.data.sub.htm
- https://github.com/web-platform-tests/wpt/blob/master/webmessaging/event.origin.sub.htm

`jsdom` currently not be able to pass the original tests in `wpt` due to 2 reasons:
- the `<script />` element executes after `<iframe />` loaded
- these tests relies on implementation for `event.source` which is not yet imeplemented

test/web-platform-tests/to-upstream/webmessaging/support/ChildWindowPostMessage.htm
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE html>
```
wkh237: This is a simplified version of https://github.com/web-platform-tests/wpt/blob/master/webmessaging/support/ChildWindowPostMessage.htm

The original file uses `event.source` which I changed to `parent` as a temp workaround.

test/web-platform-tests/to-upstream/webmessaging/event.data.sub.htm
wkh237: refer to https://github.com/web-platform-tests/wpt/issues/206

wkh237:
@domenic I’ve moved the tests to tuwpt folder, please let me know if there’s any problems, thanks.

lib/jsdom/living/post-message.js
```diff
@@ -22,11 +23,28 @@ module.exports = function (message, targetOrigin) {
     return;
   }
 
+  // Get call stack from Error object, so we can know which file (uri) invoked this method
+  const stack = new Error().stack.split("\n");
+  const prevStack = stack[2];
+
+  let sourceOrigin = this.location.origin;
+
+  // grab the filename part from stack trace and the hostname should be the origin
```
TimothyGu: This is not really true in several different ways.

1. If an HTML file on domain1.com includes a script from domain2.com (using `<script src>`), then the origin-in-effect (the [incumbent](https://html.spec.whatwg.org/multipage/webappapis.html#concept-incumbent-everything) origin, per [window post message steps](https://html.spec.whatwg.org/multipage/web-messaging.html#window-post-message-steps) step 8.2) should in fact be domain1.com.
2. The current treatment doesn't take into account port numbers, which are generally part of an [origin](https://html.spec.whatwg.org/multipage/origin.html#concept-origin-tuple).

I'm wondering if the best we could do is to always use the target origin as the source origin.

----

Also, please name your variables as similarly as possible to [the spec](https://html.spec.whatwg.org/multipage/web-messaging.html#window-post-message-steps).

wkh237: Thanks @TimothyGu for and advices, I will change the variable naming and try to figure out if there's any solution on these situations.

PR #2664, opened 2019-09-15
DOMMatrix & DOMMatrixReadOnly Implementations

dpchamps:
Couple of notes: 

Sorry that this PR is long, my desire is to start expanding jsdom's support for geometry and svg. A lot of that begins with DOMMatrix, and I didn't see a better way to spread this initial PR out.

Most of the lone additions come from the tests -- they're extensive, but I'd like to expand some of the coverage even further in future PR's. 

A utility for extracting CSS functions from string for initializing matrices via DOMStrings was introduced. It's crucial to add unit tests for this as well, but I wanted to prevent this PR from gaining even more lines out of the gate.

I've annotated what specifications certain aspects of the matrix calculations come from where I thought appropriate.

### DIFF
--- a/lib/jsdom/living/geometry/DOMMatrix.webidl
+++ b/lib/jsdom/living/geometry/DOMMatrix.webidl
@@ -1,3 +1,5 @@
+// https://www.w3.org/TR/geometry-1/#dommatrix
+
 [Constructor(optional (DOMString or sequence<unrestricted double>) init),
  Exposed=(Window,Worker),
  Serializable,
--- a/lib/jsdom/living/geometry/DOMMatrix.webidl
+++ b/lib/jsdom/living/geometry/DOMMatrix.webidl
@@ -1,0 +1,69 @@
+[Constructor(optional (DOMString or sequence<unrestricted double>) init),
+ Exposed=(Window,Worker),
+ Serializable,
+ LegacyWindowAlias=(SVGMatrix,WebKitCSSMatrix)]
+interface DOMMatrix : DOMMatrixReadOnly {
+    [NewObject] static DOMMatrix fromMatrix(optional DOMMatrixInit other);
+    [NewObject] static DOMMatrix fromFloat32Array(Float32Array array32);
+    [NewObject] static DOMMatrix fromFloat64Array(Float64Array array64);
+
+    // Mutable transform methods
+    DOMMatrix multiplySelf(optional DOMMatrixInit other);
+    DOMMatrix preMultiplySelf(optional DOMMatrixInit other);
+    DOMMatrix translateSelf(optional unrestricted double tx = 0,
+                            optional unrestricted double ty = 0,
+                            optional unrestricted double tz = 0);
+    DOMMatrix scaleSelf(optional unrestricted double scaleX = 1,
+                        optional unrestricted double scaleY,
+                        optional unrestricted double scaleZ = 1,
+                        optional unrestricted double originX = 0,
+                        optional unrestricted double originY = 0,
+                        optional unrestricted double originZ = 0);
+    DOMMatrix scale3dSelf(optional unrestricted double scale = 1,
+                          optional unrestricted double originX = 0,
+                          optional unrestricted double originY = 0,
+                          optional unrestricted double originZ = 0);
+    DOMMatrix rotateSelf(optional unrestricted double rotX = 0,
+                         optional unrestricted double rotY,
+                         optional unrestricted double rotZ);
+    DOMMatrix rotateFromVectorSelf(optional unrestricted double x = 0,
+                                   optional unrestricted double y = 0);
+    DOMMatrix rotateAxisAngleSelf(optional unrestricted double x = 0,
+                                  optional unrestricted double y = 0,
+                                  optional unrestricted double z = 0,
+                                  optional unrestricted double angle = 0);
+    DOMMatrix skewXSelf(optional unrestricted double sx = 0);
+    DOMMatrix skewYSelf(optional unrestricted double sy = 0);
+    DOMMatrix invertSelf();
+
+    [Exposed=Window] DOMMatrix setMatrixValue(DOMString transformList);
+};
+
+dictionary DOMMatrix2DInit {
+    unrestricted double a;
+    unrestricted double b;
+    unrestricted double c;
+    unrestricted double d;
+    unrestricted double e;
+    unrestricted double f;
+    unrestricted double m11;
+    unrestricted double m12;
+    unrestricted double m21;
+    unrestricted double m22;
+    unrestricted double m41;
+    unrestricted double m42;
+};
+
+dictionary DOMMatrixInit : DOMMatrix2DInit {
+    unrestricted double m13 = 0;
+    unrestricted double m14 = 0;
+    unrestricted double m23 = 0;
+    unrestricted double m24 = 0;
+    unrestricted double m31 = 0;
+    unrestricted double m32 = 0;
+    unrestricted double m33 = 1;
+    unrestricted double m34 = 0;
+    unrestricted double m43 = 0;
+    unrestricted double m44 = 1;
+    boolean is2D;
+};
--- a/lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
+++ b/lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
@@ -1,0 +1,451 @@
+/* eslint-disable no-use-before-define */
+"use strict";
+const DOMException = require("domexception");
+const {
+  matrix3dArrayProperties,
+  matrix2dArrayProperties
+} = require("../helpers/geometry/matrix");
+const {
+  validateDOMMatrixInit,
+  arrayFromDOMMatrixInit
+} = require("../helpers/geometry/dommatrix-init");
+
+class DOMMatrixReadOnlyImpl {
+  constructor(constructorArgs) {
+    let [matrixInit] = constructorArgs;
+
+    if (typeof matrixInit === "undefined") {
+      matrixInit = [1, 0, 0, 1, 0, 0];
+    }
+
+    if (typeof matrixInit === "string") {
+      matrixInit = this.parseDomString(matrixInit);
+      throw new Error(`Matrix as DOMString not implemented.`);
+    }
+
+    /**
+     * Note: An underscore indicates that the property is defined by a property accessor.
+     * These properties produce side effects upon getting or setting.
+     * To prevent a ton of property accessors, only those that produce side effects are defined as property accessors.
+     */
+    this.m11 = 0;
+    this.m12 = 0;
+    this._m13 = 0;
+    this._m14 = 0;
+    this.m21 = 0;
+    this.m22 = 0;
+    this._m23 = 0;
+    this._m24 = 0;
+    this._m31 = 0;
+    this._m32 = 0;
+    this._m33 = 0;
+    this._m34 = 0;
+    this.m41 = 0;
+    this.m42 = 0;
+    this._m43 = 0;
+    this._m44 = 0;
+    this._is2D = false;
+
+
+    if (matrixInit.length === 6) {
+      this._initialize2dMatrix(matrixInit);
+    } else if (matrixInit.length === 16) {
+      this._initialize3dMatrix(matrixInit);
+    } else {
+      throw new TypeError(`Invalid length for constructor init, expected 6 or 16, but received ${matrixInit.length}`);
+    }
+  }
+
+  parseDomString(domstring) {
+    // Not implemented
+  }
+
+  _initialize2dMatrix(init) {
+    matrix2dArrayProperties.forEach((prop, idx) => {
+      this[prop] = init[idx];
+    });
+
+    this.m33 = 1;
+    this.m44 = 1;
+    this._is2D = true;
+  }
+
+  _initialize3dMatrix(init) {
+    matrix3dArrayProperties.forEach((prop, idx) => {
+      this[prop] = init[idx];
+    });
+  }
+
+  _getCurrentMatrix() {
+    if (this.is2D) {
+      return matrix2dArrayProperties.map(prop => this[prop]);
+    }
+
+    return matrix3dArrayProperties.map(prop => this[prop]);
+  }
+
+  _updatePropsFromMatrix(matrix) {
+    matrix3dArrayProperties.forEach((prop, idx) => {
+      this[prop] = matrix[idx];
+    });
+  }
+
+  * [Symbol.iterator]() {
+    for (const i of matrix3dArrayProperties.map(prop => this[prop])) {
+      yield i;
+    }
+  }
+
+  get a() {
+    return this.m11;
+  }
+
+  set a(value) {
+    this.m11 = value;
+  }
+
+  get b() {
+    return this.m12;
+  }
+
+  set b(value) {
+    this.m12 = value;
+  }
+
+  get c() {
+    return this.m21;
+  }
+
+  set c(value) {
+    this.m21 = value;
+  }
+
+  get d() {
+    return this.m22;
+  }
+
+  set d(value) {
+    this.m22 = value;
+  }
+
+  get e() {
+    return this.m41;
+  }
+
+  set e(value) {
+    this.m41 = value;
+  }
+
+  get f() {
+    return this.m42;
+  }
+
+  set f(value) {
+    this.m42 = value;
+  }
+
+  get m13() {
+    return this._m13;
+  }
+
+  set m13(value) {
+    if (value !== 0) {
+      this.is2D = false;
+    }
+    this._m13 = value;
+  }
+
+  get m14() {
+    return this._m14;
+  }
+
+  set m14(value) {
+    if (value !== 0) {
+      this.is2D = false;
+    }
+    this._m14 = value;
+  }
+
+  get m23() {
+    return this._m23;
+  }
+
+  set m23(value) {
+    if (value !== 0) {
+      this.is2D = false;
+    }
+    this._m23 = value;
+  }
+
+  get m24() {
+    return this._m24;
+  }
+
+  set m24(value) {
+    if (value !== 0) {
+      this.is2D = false;
+    }
+    this._m24 = value;
+  }
+
+  get m31() {
+    return this._m31;
+  }
+
+  set m31(value) {
+    if (value !== 0) {
+      this.is2D = false;
+    }
+    this._m31 = value;
+  }
+
+  get m32() {
+    return this._m32;
+  }
+
+  set m32(value) {
+    if (value !== 0) {
+      this.is2D = false;
+    }
+    this._m32 = value;
+  }
+
+  get m33() {
+    return this._m33;
+  }
+
+  set m33(value) {
+    if (value !== 1) {
+      this.is2D = false;
+    }
+    this._m33 = value;
+  }
+
+
+  get m34() {
+    return this._m34;
+  }
+
+  set m34(value) {
+    if (value !== 0) {
+      this.is2D = false;
+    }
+    this._m34 = value;
+  }
+
+  get m43() {
+    return this._m43;
+  }
+
+  set m43(value) {
+    if (value !== 0) {
+      this.is2D = false;
+    }
+    this._m43 = value;
+  }
+
+  get m44() {
+    return this._m44;
+  }
+
+  set m44(value) {
+    if (value !== 1) {
+      this.is2D = false;
+    }
+    this._m44 = value;
+  }
+
+  get is2D() {
+    return this._is2D;
+  }
+
+  set is2D(value) {
+    if (value === true && !this._is2D) {
+      // see note https://www.w3.org/TR/geometry-1/#matrix-is-2d
+      return;
+    }
+    this._is2D = Boolean(value);
+  }
+
+  get isIdentity() {
+    const hasRequiredZeroProps = [
+      "m12", "m13", "m14", "m21", "m23", "m24",
+      "m31", "m32", "m34", "m41", "m42", "m43"
+    ].every(prop => this[prop] === 0);
+
+    const hasRequiredIdentityProps = ["m11", "m22", "m33", "m44"].every(prop => this[prop] === 1);
+
+    return hasRequiredZeroProps && hasRequiredIdentityProps;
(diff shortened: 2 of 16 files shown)


lib/jsdom/living/helpers/geometry/matrix.js
```diff
@@ -0,0 +1,64 @@
+"use strict";
```
dpchamps: This should be thoroughly unit tested.

test/web-platform-tests/to-upstream/geometry/dommatrix.html
```diff
@@ -16,7 +16,7 @@
     dma = dma instanceof DOMMatrix ? dma : DOMMatrix.fromMatrix(dma);
     dmb = dmb instanceof DOMMatrix ? dmb : DOMMatrix.fromMatrix(dmb);
 
-    assert_array_approx_equals(arrayFromDm(dma), arrayFromDm(dmb), 0.00000000001);
+    assert_array_approx_equals(arrayFromDm(dma), arrayFromDm(dmb), 0.000001);
```
dpchamps: If anyone has suggestions on how to intuitively preform these calculations with more precision, I'm open to hearing how to do it. Since the browsers are preforming matrix multiplication with native libraries, the precision varies in some cases but is negligible.

scripts/webidl/convert.js
```diff
@@ -30,6 +30,7 @@ addDir("../../lib/jsdom/living/mutation-observer");
 addDir("../../lib/jsdom/living/navigator");
 addDir("../../lib/jsdom/living/nodes");
 addDir("../../lib/jsdom/living/svg");
+addDir("../../lib/jsdom/living/geometry");
```
TimothyGu: I see you've added quite a lot of tests, which makes me a bit nervous as they seem to contain a whole bunch of numbers. Have you checked if the [existing tests](https://github.com/web-platform-tests/wpt/tree/master/css/geometry) in WPT already cover these cases? If so, then let's just enable those tests by modifying https://github.com/jsdom/jsdom/blob/master/test/web-platform-tests/to-run.yaml.

If there is additional coverage provided by this PR, please put the tests into `css/geometry` in order to be consistent with upstream. Additionally, make sure to run these tests in the browser to make sure they're good.

lib/jsdom/living/index.js
```diff
@@ -92,3 +92,6 @@ exports.URL = require("whatwg-url").URL;
 exports.URLSearchParams = require("whatwg-url").URLSearchParams;
 
 exports.Headers = require("./generated/Headers").interface;
+
+exports.DOMMatrixReadOnly = require("./generated/DOMMatrixReadOnly").interface;
+exports.DOMMatrix = require("./generated/DOMMatrix").interface;
```
TimothyGu: Could you put these near the SVG-related entries further up in the file?

lib/jsdom/living/geometry/DOMMatrix.webidl
```diff
@@ -0,0 +1,71 @@
+// https://www.w3.org/TR/geometry-1/#dommatrix
```
TimothyGu: Always implement the latest editor's draft: https://drafts.fxtf.org/geometry/#dommatrix

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
```diff
@@ -0,0 +1,519 @@
...
+    matrix3dArrayProperties.forEach((prop, idx) => {
+      this[prop] = matrix[idx];
+    });
+  }
+
+  * [Symbol.iterator]() {
+    for (const i of matrix3dArrayProperties.map(prop => this[prop])) {
+      yield i;
+    }
+  }
```
TimothyGu: I don't see this being used anywhere. Let's just remove it.

lib/jsdom/living/helpers/geometry/dommatrix-init.js
```diff
@@ -0,0 +1,89 @@
...
+  const fixedInit = Object.assign({}, init);
+
+  aliasedProperties.forEach(([alias, property, defaultValue]) => {
+    fixedInit[property] = init[alias] || defaultValue;
+  });
+
+  return fixedInit;
+}
+
+exports.validateDOMMatrixInit = function (init) {
```
TimothyGu: Let's call this `validateAndFixupDOMMatrixInit`, and add a link to https://drafts.fxtf.org/geometry/#matrix-validate-and-fixup.

lib/jsdom/living/helpers/geometry/dommatrix-init.js
```diff
@@ -0,0 +1,89 @@
+"use strict";
```
TimothyGu: I'd appreciate it if you could use more spec terminology in this file. For example, `getDOMMatrixDimension` really seems to be the last three steps in https://drafts.fxtf.org/geometry/#matrix-validate-and-fixup, while validateAliasedProps and fixupDOMMatrixInit together make up https://drafts.fxtf.org/geometry/#matrix-validate-and-fixup-2d. Let's try to organize the code as close to the spec as possible, and add links to the spec algorithm next to each function.

lib/jsdom/living/helpers/geometry/dommatrix-init.js
```diff
@@ -0,0 +1,89 @@
...
+  return hasProp(o, prop) && Boolean(o[prop]);
+}
+
+// If both alias and actual properties are present, they must be equal.
+function validateAliasedProps(init) {
+  aliasedProperties.forEach(([alias, prop]) => {
+    if (
+      hasProp(init, alias) &&
+      hasProp(init, prop) &&
+      init[alias] !== init[prop]
```
TimothyGu: This uses [Strict Equality Comparison](https://tc39.es/ecma262/#sec-strict-equality-comparison), but the spec (https://drafts.fxtf.org/geometry/#matrix-validate-and-fixup-2d) calls for [SameValueZero](https://tc39.es/ecma262/#sec-samevaluezero) as the comparison algorithm. They have subtle differences w.r.t. the treatment of NaN, as documented at https://drafts.fxtf.org/geometry/#matrix-validate-and-fixup-2d:

> Note: The SameValueZero comparison algorithm returns true for two ‘NaN’ values, and also for ‘0’ and ‘-0’.

Make a new function called `sameValueZero` to handle this.

lib/jsdom/living/helpers/geometry/dommatrix-init.js
```diff
@@ -0,0 +1,89 @@
...
+  });
+}
+
+function isValid2dMatrix(init) {
+  return !(matrix2dZeroProps.some(x => hasProp(init, x) && init[x] !== 0) ||
+    matrix2dOneProps.some(x => hasProp(init, x) && init[x] !== 1));
+}
+
+function getDOMMatrixDimension(init) {
+  if (hasProp(init, "id2D") && !init.is2D) {
```
TimothyGu: For any internal function, we can assume that the input is "well-behaved", since the Web IDL translation layer has already taken care of the kinks like getter functions, inheritance chain, etc. So here, you can simply do `init.is2D === false`.

Even better would be to use the same logic as https://drafts.fxtf.org/geometry/#matrix-validate-and-fixup, which deals with the following cases in order:

1. `init.is2D === true` but the matrix is not 2D: throw an error,
2. `init.is2D === undefined` and the matrix is not 2D: set `is2D` to false,
3. otherwise: set `is2D` to true.

lib/jsdom/living/geometry/DOMPointReadOnly.webidl
```diff
@@ -0,0 +1,18 @@
+// https://www.w3.org/TR/geometry-1/#dompointreadonly
```
TimothyGu: Ditto; use https://drafts.fxtf.org/geometry/.

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
```diff
@@ -0,0 +1,519 @@
...
+
+  get a() {
```
TimothyGu: I'm wondering if there's a better way to specify all of these. Maybe something like this?

```js
// Each row
for (const [
  [element, altName = undefined],
  whenNotThenSet2DToFalse = undefined
] of [
  [['m11', 'a']]
  [['m12', 'b']],
  [['m13'], 0],
  // ...
]) {
  const descriptor = {
    get() {
      return this["_" + element];
    },
    set(newValue) {
      if (whenNotThenSet2DToFalse !== undefined && newValue !== whenNotThenSet2DToFalse) {
        this.is2D = false;
      }
      this["_" + element] = newValue;
    }
    writable: true
    configurable: true
  };
  Object.defineProperty(DOMMatrixReadOnlyImpl.prototype, element, descriptor);
  if (altName !== undefined) {
    Object.defineProperty(DOMMatrixReadOnlyImpl.prototype, altName, descriptor);
  }
}
```

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
```diff
@@ -0,0 +1,519 @@
...
+    if (this.is2D) {
+      return `matrix(${matrix.join(", ")})`;
+    }
+
+    return `matrix3d(${matrix.join(", ")})`;
+  }
+}
+
+module.exports.implementation = DOMMatrixReadOnlyImpl;
+const DOMMatrix = require("../generated/DOMMatrix");
```
TimothyGu: Seems to be unused?

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
```diff
@@ -0,0 +1,519 @@
...
+    });
+    return arrayFromDOMMatrixInit(baseMatrix);
+  }
+
+  _initialize2dMatrix(init) {
+    matrix2dArrayProperties.forEach((prop, idx) => {
+      this[prop] = init[idx];
+    });
+
+    this.m33 = 1;
```
TimothyGu: Add link to https://drafts.fxtf.org/geometry/#create-a-2d-matrix?

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
```diff
@@ -0,0 +1,519 @@
...
+
+  toFloat32Array() {
+    return new Float32Array(this._getCurrentMatrix());
+  }
+
+  toFloat64Array() {
+    return new Float64Array(this._getCurrentMatrix());
+  }
+
+  toJSON() {
```
TimothyGu: File a bug against https://github.com/jsdom/webidl2js to implement `[Default] toJSON()`, and link that here.

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
```diff
@@ -0,0 +1,519 @@
...
+      "m31", "m32", "m33", "m34",
+      "m41", "m42", "m43", "m44",
+      "is2D", "isIdentity"
+    ].reduce((domMatrixInit, prop) => {
+      domMatrixInit[prop] = this[prop];
+      return domMatrixInit;
+    }, {});
+  }
+
+  toString() {
```
TimothyGu: Link to https://drafts.fxtf.org/geometry/#dommatrixreadonly-stringification-behavior.

scripts/webidl/convert.js
KimiSon: No I've not tested anything

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
```diff
@@ -0,0 +1,519 @@
...
+    const matrix = this._getCurrentMatrix();
+
+    if (matrix.some(x =>
```
TimothyGu:
```suggestion
    if (!matrix.every(Number.isFinite))
```

Generally prefer the `Number.is`… variants as they type-check the input argument.

lib/jsdom/living/geometry/DOMPoint-impl.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+// TODO: Write tuwpt tests for DOMPoint
+// eslint-disable-next-line no-undef,no-unused-vars
```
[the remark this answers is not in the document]
dpchamps: No, I was going to remove the DOMPoint implementations from this pr actually. I implemented this thinking matrix wasn't going to be too big and then, well...

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
dpchamps: This is actually a helper for using `Array.from(DOMMatrix)` it makes writing some of the transform methods a lot cleaner

lib/jsdom/living/geometry/DOMPoint-impl.js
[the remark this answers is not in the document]
TimothyGu: That works for me.

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
dpchamps: That works. I approached this pretty cold from the jsdom code base, so I wanted to avoid anything that looked too "overly-clever". 

The getters that produce side-effects will have to remain though

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
TimothyGu: Ah. In that case, I'd just call this function `_to1DArray`, and call `matrix._to1DArray()` explicitly.

If you insist on using `[Symbol.iterator]`, then you could just use an ordinary (non-generator) function:

```js
[Symbol.iterator]() {
  return matrix3dArrayProperties.map(prop => this[prop]).values();
}
```

Yet another way that's still slightly simpler is

```js
* [Symbol.iterator]() {
  yield* matrix3dArrayProperties.map(prop => this[prop]);
}
```

lib/jsdom/living/helpers/geometry/matrixDomStringParser.js
```diff
@@ -0,0 +1,397 @@
...
+  matrix: (...params) => {
+    validateParameterLength(6, params, "matrix");
+    return params.map(paramConverter.number.bind(paramConverter));
+  },
+  translate3d: (...params) => {
+    validateParameterLength(3, params, "translate3d");
+    return [
+      paramConverter.lengthPercentage(params[0]),
+      paramConverter.lengthPercentage(params[1]),
+      paramConverter.lengthPercentage(params[2])
```
TimothyGu: This seems like it should be `.length(params[2])`.

lib/jsdom/living/helpers/geometry/matrixDomStringParser.js
```diff
@@ -0,0 +1,397 @@
...
+    return ANGLE_CONVERTERS[unit || "deg"](Number(length));
+  }
+};
+
+// https://drafts.csswg.org/css-transforms-1/#two-d-transform-functions
+// https://drafts.csswg.org/css-transforms-2/#three-d-transform-functions
+
+// Contains the definitions for valid parameters per transform function.
+// Preforms conversion of parameters to expected values.
+const transformFnParameterValidation = {
```
TimothyGu: Could you sort these in a way that's easy to find where each function comes from?

```js
const transformFnParameterValidation = {
  // https://drafts.csswg.org/css-transforms-1/#two-d-transform-functions
  matrix(...params) {},
  translate(...params) {},
  // ...

  // https://drafts.csswg.org/css-transforms-2/#three-d-transform-functions
  matrix3d(...params) {},
  translate3d(...params) {},
  // ...
};
```

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
dpchamps: No need for an iterator :) I'll change it

lib/jsdom/living/helpers/geometry/matrixDomStringParser.js
```diff
@@ -0,0 +1,397 @@
...
+      paramConverter.lengthPercentage(params[0]),
+      paramConverter.lengthPercentage(params[1]),
+      paramConverter.lengthPercentage(params[2])
+    ];
+  },
+  translate: (...params) => {
+    validateParameterLength([1, 2], params, "translate");
+    return [
+      paramConverter.lengthPercentage(params[0]),
+      paramConverter.lengthPercentage(params[1])
```
TimothyGu: By spec, this should default to zero if `params[1] === undefined`. However, it doesn't _look_ like lengthPercentage handles undefined(?)

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
dpchamps: No need for an iterator :) I'll change it to a plain old method

lib/jsdom/living/helpers/geometry/matrixDomStringParser.js
```diff
@@ -0,0 +1,397 @@
...
+    params.length === expectedLength;
+
+  if (!isValidLength) {
+    throw new TypeError(`Received invalid number of parameters for function ${fnName}`);
+  }
+}
+
+// https://drafts.csswg.org/css-transforms-1/#typedef-transform-function
+// https://drafts.csswg.org/css-transforms-2/#three-d-transform-functions
+const transformFunctionNames = [
```
TimothyGu: Looks like this could just be `Object.keys(transformFnParameterValidation)`?

lib/jsdom/living/helpers/geometry/matrixDomStringParser.js
```diff
@@ -0,0 +1,397 @@
...
+  "rotate",
+  "rotateX",
+  "rotateY",
+  "rotateZ",
+  "perspective",
+  "skew",
+  "skewX",
+  "skewY"
+];
+// https://www.w3.org/TR/css-values-4/#absolute-length
```
TimothyGu: https://drafts.csswg.org/css-values-4/#absolute-lengths

Also, let's put the CSS Values and Units stuff in a separate file, to ease reuse.

lib/jsdom/living/helpers/geometry/matrixDomStringParser.js
```diff
@@ -0,0 +1,397 @@
...
+  },
+  pt(x) {
+    return this.in(x) / 72;
+  },
+  px(x) {
+    return x;
+  }
+};
+
+// https://drafts.csswg.org/css-values-3/#angles
```
TimothyGu: https://drafts.csswg.org/css-values-4/#angles

lib/jsdom/living/helpers/geometry/matrixDomStringParser.js
```diff
@@ -0,0 +1,397 @@
...
+    }
+  },
+  _validate(rules, param) {
+    rules.forEach(matcher => {
+      if (!matcher(param)) {
+        throw new TypeError(`Failed to validate a parameter, ${rules} ${param}`);
+      }
+    });
+  },
+  _getUnitParam(paramString) {
```
TimothyGu: It would be more appropriate to call this "_getDimension", considering that "dimension" is [defined](https://drafts.csswg.org/css-values-4/#dimensions) as "number with a unit attached to it".

lib/jsdom/living/helpers/geometry/matrixDomStringParser.js
```diff
@@ -0,0 +1,397 @@
...
+  radixPoint: /\./,
+  comma: /,/,
+  commentOpen: /\/\*/,
+  commentClose: /\*\//
+};
+
+/**
+ * A very slim css-parser for extracting functions from a css dom string
+ */
+class CssFunctionParser {
```
TimothyGu: Please look over the spec for how to parse a CSS "component value" (https://drafts.csswg.org/css-syntax-3/#consume-component-value) and use that to implement this parser. Let's try to make it as spec-compliant as we can make it.

The syntax spec has two representations of a grammar construct: a railroad diagram, and a set of parsing algorithms. Always use the parsing algorithms: the railroad diagram is only for informative purposes.

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
```diff
@@ -0,0 +1,519 @@
...
+    if (matrixInit.length === 6) {
+      this._initialize2dMatrix(matrixInit);
+    } else if (matrixInit.length === 16) {
+      this._initialize3dMatrix(matrixInit);
+    } else {
+      throw new TypeError(`Invalid length for constructor init, expected 6 or 16, but received ${matrixInit.length}`);
+    }
+  }
+
+  parseDomString(domstring) {
```
TimothyGu: This looks to be private, i.e., not exposed to the web. Let's prefix the name with an underscore.

Additionally, this seems to implement https://drafts.fxtf.org/geometry/#parse-a-string-into-an-abstract-matrix. Link it, and rename the function to make it clear that is the case.

lib/jsdom/living/helpers/geometry/dommatrix-init.js
dpchamps: sure thing!

lib/jsdom/living/helpers/geometry/matrixDomStringParser.js
```diff
@@ -0,0 +1,397 @@
...
+  },
+  turn(x) {
+    return x * 360;
+  }
+};
+
+const paramConverter = {
+  _rules: {
+    number(x) {
+      return (SYMBOL.radixPoint.test(x) || !isNaN(x)) && x !== null;
```
TimothyGu: For one, this doesn't take into account `e`, like `1e3`. I recommend following https://drafts.csswg.org/css-syntax-3/#consume-numeric-token closely.

dpchamps: The parser will be closer to spec with the next commit, I just wanted something to get the ball rolling here. This is very much quick and dirty in it's current state

lib/jsdom/living/geometry/DOMMatrixReadOnly-impl.js
dpchamps: That's great, first I've seen the `Number.is` functions

PR #2666, opened 2019-09-18
feat: Add PointerEvent

### LINKED ISSUE #2527: Implement `PointerEvent`
### Basic info:

- **Node.js version:** 10.15.2 
- **jsdom version:** 13.2.0 

### Minimal reproduction case

```js
const dom = new JSDOM();
new dom.window.PointerEvent('pointerup', {button: 0});
```
This throws `TypeError: dom.window.PointerEvent is not a constructor`

image

Standard: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events

eps1lon:
Closes #2527

As far as I understood it the full PointerEvents spec is not part of the living standard (yet?) but [`pointerup` is](https://html.spec.whatwg.org/multipage/interaction.html#activation). It links back to the W3C recommendation in which `PointerEvent` is defined which must be fired by `pointerup`. Therefore (though it does feel a bit like cheating) [`PointerEvent`](https://www.w3.org/TR/pointerevents/#pointerevent-interface) is part of the living standard.


TimothyGu:
It doesn't really matter which standard (HTML or otherwise) a particular interface is defined. Anything goes into `living/` these days :)

lib/jsdom/living/events/PointerEvent.webidl
```diff
@@ -0,0 +1,27 @@
+[Constructor(DOMString type, optional PointerEventInit eventInitDict), Exposed=Window]
```
TimothyGu: Could you add a comment with a link to https://w3c.github.io/pointerevents/#pointerevent-interface? Also, we generally copy the IDL over verbatim, so no need to collapse the whitespace :)

eps1lon:
@TimothyGu Thank you for looking into this.

I'm stuck with the tests though. If I copy the most basic test from wpt into to-upstream it passes. However, the same test fails if it is run from the git submodule with a non-descriptive `Error: unimplemented`. Should I update the submodule? If so to what version and should this be done in a separate PR?

lib/jsdom/living/events/PointerEvent.webidl
```diff
@@ -0,0 +1,28 @@
+// https://www.w3.org/TR/pointerevents/#pointerevent-interface
```
TimothyGu: nit: We usually always use the latest editor's draft, so https://w3c.github.io/pointerevents/#pointerevent-interface.

eps1lon:
```suggestion
// https://w3c.github.io/pointerevents/#pointerevent-interface
```

lib/jsdom/living/events/PointerEvent.webidl
eps1lon: Done

TimothyGu:
> I'm stuck with the tests though. If I copy the most basic test from wpt into to-upstream it passes. However, the same test fails if it is run from the git submodule with a non-descriptive `Error: unimplemented`. Should I update the submodule? If so to what version and should this be done in a separate PR?

Hmm. Check out https://github.com/jsdom/jsdom/blob/master/Contributing.md#running-the-tests for running tests. If that doesn't help, I'll create a separate PR that updates our WPT submodule, merge that, and then look at this.

lib/jsdom/living/events/PointerEvent.webidl
eps1lon: I'm still learning about the different spec versions. I assumed latest editor draft might also include new features like an alpha release. But I guess it only includes editorial changes which means it's perfectly safe to use as a reference for implementations?

eps1lon:
> Hmm. Check out https://github.com/jsdom/jsdom/blob/master/Contributing.md#running-the-tests [](/jsdom/jsdom/blob/HEAD@{2019-09-24T17:11:30Z}/Contributing.md#running-the-tests) for running tests. If that doesn't help, I'll create a separate PR that updates our WPT submodule, merge that, and then look at this.

The CI build has the same error as locally: https://travis-ci.org/jsdom/jsdom/jobs/588992487#L5597

The other error is the new upstream test but that shouldn't be merged anyway. I just added it to illustrate that one is working locally while the other one doesn't.

lib/jsdom/living/events/PointerEvent.webidl
TimothyGu: @eps1lon The latest editor's draft is the document that implementors use, as they are the latest versions of everything. For the same reason why we use the Living Standards for WHATWG specs, we use editor's draft for W3C specs (they are generally considered to be quite equivalent).

TimothyGu:
Thanks. I'm working on updating the WPT repo.

TimothyGu:
Okay. See how things work out after rebasing on top of #2671.

eps1lon:
Every test isn't supposed to pass anyway. I'm going through them and see what fails and passes. 

I only intended to make the constructor available anyway. The specific behavior requires a lot more work I suppose.

domenic:
It can be a bit dangerous for us to half-implement features, as sites that use feature detection then get misleading signals. So let's be careful with strategies that involve things like adding stub constructors.

eps1lon:
> It can be a bit dangerous for us to half-implement features, as sites that use feature detection then get misleading signals. So let's be careful with strategies that involve things like adding stub constructors.

Agreed. First I need to get the tests running though.

I confirmed that `action_sequence` is rejecting with `unimplemented`. It seems like this is part of `testdriver-extra.js` but I have no idea how to tell the testdriver to include this file.

eps1lon:
Error message are just not helpful at all. It seems like many of the tests are async which hides any helpful error (either `unimplemented` in `action_sequence` or `testharness should not timeout`). 

I need to manually remove code and run again to check where the failure  occurs. This will take longer than expected.

eps1lon:
Ok I'm slowly getting the hang of it. Sorry for opening prematurely. Added a notice to the initial description that this should've been be a draft PR.

test/web-platform-tests/to-run.yaml
```diff
@@ -843,6 +843,97 @@ DIR: html/webappapis/timers
...
+extension/pointerevent_touch-action-pan-right-css_touch.html: [fail-slow, touch-action not implemented]
+extension/pointerevent_touch-action-pan-up-css_touch.html: [fail-slow, touch-action not implemented]
+extension/pointerevent_touch-action-verification.html: [fail-slow, touch-action not implemented]
+
+idlharness.window.html: [fail, fetch is not defined]
+inheritance.html: [fail, TBD]
+
+parsing/touch-action-computed.html: [fail, touch-action not implemented]
+#parsing/touch-action-invalid.html: [pass]
+parsing/touch-action-valid.html: [fail, touch-action not implemented]
```
TimothyGu:
```yaml
parsing/touch-action*: [timeout, touch-action not implemented]
```

test/web-platform-tests/to-run.yaml
```diff
@@ -843,6 +843,97 @@ DIR: html/webappapis/timers
...
+extension/idlharness.window.html: [fail, fetch is not defined]
+extension/pointerevent_coalesced_events_attributes.html: [fail-slow, getCoalescedEvents not implemented]
+extension/pointerevent_constructor.html: [fail, getCoalescedEvents not implemented]
+extension/pointerevent_getCoalescedEvents_when_pointerlocked.html: [fail-slow, unimplemented]
+extension/pointerevent_pointerrawupdate.html: [fail-slow, pointerlock not implemented]
+extension/pointerevent_touch-action-pan-down-css_touch.html: [fail-slow, touch-action not implemented]
+extension/pointerevent_touch-action-pan-left-css_touch.html: [fail-slow, touch-action not implemented]
+extension/pointerevent_touch-action-pan-right-css_touch.html: [fail-slow, touch-action not implemented]
+extension/pointerevent_touch-action-pan-up-css_touch.html: [fail-slow, touch-action not implemented]
+extension/pointerevent_touch-action-verification.html: [fail-slow, touch-action not implemented]
```
TimothyGu: Group `extension/pointerevent_touch-action*` together.

test/web-platform-tests/to-run.yaml
```diff
@@ -843,6 +843,97 @@ DIR: html/webappapis/timers
...
+pointerevent_disabled_form_control.html: [fail-slow, TDB]
+pointerevent_element_haspointercapture.html: [fail-slow, TDB]
+pointerevent_lostpointercapture_for_disconnected_node.html: [fail-slow, TDB]
+pointerevent_lostpointercapture_is_first.html: [fail-slow, TDB]
+pointerevent_mouse_capture_change_hover.html: [fail-slow, TDB]
+pointerevent_on_event_handlers.html: [fail-slow, TDB]
+pointerevent_pointercancel_touch.html: [fail-slow, TDB]
+pointerevent_pointercapture_in_frame.html: [fail-slow, TDB]
+pointerevent_pointerenter_does_not_bubble.html: [fail-slow, TDB]
+pointerevent_pointerleave_after_pointercancel_touch.html: [fail-slow, TDB]
```
TimothyGu: You can group each of these by events:

```yaml
pointerevent_pointerleave*: [timeout, pointerleave not implemented]
```

lib/jsdom/living/events/PointerEvent.webidl
TimothyGu: This still doesn't seem to be fixed?

TimothyGu:
Any test that uses testdriver would fail. You could use something like "uses testdriver" as the failure reason.

lib/jsdom/living/events/PointerEvent.webidl
eps1lon: Sorry I juggle different versions of it locally and it got lost. Is there an easier way to tell the test runner to only run on a subset?

eps1lon:
> Any test that uses testdriver would fail. You could use something like "uses testdriver" as the failure reason.

Which is basically all actual tests for the pointerevents. 

Do I need to implement the `action_sequence` for jsdom or how would I go about this?

eps1lon:
Having looked through Pointer Capture I now lean more towards not implementing this. To actually use jsdom for pointer events in a test one would need an extra testdriver for jsdom to properly trigger pointer events. 

For example we need to track active pointers and their states for pointer capture to work properly. We could do this by reading from the dispatched events but that would work differently from the browser. I could not get a pointer capture in google chrome to work because you can't just register new active pointers and you can't manipulate the active pointer state either.

It would make more sense to implement https://w3c.github.io/webdriver/#pointer-actions as well but I suspect this would require a lot more work.

TimothyGu:
Okay, that sounds good to me. Is this PR ready for a final review + merge then?

test/web-platform-tests/to-run.yaml
```diff
@@ -850,6 +850,63 @@ DIR: html/webappapis/timers
...
+pointerevent_pointerleave_descendants.html: [fail-slow, Uses testdriver]
+pointerevent_pointerleave_does_not_bubble.html: [fail-slow, Uses testdriver]
+pointerevent_pointermove.html: [fail-slow, Uses testdriver]
+pointerevent_pointermove_isprimary_same_as_pointerdown.html: [fail-slow, Uses testdriver]
+pointerevent_pointermove_on_chorded_mouse_button.html: [fail-slow, Uses testdriver]
+pointerevent_pointerout_after_pointercancel_touch.html: [fail-slow, Uses testdriver]
+pointerevent_pointerout_pen.html: [fail-slow, Uses testdriver]
+pointerevent_pointerout_received_once.html: [fail-slow, Uses testdriver]
+pointerevent_releasepointercapture_*: [fail-slow, Pointer Capture not implemented]
+pointerevent_root_computed_style.html: [fail, getComputedStyle does not support inheritance]
```
eps1lon: https://github.com/jsdom/jsdom/pull/2668 would enable implementing this since it includes  the ability to define initial values of CSS properties.

eps1lon:
> Okay, that sounds good to me. Is this PR ready for a final review + merge then?

Oh I meant not implementing any of this. Apart from Safari the most popular user agents implement the spec completely. It might be confusing if you rely on `typeof PointerEvents` in your code and assume that the current spec is implemented. This could create very weird test results in JSDOM. Later calling `hasPointerCapture` would throw (or calling it from a pointer event handler).

Just wanted to emphasize this potential issue since I believe this is what @domenic was referring to.

Other than that I think this is ready for review.

jljorgenson18:
Any updates on this? This seemed close to ready minus a single change.

eps1lon:
> Any updates on this? This seemed close to ready minus a single change.

In my opinion this shouldn't be merged without PointerCapture. To test the correct behavior of it we would need a jsdom specific implementation of the testdriver. While this can be considered "a single change" it requires a bit of work.

pat-son:
Anyone still working on this?

JoeBobMiles:
What is the status on this?

I just ran into the very issue that @domenic prophesied. I'm writing unit tests for a React component that uses PointerEvents pretty extensively and also uses JSDOM to manipulate the component. After investigating why I was getting a very specific test failure, related to retrieving event data from PointerEvent instances, the rabbit hole ends here.

To keep it short, the issue I'm seeing is that when I pass optional event data into the PointerEvent constructor (i.e. `new PointerEvent("pointerdown", { pointerType: "mouse" })`), my event handler tries to access `pointerType`. When it does so, it gets `undefined` instead of `"mouse"`. Trying this with various other properties I expected to be able to pass to a PointerEvent yielded the same result.

I assume this is related to the subbed constructor?

And if it is, then could anyone explain why this issue doesn't happen in Vue? (I got curious and played with Vue to see if it ran into this issue. I don't pretend to know how Vue's test utils work, but they depend on JSDOM and don't seem to suffer from this issue like the DOM test library used by React does.)

itaylor:
I was able to work around the issue @JoeBobMiles describes with React DOM test library.
https://github.com/testing-library/dom-testing-library/issues/558#issuecomment-690951258

eps1lon:
> What is the status on this?

The issue is that we need some implementation to test pointer capture. So far JSDOM always implemented features not single interfaces. Otherwise you might rely on `Element.setPointerCapture()` being available by checking if `PointerEvent` is defined. However, `setPointerCapture` requires a bit more work since all the automated web-platform-test require a test driver. 

So far implementing `PointerEvent` wouldn't allow much with automated testing. If you just want to make sure your event handlers are called you don't need `PointerEvent`. For actual pointer lock testing you would need manual testing anyway.

itaylor:
@eps1lon, while I agree with the goal of JSDOM shipping complete features, I respectfully disagree with the assessment that just implementing `PointerEvent` wouldn't allow for much testing. 

I think the majority of people using PointerEvents are basically removing code that used to use MouseEvents and replacing with PointerEvents.  In these cases, there's really no need for `setPointerCapture`.  setPointerCapture is extremely useful when there are interactions between multiple elements using PointerEvents, but there are plenty of times where it's not needed.  Also, the assertion that you can already test event handlers are called without `PointerEvent` being implemented is mistaken, at least for React users.  If I fire a regular `Event` as a `pointerdown` it will not have the properties that are specific to PointerEvent, like `clientX`, `clientY`, `pointerType`, etc, as they are stripped from the object in the Event constructor see: https://github.com/testing-library/dom-testing-library/issues/558, making it impossible to test your PointerEvent code.

As a compromise, is it possible to put the incomplete PointerEvent support from this ticket behind some sort of flag or something that would allow those of us who need it to enable it without enabling it for all use cases?

eps1lon:
>  Also, the assertion that you can already test event handlers are called without PointerEvent being implemented is mistaken, at least for React users.

You can definitely do that: https://codesandbox.io/s/vibrant-lake-np49v?file=/src/Component.test.js

> As a compromise, is it possible to put the incomplete PointerEvent support from this ticket behind some sort of flag or something that would allow those of us who need it to enable it without enabling it for all use cases?

You can write your own polyfill for JSDOM. This should work as a starting point:
```js
class PointerEvent extends MouseEvent {
  constructor(type, { pointerType = "", ...MouseEventInit } = {}) {
    super(type, MouseEventInit);

    this.pointerType = pointerType;
    // continue with https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/PointerEvent#Arguments
  }
};
```

I think we should respect the decision of the maintainers to not implement partial features. At the end of the day they'll have to deal with issues from confused people about `PointerEvent` being defined but `setPointerCapture` not working.

jljorgenson18:
If anyone needs it, this is how we polyfilled it (with Typescript)

```ts
if (!global.PointerEvent) {
  class PointerEvent extends MouseEvent {
    public height?: number;
    public isPrimary?: boolean;
    public pointerId?: number;
    public pointerType?: string;
    public pressure?: number;
    public tangentialPressure?: number;
    public tiltX?: number;
    public tiltY?: number;
    public twist?: number;
    public width?: number;

    constructor(type: string, params: PointerEventInit = {}) {
      super(type, params);
      this.pointerId = params.pointerId;
      this.width = params.width;
      this.height = params.height;
      this.pressure = params.pressure;
      this.tangentialPressure = params.tangentialPressure;
      this.tiltX = params.tiltX;
      this.tiltY = params.tiltY;
      this.pointerType = params.pointerType;
      this.isPrimary = params.isPrimary;
    }
  }
  global.PointerEvent = PointerEvent as any;
}

```

eps1lon:
@jljorgenson18 You should mind the default values for the properties exclusive to pointer events. You can find them in https://w3c.github.io/pointerevents/#pointerevent-interface

jljorgenson18:
@eps1lon The default values aren't relevant to us with this case personally because we are recreating the entire event object. I'd appreciate it if you didn't downvote a comment that was just sharing our solution to this problem.

joebobmiles:
> The issue is that we need some implementation to test pointer capture.

@eps1lon What did you mean by "some implementation"? By implementation, are you referring to something inside jsdom? Or are you alluding to the test drivers I've seen mentioned in this thread a few times? I've paged through this thread a couple of times in a past few days trying to understand the hold up and I still don't know what the block is. Would gladly help resolve it if I knew what it was.

I realize it's been over six months since you wrote that, but I've recently returned to this thread due to still looking for a way to get PointerEvents, and in particular, pointer capture, working in jsdom. I've tried implementing a library that provided some monkey patches, but after taking a deeper look into pointer capture and how jsdom implements event dispatch, I realized that was a fools errand.

eps1lon:
> By implementation, are you referring to something inside jsdom? Or are you alluding to the test drivers I've seen mentioned in this thread a few times?

To test drivers. It's been a while so I'm not that familiar with the relevant tests anymore. I suggest you look at the web-platform-tests relating to pointer events to get an overview what web drivers would need to be implemented. 

I think I already did one pass but it would be nice to check again. Seems like every test added with "Uses testdriver" would need a testdriver implementation e.g. https://github.com/jsdom/jsdom/pull/2666/files#diff-0e843db2255ba14478947e3d59763ad2b09bb8abe6e30bdebab0ea750487eaeeR874

domenic:
PointerEvent was added in https://github.com/jsdom/jsdom/pull/3862.

PR #2668, opened 2019-09-19
feat(window): getComputedStyle inherits visibility

### LINKED ISSUE #2616: getComputedStyle not inheriting vsibility
### Basic info:

- **Node.js version:** 8.16
- **jsdom version:** 15.1

### Minimal reproduction case
```js
const { JSDOM } = jsdom;

const { window } = new JSDOM(`<!DOCTYPE html>
<html>
    <body>
        <div id="parent" style="visibility: hidden;">
            <div id="child">Hello, Dave!</div>
        </div>
    </body>
</html>`);

const parent = window.document.querySelector("#parent");
const child = window.document.querySelector("#child");

console.log(window.getComputedStyle(parent).visibility);
console.log(window.getComputedStyle(child).visibility);
```

-- https://runkit.com/eps1lon/5d0b73d5538b0a001a42239e

Prints:
```
"hidden"
""
```

https://jsbin.com/hinuhuxubi/1/edit?html,js,console

Prints:
```
"hidden"
"hidden"
```

eps1lon:
Closes #2616 

Missing behavior:
* https://drafts.csswg.org/css-cascade-4/#cascading
* https://drafts.csswg.org/css-cascade-4/#defaulting-keywords

I would recommend reviewing the first commit separately since it only includes refactoring to make the later inheritance implementation easier.

This lays the ground work for `getComputedStyle` respecting style inheritance as well as initial values. To fully implement [`getComputedStyle`](https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle) we need to implement proper specificity (code comment has a remark about this) and port https://www.w3.org/TR/CSS2/propidx.html into code.

I'd like to follow-up with implementing specificity. I'm mostly interested in visibility since it affects accessibility (`display` isn't inherited anyway and there's no layout in jsdom so `offsetParent` etc is not possible). For work on accessibility queries knowing if an element is rendered is paramount. `getComputedStyle` not supporting inheritance is a major blocker in that regard. If other work has already started to support visibility I'm happy to help.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -504,46 +504,57 @@ function Window(options) {
     const { forEach, indexOf } = Array.prototype;
     const { style } = elt;
 
-    function setPropertiesFromRule(rule) {
+    function matches(rule, element) {
       if (!rule.selectorText) {
-        return;
+        return false;
       }
 
       const cssSelectorSplitRe = /((?:[^,"']|"[^"]*"|'[^']*')+)/;
       const selectors = rule.selectorText.split(cssSelectorSplitRe);
-      let matched = false;
 
       for (const selectorText of selectors) {
-        if (selectorText !== "" && selectorText !== "," && !matched && matchesDontThrow(elt, selectorText)) {
-          matched = true;
-          forEach.call(rule.style, property => {
-            declaration.setProperty(
-              property,
-              rule.style.getPropertyValue(property),
-              rule.style.getPropertyPriority(property)
-            );
-          });
+        if (selectorText !== "" && selectorText !== "," && matchesDontThrow(element, selectorText)) {
+          return true;
         }
       }
+
+      return false;
     }
 
-    function readStylesFromStyleSheet(sheet) {
-      forEach.call(sheet.cssRules, rule => {
-        if (rule.media) {
-          if (indexOf.call(rule.media, "screen") !== -1) {
-            forEach.call(rule.cssRules, setPropertiesFromRule);
-          }
-        } else {
-          setPropertiesFromRule(rule);
-        }
-      });
+    function setPropertiesFromRule(rule) {
+      if (matches(rule, elt)) {
+        forEach.call(rule.style, property => {
+          declaration.setProperty(
+            property,
+            rule.style.getPropertyValue(property),
+            rule.style.getPropertyPriority(property)
+          );
+        });
+      }
     }
 
     if (!parsedDefaultStyleSheet) {
       parsedDefaultStyleSheet = cssom.parse(defaultStyleSheet);
     }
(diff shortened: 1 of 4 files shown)


lib/jsdom/browser/Window.js
```diff
@@ -556,6 +556,97 @@ function Window(options) {
 
     forEachSheetRuleOfElement(elt, setPropertiesFromRule);
 
+    /**
+     * properties for which `getResolvedValue` is implemented which is less than
+     * every supported property according to step 5
+     * @type Record<string, { inherited: boolean; initial: unknown; computedValue: 'as-specified' }>
```
domenic: Can you remove the jsdoc/type annotations and just have this be a normal `//` comment? Here and below.

lib/jsdom/browser/Window.js
```diff
@@ -556,6 +556,97 @@ function Window(options) {
...
+     */
+    const properties = {
```
domenic: I think this could move to the top of the file, which would be cleaner IMO, especially if the goal is to expand it.

lib/jsdom/browser/Window.js
```diff
@@ -556,6 +556,97 @@ function Window(options) {
...
+      visibility: {
+        inherited: true,
+        initial: "visible",
+        computedValue: "as-specified"
+      }
+    };
+
+    /**
+     * Naive implementation of https://www.w3.org/TR/CSS2/cascade.html#cascade
```
domenic: Can you change all the spec references to https://drafts.csswg.org/css2/ ? And make sure that the algorithms haven't changed in important ways :).

lib/jsdom/browser/Window.js
domenic: Actually, maybe it'd be best to move this, and most (all?) of the functions here, into a separate file? E.g. jsdom/lib/jsdom/living/helpers/style-rules.js

lib/jsdom/browser/Window.js
TimothyGu: This should actually now be https://drafts.csswg.org/css-cascade/#cascading.

> This module replaces and extends the rules for assigning property values, cascading, and inheritance defined in [[CSS2]](https://www.w3.org/TR/CSS2/) chapter 6.
>
> – https://drafts.csswg.org/css-cascade/#placement

eps1lon: Couldn't detect changes. Only parts that are not implemented yet.

lib/jsdom/browser/Window.js
eps1lon: done

lib/jsdom/browser/Window.js
eps1lon: done

PR #2669, opened 2019-09-19
feat: Rename `HTMLOrSVGElement` to `HTMLOrForeignElement`

ExE-Boss:
See https://github.com/whatwg/html/issues/4702 for details.

---

This has already been implemented in all major browsers:
- Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=835571
- Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1571487
- Safari: https://bugs.webkit.org/show_bug.cgi?id=200470


domenic:
Not doing this because we have no plans to implement MathML and it just causes code churn for no gain.

ExE-Boss:
It does align the mixin’s name with the upcoming spec changes.

PR #2672, opened 2019-09-26
Update dependencies

TimothyGu:
This updates all dependencies to their latest versions, and gets rid of all `npm`/`yarn audit` errors. Additionally, the Travis CI configuration is updated to use headless Chrome installed on the test machine rather than using Sauce Labs (which is not run on pull requests and could be flaky).


domenic:
Remember that browser tests don't run on Travis CI on PRs, only on master. To fix that, we really need to move to a modern mechanism for running the tests in browsers, e.g. a Docker container with Chrome installed on Travis or something.

TimothyGu:
Oh yeah, you're right. sigh…

TimothyGu:
I've updated the Karma setup to be somewhat more modern. However, it's nothing more than a quick fix, and the longer-term solution is indeed to switch to a new setup.

.mocharc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "exit": true
```
domenic: Just curious, what does this do?

package.json
```diff
@@ -58,27 +58,26 @@
...
     "eslint-plugin-jsdom-internal": "link:./scripts/eslint-plugin",
     "js-yaml": "^3.13.1",
-    "karma": "^1.7.1",
-    "karma-browserify": "^5.3.0",
-    "karma-chrome-launcher": "^2.2.0",
+    "karma": "^4.3.0",
+    "karma-browserify": "^6.1.0",
+    "karma-chrome-launcher": "^3.1.0",
     "karma-mocha": "^1.3.0",
-    "karma-mocha-webworker": "^1.3.0",
```
domenic: Do we still run the tests in a web worker?

.mocharc.json
TimothyGu: See https://mochajs.org/#-exit.

package.json
TimothyGu: On master, yes. But here I removed it.

TimothyGu:
Still doing some experimentation with perhaps removing Sauce Labs (and always use latest Chrome).

TimothyGu:
Great, things seem to be working pretty well. I guess this is now ready for review :)

domenic:
I'm worried about losing coverage for whether we work well in web workers, i.e. making sure we don't accidentally introduce dependencies on Window-only globals. Do you have any ideas there?

TimothyGu:
Figured out what was broken with web workers and fixed it.

.mocharc.json
domenic: We may want to file an issue for figuring out what in our tests is causing a hang?

.travis.yml
```diff
@@ -3,13 +3,14 @@ node_js:
...
   - 10
   - stable
-sudo: false
 
 git:
   submodules: false
 
 install:
   - yarn --frozen-lockfile
+  - yarn pretest
```
domenic: Maybe comment why this is necessary. (Because the next line needs the submodules synced.)

Or I wonder if there's a specific Travis thing for including submodules in the clone.

.travis.yml
```diff
@@ -19,92 +20,22 @@ branches:
...
+      addons:
+        apt:
+          packages:
+            - build-essential
+            - libcairo2-dev
+            - libpango1.0-dev
+            - libjpeg-dev
+            - libgif-dev
+            - librsvg2-dev
+      script: "yarn add canvas && yarn test --retries 1"
```
domenic: Nice cleanup with the packages and no longer requiring the CXX environment variable

.mocharc.json
TimothyGu: Filed #2673.

.travis.yml
TimothyGu: Per df2e6189b, the built-in submodule support was previously insufficient for our use cases as wpt had its own submodules. wpt no longer seems to have any submodules. I'm going to experiment with using the default support a bit.

TimothyGu: Looks like the built-in support works.

PR #2676, opened 2019-10-01
Remove mocha exit option

### LINKED ISSUE #2673: Investigate what keeps Mocha from exiting
#2672 adds a .mocharc.json file that sets `exit` to `true`. This is because without it, the tests would hang. See https://mochajs.org/#-exit for a more detailed description of the option.

Investigate why the tests are hanging, and remove the option ideally.

TimothyGu:
Fixes #2673.


Sebmaster:
Shouldn't we close those properly in the after()s instead?

TimothyGu:
That does sound like a better idea perhaps. Let me see if that's doable.

domenic:
I did some local experimenting with a version that actually shuts down the servers instead of using unref, and it's pretty involved. I think we can stick with unref().

domenic:
Hmm, however, waiting for the HTTP servers to shut themselves down causes a 6 second hang at the end, which is not great.

domenic:
OK, I re-did everything. Let's see if it passes on CI...

test/web-platform-tests/run-wpts.js
```diff
@@ -39,13 +39,19 @@ const minimatchers = new Map();
...
 let wptServerURL;
+let serverProcess;
 const runSingleWPT = require("./run-single-wpt.js")(() => wptServerURL);
 before({ timeout: 30 * 1000 }, () => {
-  return startWPTServer({ toUpstream: false }).then(url => {
-    wptServerURL = url;
+  return startWPTServer({ toUpstream: false }).then(({ urls, subprocess }) => {
+    wptServerURL = urls[0];
+    serverProcess = subprocess;
   });
```
TimothyGu: Make this an async function?

PR #2681, opened 2019-10-09
dispatchEvent now takes into account if an element has been disabled

### LINKED ISSUE #2665: dispatching click event executes click handler for disabled elements
### Basic info:

- **Node.js version:** 10.16.3
- **jsdom version:**  15.1.1

### Minimal reproduction case

```js

<!DOCTYPE html>
<html>
  <body><button disabled></button></body>
</html>
`);

const button = window.document.querySelector('button');
button.addEventListener('click', () => console.log('click'));

button.click();
button.dispatchEvent(new window.MouseEvent('click'))
```
https://runkit.com/eps1lon/5d80dae1e7cc46001351ca95

https://jsbin.com/kebukirute/edit?html,js,console

In the browser no click handler is called. In JSDOM the click handler is called on dispatchEvent but not click().

brendo:
Fixes #2665

### DIFF
--- a/lib/jsdom/living/nodes/HTMLElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLElement-impl.js
@@ -58,6 +58,14 @@ class HTMLElementImpl extends ElementImpl {
     }
   }
 
+  dispatchEvent(eventImpl) {
+    if (isDisabled(this)) {
+      return;
+    }
+
+    return super.dispatchEvent(eventImpl);
+  }
+
   click() {
     // https://html.spec.whatwg.org/multipage/interaction.html#dom-click
     // https://html.spec.whatwg.org/multipage/webappapis.html#fire-a-synthetic-mouse-event
--- a/test/web-platform-tests/to-upstream/html/editing/activation/click-bail-on-disabled.html
+++ b/test/web-platform-tests/to-upstream/html/editing/activation/click-bail-on-disabled.html
@@ -30,6 +30,16 @@ async_test(t => {
 
 async_test(t => {
 
+  const element = document.querySelector("#el1");
+  element.addEventListener("click", t.unreached_func("click event should never happen"));
+
+  element.dispatchEvent(new MouseEvent('click'));
+  t.step_timeout(() => t.done(), 500);
+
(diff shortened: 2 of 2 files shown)


lib/jsdom/living/nodes/HTMLElement-impl.js
```diff
@@ -58,6 +58,14 @@ class HTMLElementImpl extends ElementImpl {
     }
   }
 
+  dispatchEvent(eventImpl) {
+    if (isDisabled(this)) {
+      return;
+    }
+
+    return super.dispatchEvent(eventImpl);
```
TimothyGu: This is not correct. https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#enabling-and-disabling-form-controls:-the-disabled-attribute:concept-fe-disabled says:

> A form control that is disabled must prevent any `click` events that are queued on the user interaction task source from being dispatched on the element.

So any special processing must only apply to `click` events.

Additionally, some places call `_dispatch` directly: https://github.com/jsdom/jsdom/blob/e3744f5ceffbf06a83b8e0c11b1641f4e229ea18/lib/jsdom/living/helpers/events.js#L17 It might be better to override `_dispatch` instead of the more public `dispatchEvent`.

domenic: That can't be the right part of the spec, because the user interaction task source is never involved in jsdom, I don't think.

The click() disabled handling is in the spec at https://html.spec.whatwg.org/multipage/interaction.html#activation:concept-fe-disabled.

For dispatchEvent, I cannot find any support in the spec for this. Furthmore, running the test in #2665 in Firefox shows that it matches jsdom.

So I don't think we should change this in jsdom, but there is a spec interop issue, apparently.

domenic: This is https://github.com/whatwg/html/issues/2368

domenic:
Since jsdom is behaving according to the spec here, we won't change anything.

brendo:
Thanks for the replies! I meant to come back to this but forget. I'm glad it kicked off some discussion though 👍

PR #2686, opened 2019-10-15
Upgrade to saxes 4.0.2.

### LINKED ISSUE #2677: upgrading to saxes 4.0.0
I've just released version 4.0.0-rc1 of saxes. I can submit a PR but there are a few things that require discussion due to breaking changes. Staying on the 3.x version is not really an option in the long term. There are bugs with end-of-line (EOL) handling in all versions prior to 4.0.0-rc1 that have been fixed in 4.0.0-rc1. It is likely that most people never saw those bugs because most folks have files that use ``\n`` as the EOL character.

The new release has a few breaking changes:

1. There's a remote possibility that someone came to rely on the old incorrect way saxes handled EOL characters other than ``\n``. Unfortunately, the upgrade will be a breaking change *for them*. Tough luck though because the old behavior was just flat out incorrect. It went against the XML specification and was inconsistent with how Firefox and Chrome behave.

2. In previous versions of saxes, when the configuration option ``fileName`` was left unset, then saxes would report errors like ``undefined:1:13: foo`` because the JavaScript value of the option would be ``undefined`` and JavaScript would convert it to a string. I knew this, and was fine with it, but someone filed a bug report and I agreed that it should be changed so that if there's no ``fileName`` specified, the error is reported without a file name ``1:13: foo``.

   I can easily modify jsdom to preserve the old behavior by setting ``fileName`` to the string ``"undefined"`` in those cases where there's no other available file name to pass. I don't know whether this is the desired solution.

3. saxes 4.0.0 adds support for XML 1.1, but last time I checked the whatwg specs there was no mention of XML 1.1, and in practice both Firefox and Chrome support only XML 1.0. However, Firefox and Chrome behave differently when passed a document with an XML declaration that specify version 1.1:

   * Chrome just ignores the version number in the XML declaration. So a document which declares itself to be XML 1.1 but is otherwise well-formed as an XML 1.0 document will parse without error.

   * Firefox generates a parsing error. So a document which declares itself to be XML 1.1 will not parse without error, even if it would otherwise be well-formed XML 1.0.

   There is a note in the XML specification 1.0 edition 5 under the [`XMLDecl` production](https://www.w3.org/TR/xml/#NT-XMLDecl) that says:

   > When an XML 1.0 processor encounters a document that specifies a 1.x version number other than '1.0', it will process it as a 1.0 document. This means that an XML 1.0 processor will accept 1.x documents provided they do not use any non-1.0 features.

   This is what Chrome does. Unfortunately, the authors of the specification did not use any of the consecrated words MUST, MUST NOT, etc. They say "will" and the word is unemphasized. (I find no equivalent note in the XML 1.1 specification.)

   At any rate, saxes can be used to follow either behavior. Following what Chrome does would be consistent with how jsdom has used saxes in the past. Still, if it turns out that's undesirable, now is the time to argue for a change.

lddubeau:
Closes #2677

### DIFF
--- a/lib/jsdom/browser/parser/xml.js
+++ b/lib/jsdom/browser/parser/xml.js
@@ -47,7 +47,15 @@ function isHTMLTemplateElement(element) {
 
 
 function createParser(rootNode, globalObject, saxesOptions) {
-  const parser = new SaxesParser(saxesOptions);
+  const parser = new SaxesParser({
+    ...saxesOptions,
+    // Browsers always have namespace support.
+    xmlns: true,
+    // We force the parser to treat all documents (even documents declaring themselves to be XML 1.1 documents) as XML
+    // 1.0 documents. See https://github.com/jsdom/jsdom/issues/2677 for a discussion of the stakes.
+    defaultXMLVersion: "1.0",
+    forceXMLVersion: true
+  });
   const openStack = [rootNode];
 
   function getOwnerDocument() {
@@ -164,7 +172,6 @@ function parseFragment(markup, contextElement) {
   // Only parseFragment needs resolvePrefix per the saxes documentation:
   // https://github.com/lddubeau/saxes#parsing-xml-fragments
   const parser = createParser(fragment, _globalObject, {
-    xmlns: true,
     fragment: true,
     resolvePrefix(prefix) {
       // saxes wants undefined as the return value if the prefix is not defined, not null.
@@ -181,7 +188,6 @@ function parseIntoDocument(markup, ownerDocument) {
   const { _globalObject } = ownerDocument;
 
   const parser = createParser(ownerDocument, _globalObject, {
-    xmlns: true,
     fileName: ownerDocument.location && ownerDocument.location.href
   });
 
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,7 +5,7 @@
     "es6": true
   },
   "parserOptions": {
-    "ecmaVersion": 6
+    "ecmaVersion": 9
   },
   "plugins": [
     "jsdom-internal"
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "pn": "^1.1.0",
     "request": "^2.88.0",
     "request-promise-native": "^1.0.7",
-    "saxes": "^3.1.9",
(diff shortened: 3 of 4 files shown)


lib/jsdom/browser/parser/xml.js
```diff
@@ -51,7 +51,22 @@ function appendChild(parent, child) {
 }
 
 function createParser(rootNode, ownerDocument, saxesOptions) {
-  const parser = new SaxesParser(saxesOptions);
+  const parser = new SaxesParser({
+    ...saxesOptions,
```
TimothyGu: ESLint is complaining about this. Could you modify https://github.com/jsdom/jsdom/blob/b4bb1c5248e05762aa3c4c14faa71ef3adf10657/.eslintrc.json#L8 to say 2018 (or 9) instead? We now require Node.js v8.x so using it should be fine.

lib/jsdom/browser/parser/xml.js
```diff
@@ -51,7 +51,22 @@ function appendChild(parent, child) {
...
+    //
+    // There are no situations under which it is legitimate to set the following
+    // parameters to any other value than the ones we set here.
```
TimothyGu: Could we be a bit clearer about what this means? Like what this option does, and why it should always be true.

domenic: Ping @lddubeau for this comment in particular.

lib/jsdom/browser/parser/xml.js
lddubeau: Done.

lib/jsdom/browser/parser/xml.js
lddubeau: Sorry for the delay. I'm spread thin these days. 

It is not just the one option that must always be true but all the following parameters that must always be set the way they are hard coded there. `xmlns: false` (which is the only other possible value for `xmlns`) would turn off namespace support and cannot be correct when parsing XML as a browser does because browsers don't parse XML without namespace support. The two other options must be set the way they are because browsers support only XML 1.0, not 1.1.

The comment is there to signify that it is not a mistake if these parameters are being hard-coded there because there's no legitimate scenario in jsdom under which they should be set differently.

Issue #2687, opened 2019-10-16
Navigator interface is not expose on Window


pmdartus:
### Basic info:

- **Node.js version:** v12.9.0
- **jsdom version:** v15.2.0

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");
const dom = new JSDOM('');
console.log(dom.window.Navigator); // Expected: Navigator() { ... }, Actual: undefined
```
https://runkit.com/pmdartus/5da6b38cfc8b010013f7ed47

### How does similar code behave in browsers?

https://jsfiddle.net/ab3zy6ek/

### Details

This can be solved by exporting the generated `Navigator` interface out of [lib/jsdom/living/index.js](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/index.js).

pmdartus:
Fixed with #2691

PR #2691, opened 2019-10-23
Constructor & prototype reform

### LINKED ISSUE #1453: All classes are shared between separate jsdom windows
defineProperty on DOM Interfaces seems to survive new page load (via jsdom.jdom()).

`Object.defineProperty(HTMLAnchorElement.prototype, "href" , {configurable : false});`

After new jsdom.jsdom() this should be true:
`
Object.getOwnPropertyDescriptor(HTMLAnchorElement.prototype, "href").configurable === true;`

maybe/probably related to [1207](https://github.com/tmpvar/jsdom/issues/1207).

pmdartus:
## Changes

This PR applies the [webdidl2js constructor & prototype reform](https://github.com/jsdom/webidl2js/pull/140) to jsdom.

Fixes #1453 #2687


package.json
```diff
@@ -82,7 +82,7 @@
     "st": "^1.2.2",
     "watchify": "^3.11.1",
     "wd": "^1.11.2",
-    "webidl2js": "^10.0.0"
+    "webidl2js": "https://github.com/pmdartus/webidl2js.git#pmdartus/constructor-reform-pt2"
```
pmdartus: **⚠️Use the released version of webidl2js before merging ⚠️**
Using the WIP version to run the test suite on the CI.

lib/jsdom/living/aborting/AbortController-impl.js
```diff
@@ -1,12 +1,19 @@
...
+
 class AbortControllerImpl {
-  constructor(args, privateData) {
-    this.signal = privateData.AbortSignal.createImpl([]);
+  constructor(globalObject) {
+    this._signal = AbortSignal.createImpl(globalObject, []);
+  }
+
+  get signal() {
+    return idlUtils.wrapperForImpl(this._signal);
```
domenic: This shouldn't be necessary, in theory; the generated wrapper should unwrap the signal.

lib/jsdom/browser/Window.js
```diff
@@ -38,39 +34,62 @@ const WindowEventHandlersImpl = require("../living/nodes/WindowEventHandlers-imp
...
+  windowInstance._initGlobalEvents();
+
+  windowInstance._globalObject = windowInstance;
+}
 
 // NOTE: per https://heycam.github.io/webidl/#Global, all properties on the Window object must be own-properties.
 // That is why we assign everything inside of the constructor, instead of using a shared prototype.
 // You can verify this in e.g. Firefox or Internet Explorer, which do a good job with Web IDL compliance.
-
 function Window(options) {
```
domenic: Hmm, the weirdness where this is named `Window` but the actual `Window` constructor is created on line 47 is strange... I'm not sure what the right architecture is. I guess we can try to straighten it out later, or just wait until we move Window to webidl2js.

lib/jsdom/living/interfaces.js
```diff
@@ -0,0 +1,212 @@
...
+const DOMException = require("domexception");
+const { URL, URLSearchParams } = require("whatwg-url");
+const XMLSerializer = require("w3c-xmlserializer/lib/XMLSerializer").interface;
+
+const registerElements = require("./register-elements");
+const style = require("../level2/style");
+const xpath = require("../level3/xpath");
+const nodeFilter = require("./node-filter");
+
+const GENERATED_INTERFACES = [
```
domenic: No need to SHOUTY_CASE this.

lib/jsdom/living/xhr-utils.js
```diff
@@ -113,8 +115,9 @@ function requestErrorSteps(xhr, event, exception) {
     properties.uploadComplete = true;
 
     if (properties.uploadListener) {
-      fireAnEvent(event, xhr.upload, ProgressEvent, { loaded: 0, total: 0, lengthComputable: false });
-      fireAnEvent("loadend", xhr.upload, ProgressEvent, { loaded: 0, total: 0, lengthComputable: false });
+      const uploadWrapper = idlUtils.tryImplForWrapper(xhr.upload);
```
domenic: The name seems wrong here; should it be uploadImpl? In general not sure why this changed; was it a bug you found?

lib/jsdom/living/interfaces.js
ExE-Boss: But isn’t it a common convention that constants are in `UPPER_SNAKE_CASE`?

lib/jsdom/browser/Window.js
pmdartus: I wanted to minimize the amount of code change made to `lib/jsdom/browser/Window.js`, a clean up is overdue. Adding support for `[Global]` and `[NamedConstructor]` would greatly help. Let's do that in another PR.

lib/jsdom/living/xhr-utils.js
pmdartus: Currently, `XMLHttpRequest` doesn't uses wedidl2js. All the properties including [`upload`](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/xmlhttprequest.js#L102) are wrappers and not impls. With this PR the `fireAnEvent` function expects an impl and not a wrapper, so I need to unwrap before the invocation.

I initially undertook the task of converting the `XMLHttpRequest` interface to leverage webidl2js, however it was too much change to fit into a single PR. Once this PR is merged, I need to open couple of issues to remove some of the hacks now that all the interfaces have access to the global object.

domenic: That makes sense, I'm just confused why the variable name is uploadWrapper if you're calling (try)ImplForWrapper, which returns an impl.

pmdartus:
Yes, you are right the plan is to migrate the dependant projects as we go. I already opened a PR on [domexception](https://github.com/jsdom/domexception/pull/3), I will try to update the other one soon.

Issue #2685, opened 2019-10-15
TypeError: client.abort is not a function


kaiyoma:
- **Node.js version:** 10.16.0
- **jsdom version:** 15.1.1

I've started running into this error lately while running the unit tests for my project:

```
● Test suite failed to run

  TypeError: client.abort is not a function

    at XMLHttpRequest.abort (node_modules/jest-environment-jsdom/node_modules/jsdom/lib/jsdom/living/xmlhttprequest.js:379:16)
    at JSDOMEnvironmentGlobal.teardown (node_modules/jest-environment-jsdom-global/environment.js:13:18)
```

Before I try to come up with a minimal repro, I wanted to post just this first. Is this a known issue? Seems to happen quite regularly on Windows 10, but not on macOS/Linux.

domenic:
Not enough information to reproduce.

PR #2694, opened 2019-10-24
Add missing abort (#2685)

pwolaq:
Apparently there is one path with missing implementation of `XMLHttpRequest.abort`, which may cause errors as reported in #2685


domenic:
Thanks, this sounds interesting. Can you add tests, per the Contributing.md?

domenic:
Closing for now, until we can get tests!

PR #2697, opened 2019-10-28
Fix fromURL API to not include hash fragment in fetch request

### LINKED ISSUE #2696: Incorrect behaviour when using fromURL with URL that contains hash
### Basic info:

- **Node.js version:** 11
- **jsdom version:** 15.2.0

### Minimal reproduction case

```js

async function run () {
  const dom = await JSDOM.fromURL('http://localhost:3000/#/test')
  const passed = dom.window.location.href === 'http://localhost:3000/#/test'
  console.log({ passed, got: dom.window.location.href })
}

run()
```

`fromURL` does:
1. Parse URL with `new URL`
2. Take `parsed.href` and make a request to the server
3. Concatenate href with `parsed.hash` and create JSDOM instance with that URL

Problem: `href` already contains `hash` so URL ends up having duplicated hash (`http://localhost:3000/#/test#/test`).

Responsible code is at:
https://github.com/jsdom/jsdom/blob/7148ebbb4333a357de343fdddb57c4fc0e2a5e58/lib/api.js#L111-L140

In a browser, `hash` is not passed to the server so I believe JSDOM should make sure that the URL it uses while making request does not contain hash and then concatenate that URL with `hash` from parsed URL.

rchl:
Since browsers don't include hash when making http requests to the
server, do the same here. parsedURL.href was including hash (fragment).

This also fixes duplicate hash as parsedURL.href was concatenated
with parsedURL.hash and passed to JSDOM constructor and that resulted
in duplicate hash in actual document.

Resolves #2696

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -110,8 +110,13 @@ class JSDOM {
 
   static fromURL(url, options = {}) {
     return Promise.resolve().then(() => {
+      // Remove the hash while sending this through the research loader fetch().
+      // It gets added back a few lines down when constructing the JSDOM object.
       const parsedURL = new URL(url);
-      url = parsedURL.origin + parsedURL.pathname + parsedURL.search;
+      const originalHash = parsedURL.hash;
+      parsedURL.hash = "";
+      url = parsedURL.href;
+
       options = normalizeFromURLOptions(options);
 
       const resourceLoader = resourcesToResourceLoader(options.resources);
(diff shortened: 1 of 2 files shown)


lib/api.js
```diff
@@ -111,7 +111,7 @@ class JSDOM {
   static fromURL(url, options = {}) {
     return Promise.resolve().then(() => {
       const parsedURL = new URL(url);
-      url = parsedURL.href;
+      url = parsedURL.origin + parsedURL.pathname + parsedURL.search;
```
domenic: This doesn't seem like quite the right fix; it'd be better to strip out the hash by setting it to null. I'll push a tweak that does that.

Thanks very much for the tests though; they're the key!

PR #2700, opened 2019-11-01
fix: Focusing focused dispatching FocusEvent

### LINKED ISSUE #2621: focus() on focused element should not fire focus event
### Basic info:

- **Node.js version:** 10.15.2
- **jsdom version:** 15.1.

### Minimal reproduction case

```js

const { window: { document } } = new JSDOM(`
<!DOCTYPE html>
<html>
  <head>
    <title>Parcel Sandbox</title>
    <meta charset="UTF-8" />
  </head>

  <body>
    <div id="app"><button>Hello, Dave</button></div>
  </body>
</html>
`);

function handleEvent(event) {
  console.log(`${event.type} event`, event.target);
}

button.addEventListener('focus', handleEvent);

button.focus();
button.focus();

```

https://repl.it/@eps1lonExists/SimpleUnfoldedSequence

Repeated focus does not fire focus event. It seems like this is matching 

> If new focus target is the currently focused area of a top-level browsing context, then return.

-- https://html.spec.whatwg.org/multipage/interaction.html#focusing-steps

which isn't implemented by jsdom.

https://codesandbox.io/s/repeated-imperative-focus-2gftr

### Misc

Would it be sufficient to check if `previous === this` and return if this is true after 

https://github.com/jsdom/jsdom/blob/529b9f31839d5c48d0352f239ccd9b53a229aee2/lib/jsdom/living/nodes/HTMLOrSVGElement-impl.js#L34

? Couldn't find an existing test in https://github.com/web-platform-tests/wpt/tree/master/html/interaction/focus so I would need to add a new one to the upstream folder too, right?

eps1lon:
Closes #2621 

Implements

> If new focus target is the currently focused area of a top-level browsing context, then return.

-- https://html.spec.whatwg.org/multipage/interaction.html#focusing-steps

### DIFF
--- a/test/web-platform-tests/to-upstream/html/interaction/focus-managment/focusing-focused-noop.html
+++ b/test/web-platform-tests/to-upstream/html/interaction/focus-managment/focusing-focused-noop.html
@@ -3,10 +3,7 @@
 <head>
   <meta charset="utf-8">
   <title>Focus events fire at correct targets in correct order in simple case</title>
-  <link rel="author" title="Chris Rebert" href="http://chrisrebert.com">
-  <link rel="help" href="https://html.spec.whatwg.org/#focus-update-steps">
-  <link rel="help" href="https://html.spec.whatwg.org/#focus-chain">
-  <meta name="flags" content="dom">
+  <link rel="help" href="https://html.spec.whatwg.org/#focusing-steps">
   <script src="/resources/testharness.js"></script>
   <script src="/resources/testharnessreport.js"></script>
 </head>
--- a/test/web-platform-tests/to-upstream/html/interaction/focus-managment/focusing-focused-noop.html
+++ b/test/web-platform-tests/to-upstream/html/interaction/focus-managment/focusing-focused-noop.html
@@ -1,0 +1,43 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


test/web-platform-tests/to-upstream/html/interaction/focus-managment/focusing-focused-noop.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>Focus events fire at correct targets in correct order in simple case</title>
+  <link rel="author" title="Chris Rebert" href="http://chrisrebert.com">
```
domenic: The title and author seem inaccurate here

test/web-platform-tests/to-upstream/html/interaction/focus-managment/focusing-focused-noop.html
```diff
@@ -0,0 +1,43 @@
...
+  <link rel="help" href="https://html.spec.whatwg.org/#focus-update-steps">
+  <link rel="help" href="https://html.spec.whatwg.org/#focus-chain">
+  <meta name="flags" content="dom">
```
domenic: I'm not sure what this does; presumably you copied it from another test. Let's delete it.

domenic:
Thanks very much, and sorry for never replying directly to the issue #2621!

PR #2703, opened 2019-11-03
Bump nwsapi minimum version and add namespace selector test

### LINKED ISSUE #2028: querySelectorAll doesn't support namespaced attributes
Selectors targetting namespaced attributes (for instance, `[*|href]`) raise a SyntaxError.
### Basic info:

- **Node.js version:** v8.7.0
- **jsdom version:** v11.3.0

### Minimal reproduction case

```js

const dom = new JSDOM('<!doctype html><svg xlink:href="foo"/>');

dom.window.document.querySelectorAll('[*|href]');
```
throws:
```
    SyntaxError: The string "[*|href]", is not a valid CSS selector
```
although the selector [**is** valid](https://drafts.csswg.org/selectors-3/#attrnmsp).

It selects the `svg` element. See http://jsbin.com/relisatula/edit?html,console

See also #1587 (although I think this one is slightly different, as it not only affects XML documents but also HTML documents).

domenic:
Fixes #2028.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "domexception": "^1.0.1",
     "escodegen": "^1.11.1",
     "html-encoding-sniffer": "^1.0.2",
-    "nwsapi": "^2.1.4",
+    "nwsapi": "^2.2.0",
     "parse5": "5.1.0",
     "pn": "^1.1.0",
     "request": "^2.88.0",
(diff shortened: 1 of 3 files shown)


ExE-Boss:
**NSWAPI&nbsp;v2.2.0** also&nbsp;adds&nbsp;support for&nbsp;[unknown&nbsp;`-webkit-` pseudo‑elements](https://drafts.csswg.org/selectors-4/#compat).

PR #2706, opened 2019-11-05
feat(getComputedStyles): Improve performance

eps1lon:
`nwsapi` seems to be able to handle multiple selectors. 

Ran `yarn workspace @material-ui/core test` (on https://github.com/mui-org/material-ui/tree/a98abed2fce2fafe16f3c459d1a0de838e5736d2) 6 times with `node@v12.11.1` on Ubuntu 18.04.3 LTS.

median c3f0f2756bdbb20e235473728b562e04a67a6555: 30s
median PR: 26s

Not every test relies on `getComputedStyles` (more like 30-40% of them).

### DIFF
--- a/lib/jsdom/living/helpers/style-rules.js
+++ b/lib/jsdom/living/helpers/style-rules.js
@@ -47,24 +47,7 @@ exports.forEachMatchingSheetRuleOfElement = (element, handleRule) => {
 };
 
 function matches(rule, element) {
-  if (!rule.selectorText) {
-    return false;
-  }
-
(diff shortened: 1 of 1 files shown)


eps1lon:
> I do wonder a bit if our test coverage is sufficient here, but if it's not, probably we'll find out soon enough :).

I didn't find any test specifically testing for comma separated selectors. However, the `document.querySelectorAll` implementation isn't splitting them either. Maybe it's more prevalent to use comma-separated selectors in css compared to js but no reported issue would still be surprising.

Issue #2684, opened 2019-10-15
Update request-promise-native and request-promise-core to refer latest version of lodash


anushreemjoshi:
### Basic info:

- **Node.js version:** 
- **jsdom version:** 

### Minimal reproduction case

jsdom package requires > request-promise-native >request-promise-core > lodash.

It has been seen that request-promise-core hasn't been updated for a while to require latest version of lodash. Versions of ldash less than 4.17.12 has security vulnerability for Prototype pollution. 

Please update request-promise-core to refer to latest version of lodash and update the same in request-promise-native.  

### How does similar code behave in browsers?

(Link to a jsbin or similar suggested.)

domenic:
We don't control request-promise-core and request-promise-native, so this appears to be filed on the wrong repository.

PR #2707, opened 2019-11-06
bumping request-promise-native to 1.0.8 to include security patch

SubtleMachination:
A dependency of jsdom (request-promise-native 1.0.7) has been updated (1.0.8) to include a security patch.

See the 1.0.8 commit here: https://github.com/request/request-promise-native/commit/1592a3171e5a5c8d59da1052b6f7b84d733dd4f0
And the 1.0.8 change notes here: https://github.com/request/request-promise-native/commit/6498be1290b24e8911278df724b49670e80c684b

This is related to the jsdom issue here: https://github.com/jsdom/jsdom/issues/2684, which was closed at the time because request-promise-native needed to do their upgrade first before jsdom could take action. Now what request-promise-native is updated, it is time to update jsdom.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "parse5": "5.1.0",
     "pn": "^1.1.0",
     "request": "^2.88.0",
-    "request-promise-native": "^1.0.7",
+    "request-promise-native": "^1.0.8",
     "saxes": "^3.1.9",
     "symbol-tree": "^3.2.2",
     "tough-cookie": "^3.0.1",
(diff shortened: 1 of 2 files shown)


domenic:
Can you provide a demonstration of the attack on jsdom, so we can include it in our test suite?

SubtleMachination:
The vulnerability in lodash involves the ‘defaultsDeep’ function, which is currently uncalled in the request-promise-native module. Thus jsdom doesn’t actually have an consumer-facing exposure to the vulnerability right now.

Instead, the purpose of this patch is to close the possibility of exposing the vulnerability in the future. Just a due-dilligence measure.

Issue #2667, opened 2019-09-18
"openedRequest.abort is not a function" in Window.close()


CupOfTea696:
### Basic info:

- **Node.js version:** 11.12.0
- **jsdom version:** 15.1.1

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom")

const content = `
<html>
<head>
  ...
</head>
<body>
[...]

PR #2678, opened 2019-10-03
Fix trying to abort data URL requests

TimothyGu:
Fixes #2667.

PR #2711, opened 2019-11-14
Avoid aborting promises without abort method

gagern:
The ResourceLoader.fetch method is documented to return a promise, so assuming that promise will always have an abort method is in violation of the documentation and breaks for e.g. data URIs.  Only call abort when available (duck-typing).

Fixes https://github.com/jsdom/jsdom/issues/2667.


domenic:
This needs tests before we can consider merging it.

TimothyGu:
Superseded by #2678.

PR #2719, opened 2019-11-21
Implement Range and Selection APIs

### LINKED ISSUE #317: Implement Ranges and Traversal

### LINKED ISSUE #937: Implement DOM selection apis
https://developer.mozilla.org/en-US/docs/Web/API/Window.getSelection

Even a nonfunctioning shim would go a long way. Currently I just do the following:

window.document.getSelection = function() { return { addRange: function() {}, removeAllRanges:function() { } } };

Just to avoid "cannot call undefined" when running getSelection().

pmdartus:
## Changes

This PR implements the `Range` and the `Selection` APIs. Sorry for this large PR, however the Range WPT tests are highly coupled with the Selection WPT tests and because of this wanted to make those 2 land at the same time.

Fix #317, fix #937


lib/jsdom/browser/Window.js
```diff
@@ -401,6 +414,27 @@ function Window(options) {
     writable: true
   });
 
+  // https://dom.spec.whatwg.org/#dom-range-range
+  function Range() {
```
domenic: Why implement this in this manner instead of using webidl2js?

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -304,21 +301,6 @@ class DocumentImpl extends NodeImpl {
...
-  }
-
-  removeChild(/* Node */ arg) {
-    const ret = super.removeChild(arg);
-    if (arg === this._documentElement) {
-      this._documentElement = null;// force a recalculation
-    }
-    return ret;
-  }
-
```
domenic: Nice simplification

test/web-platform-tests/to-run.yaml
```diff
@@ -847,13 +850,22 @@ DIR: html/webappapis/timers
 
 ---
 
+DIR: selection
+
+idlharness.window.html: [fail, Depends on fetch]
+script-and-style-elements.html: [fail,
+  toString should return only the text within the selection that is visible to the user
+  (https://www.w3.org/Bugs/Public/show_bug.cgi?id=10583, https://bugzilla.mozilla.org/show_bug.cgi?id=1535392)
+]
```
domenic: Nice detective work, and thanks for documenting it.

lib/jsdom/browser/Window.js
pmdartus: This is not needed anymore, it was used to associate the `Range` with the `Document`. This is not needed anymore with the webidl2js constructor reform. The `Document` can now be extracted out of the newly introduced `globalObject`.

domenic:
I re-ran the CI three times and it gives segfaults on the latest Node stable once we start hitting the dom/ranges tests. https://travis-ci.org/jsdom/jsdom/jobs/618664359 . So probably we shouldn't merge until we either find a workaround or Node.js fixes their upstream bug.

I haven't tried reproducing locally yet; I hope it isn't hard to do so :(.

pmdartus:
I was able to reproduce the issue locally against node `13.2.0`.

pmdartus:
The segfault was fixed with node `13.3.0`. The CI is now passing and I also doubled checked locally just in case.

rharriso:
I really appreciate this work. I'm refactoring a highlighting feature, so this will be extremely valuable to me.

rpearce:
Thank you for this! Should it have also implemented [`Range.getClientRects()`](https://developer.mozilla.org/en-US/docs/Web/API/range/getClientRects)?

If range itself is fully supported, I believe this will open the door to automated color contrast accessibility testing via `axe-core`: https://github.com/dequelabs/axe-core/blob/104834fdc1fd09f5bb002ae48709f8939c0c7864/lib/rules/color-contrast-matches.js#L151.

TimothyGu:
There's probably no way for us to implement `getClientRects`, since we don't have a layout engine. See https://github.com/jsdom/jsdom#unimplemented-parts-of-the-web-platform.

rpearce:
@TimothyGu Thank you for the information!

Issue #2288, opened 2018-07-09
Input number validation error for decimal steps due to floating-point arithmetic


paradoxxxzero:
Consider this input:
```html
<input type="number" min="0" value="12.50" step="0.01" />
```
in a browser this is valid but with jsdom (current version) it sets the stepMismatch validation error and therefore I get the "Constraints not satisfied" error.
I traced it till this line: https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/nodes/HTMLInputElement-impl.js#L823
which computes 
`12.5 % -0.01 = 0.00999999999999974` even though `1250 * 0.01 = 12.5`.

oliverjam:
Yeah I have the exact same problem. `<input type="number" min="0" max="3.5" step="0.01" />` triggers a step mismatch validation error with any value I try (it's fine in browser)

Here's a minimal repro with react-testing-library and Jest

```js
const React = require("react");
const { render, fireEvent } = require("react-testing-library");

function Input() {
  return React.createElement("input", {
    "aria-label": "price",
    type: "number",
    min: "0",
    max: "3.5",
    step: "0.01"
  });
}

test("Input validation", () => {
  const { getByLabelText } = render(React.createElement(Input));
  const input = getByLabelText("price");
  fireEvent.change(input, { target: { value: "3" } });

  expect(input.value).toBe("3");
  expect(input.validity.stepMismatch).toBe(false);
});
```

[Repo here](https://github.com/oliverjam/jsdom-validityState-bug). `npm it` in that to see the failing test.

You can actually run the same test in Codesandbox and see it pass because that uses the actual DOM rather than jsdom

https://codesandbox.io/s/ko2z4jvjq7

kontomondo:
@domenic is there a higher precision arithmetic library available in jsdom ? An alternative solution is to normalise both the step and the value by multiplying by a factor that yields integer values - `multiply by 10^n` where `n` - total nr. of floating digits on the `step` and `value`.

domenic:
No, sadly, nothing available. We could try the multiplication business, although I'm sure that'll fail in some other interesting case... [Chrome apparently has a decimal library and some other decomposition into before/after the decimal point to handle this](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/html/forms/number_input_type.cc?type=cs&g=0). Very interesting.

We could pull in one of https://www.npmjs.com/search?q=decimal if it helps.

kontomondo:
[...]

PR #2724, opened 2019-11-28
use high precision modulo operation for stepMismatch constraint

kontomondo:
Fixes https://github.com/jsdom/jsdom/issues/2288

Associated tests are here: https://github.com/web-platform-tests/wpt/pull/20548


lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -827,7 +829,12 @@ class HTMLInputElementImpl extends HTMLElementImpl {
           if (typeof number !== "number" || isNaN(number)) {
             return false;
           }
-          return (number - this._stepBase) % allowedValueStep !== 0;
+
+          const isIntegralMultiple = 
+            new Decimal(this._stepBase - number)
```
kontomondo: the spec says `and that number subtracted from the step base`, so this follows the wording of the spec more closely although I find it bizzare since it would always result in a negative number - it doesn't matter for a `modulo` operation, it's odd nevertheless.

domenic:
Code looks great; just needs tests!

domenic:
I guess it's not immediately clear that you're "decimalizing" enough of the code; it seems like e.g. all the parsed values could be decimalized, or other operations besides just the step-mismatch one at the end. But I think tests will help confirm this one way or the other.

kontomondo:
@domenic Thanks for looking at the PR! The PR for tests is: https://github.com/web-platform-tests/wpt/pull/20548 . Regarding turning more operations to use decimal.js, that's what I initially did, but then then figured that it's likely not needed at this point, the stepBase is only non zero for type="week" and at that point number is always integer and unlikely to be higher than 2^53-1 (9007199254740991, that's year 287407) for any precision to be lost.  Given that, since the parsing of the float would be done at js precision, there shouldn't be any further accuracy loss incurred from a `0 - number` operation. If you believe that we should decimalise more of these things, happy to add that back.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
TimothyGu: Agreed. I'd honestly just use `number - this._stepBase` which makes more sense, and leave a comment – if that's okay with you.

domenic:
I'm having a really hard time rebasing this on master. @kontomondo would you be able to do that?

kontomondo:
@domenic yes, I'll do that shortly, thank you.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
kontomondo: that should be fine, I'll make the change shortly, thank you.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -822,15 +823,14 @@ class HTMLInputElementImpl extends HTMLElementImpl {
 
   // https://html.spec.whatwg.org/multipage/input.html#concept-input-min-zero
   get _stepBase() {
-    const parseAttribute = attributeName => this._convertStringToNumber(this.getAttributeNS(null, attributeName));
     if (this.hasAttributeNS(null, "min")) {
-      const min = parseAttribute("min");
+      const min = ths._convertStringToNumber(this.getAttributeNS(null, "min"));
```
domenic: Looks like the merge went wrong here; probably these three lines should not be changed?

kontomondo: I did want to inline, as the code itself is easy enough to not really need the extra function in there, but mispelled `this`, in the process, should be corrected now.

kontomondo:
@domenic Looks like it's all green now.

PR #2714, opened 2019-11-28
setTimeout leaks memory significantly

Sebmaster:
### Basic info:

- **Node.js version:** v13.1.0
- **jsdom version:** 13.2.0

### Minimal reproduction case

```js
const jsdom=require("jsdom");
	const { JSDOM }=jsdom;

	var options={
		url:"http://gg.com",
		pretendToBeVisual:true,
		includeNodeLocations: true,
		resources:'usable',
		runScripts:'dangerously',
		beforeParse:function(window){}
	}
	dom=new JSDOM('<script>function test1(){ var t=setTimeout(function(){ test1() },1); }; for(var i=0;i<1000;i++) test1(); </script>',options);
// best to run with: node --inspect test.js
// the leak is obvious on chrome://inspect
```

No leak

I use jsdom to emulate http://adventure.land 's gameplay, so the emulation needs to go on as long as possible, I've been hunting a memory leak for 2 days, since the game can run on the browser for weeks, I'm pretty sure the code itself doesn't have a memory leak

I believe the `setTimeout` implementation has a leak, above code easily causes the process to crash in a minute - inside `test1()` - if you allocate some variables too, I believe they are also preserved as the timeout prevents them from being garbage collected

**Off topic:** (As I don't want to create another issue) The JS performance is super low inside jsdom, is this expected - I haven't attempted to benchmark things yet

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -638,6 +638,7 @@ function startTimer(window, startFn, stopFn, timerId, callback, ms, timerStorage
     } catch (e) {
       reportException(window, e, window.location.href);
     }
+    delete timerStorage[timerId];
   };
 
   const res = startFn(callback, ms);


Sebmaster:
Hm, I attached a branch that fixes this issue and apparently that overrode the issue creator as well. Guess Github's handling this scenario worse and worse, sorry about that.

The attached patch fixes this issue though.

kaansoral:
Thank you @Sebmaster :)

lib/jsdom/browser/Window.js
```diff
@@ -659,6 +660,7 @@ function stopAllTimers(timers) {
     const timer = timers[key];
     // Need to .call() with undefined to ensure the thisArg is not timer itself
     timer[1].call(undefined, timer[0]);
+    delete timers[key];
   });
```
domenic: Would just setting `timers.length = 0` after all this make sense?

domenic:
Our timer code seems needlessly complicated and obscure... but this seems like a good patch for now.

lib/jsdom/browser/Window.js
Sebmaster: timers is a dict, so I don't think that'd work.

Sebmaster: Oh, but we're resetting them to empty in callers already, so this is unnecessary.

PR #2731, opened 2019-12-01
feat: add support for `vm.compileFunction`

SimenB:
Id like to move Jest away from the module wrapper, similar to what Node core did here: https://github.com/nodejs/node/pull/21573. Jest currently uses its own module wrapper, puts it in a `Script`, then executes it using `runVMScript`. However, if we use `compileFunction` instead, we don't need to create the wrapper.

Note that this function was added in `node@10.10`, so I've added a guard that'll throw on node 8 (or some minor of node that don't have the function). I'd prefer this to land in the current version of JSDOM without bumping the major, as otherwise we'd be unable to use this in Jest until sometime next year. I probably need to add a guard in the tests for this case, but I'd like to get some feedback on this first approach first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,7 +340,7 @@ const c = `
   return input * 2;
 `;
 
-const doubleFunction = dom.compileVMFunction(c, ['input'])
+const doubleFunction = dom.compileVMFunction(c, ["input"])
 
 doubleFunction(2); // 4
 ```
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ This is turned off by default for performance reasons, but is safe to enable.
 
 Note that we strongly advise against trying to "execute scripts" by mashing together the jsdom and Node global environments (e.g. by doing `global.window = dom.window`), and then executing scripts or test code inside the Node global environment. Instead, you should treat jsdom like you would a browser, and run all scripts and tests that need access to a DOM inside the jsdom environment, using `window.eval` or `runScripts: "dangerously"`. This might require, for example, creating a browserify bundle to execute as a `<script>` element—just like you would in a browser.
 
-Finally, for advanced use cases you can use the `dom.runVMScript(script)` or `dom.compileFunction(code)` methods, documented below.
+Finally, for advanced use cases you can use the `dom.compileVMFunction(code)` method, documented below.
 
 ### Pretending to be a visual browser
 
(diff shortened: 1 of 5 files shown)


domenic:
We won't be releasing any more majors of jsdom with the current version, so I'm sorry that won't be possible. Perhaps we could bump the minimum required Node.js version to 10 at the same time.

I'm not sure in general whether we want to add this complexity to jsdom to avoid "a wrapper", but I'm having a hard time getting a sense of what exactly the wrapper is. Can you try to explain the benefit to users in more detail? The example code you've added isn't very compelling; compileVMFunction and runVMScript seem very redundant. Perhaps we should remove runVMScript?

SimenB:
Yeah, the OP was pretty parse in this case and focused more on a solution to a problem I'm facing than why it should be in JSDOM. Sorry about that.

---

>We won't be releasing any more versions of jsdom with the current (v15) major, so I'm sorry that won't be possible. Perhaps we could bump the minimum required Node.js version to 10 at the same time.

Ok, that's fine 🙂 If you're going for a major soonish, dropping node 8 should be no issue as it's EOL 2019-12-31 anyways (https://github.com/nodejs/Release/blob/master/README.md). That neatly solves the issue of checking for support at the same time.

>Can you try to explain the benefit to users in more detail?

Sure! I'll try to, by taking a step back. Correct me if I'm wrong, but the main use case for `runVMScript` is to be able to pass some JS, have it evaluated within the JSDOM context, and pass back the result of this evaluation. Jest currently uses this by creating a `vm.Script` with e.g. this content

```js
({
  evalResult: function(arg1, arg2) {
    // do something in here
  }
})
```

After running the script, Jest then does `scriptResult.evalResult.call(null, passAsArg1, passAsArg2)` allowing it to inject `require`, `module` etc. from the outside. This has some issues when trying to map source code back, especially for code coverage.

There are more details about the general problem with the wrapper here: https://github.com/nodejs/node/issues/17396

Using `vm.compileFunction` instead allows us to get a function back directly, without the need to create a wrapper.

>compileVMFunction and runVMScript seem very redundant. Perhaps we should remove runVMScript?

Yup, totally! The advantage (in my mind) of `runVMScript` is that a `vm.Script` can take the content of a whole JS file, not just the body of a function. However, if your use case is "this whole file", you should probably use script tags, `eval` or similar, like documented in the readme today. The use case of "evaluate a function in the context of JSDOM and invoke that function from the outside" is better covered by `compileVMFunction`, like I talked about above. However, there might be use cases for `runVMScript` that I don't know about?

lib/api.js
```diff
@@ -80,6 +80,27 @@ class JSDOM {
...
+    }
+
+    if (!vm.compileFunction) {
+      throw new TypeError("The version of Node you're using does not support compileFunction, please upgrade");
+    }
+
+    options = options || {};
+
+    if (options.parsingContext) {
+      throw new TypeError("You cannot pass a parseing context to compileFunction");
```
thymikee:
```suggestion
      throw new TypeError("You cannot pass a parsing context to compileFunction");
```

SimenB:
@domenic thoughts? I'm happy to remove `runVMScript` at the same time

domenic:
Sorry for the delay!

Let's try removing runVMScript, indeed. When preparing the next release, I'll add a note to the changelog and/or readme that if compileVMFunction doesn't fit folks' use case, then we can add it back, but for now I buy your reasoning on compileVMFunction being generally more useful, and I'd like to keep the more-esoteric methods to a minimum if that works for the community.

Feel free to remove Node v8 from the matrix (bumping the node-canvas test suite to v10), and change the package.json engines field, to get the CI to pass.

test/api/methods.js
```diff
@@ -191,6 +191,83 @@ describe("API: JSDOM class's methods", () => {
...
+
+      assert.strictEqual(dom.window.ran, undefined);
+
+      compiledFunction();
+
+      assert.strictEqual(dom.window.ran, true);
+    });
+
+    it("should work when runScripts is set to \"dangerously\"", () => {
+      const dom = new JSDOM(``, { runScripts: "outside-only" });
```
ExE-Boss:
```suggestion
      const dom = new JSDOM(``, { runScripts: "dangerously" });
```

SimenB:
Thanks @domenic, I've updated this now. In addition to the changes you mentioned, I also added node 12 to the testing matrix, as it's the current LTS version.

README.md
```diff
@@ -309,32 +309,32 @@ console.log(dom.nodeLocation(imgEl));    // { startOffset: 13, endOffset: 32 }
 
 Note that this feature only works if you have set the `includeNodeLocations` option; node locations are off by default for performance reasons.
 
-### Running vm-created scripts with `runVMScript(script[, options])`
+### Compiling a function with `compileVMFunction(code[, params[, options]])`
```
SimenB: I'm not sure about the notation for multiple optional arguments

SimenB:
Actually thinking about this, I wonder what it'd mean for modules? Those can't be wrapped in functions... We'll probably (when the time comes) inject Jest's "globals" through a loader when that API is ready, but I suppose that'll be closer to the current `vm.Script` approach than this. Some discussion here, but it's basically "we don't know yet": https://github.com/nodejs/node/issues/27387. Currently `vm.Script` doesn't work with ESM anyways, so it doesn't matter much, but something to keep in mind

domenic:
Hmm. Maybe we should take a step back, and instead just directly expose something like `dom.vmContext` (which returns `this[window]`)? And have the README give a few examples of how to use it, and have the tests test a couple main-line use cases like `compileFunction` and `runScript`? It would return `null` (or `undefined`?) if `runScripts` was not set.

That should give maximum flexibility for integrators like Jest, including future things like modules.

It's a bit of a foot-gun if people start, for whatever reason, using `dom.vmContext` instead of `dom.window`, but hopefully the name would discourage them from doing that.

SimenB:
Sorry about the slow response, currently semi offline due to holidays (or, I try to be with varying luck 😛).

I agree a lower level API exposing the raw context directly makes sense, that way jsdom can be as flexible as possible for the advanced use cases.

I'd go with a `getInternalVmContext()` or something as its name to make it very explicit it's not an API you'd use normally. I'm happy to change this PR to be that if you want, although I won't get to it until the week after new years.

Issue #2659, opened 2019-09-07
input.valueAsDate is not set to Jan 1 1970 for <input type="time">


Rich-Harris:
### Basic info:

- **Node.js version:** 11.6.0
- **jsdom version:** 15.1.1

### Minimal reproduction case

```js
const { JSDOM } = require('jsdom');

const { window } = new JSDOM();
const { document } = window;

document.body.innerHTML = `
  <input type="time">
`;

const input = document.querySelector('input');

input.valueAsDate = new Date();
console.log(`the year is ${input.valueAsDate.getFullYear()}`);
```

This outputs `the year is 2019`.

### How does similar code behave in browsers?

`the year is 1970`. The date is discarded, only the time should remain.

domenic:
This is another consequence of the fact noted in https://github.com/jsdom/jsdom/issues/2658#issuecomment-529129461, which is that valueAsDate is not implemented at all in jsdom, so the code you're writing here is equivalent to

```js
input.anyPropertyName = new Date();
console.log(`the year is ${input.anyPropertyName.getFullYear()}`);
```

Issue #2658, opened 2019-09-07
Setting input.valueAsDate or input.valueAsNumber doesn't affect input.value


Rich-Harris:
### Basic info:

- **Node.js version:** 11.6.0
- **jsdom version:** 15.1.1

### Minimal reproduction case

```js
const { JSDOM } = require('jsdom');

const { window } = new JSDOM();
const { document } = window;

document.body.innerHTML = `
  <input type="date">
  <input type="time">
  <input type="week">
  <input type="month">
`;

const today = new Date();

[...document.querySelectorAll('input')].forEach(input => {
  input.valueAsDate = today;
  console.log(`${input.type}: ${input.value}`);
});

document.body.innerHTML = `
  <input type="number">
  <input type="range">
`;

[...document.querySelectorAll('input')].forEach(input => {
  input.valueAsNumber = 42;
  console.log(`${input.type}: ${input.value}`);
});
```

This results in:

```
date: 
time: 
week: 
month:
number:
range:
```

### How does similar code behave in browsers?

```
date: 2019-09-07
time: 15:26:41.457
week: 2019-W36
month: 2019-09
number: 42
range: 42
```

https://jsbin.com/fujutaqimu/edit?html,output

Rich-Harris:
(Should note that the same is true in reverse — setting `input.value` doesn't affect `input.valueAsDate` or `input.valueAsNumber`)

domenic:
We currently don't implement valueAsDate or valueAsNumber at all, so you're just seeing the effect of what happens if you set a random property with any name on a HTML element. A pull request implementing them would be very welcome.

PR #2732, opened 2019-12-01
Add missing value accessors on input elements

kraynel:
Hello!

Thanks for that project, here is a little contribution.
This should fix issues #2659, #2658.

- Implement valueAsDate
- Implement valueAsNumber
- Implement stepUp and stepDown

I probably missed some edge cases, and I am not sure of the step alignment for min and max values when `stepUp` and `stepDown` go over the max/min (step 8 and 9 of https://www.w3.org/TR/html52/sec-forms.html#dom-htmlinputelement-stepdown).

I'd be glad to take your feedbacks!


lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -406,6 +541,122 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+
+    const convertNumberToString = this._convertNumberToString;
+    this._value = convertNumberToString(v);
+  }
+
+  _stepUpdate(n, isUp) {
+    // https://www.w3.org/TR/html52/sec-forms.html#dom-htmlinputelement-stepup
+
+    if (!allowValueAsNumber(this.type)) {
+      throw new DOMException("The object is in an invalid state.", "InvalidStateError");
```
pmdartus:
```suggestion
      throw new DOMException(`Failed to invoke '${isUp ? 'stepUp' : 'stepDown' }' method on 'HTMLInputElement': This input element does not support Number values.`, "InvalidStateError");
```

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -406,6 +541,122 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+    }
+
+    const allowedValueStep = this._allowedValueStep;
+    if (allowedValueStep === null) {
+      throw new DOMException("The object is in an invalid state.", "InvalidStateError");
```
pmdartus:
```suggestion
      throw new DOMException(`Failed to invoke '${isUp ? 'stepUp' : 'stepDown' }' method on 'HTMLInputElement': This input element does not support value step.`, "InvalidStateError");
```

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -406,6 +541,122 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+        return;
+      }
+
+      const candidateStepValue = this._stepAlign(min + allowedValueStep, false);
+      if (candidateStepValue < min || candidateStepValue > max) {
+        return;
+      }
+    }
+
+    let value = this.valueAsNumber || 0;
```
pmdartus: > Step 5. If applying the algorithm to convert a string to a number to the string given by the element's value does not result in an error, then let value be the result of that algorithm. Otherwise, let value be zero.

You will probably need to conver this line to a try catch.

kraynel:
Thanks for the review @pmdartus.
I started to add some tests for the `valueAsNumber`, but there are lots of cases to cover. Do you have any scenario in mind?

I will add some `valueAsDate` nominal test cases as well.

pmdartus:
I realized that the `to-port-to-wpts` is the wrong folder to create the tests. We actually want to get rid of those tests. New tests should be created under `web-platform-tests/to-upstream` to match the rest of the WPT format in the hope one day that those tests will get integrated into the official test suite. 

I am sorry for this miscommunication, in order to redeem myself, I created a handful of tests (https://github.com/pmdartus/jsdom/commit/c1dd4dec2c8d811725bc28758e9adf923125c9f0) in the WPT format. 😉

Some of those tests are currently failing, however, I didn't spent time debugging the source of the failures yet. @kraynel let me know what you think of those.

kraynel:
Thanks, I fixed the valueAsDate getter/setter.
I need to figure out the realign to min/max.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -406,6 +580,145 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+
+    let delta = n * allowedValueStep;
+    if (!isUp) {
+      delta *= -1;
+    }
+
+    value += delta;
+
+    if (min !== null && value < min) {
+      // should be value = this._stepAlign(min, true);
```
pmdartus: Is this comment still valid?

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -406,6 +580,145 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+      value = min;
+    }
+
+    if (max !== null && value > max) {
+      // should be value = this._stepAlign(max, false);
```
pmdartus: Is this comment still valid?

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -406,6 +580,145 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+      throw new DOMException("Failed to set the 'valueAsNumber' property on 'HTMLInputElement': This input element " +
+        "does not support Number values.", "InvalidStateError");
+    }
+
+    const convertNumberToString = this._convertNumberToString;
+    this._value = convertNumberToString(v);
+  }
+
+  _stepUpdate(n, isUp) {
+    // https://html.spec.whatwg.org/multipage/input.html#dom-input-stepup
```
pmdartus: Move the spec reference above method declaration in the class body.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -406,6 +580,145 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+    if (v === null || (Object.prototype.toString.call(v) === "[object Date]" && isNaN(v))) {
+      this._value = "";
+    }
+
+    const convertDateToString = this._convertDateToString;
+    this._value = convertDateToString(v);
+  }
+
+  get valueAsNumber() {
+    // https://html.spec.whatwg.org/multipage/input.html#dom-input-valueasnumber
```
pmdartus: Move the spec reference above method declaration in the class body.

kraynel:
I fixed the two comments.

I continued looking on the step alignment for min/max overflowing.

I noticed the test in `test/web-platform-tests/tests/html/semantics/forms/the-input-element/time.html` fails in browser (see https://jsbin.com/bisoluguno/1/edit?html,css,output), so I guess the test `stop at border on stepUp` is wrong: when you try to stepUp a time value with a max value of `15:01`, a step of `120` and an initial value of `15:00`, you stay at `15:00` if you want to stay step aligned.

How should we proceed?

TimothyGu:
@kraynel Indeed, that looks like a test bug: https://wpt.fyi/results/html/semantics/forms/the-input-element/time.html?label=master&label=experimental&aligned&q=html%2Fsemantics%2Fforms%2Fthe-input-element%2Ftime.html

If that is the only subtest that fails, feel free to put the test in the ignore list for now. I'll file an issue or PR with the WPT project to fix that.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -408,6 +586,154 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
+  set valueAsDate(v) {
+    if (!allowValueAsDate(this.type)) {
+      throw DOMException.create(this._globalObject, [
+        "Failed to set the 'valueAsDate' property on 'HTMLInputElement': This input element does not support Date " +
+        "values.",
+        "InvalidStateError"
+      ]);
+    }
+
+    if (v !== null && Object.prototype.toString.call(v) !== "[object Date]") {
```
ExE-Boss: This&nbsp;should be&nbsp;using an&nbsp;`isDate(…)`&nbsp;helper, like&nbsp;what&nbsp;**WebIDL2JS** does&nbsp;with&nbsp;`isArrayBuffer(…)`: <https://github.com/kraynel/jsdom/pull/1>.

domenic:
I've found that the new to-upstream web platform tests fail in my timezone (UTC-5). It kind of sucks we don't have great CI coverage for that, but we should probably fix that before merging...

lib/jsdom/living/helpers/dates-and-times.js
```diff
@@ -212,7 +212,33 @@ function serializeWeek({ year, week }) {
...
+  const week = Math.ceil((((date - yearStart) / dayInSeconds) + 1) / 7);
+
+  return { year: date.getUTCFullYear(), week };
+}
+
+function isDate(obj) {
+  try {
+    Date.prototype.valueOf.call(obj);
+    return true;
+  } catch {
```
ycjcl868: maybe should be compatible with Node.js 8.x.

pmdartus: Node 8.x support has been dropped with version 16.0: https://github.com/jsdom/jsdom/blob/master/Changelog.md#1600

Issue #2727, opened 2019-11-30
Constructor and prototype reform tracking


domenic:
Constructor and prototype reform has landed in https://github.com/jsdom/jsdom/commit/5e39a4c60377c95ac09ea938ecbf5f1f91cb0360, but not yet been released. This issue is meant as a meta tracking issue for remaining potential work items.

**Currently I do not think any of these should block the next release, and am considering doing the release some time in the next few days.** So please speak up if you disagree! I am worried a bit about how fixing these issues could be considered a breaking change, in which case we may want to bundle it with the existing constructor and prototype reform breaking change that has already landed.

### Classes from external packages are not yet migrated

Namely, `DOMException`, `URL`, `URLSearchParams`, and `XMLSerializer`.

@pmdartus is working on this, and it shouldn't be too hard, except he ran into the following issue while doing so...

### ECMAScript spec globals are handled inconsistently

Discussed in https://github.com/jsdom/domexception/pull/3#discussion_r349866219.

Currently, if `runScripts` is set to either `"outside-only"` or `"dangerously"`, the Node.js `vm` machinery will install all the ECMAScript globals onto the JSDOM `Window`. I.e., `dom.window.TypeError`, `dom.window.Map`, etc. will exist, and will be fresh copies: `dom.window.TypeError !== global.TypeError`. If `runScripts` is left as its default `undefined`, then these will not exist.

The exception is the typed arrays, which we [alias from the Node.js outer environment](https://github.com/jsdom/jsdom/blob/3e0655fc0651f233c90117eca108b66453d21247/lib/jsdom/browser/Window.js#L443-L452). We do this regardless of the `runScripts` value. Presumably this was to cause us to pass more tests in some way, but as noted in the comment, it's an inconsistent hack. I'd like to stop doing this, one way or another.

I see a few options here:

- Remove support for disabling scripting, i.e. make the new default equivalent to `"outside-only"`. And, remove the typed array special-casing. Now, fresh copies of all ECMAScript globals will be installed, just like fresh copies of web globals are now installed. This is nice for consistency, but may hurt performance in the simple case where people don't need to run scripts.

[...]

PR #2733, opened 2019-12-01
domexception constructor reform upgrade

pmdartus:
This PR intregrates the changes made in https://github.com/jsdom/domexception/pull/3.

### DIFF
--- a/lib/jsdom/living/interfaces.js
+++ b/lib/jsdom/living/interfaces.js
@@ -1,7 +1,6 @@
 /* eslint-disable global-require */
 "use strict";
 
-const DOMException = require("domexception");
 const { URL, URLSearchParams } = require("whatwg-url");
 const XMLSerializer = require("w3c-xmlserializer/lib/XMLSerializer").interface;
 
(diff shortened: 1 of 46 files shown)


lib/jsdom/living/interfaces.js
```diff
@@ -184,9 +185,15 @@ function install(window, name, interfaceConstructor) {
 }
 
 exports.installInterfaces = function (window) {
+  // TODO: Fix me
+  Object.defineProperty(window, "Error", {
```
pmdartus: This line needs to be updated once we settle on an approach to expose ECMAScript globals (https://github.com/jsdom/jsdom/issues/2727). This is a tempory solution to make the tests pass.

domenic:
Looks like one test is still failing, Document-createElementName-namespace.html

lib/jsdom/browser/js-globals.json
```diff
@@ -0,0 +1,60 @@
+[
```
domenic: Is this file supposed to be here? Should we wait on merging this until #2743 is merged?

pmdartus: I completely messed up the rebase... I will fix that tomorrow. But yes, let wait for your PR to get in first.

domenic: I'm ok landing in either order. If this lands first then #2743 just needs to delete a few extra lines, probably. I'm worried about getting stuck on #2743, although I've made some good progress I guess.

pmdartus: Let's wait for #2743 to land, you have made good progress there. I will try to help out on this, this week.

Issue #2639, opened 2019-07-27
border not setting with rgba color


Morglod:
### Basic info:

- **Node.js version:** 12
- **jsdom version:** 15.1.1

### Minimal reproduction case

```js
const jsdom = require("jsdom");
const { JSDOM } = jsdom;

const dom = new JSDOM(`<!DOCTYPE html><p>Hello world</p>`);
const p = dom.window.document.querySelector("p");
p.style.border = '2px solid rgba(255, 255, 255, 0.5)';
console.log(p.style.border); // p.style.border === ''
p.style.border = '2px solid rgb(255, 255, 255)';
console.log(p.style.border); // p.style.border === '2px solid rgb(255, 255, 255)'
```

### How does similar code behave in browsers?

As it should;
There is two logs:
```
2px solid rgba(255, 255, 255, 0.5)  
2px solid rgb(255, 255, 255)
```

Morglod:
There are maybe similar issue leaved in 2014!
https://github.com/jsdom/jsdom/issues/825
❤️ 👍

Issue #825, opened 2014-07-07
HSL(A) styles are ignored


ebenpack:
Setting `HTMLElement.style.color` (or `backgroundColor`, or any other style declaration that expects a color value) to an hsl or hsla value has no effect. E.g.

```
var jsdom = require("jsdom").jsdom;
document = jsdom('<html><head><script></script></head><body></body></html>');
var foo = document.createElement('div');
foo.style.color === ''; // Foo doesn't have any color set to begin with
foo.style.color = 'hsl(0, 100%, 50%)';
foo.style.color === ''; // No effect
foo.style.color = 'rgb(255,255,255)';
foo.style.color === 'rgb(255,255,255)'; // rgb() works fine
foo.style.color = 'hsl(0, 100%, 50%)';
foo.style.color === 'rgb(255,255,255)'; // Again, hsl() has no effect
```

PR #2735, opened 2019-12-05
Add test for issue 2639 - setting rgba color style

kraynel:
I pinpointed the issue #2639 to a problem with cssstyle.
I submitted a PR on cssstyle (see https://github.com/jsdom/cssstyle/pull/109) and added a test here.
I can rename/move the test if you feel it is not in the right place.

In order for this test to pass, we need to merge/release a new version of cssstyle, bump the version used by jsdom and rerun the test.

Concerning issue #825, I could not reproduce the issue. It may have been fixed by a cssstyle update since 2014?

### DIFF
--- a/test/web-platform-tests/to-upstream/cssom/element-style-color.html
+++ b/test/web-platform-tests/to-upstream/cssom/element-style-color.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Computed visibility is inherited</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<!-- Test for https://github.com/jsdom/jsdom/issues/2639 -->
+<!-- Test for https://github.com/jsdom/jsdom/issues/825 -->
+
+<body>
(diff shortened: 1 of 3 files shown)


test/web-platform-tests/to-upstream/cssom/element-style-color.html
```diff
@@ -0,0 +1,49 @@
...
+
+  assert_equals(element.style.color, "");
+}, "Setting a mixed percent and value color should NOT update the style");
+
+test(() => {
+  const element = document.querySelector("#paragraph");
+  element.style.color = "";
+  element.style.color = "hsl(0, 100%, 50%)";
+
+  assert_equals(element.style.color, "hsl(0, 100%, 50%)");
```
domenic: This doesn't appear to be quite correct: https://jsbin.com/xuguyotefo/edit?html,css,output

kraynel: Hum, I could not find the spec but it seems Firefox, Chrome and Safari (at least) are converting HSL to RGB when parsing color values. I updated my PR on cssstyle to implement that.

PR #2746, opened 2019-12-11
update web-platform-tests to the latest

kontomondo:
Implementation of whatwg/dom#754 is delayed due to lingering issues with
the PR; see whatwg/dom#813.

Single-page to-upstream tests are updated to declare themselves as such;
per WPT RFC 28
(https://github.com/web-platform-tests/rfcs/blob/master/rfcs/single_test.md).

Co-authored-by: @TimothyGu

### DIFF
--- a/test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-escapes.html
+++ b/test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-escapes.html
@@ -65,7 +65,6 @@ testMatched("outOfRange\u{fffd}", "#outOfRange\\555555");
 testMatched("outOfRange\u{fffd}", "#outOfRange\\ffffff");
 
 // - escape EOF
-testMatched("eof\u{fffd}", "#eof\\");
 testNeverMatched("eof\\", "#eof\\");
 
 // - escape anythong else
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "test/web-platform-tests/tests"]
 	path = test/web-platform-tests/tests
-	url = https://github.com/w3c/web-platform-tests.git
+	url = https://github.com/web-platform-tests/wpt.git
--- a/Changelog.md
+++ b/Changelog.md
@@ -257,7 +257,7 @@ Other changes:
 * Fixed one regression (since v11.6.0) in `<style>` elements, where their `sheet` property would sometimes be `null` when it should not be.
 * Fixed a case where a `<style>` element's `sheet` property would be left as a `CSSStyleSheet` despite it not being in the document.
 
-Another regression remains where we are emitting spurious CSS-parsing `jsdomError` events; see [#2123](https://github.com/tmpvar/jsdom/issues/2123). We also discovered a large amount of preexisting brokenness around `<style>`, `<link>`, and `@import`; see [#2124](https://github.com/tmpvar/jsdom/issues/2124) for more details.
+Another regression remains where we are emitting spurious CSS-parsing `jsdomError` events; see [#2123](https://github.com/jsdom/jsdom/issues/2123). We also discovered a large amount of preexisting brokenness around `<style>`, `<link>`, and `@import`; see [#2124](https://github.com/jsdom/jsdom/issues/2124) for more details.
 
 We'll try to fix these soon, especially the regression.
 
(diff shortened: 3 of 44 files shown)


kontomondo:
A lot of new failures in validity related infra due to this change:

```
Change input.validity.* to not depend on willValidate
    
    Prior to this change, disabling (or marking readonly) an input
    would cause input.validity.* to become always true, because
    willValidate became false. The spec [1] isn't completely clear
    but appears to say that input.validity.* should give the state
    of validation, independent of whether the control will actually
    be validated. Other browsers, including Firefox, behave this
    way already.
    
    This CL also adds testing of most validity.* accessors with
    disabled=true, readonly=true, and both true. Most of these
    tests fail without this CL in place, confirming the bug.
    
    [1] https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#the-constraint-validation-api
```

kontomondo:
@domenic  This is the WPT update that brings in the new stepMismatch tests needed for the bug fix verification. There are 3 outstanding failures here 2 from tests that should be failing but they aren't getting marked as failed even though there are TypeError/errors thrown, and the customError test fails on a test for `validationMessage` (the test sets a custom validity message, disables the element and checks whether validationMessage equals the custom validity message). I checked the spec again, and the code is following it correctly, in particular, per https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-validationmessage
`If this element is not a candidate for constraint validation or if this element satisfies its constraints, then return the empty string.`
+
`A submittable element is a candidate for constraint validation except when a condition has barred the element from constraint validation. (For example, an element is barred from constraint validation if it is an object element.)`
+
`Constraint validation: If an element is disabled, it is barred from constraint validation.`
would result in the "" message, yet the updated test expects the custom error to show. I think the test itself is reasonable - if validity will reflect the actual state of constraints regardless of whether it's barred from validation or not - it's just the spec would have to change to account for that, otherwise it's in direct conflict with the test.

domenic:
>  two from tests that should be failing but they aren't getting marked as failed even though there are TypeError/errors thrown

Very curious. Maybe a bug in our jsdom testing infrastructure? I can try to debug if you don't first...

> one customError test that fails on a sub-test for validationMessage (the test sets a custom validity message, disables the element and checks whether validationMessage equals the custom validity message).

So you think the WPTs were changed incorrectly without changing the spec? Can you open an issue on whatwg/html to discuss, and ideally tag the people who committed the WPT change? We can mark the test as failing in the meantime with a pointer to that issue.

kontomondo:
> Very curious. Maybe a bug in our jsdom testing infrastructure? I can try to debug if you don't first...

The two tests that pass even though I guess the functionality isn't actually implemented are:
elementFromPoint-001.html: [fail, Not implemented]
getBoundingClientRect-empty-inline.html: [fail, document.fonts is not implemented]

I haven't had the time to look into this, perhaps a separate issue can be opened for that ? For now I removed them from expected to fail list.

> So you think the WPTs were changed incorrectly without changing the spec? Can you open an issue on whatwg/html to discuss, and ideally tag the people who committed the WPT change? We can mark the test as failing in the meantime with a pointer to that issue.

Opened https://github.com/web-platform-tests/wpt/issues/20825 and referenced it in to-run.yaml

TimothyGu:
Looks like the reason why the errors are no longer shown as errors is the following. Upstream WPT made single-page tests [declare themselves as such](https://github.com/web-platform-tests/rfcs/blob/master/rfcs/single_test.md) explicitly, and disabled implicit page error → test error conversion. This meant that JavaScript errors thrown before any tests were declared are not shown as testharness errors. However, we don't recognize testharness errors at the moment – only [timeouts](https://github.com/jsdom/jsdom/blob/97411a7fbeaac518355626b9a87b6e8964f3e92c/test/web-platform-tests/run-single-wpt.js#L116-L118), leading to missed errors.

I've fixed this issue and also made some more progress on updating the to-run file.

TimothyGu:
I rolled back the cloning steps changes, as they cause the following bug:

```js
o = new Option();
o.selected = true;
console.log(o.cloneNode().selected);
  // with cloning steps: true
  // browsers: false
```

TimothyGu:
Okay this is again ready for review. I also fixed the issue with form validation tests in upstream.

test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-escapes.html
```diff
@@ -65,7 +65,6 @@
 testMatched("outOfRange\u{fffd}", "#outOfRange\\ffffff");
 
 // - escape EOF
-testMatched("eof\u{fffd}", "#eof\\");
```
domenic: Why remove this?

TimothyGu: The same test already exists above, leading to the same test name passed to `test()` in both instances. testharness now rejects tests with the same name.

PR #2750, opened 2019-12-13
Fix linter error from Jesse's PR

Craigfis:


TimothyGu:
We try not to be misleading about what jsdom supports vs. does not support; see
https://github.com/jsdom/jsdom#unimplemented-parts-of-the-web-platform. If you would like to ignore these errors in your own use case, you could shim them out by [using the `beforeParse` hook](https://github.com/jsdom/jsdom#intervening-before-parsing).

PR #2768, opened 2020-01-01
Replace runVMScript() with getInternalVMContext()

domenic:
This allows more flexibility for vm-using consumers. Closes #2731.

@SimenB please take a look and let me know what you think!


README.md
```diff
@@ -325,16 +327,18 @@ const s = new Script(`
...
 `);
 
-dom.runVMScript(s);
-dom.runVMScript(s);
-dom.runVMScript(s);
+const vmContext = dom.getInternalVMContext();
+
+script.runInContext(vmContext);
+script.runInContext(vmContext);
+script.runInContext(vmContext);
```
ExE-Boss: Actually, the&nbsp;variable is&nbsp;named&nbsp;`s` instead&nbsp;of&nbsp;`script`.

domenic: Sigh, thanks. I'll do a force-push.

SimenB:
Just to follow up, this was a great call rather than messing with the `compileFunction` approach I started with. I've looked into running modules in `vm` and we need access to the context when constructing the module, not just when actually evaluating/running it: https://nodejs.org/api/vm.html#vm_constructor_new_vm_sourcetextmodule_code_options

So exposing the context directly was the correct call without jsdom having to add a bunch of extra apis/abstractions 👍

PR #2770, opened 2020-01-02
fix: Make `Attr` extend `Node`

### LINKED ISSUE #1641: Make Attr inherit from Node again

ExE-Boss:
Resolves #1641
Supersedes and closes #1822


ExE-Boss:
Looks&nbsp;like there’s&nbsp;some&nbsp;weird circular&nbsp;dependency&nbsp;issues.

One&nbsp;way to&nbsp;fix&nbsp;it&nbsp;would be&nbsp;to&nbsp;change&nbsp;**WebIDL2JS** to&nbsp;generate <code>exports.*&nbsp;=</code> assignments, rather&nbsp;than&nbsp;<code>module.exports&nbsp;=&nbsp;iface;</code>.

domenic:
Oh, awesome.

Note that the spec changes for this were pretty involved. I'm unsure what that translates to in terms of jsdom changes, exactly, but we should probably at least pass any of the web platform tests that got updated.... although it looks like maybe the spec changed before we had a tests-required policy in WHATWG :-/.

- https://github.com/whatwg/dom/commit/625a0747f137454c155a7b577a9e45be1aa35a34
- https://github.com/whatwg/dom/commit/0e17e05b7b0273799b853c4782b2eac9a88fdea1
- https://github.com/whatwg/dom/commit/dc3990d2678375a17fa05f3f4de99da23077c229

domenic:
> Looks like there’s some weird circular dependency issues.

Are they the same issues as https://github.com/jsdom/jsdom/pull/1247#issuecomment-149060470 ?

Could we merge attributes.js and Attr-impl.js to fix them?

ExE-Boss:
> Could we merge attributes.js and Attr-impl.js to fix them?

I’d&nbsp;rather fix&nbsp;**WebIDL2JS** to&nbsp;generate&nbsp;code that&nbsp;doesn’t completely&nbsp;break on&nbsp;circular&nbsp;`require(…)`s by&nbsp;creating&nbsp;properties on&nbsp;`exports`, rather&nbsp;than re‑assigning&nbsp;it: <https://github.com/jsdom/webidl2js/pull/154>.

domenic:
I'd rather avoid the circular requires; they're a sign of bad architecture, and I'm not willing to change our tools to make them easier.

ExE-Boss:
review?(@domenic): The&nbsp;tests are&nbsp;now&nbsp;passing.

domenic:
As stated previously I want us to remove the circular dependency before making any changes in this area.

ExE-Boss:
That&nbsp;will just&nbsp;create another&nbsp;circular&nbsp;dependency, this&nbsp;time between&nbsp;`NodeImpl` and&nbsp;`AttrImpl`.

domenic:
It seems very doable, you just have to not keep all the functions together. E.g. move `attributeListsEquals` and `setAnExistingAttributeValue` into `Node-impl.js`.

ExE-Boss:
`setAnExistingAttributeValue(…)` transiently&nbsp;depends on&nbsp;`changeAttribute(…)`, the&nbsp;latter&nbsp;of&nbsp;which is&nbsp;depended&nbsp;upon by&nbsp;`ElementImpl` and&nbsp;`setAttributeValue(…)`.

domenic:
This is looking pretty good. Do you have any thoughts on the compareDocumentPosition changes I linked to?

ExE-Boss:
@domenic Where&nbsp;can&nbsp;I&nbsp;find the&nbsp;`compareDocumentPosition(…)` changes&nbsp;you&nbsp;mentioned?

domenic:
The three commits linked in https://github.com/jsdom/jsdom/pull/2770#issuecomment-570097682

domenic:
Ping; any thoughts about `compareDocumentPosition()`? Do you think we should merge this and work on that later?

domenic:
Hmm, it seems my "style fixes" revealed that there are still some circular dependencies here. It looks like the current one is Node-impl.js -> node.js -> attributes.js -> Node-impl.js.

Since copyAttributeList is the only thing in attributes.js used by node.js, and copyAttributeList is only used in one place, this is hopefully relatively easy to fix...

ExE-Boss:
`copyAttributeList(…)`&nbsp;depends on&nbsp;`appendAttribute(…)`, which&nbsp;can’t be&nbsp;moved to&nbsp;`node.js`, due&nbsp;to&nbsp;it&nbsp;being depended&nbsp;upon by&nbsp;`attributes.js` and&nbsp;`Element‑impl.js`.

This&nbsp;is one&nbsp;of&nbsp;those&nbsp;things that&nbsp;have circular&nbsp;dependencies in&nbsp;the&nbsp;spec.

lib/jsdom/living/attributes/Attr-impl.js
```diff
@@ -1,9 +1,10 @@
 "use strict";
 
-const NodeImpl = require("../nodes/Node-impl.js");
+const NodeImpl = require("../nodes/Node-impl.js").implementation;
+const { setAnExistingAttributeValue } = require("../nodes/Node-impl.js");
```
ExE-Boss: Duplicate&nbsp;imports are&nbsp;yuck.

lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -3,7 +3,8 @@ const { addNwsapi } = require("../helpers/selectors");
 const { HTML_NS } = require("../helpers/namespaces");
 const { mixin, memoizeQuery } = require("../../utils");
 const idlUtils = require("../generated/utils");
-const NodeImpl = require("./Node-impl");
+const NodeImpl = require("./Node-impl").implementation;
+const { changeAttribute } = require("./Node-impl");
```
ExE-Boss: Ditto

test/web-platform-tests/to-upstream/dom/attributes-are-nodes.html
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Attributes are Nodes but should not be accepted outside of the `attributes` NamedNodeMap</title>
```
ExE-Boss: I’d like to move this file in a separate commit to preserve `git blame` history, like #2764.

lib/jsdom/living/attributes/Attr-impl.js
domenic: Comments like this make it rather frustrating to work with you and discourage me from doing so going forward.

domenic:
We'll consider this on hold then until I or someone else is able to make the time to fix the circular dependency issues properly.

lib/jsdom/living/attributes/Attr-impl.js
ExE-Boss: Well, they&nbsp;require searching&nbsp;through Node’s&nbsp;`require`&nbsp;cache, which&nbsp;adds an&nbsp;unnecessary performance&nbsp;hit, when&nbsp;you&nbsp;can&nbsp;just&nbsp;do:

```js
const {
	implementation: NodeImpl,
	setAnExistingAttributeValue
} = require("../nodes/Node-impl.js");
```

Which&nbsp;requires&nbsp;doing so&nbsp;only&nbsp;once.

domenic: I wish you would stop making performance claims without measurements of how much they affect real-world programs.

ExE-Boss: _**Every**_&nbsp;`require(…)`&nbsp;call has&nbsp;some performance&nbsp;overhead, since&nbsp;**Node** has&nbsp;to&nbsp;search its&nbsp;cache, which&nbsp;has an&nbsp;<code>O(log&nbsp;<var>n</var>)</code> complexity&nbsp;at&nbsp;best, so&nbsp;it&nbsp;stands to&nbsp;reason that&nbsp;avoiding duplicate&nbsp;requires will&nbsp;keep the&nbsp;startup time&nbsp;lower&nbsp;than if&nbsp;duplicate&nbsp;requires are&nbsp;used.

ExE-Boss:
Fixing&nbsp;this&nbsp;circular&nbsp;dependency is&nbsp;impossible, as&nbsp;it’s&nbsp;caused&nbsp;by:

`Node‑impl.js` → `node.clone` → `attributes.copyAttributeList` → `attrGenerated.createImpl` → `Attr‑impl.js` → `Node‑impl.js`

Moving&nbsp;`copyAttributeList` to&nbsp;`node` or&nbsp;`Node‑impl.js` won’t&nbsp;change&nbsp;anything.

ExE-Boss:
In&nbsp;fact, using&nbsp;[madge](https://www.npmjs.com/package/madge) to&nbsp;measure dependency&nbsp;cycles, I’ve&nbsp;determined that&nbsp;moving&nbsp;`copyAttributeList` to&nbsp;`living/node.js` increases&nbsp;the&nbsp;number of&nbsp;circular&nbsp;dependencies from&nbsp;9&nbsp;to&nbsp;10.

lib/jsdom/living/attributes/Attr-impl.js
domenic: I fundamentally disagree with your approach to performance, and this further increases my doubt about working with you. I can only say that this unprofessional approach would not be accepted at major software development companies.

domenic:
I am confident that the circular dependencies can be detangled. If you cannot do it, that just means we'll have to wait on someone who can.

ExE-Boss:
I’ve&nbsp;managed to&nbsp;get the&nbsp;number of&nbsp;circular&nbsp;dependencies down&nbsp;to&nbsp;7, which&nbsp;is&nbsp;less&nbsp;than&nbsp;`master`, which&nbsp;has&nbsp;8 by&nbsp;removing&nbsp;the&nbsp;following&nbsp;one, as&nbsp;reported&nbsp;by&nbsp;[madge](https://www.npmjs.com/package/madge):

1) jsdom/living/generated/Attr.js > jsdom/living/attributes/Attr-impl.js > jsdom/living/attributes.js

test/web-platform-tests/to-upstream/dom/attributes-are-nodes.html
TimothyGu: You could preserve the history here by making it a separate commit. We will then rebase while merging.

ExE-Boss: I&nbsp;don’t&nbsp;think it&nbsp;really&nbsp;matters that&nbsp;much, unlike&nbsp;`XMLHttpRequest`.

lib/jsdom/living/elements.js
```diff
@@ -0,0 +1,39 @@
+"use strict";
```
TimothyGu: I’d like to establish some kind of rule for what goes into this file and what goes into attributes.js. Like “things that have to do with attributes and elements but don’t require access to … module goes here”. Add this as a comment to both this file and attributes.js.

Additionally, make sure this rule is enforced.

For instance, if the rule is that anything that requires access to the generated Attr class needs to go into attributes.js, then you should say that and move everything that doesn’t need it into this file – though it looks like it would be better to move `setAttributeValue` _out_ of attributes.js in that case.

Looking further, we could probably just change `setAttributeValue` to instead use the new `document._createAttribute` to get rid of the Attr import. That might even eliminate the requirement of a separate file.

lib/jsdom/living/interfaces.js
```diff
@@ -15,8 +15,8 @@ const generatedInterfaces = [
   require("./generated/EventTarget"),
 
   require("./generated/NamedNodeMap"),
-  require("./generated/Attr"),
   require("./generated/Node"),
+  require("./generated/Attr"),
```
TimothyGu: Is this still necessary? If so, add a comment for why it is.

lib/jsdom/living/node.js
```diff
@@ -69,6 +73,21 @@ module.exports.clone = function (node, document, cloneChildren) {
   return copy;
 };
 
+function copyAttributeList(sourceElement, destElement, document) {
+  // Needed by https://dom.spec.whatwg.org/#concept-node-clone
```
TimothyGu: This doesn’t seem to be a separate function according to the spec. Instead, it’s folded into the cloning algorithm directly. See https://dom.spec.whatwg.org/#concept-node-clone.

Additionally, the spec just uses “clone” to create the attributes these days. So no need to use document._createAttribute below.

lib/jsdom/living/attributes/Attr-impl.js
TimothyGu: 1. In software engineering, it’s not _polite_ to impose your own programming style onto a project of which you are not the creator or maintainer. This is especially true in open source, which provides you with the _privilege_ (not right) of contributing.

2. Node.js uses a hash table as the cache, so a lookup is generally constant time. Even if a file system lookup is required, that would be O(_N_) time where _N_ is the level of file directory needed to traverse to find the resulting file. Here, _N_ is a constant anyway, so it’s O(1) there too.

3. We welcome performance improvements that have to do with requires, but we are also data-oriented. We want to see hard numbers before abandoning our established style. See https://github.com/jsdom/jsdom/pull/2783 for an excellent investigation and PR done by @pmdartus that provides a noticeable start-up speed-up.

4. It is an established style of ours to not use object pattern destructuring where the source property name is different from the bound identifier.

Due to these four reasons, we ask you to change this back. I also hope you may be more receptive to such review comments in the future, not just in our project but also in others. Feel free to contact me through IRC (I’m on Freenode #jsdom) or email if you have any questions.

lib/jsdom/living/interfaces.js
ExE-Boss: Yes, this&nbsp;is&nbsp;necessary, otherwise&nbsp;`AttrGenerated.install()` will&nbsp;throw an&nbsp;error because the&nbsp;`Node`&nbsp;constructor has&nbsp;to&nbsp;be&nbsp;installed before&nbsp;attempting to&nbsp;install the&nbsp;`Attr`&nbsp;constructor.

lib/jsdom/living/nodes/Node-impl.js
```diff
@@ -93,7 +117,7 @@ function isHostInclusiveAncestor(nodeImplA, nodeImplB) {
   return false;
 }
 
-class NodeImpl extends EventTargetImpl {
+exports.implementation = class NodeImpl extends EventTargetImpl {
```
TimothyGu: Revert this.

```suggestion
class NodeImpl extends EventTargetImpl {
```

lib/jsdom/living/nodes/Node-impl.js
```diff
@@ -1078,8 +1132,4 @@ class NodeImpl extends EventTargetImpl {
       this._childTextContentChangeSteps();
     }
   }
-}
-
-module.exports = {
-  implementation: NodeImpl
```
TimothyGu: Revert this.

lib/jsdom/living/nodes/Node-impl.js
```diff
@@ -336,15 +362,43 @@ class NodeImpl extends EventTargetImpl {
...
+      node1 = attr1._element;
+    }
+
+    if (node2.nodeType === NODE_TYPE.ATTRIBUTE_NODE) {
+      attr2 = node2;
+      node2 = attr2._element;
+
+      if (attr1 !== null && node2 === node1) {
+        for (const attr of node2._attributeList) {
+          if (attr === attr1) {
```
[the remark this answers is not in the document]
ExE-Boss: Well, `SymbolTree`&nbsp;uses strict&nbsp;equality for&nbsp;the&nbsp;other&nbsp;nodes.

lib/jsdom/living/nodes/Node-impl.js
ExE-Boss: Well, 57&nbsp;instances use&nbsp;`exports.implementation`, the&nbsp;rest&nbsp;re‑assigns&nbsp;`module.exports`.

I&nbsp;prefer&nbsp;avoiding re‑assigning&nbsp;`module.exports` wherever&nbsp;possible and&nbsp;this&nbsp;was&nbsp;necessary before&nbsp;<https://github.com/jsdom/jsdom/commit/59b7dcf67f00aaa9a69dc81858aa3090265f3e6e>.

TimothyGu: Well arguably all intermediate parts of this PR are "broken" in some ways – and that's why they are not merged; it's just that clicking the "commit suggestion" button makes it more broken.

If you are so concerned about this, why don't you apply the change manually then?

domenic:
Regardless of my frustration about the process of working with you on this, I do want to say thank you for taking it on, and that I am happy with the end result. It's a pretty good improvement that solves a ~3 year old bug report.

Thanks as well to @TimothyGu for stepping in when I was burning out on the process.

PR #2771, opened 2020-01-02
fix(Window): Avoid re‑assigning `module.exports`

ExE-Boss:
This&nbsp;also&nbsp;allows us&nbsp;to&nbsp;move the&nbsp;`require("../living/interfaces")`&nbsp;call to&nbsp;the&nbsp;top.

Note&nbsp;that this&nbsp;will once&nbsp;again cause&nbsp;issues once&nbsp;**Window** is&nbsp;converted to&nbsp;**WebIDL** unless&nbsp;<https://github.com/jsdom/webidl2js/pull/154> is&nbsp;merged&nbsp;first.

---

Extracted&nbsp;from&nbsp;<https://github.com/jsdom/jsdom/pull/2548>.

---

review?(@pmdartus, @TimothyGu)

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -33,7 +33,7 @@ class JSDOM {
 
     options = transformOptions(options, encoding, mimeType);
 
-    this[window] = new Window(options.windowOptions);
+    this[window] = Window.createWindow(options.windowOptions);
 
     const documentImpl = idlUtils.implForWrapper(this[window]._document);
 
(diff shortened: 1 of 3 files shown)


domenic:
I'd rather figure out a way to break the cycle. In my opinion this just replaces one hack around circular requires with another hack around circular requires.

ExE-Boss:
Adding&nbsp;properties to&nbsp;`exports` isn’t&nbsp;a&nbsp;hack around&nbsp; circular&nbsp;requires, as&nbsp;that's&nbsp;also what’s&nbsp;closer to&nbsp;how&nbsp;**ECMAScript** modules&nbsp;behave.

domenic:
I disagree.

PR #2777, opened 2020-01-04
Add live collection tests

### LINKED ISSUE #77: Live NodeList isn't updated after DOM manipulation
Run the following snippet to reproduce the bug:

```
// jsdom bug: Live NodeList isn't updated after DOM manipulation
// node.js v0.2.4
// jsdom@0.1.20

var jsdom = require('jsdom');
var html = '<html><body><p id="p1"></p><p id="p2"></p></body></html>';
var window = jsdom.jsdom(html).createWindow();
var document = window.document;

var all = document.getElementsByTagName('*');
var i = 2;
var node = all[i];
console.log(''+node); //P#p1
node.parentNode.removeChild(node);

console.log(''+all[i]); //still P#p1. the live NodeList wasn't updated properly
all.length; //trigger a refresh
console.log(''+all[i]); //P#p2 OK
```

Interestingly, accessing the length property will trigger a "refresh".

pmdartus:
Add tests for live `NodeList` and `HTMLCollection` to ensure that the collections are always in sync when the DOM gets updated.
Fixes #77.

### DIFF
--- a/test/web-platform-tests/to-upstream/dom/nodes/HTMLCollection-live-mutations.html
+++ b/test/web-platform-tests/to-upstream/dom/nodes/HTMLCollection-live-mutations.html
@@ -1,0 +1,103 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>HTMLCollection live mutations</title>
+<link rel="help" href="https://dom.spec.whatwg.org/#ref-for-concept-collection-live">
+<script src="/resources/testharness.js"></script>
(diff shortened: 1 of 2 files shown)


domenic:
Checked that they pass in browsers via my time-honored technique of copying-and-pasting the file into jsbin.com, and then prefixing the `<script src="">s` with `https://wpt.live/`.

pmdartus:
The PR on the upstream is opened: https://github.com/web-platform-tests/wpt/pull/21125

Issue #2390, opened 2018-10-11
Reduce the size of jsdom (and its dependencies)


Zirro:
I occasionally see the size of jsdom cited as a reason to avoid it and prefer lighter packages instead. While an extensive implementation of the web standards will never be tiny, I believe that the total installation size of jsdom (13 MB / 21,9 MB) could be cut in half by merely reducing bloat.

The sizes shown within parentheses are with and without disk compression, and accounting for the removal of subdependencies as well. These are the three most important changes we could make to reduce our installed size (with disk compression):

**request/request-promise-core - (5,4 MB / 10,6 MB)**

The big one. `request` includes many large dependencies which we have no need for. `request-promise-core` contributes a significant chunk on its own, since it depends on the whole `lodash` package (1,4 MB / 5,1 MB) and has yet to merge https://github.com/request/promise-core/pull/7.

Replacing these with either a lighter alternative dependency or a proper implementation of the Fetch Standard would be a large win.

**acorn/acorn-globals/escodegen - (2,7 MB / 3 MB)**

The `acorn`, `acorn-globals` and `escodegen` dependencies are only used when jsdom is browserified.  While we should continue to support browserification, it is an uncommon way to use jsdom.

I believe it makes sense to make these dependencies optional the same way the same canvas-related dependencies are.

**jsdom/living/generated/ - (1,2 MB / 1,5 MB)**

With all the dependencies above removed, the largest contributor to size is actually jsdom's `generated` directory. While we obviously need this code in some form, there seems to be a lot of room for deduplication of code between these files.

domenic:
> Replacing these with either a lighter alternative dependency or a proper implementation of the Fetch Standard would be a large win.

For sure. Just a lot of work :).

> I believe it makes sense to make these dependencies optional the same way the canvas-related dependencies are.

I would rather find a way to use the `browser` package.json field so that these don't get loaded at runtime for Node consumers. I think I'm OK with Node consumers _downloading_ extra packages, in order to give browser consumers a better out of the box experience, as long as it doesn't impact startup time.

> there seems to be a lot of room for deduplication between these files.

I'd be curious to see what concrete proposals there are, but in theory this seems doable. I'm not sure we'd want to e.g. minify the output though...

Zirro:
> I would rather find a way to use the browser package.json field so that these don't get loaded at runtime for Node consumers.

[...]

Issue #1724, opened 2017-01-31
Add GlobalFetch API - Headers, Request, Response


piuccio:
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
https://developer.mozilla.org/en-US/docs/Web/API/Headers
https://developer.mozilla.org/en-US/docs/Web/API/Request
https://developer.mozilla.org/en-US/docs/Web/API/Response

jods4:
+1 would be nice as General availability in browsers means more web apps use it and fail in JSDOM.
Given that JSDOM implements `XMLHttpRequest`, it could be as simple as running `whatwg-fetch` polyfill at startup, right?

domenic:
No, unfortunately the whatwg-fetch polyfill is not very spec compliant. Someone needs to write an implementation of the fetch algorithm in the standard: https://fetch.spec.whatwg.org/#fetching

jods4:
Yes, for now I'm using the whatwg polyfill in JSDOM.
I'm not sure how much "non-compliant" it is, but it seems to work well enough for my case.

It would be nice to see something built-in, though. Especially since the `Fetch` API has become commonplace.

domenic:
Yeah, it definitely would be nice! I hope someone is able to pull-request an implementation of the spec.

jhabdas:
Here's a tool which might help with the tests for the PR: https://www.npmjs.com/package/fetch-inject. Of course it's not unit tested because it requires this issue to be closed before it can be. 😉

KyleGalvin:
My application tests run in node rather than the browser. The lack of server-side fetch API (and typescript bindings to match) has become a show-stopper.

This issue looks like the most promising solution.

Does anyone else know of a workaround for the mean-time? I can use fetch in the browser. I can use node-fetch for tests, but I can't switch on the two packages dynamically. One environment or the other will always fail.

ghostd:
Is this issue still relevant since v15 supports `fetch`?

domenic:
v15 does not support fetch.

ghostd:
Sorry, i misread the Changelog about the 15.1, only Headers was provided

thiscompsciguy:
+1 to this. fetch is natively [supported in nearly all modern browser](https://caniuse.com/#search=fetch)

Dscheksn:
> Does anyone else know of a workaround for the mean-time? 

My workaround looks like this:

```
    const firstTodo: Todo = { title: 'Todo 1' };
[...]

Issue #1454, opened 2016-04-19
Improve startup performance.


cpojer:
jsdom's require performance is quite slow. It takes about 500ms on my MBP" and it is the slowest part in Jest's startup. If we could reduce this to 100ms, that would be super awesome.

Example:
```js
console.time('jsdom');
require('jsdom');
console.timeEnd('jsdom');
jsdom: 490.010ms
```

Could we try to compile jsdom with inline-requires ( https://github.com/facebook/fbjs/blob/master/scripts/babel-6/inline-requires.js )? It has helped facebook.com's startup performance. This of course is only useful if a lot of things in jsdom are required but not used when creating an instance of jsdom.

domenic:
The one thing I worry about with inline requires, and lazy requiring in general, is that it's essentially adding a *bunch* of statSync/readFileSync calls inside methods which are generally not expected to do I/O or to block the main thread. Startup time is expected to do I/O, but e.g. doing `jsdom.jsdom()` is not.

I would be interested in doing some profiling here and figuring out who the worst offenders are. I feel like some other issue did some investigation and a big chunk of the time was in parse5's entity table...

cpojer:
Yeah, that makes sense. As said, it only works if most of the dependencies are *not* actually used during normal use of jsdom.

Is this something you are going to look into any time soon? If you know of one or two potential slowdowns during startup and can fix those, that would be very helpful.

Sebmaster:
Require times:
```
total: 506.276ms
505ms lib/jsdom.js
  408ms lib/jsdom/browser/Window.js
    153ms node_modules/cssstyle/lib/CSSStyleDeclaration.js
      7ms node_modules/cssom/lib/index.js
        6ms node_modules/cssom/lib/CSSStyleDeclaration.js
          5ms node_modules/cssom/lib/parse.js
            2ms node_modules/cssom/lib/CSSStyleSheet.js
              1ms node_modules/cssom/lib/CSSStyleRule.js
                1ms node_modules/cssom/lib/CSSRule.js
            1ms node_modules/cssom/lib/CSSImportRule.js
            1ms node_modules/cssom/lib/CSSKeyframesRule.js
            1ms node_modules/cssom/lib/CSSDocumentRule.js
        1ms node_modules/cssom/lib/clone.js
      3ms node_modules/cssstyle/lib/properties.js
        3ms node_modules/cssstyle/lib/properties/colorRendering.js
[...]

PR #2783, opened 2020-01-08
Remove unnecessary dependencies

pmdartus:
This PR drops the following dependencies to speedup jsdom evaluation:
- `array-equal`
- `pn` 
- ~`request-promise-native`~ See comment below.
- `q`

~We can measure an 18% decrease in jsdom evaluation time with the proposed changes. It can be measured using the following command: `node -e "console.time('eval');require('./lib/api.js');console.timeEnd('eval');"`~
- ~master branch: 392.369ms~
- ~PR branch: 320.972ms~

Related to #1454 and #2390.

---- 

As a side note, I also measured jsdom evaluation time in other scenarios:
1. above changes + replacing `request` with `http`, `https` and `zlib`: 255.997ms
2. above changes + lazy-evaluating `w3c-hr-time`: 241.005ms
3. above change + lazy-evaluating `cssstyle`: 228.146ms
4. above change + lazy-evaluating jsdom interfaces: 161.444ms

Scenario 1. brings the most meaningful improvement with the lowest risk. It might be interesting to look into dropping `request` when implementing #1724. Lazily evaluating the interfaces (2., 3. and 4.) as they are accessed on the global object is also something worth investigating, but it feels way more complex to get it right compared to scenario 1.


domenic:
The one I'm unsure about here is removing request-promise-native. The others are clear wins.

pmdartus:
The bulk of the performance gain comes from the request-promise-native removal. Here is the profiling of jsdom evaluation.

Screen Shot 2020-01-09 at 7 30 58 AM

On this profile, we can observe that request is evaluated twice. This comes from the fact that request-promise-native bypasses the node require cache when evaluating request: https://github.com/request/request-promise-native/blob/master/lib/rp.js#L6-L12. The removal of the other dependencies doesn't bring much in terms of evaluation performance.

Still trying to understand why some unrelated tests started to fail.

pmdartus:
After some debugging, I finally was able to find the origin of the test failure. This is indirectly related to the removal of request-promise-native. The following code snippet exhibit the bug in request:

```js
"use strict";
const http = require("http");
const request = require("request");

const server = http.createServer((req, res) => {
  res.writeHead(200, [["date", "0"]]);
  res.end("<body></body>");
});

server.listen(8000, () => {
  request("http://localhost:8000", {
    strictSSL: false,
  }, (err, response) => {
    console.log('>> Connection is expected to close', response.headers.connection);

    request("http://localhost:8000", {
      forever: true,
    }, (err, response) => {
      console.log('>> Connection is expected to keep alive', response.headers.connection);

      request("http://localhost:8000", {
      }, (err, response) => {
        console.log('>> Connection is expected to close', response.headers.connection);
        server.close();
      });
    });
  });
});
```

The expected output is:

```
>> Connection is expected to close close
>> Connection is expected to keep alive keep-alive
>> Connection is expected to close close
```

While the actual output is:

```
>> Connection is expected to close close
>> Connection is expected to keep alive close
>> Connection is expected to close close
```

The issue comes from the fact that request is reusing the same `http.Agent` instance between the invocations. The first request dictates the `keepAlive` setting for all the subsequent requests. This issue only appears now, because of the request package being evaluated twice (see the previous comment).

lib/api.js
```diff
@@ -145,11 +145,15 @@ class JSDOM {
   }
 
   static fromFile(filename, options = {}) {
-    return Promise.resolve().then(() => {
+    return new Promise((resolve, reject) => {
       options = normalizeFromFileOptions(filename, options);
 
-      return fs.readFile(filename).then(buffer => {
-        return new JSDOM(buffer, options);
+      fs.readFile(filename, (err, buffer) => {
```
domenic: You should be able to use `fs.promises` now.

lib/api.js
```diff
@@ -145,11 +145,15 @@ class JSDOM {
   }
 
   static fromFile(filename, options = {}) {
-    return Promise.resolve().then(() => {
+    return new Promise((resolve, reject) => {
```
domenic: I think this will break cause an error to be thrown, instead of a promise to be rejected, if an invalid filename is given. (I.e. one that causes URL parsing failures.) I'm surprised this wasn't tested.

A better solution here would be to move to async functions.

pmdartus:
You are right, I think it would be better to get the right approach. My first impression would be to create a new pool per jsdom instance. I will create a new issue to get some feedback first.

In the meantime, I have reverted the `request-promise-native` to first merge the non-controversial package removal. 

> But I'm not feeling super good about the result. We've gone from being able to use a nice, maintained, third-party promise-based request library, to creating our own manual wrapper in two separate parts of the code, including a confusing temporary workaround.

On this topic, request is currently in maintenance mode (https://github.com/request/request/issues/3142), and the probability of this [agent issue](https://github.com/request/request/issues/3257) getting fixed is extremely low looking at the number of opened PRs. IMO, it would be interesting to evaluate how much effort it would be to use the standard `http`, `https` and `zlib` modules instead of using `request`.

domenic:
>  IMO, it would be interesting to evaluate how much effort it would be to use the standard http, https and zlib modules instead of using request.

I think this is probably the right way to go for the project in the end.

In the past we've talked about doing this as part of implementing the Fetch standard from scratch. Then we'd use that to underlie both our XHR implementation and any resource fetching we do---exactly like how the spec is layered.

However, perhaps that ambition has just caused this project to be put off for too long. I'd be open to implementing our own internal "request-lite" that doesn't implement full Fetch semantics and just gives us the parts of request that we need.

I'd also be curious on @TimothyGu's thoughts on using node-fetch. I believe in the past we've decided it wouldn't be a great fit to implement `window.fetch()` in the jsdom itself, but as a replacement for how we currently use request, maybe it'd be OK?

TimothyGu:
> as a replacement for how we currently use request, maybe it'd be OK?

Hmm, this sounds possible. Filed #2792 to further investigate.

lib/api.js
ExE-Boss: Actually, a thrown error inside a `Promise` constructor’s `executor` callback will still result in a rejected `Promise`: https://tc39.es/ecma262/#sec-promise-executor

This only has bearing on the execution timing (`executor` is executed synchronously, whereas `then` callbacks are executed as part of the job queue).

PR #2786, opened 2020-01-09
More <input> clean-ups

TimothyGu:
Most of these changes are not observable and for consistency benefits only. Two more substantive changes are implementing `<input>` value rewrite during `type` change, and proper validation support for `<input type=time min max>` when `min > max`.

Tests for reversed range are in https://github.com/web-platform-tests/wpt/pull/21103, which is included in this roll.


lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -95,6 +95,11 @@ function valueAttributeMode(type) {
   return "value";
 }
 
+function getTypeFromAttribute(typeAttribute) {
+  const type = typeAttribute && typeAttribute.toLowerCase();
```
domenic: This is probably a preexisting problem elsewhere in the file, but I believe this will give the wrong result for "ſearch" where the "ſ" is "LATIN SMALL LETTER LONG S", or "weeK" where the "K" is "KELVIN SIGN".

Other k or s-containing types are checkbox, reset, password, submit... I think that's all of them.

TimothyGu: Indeed, I got the code from the original `type` getter. I'll fix it.

test/web-platform-tests/to-upstream/html/semantics/forms/the-input-element/default-type-dont-upstream.html
```diff
@@ -1,33 +1,87 @@
 <!DOCTYPE html>
-<meta charset="utf-8" />
+<meta charset=utf-8>
 <!-- This test should not be upstreamed, as it tests the same thing as html/dom/reflection-forms.html. -->
```
domenic: I'm not sure this is true anymore as I don't think the reflection helper tests Ks and Ss.

This also doesn't seem to test Ss, or test all the possible cases, just one with week, hmm.

TimothyGu: It looks like `"ſ".toLowerCase()` is still `"ſ"`, but I added U+0130 LATIN CAPITAL LETTER I WITH DOT ABOVE (İ).

TimothyGu: Also tested all possible cases.

PR #2789, opened 2020-01-10
Final clean-ups in <input>

TimothyGu:
This is rebased on top of #2786. Creating a draft for Travis to take a look at it first.

### DIFF
--- a/lib/jsdom/living/helpers/strings.js
+++ b/lib/jsdom/living/helpers/strings.js
@@ -1,5 +1,6 @@
 "use strict";
 
+// https://infra.spec.whatwg.org/#ascii-whitespace
 const asciiWhitespaceRe = /^[\t\n\f\r ]$/;
 exports.asciiWhitespaceRe = asciiWhitespaceRe;
 
@@ -53,9 +54,17 @@ const floatingPointNumRe = /^-?(?:\d+|\d*\.\d+)(?:[eE][-+]?\d+)?$/;
 exports.isValidFloatingPointNumber = str => floatingPointNumRe.test(str);
 
 // https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#rules-for-parsing-floating-point-number-values
+// Error is represented as null.
 exports.parseFloatingPointNumber = str => {
+  // The implementation here is slightly different from the spec's. We need to use parseFloat() in order to retain
+  // accuracy, but parseFloat() trims Unicode whitespace in addition to just ASCII ones, so we make sure that the
+  // trimmed prefix contains only ASCII whitespace ourselves.
+  const numWhitespace = str.length - str.trimStart().length;
+  if (/[^\t\n\f\r ]/.test(str.slice(0, numWhitespace))) {
+    return null;
+  }
   const parsed = parseFloat(str);
-  return isFinite(parsed) ? parsed : NaN;
+  return isFinite(parsed) ? parsed : null;
 };
 
 // https://infra.spec.whatwg.org/#split-on-ascii-whitespace
--- a/lib/jsdom/living/nodes/HTMLInputElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLInputElement-impl.js
@@ -478,8 +478,8 @@ class HTMLInputElementImpl extends HTMLElementImpl {
       return NaN;
     }
 
-    const parsedValue = this._parsedValue;
-    return typeof parsedValue === "number" ? parsedValue : NaN;
+    const parsedValue = this._convertStringToNumber(this._value);
+    return parsedValue !== null ? parsedValue : NaN;
   }
 
   set valueAsNumber(v) {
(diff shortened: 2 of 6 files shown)


lib/jsdom/living/helpers/strings.js
```diff
@@ -1,8 +1,13 @@
 "use strict";
 
+// https://infra.spec.whatwg.org/#ascii-whitespace
 const asciiWhitespaceRe = /^[\t\n\f\r ]$/;
 exports.asciiWhitespaceRe = asciiWhitespaceRe;
 
+// https://infra.spec.whatwg.org/#ascii-digit
+const asciiDigitRe = /^[0-9]$/;
+exports.asciiDigitRe = asciiDigitRe;
```
domenic: This appears to be unused?

lib/jsdom/living/nodes/HTMLProgressElement-impl.js
```diff
@@ -10,38 +10,58 @@ class HTMLProgressElementImpl extends HTMLElementImpl {
...
+        return parsedMax;
+      }
     }
-
     return 1.0;
   }
   set max(value) {
-    this.setAttributeNS(null, "max", value);
+    if (value > 0) {
+      this.setAttributeNS(null, "max", String(value));
```
domenic: The string conversion is not necessary due to https://github.com/jsdom/jsdom/commit/4285e7074367c56dda07d3c08d6528ff0651cd4d#diff-802e42090f68edfaab3d0f57d32de76bR249

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -144,27 +144,22 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
-      if (v > candidate) {
-        candidate += this._allowedValueStep;
-      }
-    } else if (v < candidate) {
-      candidate -= this._allowedValueStep;
-    }
-    return candidate + this._stepBase;
+    const allowedValueStep = this._allowedValueStep;
+    const stepBase = this._stepBase;
+    return new Decimal(v).minus(stepBase)
```
domenic: I think these changes should be tested.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -967,7 +953,8 @@ class HTMLInputElementImpl extends HTMLElementImpl {
       // than the minimum, the element is simultaneously suffering from an underflow and suffering
       // from an overflow.
       const reversedRangeSufferingOverUnderflow = () => {
-        return this._parsedValueIsNumber && this._parsedValue > this._maximum && this._parsedValue < this._minimum;
+        const parsedValue = this._convertStringToNumber(this._value);
+        return typeof parsedValue === "number" && parsedValue > this._maximum && parsedValue < this._minimum;
```
domenic: Can we replace the typeof === number with !== null?

lib/jsdom/living/helpers/strings.js
TimothyGu: Indeed. Removed.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
TimothyGu: Done.

lib/jsdom/living/nodes/HTMLProgressElement-impl.js
TimothyGu: Sure. I'd want to eventually remove the type cast though. Maybe after reflection gets re-done.

domenic: I wouldn't be comfortable doing that unless we had some sort of static enforcement (e.g. lint rule) or auto-generated testing infrastructure to catch future errors... it's just too easy to forget that the useful behavior from the wrapper is omitted in the impl...

domenic:
The tests are failing here, and I still think that we need tests for the decimal changes.

TimothyGu:
I spent quite a bit of time trying to make a test case for the decimal change but so far I haven’t been able to. I did uncover some discrepancies between Firefox and Chrome with regards to precision though, even they are both using a big decimal implementation. (The same test has the results in jsdom before/after the change.) The tricky part is that the spec doesn’t seem to mandate a specific level of precision, so it’s unclear who exactly is correct.

I’ll fix the test failure when I get to it.

TimothyGu:
In light of https://github.com/whatwg/html/issues/5207, I'd rather not add tests for the decimal change. We already know that it "works" in the sense that it doesn't regress what used to work. Otherwise, fixed the test failures.

domenic:
But, why include it at all, if it doesn't help us in a testable observable way? What if we regress, with no tests to catch us? I'd rather remove that change, if we don't think it's worth testing.

domenic:
I'm going to pull the other two commits into master :)

domenic:
Oh, I guess that isn't so straightforward, since the third depends on the under-discussion second. Nevermind, for now!

TimothyGu:
Okay, I've removed the controversial change to `_stepAlign` and the commit is now a simple refactoring. See if this works better.

PR #2794, opened 2020-01-13
[fix] noscript child should be considered as text when runScripts is set to "dangerously"

ACHP:
This PR aims to address the longstanding issues with noscript nodes.(https://github.com/jsdom/jsdom/issues/1611, https://github.com/jsdom/jsdom/issues/1802).
By setting the `scriptingEnabled` flag to `false` we can now query nodes inside `<noscript/>`

I didn't find any places in the code where this feature could have been implemented so I did it ( I tried at least)
Hope everything is in the right place and nothing is missing.

Thx for this awesome library ;)


ACHP:
> Also, can you add a to-upstream web platform test? Those are run with `runScripts: "dangerously"` so you'd be adding a test that noscript children end up as text.
> 
> Thanks!

I have made the changes but I'm really not sure about my test, tell me if it's not in the right place

test/api/options-run-scripts.js
```diff
@@ -95,6 +95,25 @@ describe("API: runScripts constructor option", () => {
...
+      assert.strictEqual(document.querySelector("noscript").children.length, 0);
+      assert.strictEqual(document.querySelector("noscript").innerHTML, "<div></div>");
+    });
+    it("should be reachable when runScripts is not set to \"dangerously\"", () => {
+      const dom = new JSDOM(`<body><noscript><div></div></noscript></body>`).window;
+      const { document } = dom.window;
+
+      assert.strictEqual(document.querySelector("noscript").children.length, 1);
+      assert.instanceOf(document.querySelector("noscript").children[0], dom.window.HTMLDivElement);
+    });
```
TimothyGu: Could you add a test for `outside-only` also? Should be the same as "not set to `dangerously`".

domenic: I added and pushed :)

PR #2796, opened 2020-01-15
test(wpt): Improve “expected fail” message and “needs‑node*” code

ExE-Boss:
Currently, each&nbsp;**Node.js**&nbsp;version needs&nbsp;to&nbsp;have its&nbsp;own&nbsp;`needs‑node*`&nbsp;reason added&nbsp;to&nbsp;`run‑wpts.js`.

This&nbsp;is&nbsp;problematic, as&nbsp;the&nbsp;code currently&nbsp;performs the&nbsp;`Number(process.versions.node.split(".")[0])`&nbsp;conversion for&nbsp;every&nbsp;supported&nbsp;`needs‑node*`&nbsp;entry, and&nbsp;each&nbsp;`needs‑node*`&nbsp;entry has&nbsp;to&nbsp;be&nbsp;manually&nbsp;added.

This&nbsp;simplifies&nbsp;that to&nbsp;only&nbsp;need to&nbsp;perform that&nbsp;conversion&nbsp;once.

I’ve&nbsp;also&nbsp;added a&nbsp;fail&nbsp;reason to&nbsp;<code>[expected&nbsp;fail]</code> when&nbsp;the&nbsp;reason is&nbsp;`needs‑node*` or&nbsp;`fail‑with‑canvas`.

### DIFF
--- a/test/web-platform-tests/run-wpts.js
+++ b/test/web-platform-tests/run-wpts.js
@@ -56,7 +56,7 @@ describe("web-platform-tests", () => {
           const reason = matchingPattern && toRunDoc[matchingPattern][0];
           const shouldSkip = ["fail-slow", "timeout", "flaky", "mutates-globals"].includes(reason) ||
                              (["fail-with-canvas", "needs-canvas"].includes(reason) && !hasCanvas);
-          const needsNodeVersion = reason.startsWith("needs-node") ?
+          const needsNodeVersion = typeof reason === "string" && reason.startsWith("needs-node") ?
             Number(reason.substring(10 /* "needs-node".length */)) :
             null;
           const expectFail = (reason === "fail") ||
--- a/test/web-platform-tests/run-wpts.js
+++ b/test/web-platform-tests/run-wpts.js
@@ -18,7 +18,6 @@ const validReasons = new Set([
   "needs-node10",
   "needs-node11",
   "needs-node12",
-  "needs-node13",
   "needs-canvas"
 ]);
 
@@ -26,7 +25,6 @@ const nodeMajor = Number(process.versions.node.split(".")[0]);
 const hasNode10 = nodeMajor >= 10;
 const hasNode11 = nodeMajor >= 11;
 const hasNode12 = nodeMajor >= 12;
-const hasNode13 = nodeMajor >= 13;
 const hasCanvas = Boolean(Canvas);
 
 const manifestFilename = path.resolve(__dirname, "wpt-manifest.json");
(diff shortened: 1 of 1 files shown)


domenic:
I find this significantly less clear than the original code. I'd like to continue maintaining the separate needs-node code, so please revert all of that.

Improving the messages seems reasonable though.

(In general, it's better etiquette when working on an open source project to open an issue to discuss a proposed change, than to immediately send a PR. Especially if the proposed change involves subjective judgments about "improving" the code.)

test/web-platform-tests/run-wpts.js
```diff
@@ -64,12 +67,19 @@ describe("web-platform-tests", () => {
...
 
           if (matchingPattern && shouldSkip) {
             specify.skip(`[${reason}] ${testFile}`);
           } else if (expectFail) {
-            runSingleWPT(testFilePath, `[expected fail] ${testFile}`, expectFail);
+            let failReason = "";
+            if (reason.startsWith("needs-node")) {
+              failReason = `: ${reason}`;
+            } else if (reason === "fail-with-canvas") {
+              failReason = ": canvas bug";
```
domenic: Hmm, do you think "canvas bug" is more helpful than "fail-with-canvas"? I like the idea of people reading the results being able to directly map the output to the .yml file contents, so it'd be good to align on one of those two, instead of translating between them. (Or a third alternative.)

test/web-platform-tests/run-wpts.js
```diff
@@ -18,12 +18,15 @@ const validReasons = new Set([
   "needs-node10",
   "needs-node11",
   "needs-node12",
+  "needs-node13",
```
domenic: Please don't add a new fail reason that's not in use. We should be trying to drive down the number of Node-version-specific failures, not increase them.

test/web-platform-tests/run-wpts.js
```diff
@@ -69,7 +70,11 @@ describe("web-platform-tests", () => {
           if (matchingPattern && shouldSkip) {
             specify.skip(`[${reason}] ${testFile}`);
           } else if (expectFail) {
-            runSingleWPT(testFilePath, `[expected fail] ${testFile}`, expectFail);
+            let failReason = "";
+            if (reason !== "fail") {
+              failReason = `: ${reason}`;
+            }
```
domenic: Could you change this to

```js
            const failReason = reason !== "fail" ? `: ${reason}` : "";
```

? I tried to do so myself but it said

```
! [remote rejected]   EB-Forks-test/wpt/improve-expected-fail-code -> refs/pull/2796/head (deny updating a hidden ref)
```

test/web-platform-tests/run-wpts.js
```diff
@@ -69,7 +70,13 @@ describe("web-platform-tests", () => {
           if (matchingPattern && shouldSkip) {
             specify.skip(`[${reason}] ${testFile}`);
           } else if (expectFail) {
-            runSingleWPT(testFilePath, `[expected fail] ${testFile}`, expectFail);
+            runSingleWPT(
+              testFilePath,
+              `[expected fail${
+                reason !== "fail" ? `: ${reason}` : ""
+              }] ${testFile}`,
```
domenic: Could you un-inline this back to the code I previously asked you to change it to?

Issue #2787, opened 2020-01-10
jsdom does not update selectionStart when `value` of `<input type=text>` is set


Matthew-Goldberg:
### Basic info:

- **Node.js version:** 10.15.3
- **jsdom version:** 15.2.1

jsdom does not update selectionStart when `value` of `<input type=text>` is set.

Actual: selectionStart is 0
Expected: selectionStart should be at the end of the set text

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const dom = new JSDOM(`<input type="text" value="4444" />`);
const { document } = dom.window;
const input = document.querySelector("input");
console.log(`Before: selectionStart=${input.selectionStart}, Value=${input.value}`);
// Set the value
input.value = 55555;
console.log(`After: selectionStart=${input.selectionStart}, Value=${input.value}`);
```

Result:
```
Before: selectionStart=0, Value=4444
After: selectionStart=0, Value=55555
```

### How does similar code behave in browsers?
In browser, this behaves differently.  The selectionStart after setting the value to `55555` is `5`, not 0.

See codepen: https://codepen.io/matthewg0/pen/YzPLyye

Result:
```
Before: selectionStart=0, Value=4444
After: selectionStart=5, Value=55555
```

You can also see this expected behavior in headless browser, using the following code:
```js
const puppeteer = require("puppeteer");

(async () => {
  // Now test in browser
  const htmlContent = `<input type="text" value="4444">`;
  const browser = await puppeteer.launch();
  const page = await browser.newPage();
  await page.setContent(htmlContent);
  const content = await page.content();

  console.log("--- Chrome ---");
  console.log(content);
  let input = await page.$eval("input", el => ({ selectionStart: el.selectionStart, value: el.value }));
  console.log(`Before: selectionStart=${input.selectionStart}, Value=${input.value}`);
  await page.$eval("input", el => (el.value = 55555));
  input = await page.$eval("input", el => ({ selectionStart: el.selectionStart, value: el.value }));
  console.log(`After: selectionStart=${input.selectionStart}, Value=${input.value}`);
  await browser.close();
})();
```

Result:
```
--- Chrome ---
<html><head></head><body><input type="text" value="4444"></body></html>
Before: selectionStart=0, Value=4444
After: selectionStart=5, Value=55555
```

domenic:
Great bug report!

[...]

PR #2799, opened 2020-01-16
Selection set to end on input value change

Matthew-Goldberg:
Fix for https://github.com/jsdom/jsdom/issues/2787

When setting the value of a `textarea` or `input type=text`, the selection should be set to the end of the new text.

Also fixed an issue in `textarea`, the selection start/end should initially be set to 0.


Matthew-Goldberg:
`set-value-reset-selection.html` Failed in "textarea setting value should reset selection indexes back to defaults"
Seems to me this test is incorrect, the browser does not appear to behave this way.

Matthew-Goldberg:
Yes it appears that `set-value-reset-selection.html` is no longer valid, spec updated in https://github.com/whatwg/html/pull/3472
Test can be updated or deleted.  I think it can be deleted since that test is covered now in `html/semantics/forms/textfieldselection`

lib/jsdom/living/nodes/HTMLTextAreaElement-impl.js
```diff
@@ -69,12 +71,14 @@ class HTMLTextAreaElementImpl extends HTMLElementImpl {
...
 
   set value(val) {
+    const oldValue = this._rawValue;
     this._rawValue = val;
     this._dirtyValue = true;
 
-    this._selectionStart = 0;
-    this._selectionEnd = 0;
-    this._selectionDirection = "none";
+    if (oldValue !== this._rawValue) {
```
domenic: This doesn't seem to match the spec; see https://html.spec.whatwg.org/multipage/form-elements.html#dom-textarea-value. You need to use API value instead of raw value.

test/web-platform-tests/to-run.yaml
```diff
@@ -574,6 +574,17 @@ DIR: html/semantics/forms/resetting-a-form
...
 
+DIR: html/semantics/forms/textfieldselection
+
+select-event.html: [fail, not yet implemented]
+selection-after-content-change.html: [fail, not yet implemented]
+selection-not-application-textarea.html: [fail, not yet implemented]
+selection-start-end-extra.html: [fail, not yet implemented]
+textfieldselection-setRangeText.html: [fail, not yet implemented]
+textfieldselection-setSelectionRange.html: [fail, not yet implemented]
+
```
domenic: Which tests in this directory fail before your change but pass afterward?

Matthew-Goldberg: The following now pass:
```
selection-not-application.html
selection-start-end.html
selection-value-interactions.html
selection.html
setSelectionRange.html
textarea-selection-while-parsing.xhtml
```

lib/jsdom/living/nodes/HTMLTextAreaElement-impl.js
Matthew-Goldberg: Thank you this is very helpful.  Looks like this may fix an additional test as well.  Working on it.

Matthew-Goldberg: This has been fixed, and now got `selection-after-content-change` to pass.

lib/jsdom/living/nodes/HTMLTextAreaElement-impl.js
```diff
@@ -69,12 +76,14 @@ class HTMLTextAreaElementImpl extends HTMLElementImpl {
   }
 
   set value(val) {
+    const oldValue = this._getValue();
```
domenic: This still doesn't seem to match the spec, which uses API value (obtained in jsdom via `_getAPIValue()`).

lib/jsdom/living/nodes/HTMLTextAreaElement-impl.js
```diff
@@ -44,7 +46,12 @@ class HTMLTextAreaElementImpl extends HTMLElementImpl {
     super._childTextContentChangeSteps();
 
     if (this._dirtyValue === false) {
+      const oldValue = this._getValue();
       this._rawValue = childTextContent(this);
+      if (oldValue !== this._getValue()) {
+        this._selectionStart = this._selectionEnd = 0;
+        this._selectionDirection = "none";
+      }
```
domenic: I don't see this in the spec. Is this change necessary to pass any tests?

test/web-platform-tests/to-run.yaml
Matthew-Goldberg: `selection-after-content-change.html` also passes after the change to use `_getAPIValue` to match spec.

lib/jsdom/living/nodes/HTMLTextAreaElement-impl.js
Matthew-Goldberg: Fixed

lib/jsdom/living/nodes/HTMLTextAreaElement-impl.js
Matthew-Goldberg: You are right, this is not needed for these tests.  I have removed this part of the change.

lib/jsdom/living/nodes/HTMLTextAreaElement-impl.js
```diff
@@ -69,12 +71,14 @@ class HTMLTextAreaElementImpl extends HTMLElementImpl {
   }
 
   set value(val) {
+    const oldValue = this._getAPIValue();
```
TimothyGu: nit: the spec calls it _oldAPIValue_.

Matthew-Goldberg: Thanks, I renamed this to `oldAPIValue` and added a comment linking to the spec.

PR #2803, opened 2020-01-18
Re-do animation frame and timer callbacks

### LINKED ISSUE #2617: setInterval returns undefined after the window closes
### Basic info:

Node 8
jsdom 15.1.1

### Minimal reproduction case

```js

const dom = new JSDOM("");
let callback = () => {
  let value = dom.window.setInterval(() => {}, 1000);
  console.log("this should not be undefined:", value);
};
dom.window.close();
process.nextTick(callback);
```

The problem is that in jsdom, `setInterval` returns undefined after the window closes. However, normally setInterval will never return undefined, either in Node or in browsers. So it's reasonable for some code to be nested in a bunch of callbacks, and run after the jsdom window has been closed, but to expect a non-undefined return value for `setInterval`.

Specifically, some code in the `simple-peer` library does something like `let returnValue = setInterval(...); if (returnValue.unref) { ... }` to handle the case where timers should be unref'd to let Node exit. That code is correct according to either browsers or Node, but fails in a jsdom environment.

Instead, perhaps this could return `0` or some other number? That is normally a valid return value for `setInterval`.

Edit: Chrome just returns 0. It *is* possible to call `setInterval` on a window after it closes, by calling it from another window.

### LINKED ISSUE #2800: window.close() does not stop the loop in the opened HTML
### Basic info:

- **Node.js version:** 13.0.1
- **jsdom version:** 16.0.0
- Seen on Windows 10

### Minimal reproduction case

```js
var JSDOM = require('jsdom').JSDOM;

JSDOM.fromFile('loop.html', {
  resources: 'usable',
  runScripts: 'dangerously',
}).then(function(dom) {
  setTimeout(function() { dom.window.close(); }, 5000);
});
```

```html
<!DOCTYPE html>
<html lang=en>
<head>
<title>JSDOM test</title>
</head>

<body>

<script><!--
setInterval(function() { console.log('*'); }, 1000);
--></script>

</body>
</html>
```
This script runs forever.

domenic:
This setup follows the spec more, is a lot simpler to follow, fixes #2800, and fixes #2617.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -351,14 +351,18 @@ function Window(options) {
     return timerInitializationSteps(handler, timeout, args, { methodContext: window, repeat: true });
   };
 
-  this.clearTimeout = function (handle) {
+  this.clearTimeout = function (handle = 0) {
+    handle = webIDLConversions.long(handle);
+
     const nodejsTimer = listOfActiveTimers.get(handle);
     if (nodejsTimer) {
       clearTimeout(nodejsTimer);
       listOfActiveTimers.delete(handle);
     }
   };
-  this.clearInterval = function (handle) {
+  this.clearInterval = function (handle = 0) {
+    handle = webIDLConversions.long(handle);
+
     const nodejsTimer = listOfActiveTimers.get(handle);
     if (nodejsTimer) {
       // We use setTimeout() in timerInitializationSteps even for this.setInterval().
@@ -449,8 +453,10 @@ function Window(options) {
     };
 
     function runAnimationFrameCallbacks(now) {
-      const callbackHandles = mapOfAnimationFrameCallbacks.keys();
+      // Converting to an array is important to get a sync snapshot and thus match spec semantics.
+      const callbackHandles = [...mapOfAnimationFrameCallbacks.keys()];
       for (const handle of callbackHandles) {
+        // This has() can be false if a callback calls cancelAnimationFrame().
         if (mapOfAnimationFrameCallbacks.has(handle)) {
           const callback = mapOfAnimationFrameCallbacks.get(handle);
           mapOfAnimationFrameCallbacks.delete(handle);
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -333,7 +333,6 @@ function Window(options) {
   // we can call Node.js-side clearTimeout() when clearing, and thus allow process shutdown faster.
   const listOfActiveTimers = new Map();
   let latestTimerId = 0;
-  let timerNestingLevel = 0;
 
   this.setTimeout = function (handler, timeout = 0, ...args) {
     if (typeof handler !== "function") {
@@ -374,6 +373,8 @@ function Window(options) {
       return 0;
     }
 
+    // TODO: implement timer nesting level behavior.
+
     const methodContextProxy = methodContext._globalProxy;
     const handle = previousHandle !== undefined ? previousHandle : ++latestTimerId;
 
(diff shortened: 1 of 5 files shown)


lib/jsdom/browser/Window.js
```diff
@@ -330,55 +327,136 @@ function Window(options) {
...
+          try {
+            callback(now);
+          } catch (e) {
+            reportException(window, e, window.location.href);
+          }
+        }
+      }
+    }
+
+    animationFrameNodejsInterval = setInterval(() => {
```
domenic: Although this is a lot simpler and closer to the spec, it probably wouldn't be appreciated by our consumers since it would mean you need to call `window.close()` on any `pretendToBeVisual` `JSDOM`, even with no `requestAnimationFrame()` calls in play. So this should probably change.

domenic: Yeah, just pushed a commit to do that

lib/jsdom/browser/Window.js
```diff
@@ -330,55 +327,151 @@ function Window(options) {
...
+  this.setInterval = function (handler, timeout = 0, ...args) {
+    if (typeof handler !== "function") {
+      handler = webIDLConversions.DOMString(handler);
+    }
+    timeout = webIDLConversions.long(timeout);
+
+    return timerInitializationSteps(handler, timeout, args, { methodContext: window, repeat: true });
+  };
+
+  this.clearTimeout = function (handle) {
```
TimothyGu: https://html.spec.whatwg.org/multipage/webappapis.html#windoworworkerglobalscope has the following IDL:

```webidl
void clearTimeout(optional long handle = 0);
```

We should call `webIDLConversions.long` on and provide the default value for `handle`.

lib/jsdom/browser/Window.js
```diff
@@ -330,55 +327,151 @@ function Window(options) {
...
-    timers = Object.create(null);
-    animationFrameCallbacks = Object.create(null);
-  };
+      mapOfAnimationFrameCallbacks.delete(handle);
+    };
+
+    function runAnimationFrameCallbacks(now) {
+      const callbackHandles = mapOfAnimationFrameCallbacks.keys();
+      for (const handle of callbackHandles) {
+        if (mapOfAnimationFrameCallbacks.has(handle)) {
```
TimothyGu: Can this ever be false? `handle` came from `mapOfAnimationFrameCallbacks` anyway.

It be look simpler to do `for (const [handle, callback] of mapOfAnimationFrameCallbacks) {}` in either case.

lib/jsdom/browser/Window.js
pmdartus: I also realized the same thing applies to `clearInterval`.

lib/jsdom/browser/Window.js
domenic: It can, if the callback calls cancelAnimationFrame. That's why the spec is written as-is. This was fixed in https://github.com/whatwg/html/pull/4375.

TimothyGu: Ah okay. It also looks like `.delete` doesn’t invalidate existing iterators so this is fine.

domenic: Hmm, but the iterator skips deleted keys, so maybe this is not necessary?

But, the iterator will continue to get new keys that are added during iteration, which does not match spec behavior... so I think I should snapshot the iterator, and maybe add tests for this...

lib/jsdom/browser/Window.js
TimothyGu: This doesn't seem to have been fixed.

lib/jsdom/browser/Window.js
TimothyGu: This doesn't seem to have been fixed either. The iterator is indeed "snapshotted", but not its values…

domenic:
I forgot to push; sorry about that!

PR #2806, opened 2020-01-20
test(wpts): Rename `Attr` extends `Node` test

ExE-Boss:
To&nbsp;preserve <code>git&nbsp;blame</code>&nbsp;history after&nbsp;<https://github.com/jsdom/jsdom/pull/2770> is&nbsp;merged.

---

See&nbsp;also: <https://github.com/jsdom/jsdom/pull/2764>


domenic:
I don't think this is worth the extra commit. Retaining revision history for a test, especially one that had its expectations flipped so it's basically a separate test, is not a priority.

PR #2808, opened 2020-01-23
Fix syntax error which halts execution

jpuffer:
fix the following syntax error:
```
.../jsdom/lib/jsdom/living/helpers/dates-and-times.js:235
  } catch {
          ^

SyntaxError: Unexpected token {
```


domenic:
This is not a syntax error; this is valid JavaScript syntax. Note that it is only supported in Node.js 10+.

PR #2811, opened 2020-01-25
Fix catch syntax error in date-and-times.js

ddrulez:
In date-and-times.js there is a problem with a incorrect syntax of catch. 
This generates eslint errors, in particular with firebase a aws functions.


domenic:
This is not incorrect syntax. It is supported in Node.js v10, which is the minimum required version to run jsdom v16.

Issue #2647, opened 2019-08-16
Allow SVG edition using jsdom + SVG.js


aurium:
### Basic info:

- **Node.js version:** 11.13.0
- **jsdom version:** 15.1.1

### The Problem

To easily edit SVG on node using jsdom + SVG.js, we need jsdom to implement (at least) this methods:

* `SVGGraphicsElement::getBBox()`
* `SVGSVGElement::createSVGRect()`
* `SVGSVGElement::createSVGPoint()`
* `SVGSVGElement::createSVGMatrix()`

### Minimal reproduction case (pure jsdom)

```js
const { JSDOM } = require('jsdom')

dom = new JSDOM(`
<svg id="test" xmlns="http://www.w3.org/2000/svg" width="50px" height="50px">
<path d="M0,0 L50,50"/>
</svg>
`)
elSVG = dom.window.document.body.children[0]

// All this will fail:
elSVG.firstElementChild.getBBox()
elSVG.createSVGRect()
elSVG.createSVGPoint()
elSVG.createSVGMatrix()
```

### Minimal reproduction case (With svg.js)

This is going to crash by the same reason, but it is a (near to) real use case.
```js
const { JSDOM } = require('jsdom')
const svgFactory = require('svg.js')

dom = new JSDOM(`
<svg id="test" xmlns="http://www.w3.org/2000/svg" width="50px" height="50px">
<path d="M0,0 L50,50"/>
</svg>
`)

SVG = svgFactory(dom.window)
elSVG = dom.window.document.body.children[0]
draw = SVG(elSVG)

line = SVG.select('#test path').members[0]
line.move(10, 20)
```

### How does similar code behave in browsers?

It simply retun values from `getBBox()`, `createSVGRect()`, `createSVGPoint()`, `createSVGMatrix()` as expected.

Sayan751:
Is there any plan for this?

TimothyGu:
https://twitter.com/slicknet/status/782274190451671040

jcubic:
Any plans to implement this? I was trying to run Raphael (as asked in DmitryBaranovskiy/raphael#1074)  but got error that createSVGMatrix is not a function.

jcubic:
I've just created basic SVGMatrix class (function constructor) and createSVGMatrix function, if you're interested I can share the code. I'm not sure how to integrate my code with jsdom, I'm not sure where to put it.

I've checked the code and it seems that this should be ES6 class it would not be that hard to convert.

[...]

PR #2812, opened 2020-01-28
Mock SVGSVGElement methods

m1gu3l:
May help with #2647.
Will allow using [Leaflet ](https://github.com/Leaflet/Leaflet/blob/472407cd2ec8a5f05f0ea0e060c0e41b61778c57/src/core/Browser.js#L143) in jsdom.

### DIFF
--- a/lib/jsdom/living/nodes/SVGSVGElement-impl.js
+++ b/lib/jsdom/living/nodes/SVGSVGElement-impl.js
@@ -4,6 +4,7 @@ const { mixin } = require("../../utils");
 const SVGNumber = require("../generated/SVGNumber");
 const SVGGraphicsElementImpl = require("./SVGGraphicsElement-impl").implementation;
 const WindowEventHandlersImpl = require("./WindowEventHandlers-impl").implementation;
+const notImplemented = require("../../browser/not-implemented");
 const { domSymbolTree } = require("../helpers/internal-constants");
 const { ELEMENT_NODE } = require("../node-type");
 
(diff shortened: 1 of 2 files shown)


domenic:
It's not clear why this belongs in jsdom. Code that uses jsdom could add these mocks themselves...

m1gu3l:
Of course, but since those methods are part of the living standard I thought it would be nice if jsdom had them. Other libraries (modernizr f.e.) also use those to detect whether environment supports SVG and it may cause not so obvious errors in tests. It's up to you though.

domenic:
Well, but these implementations that you've provided are not part of the SVG standard ;).

m1gu3l:
You're right, those are stubs - but the error would be meaningful.

TimothyGu:
This actually makes it more difficult to feature-detect SVG. Previously one could have done `method in document`, but now they have to use a `try`-`catch` to do that correctly.

TimothyGu:
Actually, I misunderstand – in the case of Leaflet you actually want to have jsdom treated as an SVG-capable user agent. I don't know how Leaflet is using the SVG APIs, but it's dangerous for us to say the functions are available since it's possible that they end up using those mocked functions and weird errors start popping up. Best to leave the decision of mocking or not to the user. :)

m1gu3l:
Leaflet doesn't use those methods, it's just for capability check. But let's say some other lib does. Is it better to get "undefined is not a function" or "createSVGRect is not implemented"?

PR #2813, opened 2020-01-29
Better reflection support: take 2

TimothyGu:
Same goal as https://github.com/jsdom/jsdom/pull/2790, but uses an approach based on IDL extended attributes. So far works pretty well but the real test will be enumerated attributes, which I'll look at later.

This depends on https://github.com/jsdom/webidl2js/pull/161 and #2822.

### DIFF
--- a/scripts/webidl/convert.js
+++ b/scripts/webidl/convert.js
@@ -69,11 +69,12 @@ const transformer = new Webidl2js({
       };
     }
 
-    if (isSimpleIDLType(idl.idlType, "DOMString")) {
+    if (isSimpleIDLType(idl.idlType, "DOMString") || isSimpleIDLType(idl.idlType, "USVString")) {
+      const isUSV = isSimpleIDLType(idl.idlType, "USVString");
       return {
         get: `
           const value = ${implObj}.getAttributeNS(null, "${attrName}");
-          return value === null ? "" : value;
+          return value === null ? "" : ${isUSV ? "conversions.USVString(value)" : "value"};
         `,
         set: `
           ${implObj}.setAttributeNS(null, "${attrName}", V);
--- a/scripts/webidl/convert.js
+++ b/scripts/webidl/convert.js
@@ -43,6 +43,32 @@ const transformer = new Webidl2js({
     const reflectAttr = idl.extAttrs.find(attr => attr.name === "Reflect");
     const attrName = (reflectAttr && reflectAttr.rhs && JSON.parse(reflectAttr.rhs.value)) || idl.name.toLowerCase();
 
+    if (idl.extAttrs.find(attr => attr.name === "ReflectURL")) {
+      // Allow DOMString also due to https://github.com/whatwg/html/issues/5241.
+      if (!isSimpleIDLType(idl.idlType, "USVString") && !isSimpleIDLType(idl.idlType, "DOMString")) {
+        throw new Error("[ReflectURL] specified on non-USV/DOMString attribute");
+      }
+      const parseURLToResultingURLRecord =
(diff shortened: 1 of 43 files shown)


domenic:
Do you know of any cases where https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#rules-for-parsing-integers would give different results than `parseInt(x, 10)`? I'd rather reuse that than reimplement the whole algorithm ourselves. It looks like the whitespace trimming is a bit different, `parseInt()` can return a `-0`, and `parseInt()` will ignore trailing garbage, but we could just take care of those cases ourselves? Tests for the changes would be a good idea here.

Also the first couple of commits could be their own independent PR.

TimothyGu:
We could just adapt the whitespace trimming part of 24d536492f46ade86ecc8a36e408b0fe647dcc74 to use with parsing integers, yes. I'm more interested in comments on the approach right now than treating this as something to be merged directly. :)

domenic:
> I'm more interested in comments on the approach right now than treating this as something to be merged directly. :)

I guess I gave those over in https://github.com/jsdom/jsdom/pull/2790#issuecomment-581074317, oops.

Do you think it's worth proceeding by trying to merge a v1 without enumerated attribute support? I'd be hesitant to do that in the HTML Standard, but for jsdom it's a definite improvement over our current situation, so I'd be fine with that here. Still, it'd be awesome if we could figure out enumerated attributes too.

Otherwise, hmm. I wonder if we should abstract the code we see repeated here a few times for URL handling. E.g. in jsdom (not webidl2js), a util like `getURLStringFromAttribute(this, "src")` or `getURLStringFromAttribute(this, "src", { disallowEmpty: true })`

TimothyGu:
Updated for the latest webidl2js change. Most of the concerns raised in #2790 should now be fixed (with the exception of state-related stuff, which we'll punt into the future for now).

scripts/webidl/convert.js
```diff
@@ -30,6 +37,104 @@ const transformer = new Webidl2js({
     return `
       return ${identifier}(globalObject, interfaceName, new.target);
     `;
+  },
+  // https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflecting-content-attributes-in-idl-attributes
+  processReflect(idl, implObj) {
+    const reflectAttr = idl.extAttrs.find(attr => attr.name === "Reflect");
+    const attrName = (reflectAttr && reflectAttr.rhs && reflectAttr.rhs.value.replace(/_/g, "-")) || idl.name.toLowerCase();
```
domenic: As mentioned, I would like to move this to strings instead of identifiers. But, we can do that in a followup.

scripts/webidl/convert.js
```diff
@@ -30,6 +37,104 @@ const transformer = new Webidl2js({
...
+        `
+      };
+    }
+
+    if (isSimpleIDLType(idl.idlType, "DOMString") || isSimpleIDLType(idl.idlType, "USVString")) {
+      const isUSV = isSimpleIDLType(idl.idlType, "USVString");
+      return {
+        get: `
+          const value = ${implObj}.getAttributeNS(null, "${attrName}");
+          return value === null ? "" : ${isUSV ? "conversions.USVString" : ""}(value);
```
domenic:
```suggestion
          return value === null ? "" : ${isUSV ? "conversions.USVString(value)" : "value"};
```

(a bit easier to tell what's going on, IMO.)

TimothyGu: Done.

scripts/webidl/convert.js
TimothyGu: Done.

PR #2744, opened 2019-12-08
Revert `HTMLDocument` removal

ExE-Boss:
Turns out, trying to make `HTMLDocument` an alias of `Document` [wasn’t](https://github.com/whatwg/html/issues/4792) [exactly](https://github.com/whatwg/dom/issues/221) [web‑compatible](https://github.com/whatwg/dom/issues/278).

The main reasons were the legacy `[LegacyOverrideBuiltins]` extended attribute and `document instanceof HTMLDocument` checks expecting to fail for `XMLDocument`.

## Depends on:
- [x] <https://github.com/jsdom/jsdom/pull/2753> (merged)
- [x] <https://github.com/jsdom/jsdom/pull/2841> (merged)

---

This PR implements <https://github.com/whatwg/html/pull/5104> and <https://github.com/whatwg/dom/pull/799>.

See also: <https://github.com/web-platform-tests/wpt/pull/20676>

PR #2814, opened 2020-01-29
fix(Document): Correctly create and clone `XMLDocument` instances

ExE-Boss:
`DOMImplementation.createDocument(…)` is&nbsp;specified to&nbsp;return a&nbsp;`XMLDocument`, so&nbsp;this&nbsp;fixes&nbsp;that and&nbsp;adds&nbsp;an&nbsp;internal `_createDocument(…)`&nbsp;helper to&nbsp;`DOMImplementationImpl`.

---

Pre‑requisite&nbsp;for: <https://github.com/jsdom/jsdom/pull/2744>.
Supersedes and closes <https://github.com/jsdom/jsdom/pull/2841>


lib/jsdom/living/documents.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
+const XMLDocument = require("../living/generated/XMLDocument");
+const Document = require("../living/generated/Document");
+
+exports.createWrapper = (globalObject, options = {}, type = options.parsingMode || null) => {
```
domenic: I'm unsure about the defaulting of type to options.parsingMode. It seems weird that you have code to synchronize the two both in the XMLDocumentImpl constructor, and here. The function signature gives the false impression that they can be decoupled, e.g. you could pass a parsingMode of HTML with a type of XML.

I'm not sure what the best function signature would be, but maybe something like `(globalObject, options = {}, { useParsingModeBasedSubclass = false } = {})`? Or some shorter variant of that.

lib/jsdom/living/documents.js
```diff
@@ -0,0 +1,31 @@
...
+      // TODO: Implement 'HTMLDocument' (https://github.com/jsdom/jsdom/pull/2744)
+      // eslint-disable-next-line no-fallthrough
+    case null:
+      return Document.create(globalObject, [], { options });
+    default:
+      throw new Error(`Internal error: Document type ${String(type)} is not one of null, "xml" or "html".`);
+  }
+};
+
+exports.createImpl = (globalObject, options = {}, type = options.parsingMode || null) => {
```
domenic: One of these two functions should be replaced with just calling the other one and then using implFromWrapper/wrapperFromImpl.

lib/jsdom/living/nodes/XMLDocument-impl.js
```diff
@@ -1,4 +1,16 @@
...
+  constructor(globalObject, args, privateData = {}) {
+    let options;
+    if (!(options = privateData.options)) {
+      options = privateData.options = {};
+    }
+
+    // XMLDocument is always in "xml" mode:
+    options.parsingMode = "xml";
+
+    super(globalObject, args, privateData);
```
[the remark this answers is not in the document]
ExE-Boss: That&nbsp;will&nbsp;break if&nbsp;`privateData.options` is&nbsp;`undefined` or&nbsp;`null`.

lib/jsdom/living/nodes/DOMImplementation-impl.js
```diff
@@ -60,14 +62,15 @@ class DOMImplementationImpl {
...
+    const document = documents.createImpl(this._globalObject, {
+      parsingMode: "html",
+      encoding: "UTF-8"
     });
 
     // Create a doctype, with "html" as its name and with its node document set
     // to doc. Append the newly created node to doc.
     const doctype = DocumentType.createImpl(this._globalObject, [], {
-      ownerDocument: this,
+      ownerDocument: document,
```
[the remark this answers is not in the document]
ExE-Boss: It’s&nbsp;unobservable, but&nbsp;it&nbsp;should make&nbsp;adopting the&nbsp;node&nbsp;faster, as&nbsp;it&nbsp;doesn’t&nbsp;need to&nbsp;iterate over&nbsp;the&nbsp;`DocuementType`&nbsp;descendants in&nbsp;`DocumentImpl._adoptNode(…)`:

https://github.com/jsdom/jsdom/blob/df2018f2b25b4663d10218fa381246a8628b01bb/lib/jsdom/living/nodes/Document-impl.js#L827-L848

lib/jsdom/living/nodes/XMLDocument-impl.js
```diff
@@ -1,4 +1,16 @@
...
-exports.implementation = class XMLDocumentImpl extends DocumentImpl {};
+exports.implementation = class XMLDocumentImpl extends DocumentImpl {
+  constructor(globalObject, args, privateData = {}) {
+    let options;
+    if (!(options = privateData.options)) {
+      options = privateData.options = {};
+    }
+
+    // XMLDocument is always in "xml" mode:
+    options.parsingMode = "xml";
```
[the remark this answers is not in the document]
ExE-Boss: Well, it’s&nbsp;impossible to&nbsp;create a&nbsp;`XMLDocument` with&nbsp;an&nbsp;`"html"` parsing&nbsp;mode in&nbsp;user&nbsp;code.

lib/jsdom/living/node.js
```diff
@@ -15,13 +15,7 @@ exports.clone = (node, document, cloneChildren) => {
...
   switch (node.nodeType) {
     case NODE_TYPE.DOCUMENT_NODE:
-      // Can't use a simple Document.createImpl because of circular dependency issues :-/
-      copy = document.implementation.createDocument(null, "", null);
-      copy._encoding = node._encoding;
-      copy.contentType = node.contentType;
-      copy._URL = node._URL;
-      copy._origin = node._origin;
-      copy._parsingMode = node._parsingMode;
+      copy = node._cloneDocument();
```
pmdartus: IMO, we should keep the note about the circular dependency here. This would make it obvious why the indirection is needed here.

lib/jsdom/living/nodes/XMLDocument-impl.js
```diff
@@ -1,4 +1,9 @@
 "use strict";
 const DocumentImpl = require("./Document-impl").implementation;
 
-exports.implementation = class XMLDocumentImpl extends DocumentImpl {};
+exports.implementation = class XMLDocumentImpl extends DocumentImpl {
+  constructor(globalObject, args, privateData = {}) {
+    // XMLDocument is always in "xml" parsing mode:
+    super(globalObject, args, { ...privateData, options: { ...privateData.options, parsingMode: "xml" } });
```
domenic: I'm still a bit unclear why this is necessary. There is no public constructor in the IDL for `XMLDocument`, so the only way this could ever have the wrong parsing mode is if some other part of the code passed it. But, then we should fix the other part of the code, right?

So I think we can just undo all changes to this file. I will push a commit doing so momentarily.

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -894,6 +895,23 @@ class DocumentImpl extends NodeImpl {
...
+  // Needed to ensure that the resulting document has the correct prototype chain:
+  _cloneDocument() {
+    const copy = documents.createImpl(
+      this._globalObject,
+      {
+        contentType: this.contentType,
+        encoding: this._encoding,
+        parsingMode: this._parsingMode
+      },
+      Object.getPrototypeOf(this) !== DocumentImpl.prototype
```
domenic: This has no effect as far as I can tell. It says to don't use the XMLDocument subclass, in a case when we would already not be using it.

domenic:
I can't push my fixups to this PR branch again, so I guess I'll start a new PR.

lib/jsdom/living/nodes/Document-impl.js
ExE-Boss: There’s some cases where browsers create a base `Document` instead of the subclasses.

domenic: This particular line still has no effect, and removing it causes the same tests to pass before and after.

ExE-Boss: Well, it will become necessary with <https://github.com/jsdom/jsdom/pull/2744>, so I guess I can hold off on introducing it until then.

PR #2816, opened 2020-01-31
this._resourceLoader == null => return null

rafal83:
Fix empty _resourceLoader


domenic:
Thanks, but this is not an acceptable change. It does not come with tests, for one.

Please open an issue using the issue template if you are encountering a problem with jsdom.

PR #2820, opened 2020-02-02
Attr-as-Node fix-ups

TimothyGu:
This PR carries two commits that fixes up some issues from #2770. In particular, this fixes #2819 and #2832.

Tests are in https://github.com/web-platform-tests/wpt/pull/21540, which will be available here pending it being merged and a WPT roll.

PR #2821, opened 2020-02-02
Update dependencies and roll WPT

TimothyGu:
Upstream WPT has made several changes that we needed to account for here.

- The JSON manifest format was bumped to version 8, which brings a   trie-like path format. Adapt wpt-manifest-utils to handle that correctly. See   https://github.com/web-platform-tests/rfcs/blob/master/rfcs/manifest-path-trie.md.

- Override `assert_throws_js`/`promise_rejects_js` to not check for the   realm that errors belong to. This is being tracked as #2727.

WPT roll in preparation for #2820.


domenic:
Is it intentional that the first commit does not update package.json? I.e., only indirect dependencies got updates?

TimothyGu:
Hmm, I thought `yarn upgrade` did that automatically but apparently not… Anyway a `syncyarnlock -skl` fixed it up.

domenic:
Still only seeing `karma-browserify` changed in the package.json...

TimothyGu:
Unfortunately, recent WPT changes made a lot of tests that use `assert_throws` check for the Error type using `instanceof`, which isn't something that we are prepared to deal with right now. I have pushed a change that adapts our wpt tooling to use the new manifest format though.

domenic:
Hmm, yeah, I think we fixed most of the DOMExceptions to be created in the relevant realm (still not perfectly correct, but should pass). But all the TypeErrors and such... including those thrown by webidl2js-generated code... urgh.

Maybe we should override assert_throws and friends for now.

TimothyGu:
Okay, did what you suggested and monkey-patched the offending `assert_throws_js`. This is now good for review.

PR #2822, opened 2020-02-03
Add rules for parsing integers and non-negative integers

TimothyGu:
Split out from #2813.

### DIFF
--- a/lib/jsdom/living/helpers/strings.js
+++ b/lib/jsdom/living/helpers/strings.js
@@ -49,6 +49,39 @@ exports.asciiCaseInsensitiveMatch = (a, b) => {
   return true;
 };
 
+// https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#rules-for-parsing-integers
+// Error is represented as null.
+const parseInteger = exports.parseInteger = input => {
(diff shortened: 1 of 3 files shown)


lib/jsdom/living/helpers/strings.js
```diff
@@ -49,6 +49,37 @@ exports.asciiCaseInsensitiveMatch = (a, b) => {
...
+// Error is represented as null.
+const parseInteger = exports.parseInteger = input => {
+  // The implementation here is slightly different from the spec's. We want to use parseInt(), but parseInt() trims
+  // Unicode whitespace in addition to just ASCII ones, so we make sure that the trimmed prefix contains only ASCII
+  // whitespace ourselves.
+  const numWhitespace = input.length - input.trimStart().length;
+  if (/[^\t\n\f\r ]/.test(input.slice(0, numWhitespace))) {
+    return null;
+  }
+  const value = parseInt(input, 10); // eslint-disable-line radix
```
domenic: We should preserve the comment stating that we are intentionally avoiding the 0x -> hexadecimal behavior here.

TimothyGu: Done.

Issue #2757, opened 2019-12-24
Heap Memory Leakage


yashimversion:
Hi,
I'm writing a server-side renderer for react application using JSDOM where it receives HTML as input and JSDOM executed javascript to build the page and spits out a rendered response (For many reasons we aren't using ReactDOMServer for server-side rendering). Although the page seems to render fine, the heap memory keeps increasing on every request until it crashes. I pretty much tried all the solutions listed in this Github issues related to memory leak i.e. usage of `window.close()` or custom collect garbage using `global.gc()`, but none of them seems to work. Also, I observed even if there are no requests for several hours, heap memory keeps increasing in small amounts until it crashed. Let me know how I can fix this. Thanks !!

### Basic info:

- **Node.js version: v10.16.3** 
- **jsdom version: 15.2.1** 

### Minimal reproduction case

```js
const resourceLoader = new jsdom.ResourceLoader( { strictSSL: false } );
const virtualConsole = new jsdom.VirtualConsole();

// Define JSDOM configurations.
const options = {
	contentType:    'text/html',
	referrer:       request.body.originUrl,
	resources:      resourceLoader,
	runScripts:     'dangerously',
	url:            ( request.body.originUrl + request.body.relativeUrl ),
	virtualConsole: virtualConsole.sendTo( console ),
	beforeParse( window ) {
		// Add event listener to know when page is fully loaded.
		window.addEventListener( 'load', function ( e ) {

		    const renderedHtml = dom.serialize(); // eslint-disable-line no-use-before-define

		    //Return response after page load.
		    response.send( renderedHtml );
	        } )
        },
};

// Render HTML and Javascript using JSDOM.
let dom = new JSDOM( request.body.html, options );
```

Sebmaster:
You'll need to get us a better repro, as-is this doesn't help us get to the bottom of the leaks unfortunately. Presumably if you turn off runScripts it'd work, indicating that some on-page scripts trigger a codepath that causes a leak.

domenic:
Closing since this does not contain enough information for us to reproduce. (However, fixes will be merged soon for a leak which may be related; we just have no way to tell from the code you've given.)

PR #2825, opened 2020-02-04
Fix issue leaking memory in parser of html

lamhieu-vk:
Resolved #2757 

Add a step to clean up the reference object after use.
Please review it, thanks.

PR #2831, opened 2020-02-08
un-monkey-patch OpenElementStack to avoid leaks

terite:
A monkey-patch as modified by commit 75a921eab52c239b2468e08be9f547f46c7f86bd
maintains a reference to the most recent JSDOMParse5Adapter (and all its
options) through the push/pop closures. Because the monkey-patches are
never cleaned up, there is always a reference to the most recent
JSDOMParse5Adapter and its related objects.

This fixes https://github.com/facebook/jest/issues/9507
This may fix https://github.com/jsdom/jsdom/issues/2757


terite:
I just noticed #2825 beat me to the punch by a few days. I'll happily close this if y'all maintainers prefer the other approach.

I assume the unintentional test fix is due to how this code stacks parser adapters rather than keeping and overwriting one reference.

lib/jsdom/browser/parser/html.js
```diff
@@ -16,6 +16,29 @@ const OpenElementStack = require("parse5/lib/parser/open-element-stack");
 const OpenElementStackOriginalPop = OpenElementStack.prototype.pop;
 const OpenElementStackOriginalPush = OpenElementStack.prototype.push;
 
+const adapterStack = [];
+
+// Horrible monkey-patch to implement https://github.com/inikulin/parse5/issues/237
+OpenElementStack.prototype.push = function (...args) {
+  OpenElementStackOriginalPush.apply(this, args);
+  adapterStack[adapterStack.length - 1]._currentElement = this.current;
```
TimothyGu: This should have a check for `adapterStack` not empty, just in case parse5 is used elsewhere not in jsdom.

PR #2839, opened 2020-02-14
test(api/encoding): Use `fs.promises`

ExE-Boss:
Currently, `test/api/encoding.js` depends&nbsp;on&nbsp;the&nbsp;undeclared `pn/fs`&nbsp;dependency.

https://github.com/jsdom/jsdom/blob/b4b5a542380316a0d7296f7f41a5909bf5e097db/test/api/encoding.js#L2

This&nbsp;is&nbsp;disallowed by&nbsp;[**pnpm**](https://pnpm.js.org) and&nbsp;will&nbsp;be&nbsp;disallowed by&nbsp;[**Yarn&nbsp;v2**](https://yarnpkg.com/).

---

See also: <https://github.com/jsdom/jsdom/pull/2783>

### DIFF
--- a/test/api/encoding.js
+++ b/test/api/encoding.js
@@ -1,5 +1,5 @@
 "use strict";
-const fs = require("pn/fs");
+const fs = require("fs");
 const path = require("path");
 const { assert } = require("chai");
 const { describe, it, before, after } = require("mocha-sugar-free");
@@ -12,19 +12,21 @@ function fixturePath(fixture) {
 }
 
 function readFixture(fixture) {
-  return fs.readFile(fixturePath(fixture));
+  return fs.promises.readFile(fixturePath(fixture));
 }
 
 const factories = {
   Buffer: fixture => readFixture(fixture),
   Uint8Array: fixture => readFixture(fixture).then(buffer => Uint8Array.from(buffer)),
-  ArrayBuffer: fixture => readFixture(fixture).then(buffer => buffer.buffer),
-  DataView: fixture => readFixture(fixture).then(buffer => new DataView(buffer.buffer)),
+  ArrayBuffer: fixture => readFixture(fixture).then(buffer => Uint8Array.from(buffer).buffer),
+  DataView: fixture => readFixture(fixture).then(buffer =>
+    new DataView(buffer.buffer, buffer.byteOffset, buffer.byteLength)),
   Int8Array: fixture => readFixture(fixture).then(buffer => {
-    // Test a view that is indexing into a larger ArrayBuffer
-    const target = new Int8Array(buffer.buffer);
+    // Test a view that is indexing into a larger ArrayBuffer. (buffer may already be such a view, but make sure we
+    // test at least one in case it's not.)
+    const target = new Int8Array(buffer.buffer, buffer.byteOffset, buffer.byteLength);
     const bigger = new Int8Array([0, 0, 0, ...target, 0, 0, 0]);
-    const subView = new Int8Array(bigger.buffer, 3, buffer.buffer.byteLength);
+    const subView = new Int8Array(bigger.buffer, 3, buffer.byteLength);
 
     return subView;
   })


domenic:
The intention in https://github.com/jsdom/jsdom/commit/d240291edbe7d4180a5152993ced7950c834dc57 was to remove these by using `require("fs").promises` instead.

TimothyGu:
What Node.js version un-flagged `fs.promises`?

test/api/encoding.js
```diff
@@ -11,9 +12,9 @@ function fixturePath(fixture) {
   return path.resolve(__dirname, "fixtures/encoding", fixture);
 }
 
-function readFixture(fixture) {
-  return fs.readFile(fixturePath(fixture));
-}
+// This can't use `fs.promises`, as it has a bug with non-UTF-8 encodings:
+const readFixture = (readFile => fixture =>
+  readFile(fixturePath(fixture)))(util.promisify(fs.readFile));
```
TimothyGu: Can you just make `readFile` its own module-global variable?

ExE-Boss:
> What Node.js version un-flagged `fs.promises`?

[**Node v10.0.0**](https://nodejs.org/api/fs.html#fs_fspromises_writefile_file_data_options), but it doesn’t deal well with non‑**UTF‑8** encoded files.

And `util.promisify` isn’t supported by the version of `util` used by the version of **Browserify** used for testing.

test/api/encoding.js
```diff
@@ -11,8 +12,14 @@ function fixturePath(fixture) {
   return path.resolve(__dirname, "fixtures/encoding", fixture);
 }
 
+// This can't use `fs.promises`, as it has a bug with non-UTF-8 encodings:
```
domenic: Can you link to the bug?

test/api/encoding.js
```diff
@@ -11,8 +12,14 @@ function fixturePath(fixture) {
...
+let readFile;
 function readFixture(fixture) {
-  return fs.readFile(fixturePath(fixture));
+  if (!readFile) {
```
domenic: The extra code complexity here of caching detracts from readability too much. Just do `return util.promisify(fs.readFile)(...)`.

test/api/encoding.js
```diff
@@ -12,14 +12,11 @@ function fixturePath(fixture) {
   return path.resolve(__dirname, "fixtures/encoding", fixture);
 }
 
-// This can't use `fs.promises`, as it has a bug with non-UTF-8 encodings:
-let readFile;
+// This can't use `fs.promises`, as it leads to non-UTF-8 files getting interpreted
+// as UTF-8 when passed to `ArrayBuffer`, `DataView` or `Int8Array` due to
+// buffer overflow issues.
```
TimothyGu: Is there a bug filed for this anywhere? If so, please link it here. (If not you should probably file a bug with Node.js.)

TimothyGu: Or is this only a bug with old versions of Node.js? If so, also put that info in the comment.

ExE-Boss:
Now that <https://github.com/jsdom/webidl2js/pull/175> has been merged, as `pn` will no longer be in `node_modules` once a new version of **WebIDL2JS** is released.

test/api/encoding.js
domenic: Ping on adding a comment pointing to the Node.js bug report.

ExE-Boss: I’ve opened <https://github.com/nodejs/node/issues/32302>.

domenic: Unfortunately it's still not clear from that bug report whether the problem is a Node.js bug or a bug in your code/in jsdom. As I commented over there, we need a minimal reproduction case (not involving jsdom) before we can accept this as a Node bug.

ExE-Boss: Well, <https://github.com/nodejs/node/issues/32302> has now been accepted as a **Node** bug.

domenic: That does not appear to be what has happened.

ExE-Boss:
The tests are passing locally, [but **Travis** currently has an outage](https://www.traviscistatus.com/).

ExE-Boss:
@domenic The **Travis CI** outage is over and the tests passed with the `Buffer` usage bugs fixed.

test/api/encoding.js
```diff
@@ -12,19 +12,24 @@ function fixturePath(fixture) {
...
+  return fs.promises.readFile(fixturePath(fixture));
 }
 
 const factories = {
   Buffer: fixture => readFixture(fixture),
-  Uint8Array: fixture => readFixture(fixture).then(buffer => Uint8Array.from(buffer)),
-  ArrayBuffer: fixture => readFixture(fixture).then(buffer => buffer.buffer),
-  DataView: fixture => readFixture(fixture).then(buffer => new DataView(buffer.buffer)),
+  Uint8Array: fixture => readFixture(fixture).then(buffer =>
+    new Uint8Array(buffer, buffer.byteOffset, buffer.length)),
```
TimothyGu:
```suggestion
    new Uint8Array(buffer.buffer, buffer.byteOffset, buffer.byteLength)),
```

Also as a nit: could you change all the `buffer.length`s to `buffer.byteLength`s for symmetry with `byteOffset`?

ExE-Boss: **Node**’s `Buffer` documentation says to use `buffer.length`: <https://nodejs.org/api/buffer.html#buffer_buf_byteoffset>.

TimothyGu: They are the same thing for `Buffer` since it's really a `Uint8Array`, but I think stylistically `byteLength` fits better.

PR #2840, opened 2020-02-16
Fix leaking memory during parsing

domenic:
The monkey-patch for maintaining the open element stack, as modified by commit 75a921eab52c239b2468e08be9f547f46c7f86bd, maintained a reference to the most recent `JSDOMParse5Adapter` (and all its options) through the push/pop closures. This means there was always a reference from the rooted `OpenElementStack.prototype.{push,pop}` functions to the most recent `JSDOMParse5Adapter` and its related objects.

This commit instead uses the `this.treeAdapter` pointer, which already exists in parse5. Doing so also allows us to move the monkey-patch back into the top level, instead of establishing it on construction of each tree adapter.

Closes #2831. Closes #2825. Fixes https://github.com/facebook/jest/issues/9507.

---

Hi @lamhieu-vk and @terite; thank you very much for your pull requests to fix this problem, and sorry for not getting to them sooner. I think I found a slightly more elegant way of making this work, by using the `this.treeAdapter` pointer. However, I don't have any way of confirming that this fix works as well as yours do. Which brings up another problem, which is that neither of your pull requests added a regression test we could use to prevent this in the future.

Although I'm willing to merge this (or, if this doesn't work, one of your PRs) without a regression test, I'd like to get confirmation that it fixes the problem first, so if you could help verify that, it would be much appreciated. And if you were able to suggest some way of adding code to the jsdom test suite to prevent the problem in the future, that would be even better.

Thank you!

### DIFF
--- a/lib/jsdom/browser/parser/html.js
+++ b/lib/jsdom/browser/parser/html.js
@@ -12,9 +12,32 @@ const nodeTypes = require("../../living/node-type");
 
 const serializationAdapter = require("../../living/domparsing/parse5-adapter-serialization");
 
+// Horrible monkey-patch to implement https://github.com/inikulin/parse5/issues/237 and
+// https://github.com/inikulin/parse5/issues/285.
 const OpenElementStack = require("parse5/lib/parser/open-element-stack");
-const OpenElementStackOriginalPop = OpenElementStack.prototype.pop;
-const OpenElementStackOriginalPush = OpenElementStack.prototype.push;
+
+const openElementStackOriginalPush = OpenElementStack.prototype.push;
+OpenElementStack.prototype.push = function (...args) {
+  openElementStackOriginalPush.apply(this, args);
+  this.treeAdapter._currentElement = this.current;
+
+  const after = this.items[this.stackTop];
+  if (after._pushedOnStackOfOpenElements) {
+    after._pushedOnStackOfOpenElements();
+  }
+};
+
+const openElementStackOriginalPop = OpenElementStack.prototype.pop;
+OpenElementStack.prototype.pop = function (...args) {
+  const before = this.items[this.stackTop];
+
+  openElementStackOriginalPop.apply(this, args);
+  this.treeAdapter._currentElement = this.current;
+
+  if (before._poppedOffStackOfOpenElements) {
+    before._poppedOffStackOfOpenElements();
+  }
+};
 
 class JSDOMParse5Adapter {
   constructor(documentImpl) {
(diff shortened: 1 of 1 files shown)


domenic:
Another thing that was puzzling us, is that as described this would be a limited leak, right? Because it only ever references the most-recent adapter and related objects, the leak would not continue growing; it would always be about the size of one JSDOM. So #2757 is probably not related, is our thinking.

lamhieu-vk:
In the teardown of jest jsdom environment, it calling to close window but the references of jsdom still exsiting in parser5 (OpenElementStack) and gc is cannot clear it.

domenic:
Sorry, that doesn't really clarify anything.

lamhieu-vk:
In my PR, I passed tests of jsdom and checked it not leak in jest jsdom environment.

lamhieu-vk:
I think before you merge this PR. Last check, please clone [this repo](https://github.com/dylanwulf/jest-25-detect-leak), link new jsdom to `jest-jsdom-environment` and run tests.

domenic:
I can't figure out any way to test this.

@lamhieu-vk's repository says "your test suite is leaking memory" with every version of jsdom; it does not distinguish between jsdom before https://github.com/jsdom/jsdom/commit/75a921eab52c239b2468e08be9f547f46c7f86bd, jsdom 16, jsdom with his patch, or jsdom with my patch.

The following script always outputs "leaking? true" for me, no matter whether I use my version of html.js, @lamhieu-vk's version, or @terite's version:

```js
"use strict";
const weak = require("weak-napi");
const { setFlagsFromString } = require("v8");
const { runInNewContext } = require("vm");
const { JSDOM } = require(".");

let dom = new JSDOM();

let isReferenceBeingHeld = true;
weak(dom.window, () => {
  isReferenceBeingHeld = false;
});

dom = null;

setFlagsFromString("--expose-gc");
runInNewContext("gc()");

setImmediate(() => {
  console.log("Leaking?", isReferenceBeingHeld);
});
```

SimenB:
@domenic by copying your patch to the reproduction repo it seemed to fix it.

```sh-session
$ cp ../jsdom/lib/jsdom/browser/parser/html.js node_modules/jest-environment-jsdom-sixteen/node_modules/jsdom/lib/jsdom/browser/parser

domenic:
Thanks; I'm not sure what was going wrong with my testing process before. That does indeed change before and after the patch, and is good enough that I'll do a merge.

It'd still be nice to find some way to create a test in isolation, that could be checked in to jsdom to prevent regressions, but that can wait.

SimenB:
I updated my comment with an updated script you can maybe use to test this is jsdom?

SimenB:
We've updated to jsdom@16 in Jest now, and as mentioned above include a memory leak smoke test

PR #2841, opened 2020-02-16
Return XMLDocument from document.implementation.createDocument()

domenic:
Currently in the spec document.implementation.createDocument() is specified to return an XMLDocument, not a Document. Additionally, once XMLDocuments can be created, they need to be cloned properly.

Co-Authored-By: ExE Boss <3889017+ExE-Boss@users.noreply.github.com>

Closes #2814 by superseding it.


lib/jsdom/browser/Window.js
```diff
@@ -119,21 +119,19 @@ function Window(options) {
...
+    encoding: options.encoding,
+    cookieJar: options.cookieJar,
+    url: options.url,
+    lastModified: options.lastModified,
+    referrer: options.referrer,
+    concurrentNodeIterators: options.concurrentNodeIterators,
+    parseOptions: options.parseOptions,
+    defaultView: this._globalProxy,
+    global: this
+  }, { alwaysUseDocumentClass: true });
```
ExE-Boss: Windows with XML documents in them create `XMLDocument` instances.

```suggestion
  });
```

lib/jsdom/living/documents.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+const XMLDocument = require("../living/generated/XMLDocument.js");
+const Document = require("../living/generated/Document.js");
+const { wrapperForImpl } = require("./generated/utils.js");
+
+exports.createImpl = (globalObject, options, { alwaysUseDocumentClass = false } = {}) => {
+  if (options.parsingMode === "xml" && !alwaysUseDocumentClass) {
```
ExE-Boss: If you forget to pass the `options` argument, then you will get a `TypeError`,

lib/jsdom/browser/Window.js
domenic: That is incorrect per spec.

lib/jsdom/living/documents.js
domenic: That seems good.

lib/jsdom/browser/Window.js
ExE-Boss: Well, this does result in `true` in all browsers:

```xml
<doc xmlns:html="http://www.w3.org/1999/xhtml">
<html:script><![CDATA[
document.documentElement.append(`${document instanceof XMLDocument}`);
]]></html:script>
</doc>
```

This is necessary for webcompat.

domenic: We won't implement that in jsdom without a spec.

ExE-Boss: Well, <https://github.com/whatwg/html/pull/5104> does fix that.

domenic: If/when that moves forward we can consider it in jsdom.

ExE-Boss:
Well, <https://github.com/jsdom/jsdom/pull/2814> is now equal to this, except it’s been rebased on top of `master`, including the creation of non‑`XMLDocument` instances for XML document windows.

ExE-Boss:
The build fails due to circular dependency issues caused by the execution order placing `XMLDocument‑impl.js` before `Document‑impl.js`, resulting in `DocumentImpl` being `undefined` when `XMLDocument‑impl.js` executes.

I’ve managed to fix this in <https://github.com/jsdom/jsdom/pull/2814>.

PR #2852, opened 2020-02-19
feat(events): Use `vm.compileFunction(…)`

ExE-Boss:
**Node v10** introduced [the `vm.compileFunction(…)` method](https://nodejs.org/api/vm.html#vm_vm_compilefunction_code_params_options), which compiles a function in the correct context.

This also moves the `with (window) { … }` wrapper into `vm‑shim.js`.

---

Extracted from: #2835


domenic:
What does this improve?

ExE-Boss:
It removes the need to use `with (window) { … }` in Node.

ExE-Boss:
Closing this as there are some dependencies on some other changes from <https://github.com/jsdom/jsdom/pull/2835>.

Issue #2408, opened 2018-10-31
"Error: Headers Authorization forbidden" in CORS


jonathanadler:
Hey,

I have a test running with `jsdom`, in which i'm sending a cross-origin request with an `Authorization` header (i'm using `nock` to mock the server).
As expected, an `OPTIONS` preflight request is sent.
However i'm receiving an exception, saying that the `Authorization` header is forbidden in the preflight request.
Since i'm not controlling which headers are being sent in the preflight request, I have a feeling that this might be a bug.

So I searched the code a bit, and tried to debug this, and I encountered this line of code:
https://github.com/jsdom/jsdom/blob/4bc7f53fed9a5387275f52ae07a85e7d4b790391/lib/jsdom/living/xhr-utils.js#L91

It seems that in the `validCORSPreflightHeaders` function, the original request headers are being validated instead of the preflight headers.

Is this correct? Or am I missing something?

domenic:
Can you provide a minimal example to reproduce this issue, per the issue template? Including important information like the versions in use, etc. We can't really debug based on this description.

jonathanadler:
@domenic Of course

I have a `package.json` file of:
```json
{
  "name": "testing",
  "version": "1.0.0",
  "description": "",
  "main": "",
  "scripts": {
    "test": "jest test.js"
  },
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "axios": "^0.18.0",
    "jest": "^23.6.0",
    "nock": "^10.0.1"
  },
  "jest": {
    "testEnvironment": "jsdom"
  }
}
```

And a `test.js` file like this:
```javascript
const nock = require('nock');
const axios = require('axios');

const HOST = 'http://localhost';
const PORT = '3333';
const somePath = '/some/path';

describe('jsdom cors', function () {
    const response = {foo: 'bar'};
    const headers = {
        'Authorization': 'Bearer someAccessToken',
    };

    beforeEach(() => {
        nock(`${HOST}:${PORT}`)
        .options(somePath)
        .reply(200, null, {'Access-Control-Allow-Origin': HOST})
        .get(somePath)
        .matchHeader('Authorization', headers['Authorization'])
        .reply(200, response, {'Access-Control-Allow-Origin': HOST});
    });

    afterEach(() => {
        nock.cleanAll();
    });

    it('should pass', done => {
        var request = {
            method: 'GET',
            url: `${HOST}:${PORT}${somePath}`,
            headers,
            timeout: 0
        };

[...]

PR #2867, opened 2020-02-25
fix: support wildcard for access-control-allow-headers

matveyok:
Regarding this issue: https://github.com/jsdom/jsdom/issues/2408
If `access-control-allow-headers` has `*` jsdom should allow all headers.


TimothyGu:
There seems to be some linting issues with this PR. Could you fix them? (See Travis CI for more info.)

We should also ideally have some tests for this, but it might be tricky as we need an HTTP server. Could you look at https://github.com/jsdom/jsdom/blob/master/Contributing.md#web-platform-feature-tests (you could add your test to to-upstream) and also maybe existing tests in WPT that use [Python handlers](https://web-platform-tests.org/writing-tests/python-handlers/index.html)?

PR #2868, opened 2020-02-26
fixed syntax error

mauzepeda:
The code wasn't working in some strict environments because it had a minor syntax error.

### DIFF
--- a/lib/jsdom/living/helpers/dates-and-times.js
+++ b/lib/jsdom/living/helpers/dates-and-times.js
@@ -232,7 +232,7 @@ function isDate(obj) {
   try {
     Date.prototype.valueOf.call(obj);
     return true;
-  } catch {
+  } catch (err) {
     return false;
   }
 }


domenic:
This is not a syntax error, this is valid JavaScript.

mauzepeda:
Hi @domenic when I try to deploy this package via npm into production it fails and it tell me that there is a syntax error, but this doesn't happen in my development environment. If you copy/paste the file source code into this checker you can see the error: https://esprima.org/demo/validate.html I believe it's related to the production environment using strict syntax rules and requiring an object in the catch. 

`try {
...
} catch (object) {
...
}`

Can you help me add this to the repo? So that I can deploy via npm?

Thank you,

Kind Regards,
Mauricio

domenic:
I don't know what "deploy via npm" means or what your "production environment" is.

Note from the readme

> The latest versions of jsdom require Node.js v10 or newer. (Versions of jsdom below v16 still work with previous Node.js versions, but are unsupported.)

mauzepeda:
@domenic as you said, this is not a syntax error. Sorry for my vague description. My production environment was running node 8 and my development environment was running node 12 hence the discrepancy. Node 10.3.0 and higher implemented [optional catch binding](https://github.com/tc39/proposal-optional-catch-binding) as per the ES2019 specification and that is why I was getting a syntax error on my production environment running node 8. I downgraded to jsdom@15.2.0 and I'm no longer getting the syntax error on node 8. Thank you!

PR #2877, opened 2020-03-01
Upgrade saxes to version 5.0.0.

lddubeau:
This PR upgrades saxes to version 5.0.0.

Why upgrade? Other than keeping up with the latest version, there's a BOM bug that is fixed in saxes 5.0.0 which in theory could be triggered in jsdom as long as jsdom remains on 4.x. (The other bug fixes in 5.0.0 fix issues that don't affect jsdom.)


TimothyGu:
Thanks as always!

PR #2878, opened 2020-03-01
fix: script element resoure not load

ZakaryCode:


domenic:
This is not correct. If canRunScript is false, we do not run the script.

PR #2881, opened 2020-03-06
add exception param to catch block to fix syntax error

strttn:
The missing "(e)" parameter on the catch block is causing a syntax error when attempting deploy the module to Google Cloud Functions.


domenic:
Only Node.js 10 and above are supported.

strttn:
Ah, ok. Apologies.
Thanks for the clarification.

Issue #2869, opened 2020-02-26
document.addEventListener lacks support for registering CustomElements as eventListener


Saegrov:
### Basic info:

- **Node.js version: 10.15.2** 
- **jsdom version: 16.2** 

### Minimal reproduction case

```js
const jsdom = require('jsdom');

const { JSDOM } = jsdom;

const dom = new JSDOM(
    `
    <!DOCTYPE html>
    <html lang="en">
      <head>
        <meta charset="UTF-8" />
        <title>No</title>
      </head>

      <body>
        <my-custom-click id="click">Click me!</my-custom-click>
        <script>
          class MyCustomClick extends HTMLElement {
            connectedCallback() {
              document.addEventListener('click', this);
              
              // Replace the line above with the line below and it works
              //document.addEventListener('click', (e) => this.handleEvent(e));
            }

            handleEvent(event) {
              if (event.target == this) {
                this.innerHTML = 'It worked!';
              }
            }
          }

          window.customElements.define('my-custom-click', MyCustomClick);
        </script>
      </body>
    </html>
  `,
    { runScripts: 'dangerously' }
);

const {
    window: { document },
} = dom;

const customElement = document.getElementById('click');
customElement.click();
if(customElement.innerHTML === 'It worked!') {
    console.log('Success!');
} else {
    throw Error('This did not work!')
}

```

### How does similar code behave in browsers?

This works in the browser. See example:
https://codesandbox.io/s/eloquent-flower-t5w08?fontsize=14&hidenavigation=1&theme=dark

Saegrov:
I've done some digging and in `jsdom/living/events/EventTarget.js` there is the following code:
```
      {
        let curArg = arguments[1];
        if (curArg === null || curArg === undefined) {
          curArg = null;
        } else {
          curArg = utils.tryImplForWrapper(curArg);
        }
        args.push(curArg);
      }
```

If I comment out `utils.tryImplForWrapper(curArg);` the example runs fine.
It seems that when adding `MyCustomClick` as an event listener, the code above replaces `MyCustomClick` with an `HTMLElementImpl`.


In `jsdom/living/EventTarget-impl` there is the following code for handling object as an eventListener if they have an `handleEvent` property:
```
      if (typeof listener.callback === "object") {
        if (typeof listener.callback.handleEvent === "function") {
          listener.callback.handleEvent(idlUtils.wrapperForImpl(eventImpl));
        }
      }
```

[...]

PR #2884, opened 2020-03-08
Convert EventListener and NodeFilter to use callback interfaces

ExE-Boss:
This updates **JSDOM** to use [callback interfaces](https://heycam.github.io/webidl/#dfn-callback-interface), for which I’m adding support to **WebIDL2JS** in <https://github.com/jsdom/webidl2js/pull/172>.

## Depends on:
- [x] <https://github.com/jsdom/webidl2js/pull/172> (merged)
- [x] <https://github.com/jsdom/webidl2js/pull/189> (merged)

---

Part of <https://github.com/jsdom/jsdom/issues/2727> (`NodeFilter`)
Fixes <https://github.com/jsdom/jsdom/issues/2869>


package.json
```diff
@@ -82,7 +82,7 @@
     "st": "^2.0.0",
     "watchify": "^3.11.1",
     "wd": "^1.12.1",
-    "webidl2js": "^14.0.0"
+    "webidl2js": "EB-Forks/webidl2js#feat/constructs/callback-interface"
```
ExE-Boss: Remember to use the production release of **WebIDL2JS**.

lib/jsdom/living/traversal/helpers.js
```diff
@@ -35,11 +35,7 @@ exports.filter = (nodeIteratorOrTreeWalkerImpl, nodeImpl) => {
 
   // https://github.com/whatwg/dom/issues/494
   try {
-    if (typeof filter === "function") {
-      result = filter(idlUtils.wrapperForImpl(nodeImpl));
-    } else {
-      result = filter.acceptNode(idlUtils.wrapperForImpl(nodeImpl));
-    }
+    result = filter(idlUtils.wrapperForImpl(nodeImpl));
```
TimothyGu, line 38: The `wrapperForImpl` would no longer be necessary, right?

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -322,13 +315,7 @@ function innerInvokeEventListeners(eventImpl, listeners, phase) {
...
 
     try {
-      if (typeof listener.callback === "object") {
-        if (typeof listener.callback.handleEvent === "function") {
-          listener.callback.handleEvent(idlUtils.wrapperForImpl(eventImpl));
-        }
-      } else {
-        listener.callback.call(eventImpl.currentTarget, idlUtils.wrapperForImpl(eventImpl));
-      }
+      listener.callback.call(eventImpl.currentTarget, idlUtils.wrapperForImpl(eventImpl));
```
TimothyGu, line 318: The `wrapperForImpl` would no longer be necessary, right?

lib/jsdom/living/interfaces.js
```diff
@@ -17,6 +15,7 @@ const generatedInterfaces = {
   URLSearchParams: require("whatwg-url/webidl2js-wrapper").URLSearchParams,
 
   EventTarget: require("./generated/EventTarget"),
+  NodeFilter: require("./generated/NodeFilter"),
```
TimothyGu, line 18: Maybe this should instead be grouped with `NamedNodeMap`?

lib/jsdom/living/events/EventTarget-impl.js
ExE-Boss: Apparently, that caused some weird issues with the way things interact in **JSDOM**: <https://travis-ci.com/github/jsdom/jsdom/builds/158224937>.

Probably something somewhere registers a listener that expects an event wrapper directly on the implementation, rather than going through the wrapper.

lib/jsdom/living/interfaces.js
ExE-Boss: I grouped it with `NodeIterator` and `TreeWalker`, which use `NodeFilter`.

lib/jsdom/living/events/EventTarget-impl.js
TimothyGu: Looks like it's https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/helpers/create-event-accessor.js#L11. `el` is an impl, so `event` there is already an impl.

ExE-Boss: It’s something else because I fixed that in <https://github.com/jsdom/jsdom/pull/2884/commits/6bfe5ab875819df23e055523cd026290196fb596#diff-32bfa972748b97cd78336775200b2e64>, but had to roll it back in <https://github.com/jsdom/jsdom/pull/2884/commits/3a2ca8285078c602e14e509006512be405f692ee> due to test failures.

TimothyGu: Could you then add a `if (!event) throw new Error();` to after https://github.com/jsdom/jsdom/pull/2884/commits/3a2ca8285078c602e14e509006512be405f692ee#diff-32bfa972748b97cd78336775200b2e64R11, so that you can catch that?

ExE-Boss: It’d be so much easier if we had the codebase annotated with **JSDoc**/**TypeScript** types, because then static analysis would make it possible to find the root of the issue much quicker.

domenic:
Please update this to use the latest webidl2js (and ideally webidl-conversions), then we can merge.

And please stop sending pull requests from separate organizations in the future, so that we can push changes to them. (No, I am not interested in joining your organization to work on my own project.)

PR #2889, opened 2020-03-10
refactor(StyleSheetList): Use implementation and generated wrapper

ExE-Boss:
Part of https://github.com/jsdom/jsdom/issues/2727.

### DIFF
--- a/lib/jsdom/living/cssom/StyleSheetList-impl.js
+++ b/lib/jsdom/living/cssom/StyleSheetList-impl.js
@@ -35,8 +35,4 @@ exports.implementation = class StyleSheetList {
       _list.splice(index, 1);
     }
   }
-
-  _values() {
-    return this._list.values();
-  }
 };
--- a/lib/jsdom/living/cssom/StyleSheetList-impl.js
+++ b/lib/jsdom/living/cssom/StyleSheetList-impl.js
@@ -1,0 +1,42 @@
+"use strict";
+
+const idlUtils = require("../generated/utils.js");
+
+exports.implementation = class StyleSheetList {
+  constructor() {
+    this._list = [];
+  }
+
+  get length() {
+    return this._list.length;
+  }
+
+  item(index) {
+    const result = this._list[index];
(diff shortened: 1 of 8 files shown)


lib/jsdom/living/cssom/StyleSheetList-impl.js
```diff
@@ -0,0 +1,42 @@
...
+  _remove(sheet) {
+    const { _list } = this;
+
+    const index = _list.indexOf(sheet);
+    if (index >= 0) {
+      _list.splice(index, 1);
+    }
+  }
+
+  _values() {
```
domenic: What is this for?
```suggestion
  _values() {
```
```suggestion
  _values() {
```
```suggestion
  _values() {
```

lib/jsdom/living/cssom/StyleSheetList.webidl
```diff
@@ -0,0 +1,6 @@
+[Exposed=Window]
+interface StyleSheetList {
+  [WebIDL2JSValueAsUnsupported=_null]
+  getter CSSStyleSheet? item(unsigned long index);
```
domenic: Can you put the extended attribute on the same line?

lib/jsdom/living/cssom/StyleSheetList.webidl
```diff
@@ -0,0 +1,6 @@
+[Exposed=Window]
```
domenic: Can you add a spec link? https://drafts.csswg.org/cssom/#the-stylesheetlist-interface

lib/jsdom/living/cssom/StyleSheetList-impl.js
ExE-Boss: That’s a leftover from what I think was an attempt at refactoring `forEachMatchingSheetRuleOfElement(…)` in [`style‑rules.js`](https://github.com/jsdom/jsdom/blob/master/jsdom/living/helpers/style-rules.js).

PR #2890, opened 2020-03-10
Convert skipped inline event handlers on body test to WPT

ExE-Boss:
It’s been bothering me that this is the only non‑**WPT** test that is skipped, so I decided to take a crack at fixing it.

---

This still fails in the browser, this time with a `TypeError` stack trace:
```
Error: Uncaught TypeError: Cannot read property '_ownerDocument' of undefined
    /tmp/lib/jsdom/living/events/EventTarget-impl.js:340:1 <- /tmp/d02cce143d09f2d44600f80c43efc350.browserify.js:6911
```

The line is in: <https://github.com/jsdom/jsdom/blob/19884d85f9c8f3e7ce2f3ad68c7b86d00674bef8/lib/jsdom/living/events/EventTarget-impl.js#L333-L343>

### DIFF
--- a/test/to-port-to-wpts/misc2.js
+++ b/test/to-port-to-wpts/misc2.js
@@ -385,37 +385,6 @@ describe("jsdom/miscellaneous", () => {
     assert.ok(a.getAttribute("style").match(/^\s*width\s*:\s*100%\s*;?\s*$/), "style attribute must contain width");
   });
 
-  // Test inline event handlers set on the body.
-  // TODO this currently fails!?
-  specify.skip("test_body_event_handler_inline", { skipIfBrowser: true, async: true }, t => {
-    // currently skipped in browsers because of an issue:
-    // TODO: https://github.com/jsdom/jsdom/issues/1379
-    const html = `
-      <html>
-        <head>
-          <script>
-            function loader () {
-              window.loader_called = true;
-            }
-          </script>
-        </head>
-        <body onload="loader()"></body>
-      </html>`;
-    const doc = (new JSDOM(html, { runScripts: "dangerously" })).window.document;
-    const window = doc.defaultView;
-    // In JSDOM, listeners registered with addEventListener are called before
-    // "traditional" listeners, so listening for "load" will fire before our
-    // inline listener.  This means we have to check the value on the next
-    // tick.
-    window.addEventListener("load", () => {
-      process.nextTick(() => {
-        assert.equal(window.loader_called, true);
-        t.done();
-      });
-    });
-    doc.close();
-  });
-
   specify("get_element_by_id", () => {
     const doc = (new JSDOM()).window.document;
     const el = doc.createElement("div");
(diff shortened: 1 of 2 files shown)


domenic:
Thanks for taking this on; it does seem bad. However instead of fixing our non-WPT tests, it'd be better to just port them to WPTs. What do you think of doing that instead? Hopefully the logic could become cleaner at the same time... all these random promises is pretty confusing.

ExE-Boss:
I tried porting this to WPTs, but I just can’t get the test to work at all in WPT format.

test/to-port-to-wpts/misc2.js
```diff
@@ -416,19 +421,30 @@ describe("jsdom/miscellaneous", () => {
...
+    let reject;
+    const result = new Promise((res, rej) => {
+      resolve = res;
+      reject = rej;
+    });
+
     // In JSDOM, listeners registered with addEventListener are called before
     // "traditional" listeners, so listening for "load" will fire before our
     // inline listener.  This means we have to check the value on the next
     // tick.
```
TimothyGu: Well, this just seems to be wrong. It looks like to me that per [spec](https://html.spec.whatwg.org/C/#event-handler-attributes), the `onload` content/traditional attribute should be called before ones added through JavaScript, if it is added first. Since the HTML one is added during parsing, it should be added pretty early I assume. Am I missing anything?

ExE-Boss: Turns out, this has already been fixed.

The test was just not updated.

TimothyGu:
I still think we can convert this to a WPT. What about something like this:

```html
<!doctype html>
<script src=/resources/testharness.js></script>
<script src=/resources/testharnessreport.js></script>
<script>
  setup({ single_test: true });
  function onloadHandler() {
    window.loader_called = true;
  }
</script>
<body onload="onloadHandler()">
  <div id=log></div>
  <script>
      assert_true(window.loader_called);
      done();
    });
  </script>
</body>
```

(Also, I'd be surprised if something like this doesn't already exist in WPT…)

joeyguerra:
I just ran into this situation. How can I help move this along?

domenic:
@joeyguerra what situation are you referring to? This pull request isn't really about a behavior of jsdom; it's about an internal refactoring to jsdom's test suite.

joeyguerra:
I ran into this error today. The scenario was client side script threw an exception and `wrapper` was undefined so Node threw an exception that hid the client side exception. `      } else if (wrapper && wrapper._ownerDocument) {` kept Node from throwing an exception so that I could see the client side exception.

domenic:
I still don't really know what error you're talking about, or what it has to do with this internal refactoring of the test suite. I'd suggest you file a new issue, following the issue template.

joeyguerra:
Hey @domenic, thanks for responding. I apologize, I didn't realize that this was about a test.

The situation I ran across is loading an HTML document into JSDOM where there was javascript executing in the window load event that had an error and JSDOM threw this exact error:

Error: Uncaught TypeError: Cannot read property '_ownerDocument' of undefined

What I noticed is that https://github.com/jsdom/jsdom/blob/19884d85f9c8f3e7ce2f3ad68c7b86d00674bef8/lib/jsdom/living/events/EventTarget-impl.js#L340 was where the error was thrown because `wrapper` was not defined.

domenic:
I'd suggest you file a new issue, following the issue template. We can't really help based on "loading an HTML document into jsdom"; we need a minimal reproduction case with no external dependencies involved.

joeyguerra:
Thanks @domenic

domenic (approved):

PR #2892, opened 2020-03-10
Support running WPTs in the browser

ExE-Boss:
This is an attempt to support running **Web Platform Tests** in the browser using [express](https://expressjs.com/) and [karma‑child‑process](https://www.npmjs.com/package/karma-child-process).

Note that as `test‑tuwpts.js` requires generating `tuwpt‑manifest.json`, I’ve left it as a **TODO**.


ExE-Boss:
So, for whatever reason, **Chrome** doesn’t seem to be receiving the response from `karma‑server.js`.

PR #2894, opened 2020-03-10
Update dates-and-times.js try catch with true syntax

Buminta:


domenic:
This is not correct. The syntax is correct as-is.

PR #2897, opened 2020-03-11
fix: setTimeout leaking handles

AviVahl:
When *not* in repeat mode (setInterval),  nothing removes `handle` from `listOfActiveTimers`.

We noticed this memory leak while using jest with jsdom.

Our code calls `setTimeout` with a large object in its closure, and listOfActiveTimers kept holding on to the handle, even though its `task()` completed.

Calling `stopAllTimers()` manually fixed it, as it cleared the entire map.


domenic:
Thanks! This will need tests before it can be merged though.

AviVahl:
Not sure how to test it without going into internals.
Other than leaking memory, this didn't affect behaviour.
Any chance you could point me to a relevant test file which tests `setTimeout`?

Trying to execute `yarn test` locally resulted in:
```
File "/home/avi/projects/jsdom/test/web-platform-tests/tests/tools/manifest/download.py", line 58, in merge_pr_tags
    for line in gitfunc("log", "--format=%D", "--max-count=%s" % max_count).split("\n"):
TypeError: a bytes-like object is required, not 'str'
```

Is it python 2 specific? Fedora 31 comes with python 3 by default, so I'm not sure how I'll run the tests.

domenic:
This seems like the same issue as https://github.com/jsdom/jsdom/issues/2886, which may give hints on testing.

I believe we do use Python 2, yes. I'm not sure how to install that on your system, but I hope you can figure it out.

AviVahl:
@domenic any further suggestions regarding how to test this?

I mean, the behaviour of `setTimeout`/`setInterval` is not changing. That `Map` is an internal implementation detail, and is not exposed to user-land.
I don't know the rest of the implementation very well, but I don't seem to find a way to test this from the outside. Any suggestions?

One possible way would be running node with `--expose-gc`, cleaning garbage collector, and logging heap usage size. It feels way too excessive.

AviVahl:
Apparently, web platform tests require `python` `python-pip` `virtualenv`, all of the python@2 variants.
This makes things way more complicated. I played around re-linking python to python2, but I'm out of my depth here.

I've looked at the linked `requestAnimationFrame` issue (#2886), and it appears to be a separate issue. It leaves the interval running, stopping node from ever closing.

In the case of `setTimeout`, it doesn't keep anything open and node properly closes. It just leaks memory for extended runs using setTimeout.

domenic:
Alright. One of the maintainers can take a look at working on this when they have time. (Usually weekends.)

AviVahl:
@domenic I added a test that spawns node with `--expose-gc` and exposes memory usage. It fails without the patch, and passes with it.

AviVahl:
A bit saner now. :) Added the test to from-outside.js and marked it to be skipped in browser.

PR #2901, opened 2020-03-13
fix(vm‑shim): Use `js‑globals.json`

ExE-Boss:
This should’ve been done in <https://github.com/jsdom/jsdom/pull/2743>.

### DIFF
--- a/lib/jsdom/vm-shim.js
+++ b/lib/jsdom/vm-shim.js
@@ -7,6 +7,7 @@ const jsGlobals = require("./browser/js-globals.json");
 
 // We can't use the default browserify vm shim because it doesn't work in a web worker.
 
+// "eval" is skipped because it's set to a function that calls `runInContext`:
 const jsGlobalEntriesToInstall = Object.entries(jsGlobals).filter(([name]) => name !== "eval" && name in global);
 
 exports.createContext = function (sandbox) {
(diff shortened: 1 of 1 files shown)


lib/jsdom/vm-shim.js
```diff
@@ -3,45 +3,24 @@
...
-const builtInConsts = ["Infinity", "NaN", "undefined"];
-const otherBuiltIns = [
-  "isFinite", "isNaN", "parseFloat", "parseInt", "decodeURI", "decodeURIComponent",
-  "encodeURI", "encodeURIComponent", "Array", "ArrayBuffer", "Boolean", "DataView", "Date", "Error", "EvalError",
-  "Float32Array", "Float64Array", "Function", "Int8Array", "Int16Array", "Int32Array", "Map", "Number", "Object",
-  "Proxy", "Promise", "RangeError", "ReferenceError", "RegExp", "Set", "String", "Symbol", "SyntaxError", "TypeError",
-  "Uint8Array", "Uint8ClampedArray", "Uint16Array", "Uint32Array", "URIError", "WeakMap", "WeakSet", "JSON", "Math",
-  "Reflect", "escape", "unescape"
-];
+const jsGlobalEntriesToInstall = Object.entries(jsGlobals).filter(([name]) => name !== "eval" && name in global);
```
domenic: Why skip eval? Maybe add a comment?

ExE-Boss: This is because `eval` is set to a custom function: <https://github.com/jsdom/jsdom/blob/713f97aeebf63e959f9682bd9fe5b8fd56b80cf7/lib/jsdom/vm-shim.js#L28-L30>

Issue #2830, opened 2020-02-07
`[ImplicitThis]` behaviour is incorrect


ExE-Boss:
### Basic info:

- **Node.js version:** v13.7.0
- **jsdom version:** v16.1.0

### Minimal reproduction case

```js
"use strict";
const jsdom = require("jsdom");

const options = {};
const dom = new jsdom.JSDOM("", options);

function log() {
	console.log(this, ...arguments);
}

const kCustom = "custom-event";
const customEvent = new dom.window.CustomEvent(kCustom, {
	detail: "foo",
	bubbles: true,
});

dom.window.addEventListener.call(dom.window.document.body, "custom-event", log);
dom.window.document.body.dispatchEvent(customEvent);
// Firefox logs: <body> CustomEvent { detail: "foo" }
// JSDOM logs: Window {...} CustomEvent { detail: "foo" }

try {
  dom.window.addEventListener.call(123, "custom-event", log);
  console.error("`window.addEventListener` allows any this parameter");
  // JSDOM logs: "`window.addEventListener` allows any this parameter"
} catch (err) {
  console.log(err);
  // Firefox logs: TypeError: number doesn't implement EventTarget
}

// Clear for part 2:
dom.window.removeEventListener.call(dom.window, "custom-event", log);
dom.window.removeEventListener.call(dom.window.document.body, "custom-event", log);

try {
  dom.window.EventTarget.prototype.addEventListener.call(undefined, "custom-event", log);
  dom.window.document.body.dispatchEvent(customEvent);
  // Firefox logs: Window {...} CustomEvent { detail: "foo" }
} catch (err) {
  console.error(err);
  // JSDOM logs: TypeError: Illegal invocation
}
```

### How does similar code behave in browsers?

See the `// Firefox logs:` comments in the above code snippet.

domenic:
Could you add a to-upstream WPT for these issues, so we can make sure it passes with https://github.com/jsdom/webidl2js/pull/166 ? Also a basic thing to test is that `window.addEventListener === EventTarget.prototype.addEventListener` (and same for the other methods).

PR #2902, opened 2020-03-14
fix(window): Use inherited methods from `EventTarget`

ExE-Boss:
Fixes <https://github.com/jsdom/jsdom/issues/2830>

## Depends on:
- [x] <https://github.com/jsdom/webidl2js/pull/166> (merged)
- [x] <https://github.com/jsdom/jsdom/pull/2839> (merged) – Because otherwise the tests won’t even run in CI, since **WebIDL2JS** no longer has `pn` among its dependencies.

---

review?(@domenic, @TimothyGu)


package.json
```diff
@@ -82,7 +82,7 @@
     "st": "^2.0.0",
     "watchify": "^3.11.1",
     "wd": "^1.12.1",
-    "webidl2js": "^14.0.0"
+    "webidl2js": "EB-Forks/webidl2js.git#feat/global/implicit-this"
```
ExE-Boss: Remember to use the production **WebIDL2JS** release.

domenic:
Note that #2839 is still waiting on your action to add a comment pointing to the Node.js bug.

TimothyGu:
Looks good, though we have to wait till a new release is made for webidl2js.

PR #2904, opened 2020-03-16
test(web‑platform‑tests): Fix `run‑tuwpts.js` on Windows

ExE-Boss:
On **Windows**, `python.exe` is an alias for `python3.exe`, which differs from ~~**Linux**~~ **Ubuntu**, where `python` is an alias for `python2`.

`run‑tuwpts.js` therefore fails to run on **Windows** because `wpt manifest` doesn’t currently work in **Python 3**: <https://github.com/web-platform-tests/wpt/issues/19836>.

### DIFF
--- a/test/web-platform-tests/run-tuwpts.js
+++ b/test/web-platform-tests/run-tuwpts.js
@@ -13,7 +13,7 @@ const manifestFilename = path.resolve(__dirname, "tuwpt-manifest.json");
 const testsRootArg = path.relative(wptPath, testsPath);
 const pathArg = path.relative(wptPath, manifestFilename);
 const args = ["./wpt.py", "manifest", "--tests-root", testsRootArg, "--path", pathArg];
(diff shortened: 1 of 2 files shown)


domenic:
Hmm. My Windows setup has `python` giving Python 2, `python3` giving Python 3, and no `python2` in the PATH. So I'm not sure what to do here... maybe update the docs to indicate `python` should point to Python 2 on one's system, like it does on Travis etc.?

ExE-Boss:
Well, **Travis** has `python2`, so I’d say we can merge this once **CI** passes.

domenic:
I'm not very excited about merging something that makes it harder to develop on my system, given that CI is fine either way.

TimothyGu:
To make things more confusing, on my [Arch](https://wiki.archlinux.org/index.php/Python#Python_2)-based Linux system, `python` in fact `python3` by default. (I have long ago symlinked `python` to `python2` on a per-user basis, but `python3` remains the default system setting.)

PR #2905, opened 2020-03-16
Make WPT server startup non‑blocking

ExE-Boss:
This makes the **WPT** server startup not block non‑WPT tests, allowing the tests to start faster.

~~It also closes the **WPT** server process as soon as it’s no longer needed.~~ (See <https://github.com/jsdom/jsdom/pull/2905#discussion_r394082401> for the reason why this doesn’t quite work.

---

Extracted from <https://github.com/jsdom/jsdom/pull/2892>.

### DIFF
--- a/test/web-platform-tests/run-tuwpts.js
+++ b/test/web-platform-tests/run-tuwpts.js
@@ -19,22 +19,22 @@ const manifest = readManifest(manifestFilename);
 const possibleTestFilePaths = getPossibleTestFilePaths(manifest);
 
 let wptServerURL;
-let serverProcess;
 const runSingleWPT = require("./run-single-wpt.js")(() => wptServerURL);
-const wptServerPromise = (async () => {
-  const { urls, subprocess } = await startWPTServer({ toUpstream: true });
-  wptServerURL = urls[0];
-  serverProcess = subprocess;
-})();
+const wptServerPromise = startWPTServer({ toUpstream: true });
+const serverProcess = wptServerPromise.child;
 
 describe("Local tests in web-platform-test format (to-upstream)", () => {
-  before({ timeout: 30 * 1000 }, () => wptServerPromise);
+  before({ timeout: 30 * 1000 }, async () => {
+    wptServerURL = await wptServerPromise;
+  });
 
   for (const test of possibleTestFilePaths) {
     runSingleWPT(test);
   }
+});
 
-  after(() => {
-    serverProcess.kill();
-  });
+after(() => {
+  // Python doesn't register a default handler for SIGTERM and it doesn't run __exit__() methods of context
+  // managers when it gets that signal. Using SIGINT avoids this problem.
+  serverProcess.kill("SIGINT");
 });
--- a/test/web-platform-tests/run-tuwpts.js
+++ b/test/web-platform-tests/run-tuwpts.js
@@ -21,18 +21,20 @@ const possibleTestFilePaths = getPossibleTestFilePaths(manifest);
 let wptServerURL;
 let serverProcess;
 const runSingleWPT = require("./run-single-wpt.js")(() => wptServerURL);
-before({ timeout: 30 * 1000 }, async () => {
+const wptServerPromise = (async () => {
   const { urls, subprocess } = await startWPTServer({ toUpstream: true });
   wptServerURL = urls[0];
   serverProcess = subprocess;
-});
-
-after(() => {
-  serverProcess.kill();
-});
+})();
 
 describe("Local tests in web-platform-test format (to-upstream)", () => {
+  before({ timeout: 30 * 1000 }, () => wptServerPromise);
+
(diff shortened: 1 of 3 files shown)


test/web-platform-tests/run-wpts.js
```diff
@@ -41,17 +41,15 @@ checkToRun();
 let wptServerURL;
 let serverProcess;
 const runSingleWPT = require("./run-single-wpt.js")(() => wptServerURL);
-before({ timeout: 30 * 1000 }, async () => {
+const wptServerPromise = (async () => {
   const { urls, subprocess } = await startWPTServer({ toUpstream: false });
   wptServerURL = urls[0];
   serverProcess = subprocess;
```
ExE-Boss: Arguably, this should be turned back into a `Promise.then(…)` callback, instead of an **IIAAFE** (Immediately‑invoked async arrow function expression).

TimothyGu: I tend to disagree. I generally prefer using `async` functions as much as possible.

test/web-platform-tests/run-tuwpts.js
```diff
@@ -21,18 +21,20 @@ const possibleTestFilePaths = getPossibleTestFilePaths(manifest);
...
+})();
 
 describe("Local tests in web-platform-test format (to-upstream)", () => {
+  before({ timeout: 30 * 1000 }, () => wptServerPromise);
+
   for (const test of possibleTestFilePaths) {
     runSingleWPT(test);
   }
+
+  after(() => {
```
TimothyGu: The `after` should still be outside the `describe`, as we are starting the WPT server unconditionally regardless if `to-upstream` tests are actually run. This bug might be reproducible if you try `mocha --grep=something-thats-not-wpt` or something similar.

When changing that, you might want to further make sure `serverProcess` is not undefined before you try to call `kill()` on it.

ExE-Boss: I decided to make the WPT subprocess available synchronously in the same way as `util.promisify(childProcess.exec)` behaves.

TimothyGu: Works for me.

test/web-platform-tests/start-wpt-server.js
```diff
@@ -30,39 +30,47 @@ module.exports = ({ toUpstream = false } = {}) => {
...
+    dnsLookup("web-platform.test")
+      .then(
+        () => {
+          return Promise.all([
+            pollForServer(`http://${config.browser_host}:${config.ports.http[0]}/`),
+            pollForServer(`https://${config.browser_host}:${config.ports.https[0]}/`),
+            pollForServer(`http://${config.browser_host}:${config.ports.ws[0]}/`),
+            pollForServer(`https://${config.browser_host}:${config.ports.wss[0]}/`)
+          ]);
+        },
```
TimothyGu:
```suggestion
        () => Promise.all([
          pollForServer(`http://${config.browser_host}:${config.ports.http[0]}/`),
          pollForServer(`https://${config.browser_host}:${config.ports.https[0]}/`),
          pollForServer(`http://${config.browser_host}:${config.ports.ws[0]}/`),
          pollForServer(`https://${config.browser_host}:${config.ports.wss[0]}/`)
        ]),
```

TimothyGu:
Looks good otherwise.

test/web-platform-tests/start-wpt-server.js
```diff
@@ -30,39 +30,45 @@ module.exports = ({ toUpstream = false } = {}) => {
...
-          reject(new Error("Error starting python server process:", e.message));
-        });
+    process.on("exit", () => {
+      // Python doesn't register a default handler for SIGTERM and it doesn't run __exit__() methods of context
+      // managers when it gets that signal. Using SIGINT avoids this problem.
+      subprocess.kill("SIGINT");
+    });
 
-        resolve(Promise.all([
+    dnsLookup("web-platform.test")
```
domenic: Why did this get rearranged to happen after server startup? I prefer surfacing these misconfiguration errors as soon as possible.

ExE-Boss: This is so that `subprocess` can be accessed synchronously as the `child` property of the returned `Promise` (same as `util.promisify(childProcess.exec`).

domenic: Hmm, that doesn't seem like a worthwhile tradeoff to me. It makes the test running experience worse just to make our code a bit simpler.

ExE-Boss: Well, if I don’t do this, then if the tests finish before DNS lookup completes, then the server will still start, but won’t be terminated in the `after` hook, because `serverProcess` will be `undefined` in the `after` hook: <https://github.com/jsdom/jsdom/pull/2905#discussion_r394082401>.

PR #2906, opened 2020-03-16
test(start‑wpt‑server): Make logging atomic

ExE-Boss:
The current **WPT** server polling implementation only synchronises after all **WPT** server ports are confirmed available, which results in the messages slowly getting out of sync, and with <https://github.com/jsdom/jsdom/pull/2905>, several **WPT** server polling logs end up interspersed between **Mocha** test logs, this lowers that to just one.

## Depends on:
- [ ] <https://github.com/jsdom/jsdom/pull/2905>

### DIFF
--- a/test/web-platform-tests/start-wpt-server.js
+++ b/test/web-platform-tests/start-wpt-server.js
@@ -41,14 +41,14 @@ module.exports = ({ toUpstream = false } = {}) => {
 
       return new Promise((resolve, reject) => {
         subprocess.on("error", e => {
-          reject(new Error("Error starting python server process:", e.message));
+          reject(new Error(`Error starting python server process: ${e.message}`));
         });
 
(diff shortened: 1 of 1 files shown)


TimothyGu:
I tend to prefer the original output of one-server-per-line. I suppose I ask for two things:

1. Could you make it so that all four lines are displayed in one `console.log` call? This will make logging atomic as well.

2. An improvement I'd be happy with is not printing duplicated "… is up" messages:

   ```
   A is up
   B isn't up yet
   ...
   B is up
   ```

   rather than

   ```
   A is up
   B isn't up yet
   ...
   A is up
   B is up
   ```

ExE-Boss:
I’ll do that after <https://github.com/jsdom/jsdom/pull/2905> is merged, because there will be merge conflicts.

PR #2909, opened 2020-03-19
fix(HTMLSelectElement): Update version in value setter

### LINKED ISSUE #2908: BUG with `select.selectedOptions`. Probably caching.
### Basic info:

- **Node.js version:** 13.11.0
- **jsdom version:** 16.2.1

### Minimal reproduction case

```js
const jsdom = require('jsdom')
const dom = new jsdom.JSDOM(`<select>
  <option value="x" data-foo="a">1st opt</option>
  <option value="y" data-foo="b">2nd opt</option>
</select>`)
const select = dom.window.document.querySelector('select')
select.value = 'x'
console.log(select.value, select.selectedOptions[0].dataset.foo) // comment this out and it works
select.value = 'y'
console.log(select.value, select.selectedOptions[0].dataset.foo)
```

If you set a value on a `select` field and then you mention `select.selectedOptions` and then you change the `select` field value again, the value does change, but the `selectedOptions` object stays the same.

Once you comment out the `select.selectedOptions` code between the 2 value setters, then it works correctly, showing the correct `selectedOptions` dataset value.

https://jsbin.com/dufivogoxe/edit?html,js,console,output
As you can see on jsbin, the browser does show the correct `selectedOptions` object even if you mention the `selectedOptions` in between the setters

ExE-Boss:
Fixes #2908, which happened because changing the `select.value` was not correctly updating the internal `_version` field, which is used to optimise `HTMLCollection`, `HTMLOptionCollection`, `NodeList` and several other things.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLSelectElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLSelectElement-impl.js
@@ -171,7 +171,9 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
   }
 
   set value(val) {
+    const seen = new Set([this]);
     let updated = false;
+
     for (const option of this.options) {
       updated = true;
       if (option.value === val) {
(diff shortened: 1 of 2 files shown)


lib/jsdom/living/nodes/HTMLSelectElement-impl.js
```diff
@@ -171,14 +171,20 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
...
   set value(val) {
+    let updated = false;
     for (const option of this.options) {
+      updated = true;
       if (option.value === val) {
         option._selectedness = true;
         option._dirtyness = true;
       } else {
         option._selectedness = false;
       }
```
TimothyGu: It seems better to call `option._modified()` here, rather than mucking with `_version` ourselves. This is particularly important as it updates `option._version` as well. See the `selected` setter of `HTMLOptionElementImpl`. (The `askForAReset` bit doesn't seem necessary however.)

ExE-Boss: `node._modified()` inherits `O(n)` complexity from `node._clearMemoizedQueries()`, where `n` is the number of ancestor nodes.

This results in calling `option._modified()` having `O(n*m)` complexity, where `n` is the number of option elements and `m` is the number of ancestor nodes from the `option` element.

TimothyGu: We can assume that `n` is small enough so that we don't have to worry about it. I don't believe the complexity tradeoff is worth it here.

Issue #1928, opened 2017-08-05
Don't support pseudo-element ?


dwqs:
### Basic info:

- **Node.js version:** v8.1.3
- **jsdom version:** v11.1.0

### Minimal reproduction case

```js
let jsdom = require('jsdom');
const JSDOM = jsdom.JSDOM;

const domString = `
    <!DOCTYPE html>
    <html>
        <head>
            <title>Test</title>
            <style>
                   .test{
                    color: red;
                    width: 50px;
                    height: 100px;
                }
                .test::before{
                    position: absolute;
                    content: 'sdas',
                    width: 20px;
                    height: 20px;
                }
                .test::after{
                    position: absolute;
                    content: 'sdas',
                    width: 40px;
                    height: 40px;
                }
           </style>
        </head>
        <body>
            <p class="test">test content</p>
        </body>
    </html>
`;
const dom = new JSDOM(domString);
const window = dom.window;
const document = window.document;
const p = document.querySelector('.test');

// get 50px but expected 40px
console.log('after', window.getComputedStyle(p, '::after').width);  
// get 50px but expected 20px
console.log('before', window.getComputedStyle(p, '::before').width); 
// get empty but expected sdas
console.log('content', window.getComputedStyle(p, '::before').content); 
```
Here is a [link](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) about `Window.getComputedStyle()` on MDN.

domenic:
Yeah, I can imagine we haven't implemented this. /cc @Zirro

Zirro:
The current `getComputedStyle`-implementation only supports the first argument, indeed. I'll see what I can do when I start reimplementing it in my package, but it will take a while before I can get there.

dwqs:
Thanks and look forward to this feature.

TMG-MattNewman:
I'm having a sort of similar problem, found this and just wanted to let you know, your CSS is invalid, it won't work even in a browser, it's the comma after content `content: 'sdas',`

3rd-Eden:
Is there a roadmap for this? `:before` and `:after` are commonly used in today's web.

chrifmarwen:
It would be very interessting if jsdom could support this.

getComputedStyle(element[, pseudoElt]);

milahu:
how can i help?

any advice on how/where to implement pseudo elements? @domenic @Zirro (etc)

my idea is:
[...]

PR #2918, opened 2020-03-25
fix(getComputedStyle): throw on Shadow DOM pseudo‑elements

ExE-Boss:
Extracted from <https://github.com/jsdom/jsdom/pull/2835>.

This makes [`getComputedStyle(…)`](https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle) call `notImplemented(…)` when trying to use the `pseudoElt` parameter, and throw when trying to get the computed style of a **Shadow DOM** pseudo‑element.

See also <https://github.com/jsdom/jsdom/issues/1928>

## Spec:
- https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle

---

review?(@TimothyGu): The build passed

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -26,8 +26,8 @@ const Selection = require("../living/generated/Selection");
 const reportException = require("../living/helpers/runtime-script-errors");
 const { fireAnEvent } = require("../living/helpers/events");
 const SessionHistory = require("../living/window/SessionHistory");
-const { forEachMatchingSheetRuleOfElement, getResolvedValue, propertiesWithResolvedValueImplemented } =
-  require("../living/helpers/style-rules");
+const { forEachMatchingSheetRuleOfElement, getResolvedValue, propertiesWithResolvedValueImplemented,
+  SHADOW_DOM_PSEUDO_REGEXP } = require("../living/helpers/style-rules.js");
 const CustomElementRegistry = require("../living/generated/CustomElementRegistry");
 const jsGlobals = require("./js-globals.json");
 
@@ -642,6 +642,20 @@ function Window(options) {
 
   this.getComputedStyle = function (elt) {
     elt = Element.convert(elt);
+    let pseudoElt = arguments[1];
+    if (pseudoElt !== undefined && pseudoElt !== null) {
+      pseudoElt = webIDLConversions.DOMString(pseudoElt);
+    }
+
+    if (pseudoElt !== undefined && pseudoElt !== null && pseudoElt !== "") {
+      // TODO: Parse pseudoElt
+
+      if (SHADOW_DOM_PSEUDO_REGEXP.test(pseudoElt)) {
+        throw new TypeError("Tried to get the computed style of a Shadow DOM pseudo-element.");
+      }
+
+      notImplemented("window.computedStyle(elt, pseudoElt)", this);
+    }
 
     const declaration = new CSSStyleDeclaration();
     const { forEach } = Array.prototype;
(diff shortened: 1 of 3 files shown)


TimothyGu:
We can't really merge this without tests.

ExE-Boss:
I thought that there were tests for this in WPT already as part of a bigger test that we fail, but it turns out that there currently aren’t any.

lib/jsdom/browser/Window.js
```diff
@@ -642,6 +642,20 @@ function Window(options) {
...
+    }
+
+    if (pseudoElt !== undefined && pseudoElt !== null && pseudoElt !== "") {
+      // TODO: Parse pseudoElt
+
+      notImplemented("window.computedStyle(elt, pseudoElt)", this);
+
+      if (SHADOW_DOM_PSEUDO_REGEXP.test(pseudoElt)) {
+        throw new TypeError("Tried to get the computed style of a Shadow DOM pseudo-element.");
+      }
```
TimothyGu: I don't really understand this. Doesn't `notImplemented` already throw an error, and thus make this if-statement dead code?

ExE-Boss: No, `notImplemented` simply logs a warning to the console: <https://github.com/jsdom/jsdom/blob/7242c4b238694a75b1a975572b767626dd550873/lib/jsdom/browser/not-implemented.js#L3-L13>

lib/jsdom/browser/Window.js
```diff
@@ -642,6 +642,20 @@ function Window(options) {
...
     elt = Element.convert(elt);
+    let pseudoElt = arguments[1];
+    if (pseudoElt !== undefined && pseudoElt !== null) {
+      pseudoElt = webIDLConversions.DOMString(pseudoElt);
+    }
+
+    if (pseudoElt !== undefined && pseudoElt !== null && pseudoElt !== "") {
+      // TODO: Parse pseudoElt
+
+      notImplemented("window.computedStyle(elt, pseudoElt)", this);
```
domenic, line 653: Let's move this after the throw since we do in fact implement that tiny part of the algorithm.

radar:
I'm getting a bunch of these `console.error` messages which I think are related. About 14k lines of these happen in my CI atm:

```
  console.error
    Error: Not implemented: window.computedStyle(elt, pseudoElt)
        at module.exports (/Users/ryan/code/covidence/covidence/node_modules/jsdom/lib/jsdom/browser/not-implemented.js:9:17)
        at Window.getComputedStyle (/Users/ryan/code/covidence/covidence/node_modules/jsdom/lib/jsdom/browser/Window.js:657:7)
        at computeMiscTextAlternative (/Users/ryan/code/covidence/covidence/node_modules/dom-accessibility-api/sources/accessible-name.ts:347:24)
        at computeTextAlternative (/Users/ryan/code/covidence/covidence/node_modules/dom-accessibility-api/sources/accessible-name.ts:576:11)
        at computeAccessibleName (/Users/ryan/code/covidence/covidence/node_modules/dom-accessibility-api/sources/accessible-name.ts:607:3)
        at /Users/ryan/code/covidence/covidence/node_modules/@testing-library/dom/dist/queries/role.js:96:82
        at Array.filter (<anonymous>)
        at queryAllByRole (/Users/ryan/code/covidence/covidence/node_modules/@testing-library/dom/dist/queries/role.js:90:6)
        at /Users/ryan/code/covidence/covidence/node_modules/@testing-library/dom/dist/query-helpers.js:82:17
        at /Users/ryan/code/covidence/covidence/node_modules/@testing-library/dom/dist/query-helpers.js:62:17 undefined
```

I am using `jsdom@16.4.0`, which I believe is the latest release.

Could we please get a new release with this commit included?

domenic:
jsdom 16.4.0 includes this commit, which introduces the warning for when you do something which will work differently in the browser vs. jsdom.

radar:
Rightio then. I bet this is something upstream that needs fixing. I'll go bother `dom-accessibility-api` folks then.

ExE-Boss:
@radar That has already been fixed in `dom‑accessibility‑api` [v0.5.0](https://github.com/eps1lon/dom-accessibility-api/blob/main/CHANGELOG.md#050).

PR #2921, opened 2020-03-26
Better activation behavior for <label> elements

### LINKED ISSUE #2717: Click events blow up call stack in nested <label>s
### Basic info:

- **Node.js version:** 8.16.0
- **jsdom version:** 15.1.1

### Description

When clicking an `<option>` in a `<select>`, there is a scenario where the click events are recursive, and never stop. It seems to stem from clicking the "associated" label element.

Here's the callstack frozen at a certain point:

image

Here is what the markup looks like:

```html
<div>
  <label for="gender-0">
    <span>Gender</span><span></span>
  </label>
  <div>
    <label>
      <select id="gender-0">
        <option value="">Select one</option>
        <option value="M">Male</option>
        <option value="F">Female</option>
        <option value="U">Unknown</option>
      </select>
    </label>
    </div>
</div>
```

**Edit:**

I realize the weird nature of having two `<label>` elements associated with the `<select>` (this is an effect of a component library I'm consuming). When I remove the second `<label>` from the example above, this problem goes away.

My question is at this point: Is this still a bug with `jsdom`, or is this invalid HTML? Even if it is the latter, maybe a better warning could be output in this scenario, and short-circuit to avoid the call stack problem.

TimothyGu:
This commit deals with two cases where the current activation behavior for `<label>` elements is deficient:

- When the target of the event that triggered said activation behavior is in fact part of some interactive content descendant of the `<label>`, then nothing should be done.

- When the target of the event is an inclusive descendant of the labeled control, then the control presumably has already received this event, so nothing more should be done by the `<label>` element.

The former case is specifically treated by the spec as a no-op case, under https://html.spec.whatwg.org/C/#the-label-element. This applies to such trees as

```html
<label>
  <button>Click me!</button>
</label>
```

Since the `<button>` is labelable, if upon clicking the button the `<label>` element tries to click the button once again, then an infinite recursion would result. This is indeed the essence of #2717.

The latter case is not mentioned in the spec directly, but is implemented by all major browsers. Consider instead the following tree:

```html
<label>
  <progress>Click me too!</progress>
</label>
```

The `<progress>` element is not interactive, and therefore does not fall under the jurisdiction of the previous case. But it is still labelable, and is the labeled control of the `<label>` element. So when the `<progress>` element is clicked, the `<label>` should also avoid trying to click the `<progress>` in a way that would result in infinite recursion.

It may first appear that the first case is a special case of the second; indeed, the example above with a single button would fall into the second case also. But this is not true in general. Consider instead:

```html
<label>
  <button>Don't click me!</button>
  <button>Click me!</button>
</label>
```

If the second button is clicked, then this falls into the first case but not the second. But we still need to prevent the `<label>` from clicking the first button (the actual labeled control) instead.

Fixes #2717.

---

Tests will be upstreamed through the Chromium project.

### DIFF
--- a/lib/jsdom/living/helpers/form-controls.js
+++ b/lib/jsdom/living/helpers/form-controls.js
@@ -80,22 +80,63 @@ exports.normalizeToCRLF = string => {
     .replace(/^\n/, "\r\n");
 };
 
-exports.isLabelable = node => {
-  // labelable logic defined at: https://html.spec.whatwg.org/multipage/forms.html#category-label
+// https://html.spec.whatwg.org/multipage/dom.html#interactive-content-2
+exports.isInteractiveContent = node => {
   if (node.nodeType !== NODE_TYPE.ELEMENT_NODE) {
     return false;
   }
+  if (node.namespaceURI !== HTML_NS) {
+    return false;
+  }
+  if (node.hasAttributeNS(null, "tabindex")) {
+    return true;
+  }
+  switch (node.localName) {
+    case "a":
+      return node.hasAttributeNS(null, "href");
 
-  switch (node.tagName) {
-    case "BUTTON":
-    case "METER":
-    case "OUTPUT":
-    case "PROGRESS":
-    case "SELECT":
-    case "TEXTAREA":
+    case "audio":
+    case "video":
+      return node.hasAttributeNS(null, "controls");
+
+    case "img":
+    case "object":
+      return node.hasAttributeNS(null, "usemap");
+
+    case "input":
+      return node.type !== "hidden";
+
+    case "button":
+    case "details":
+    case "embed":
+    case "iframe":
+    case "label":
+    case "select":
+    case "textarea":
+      return true;
+  }
+
+  return false;
+};
+
+// https://html.spec.whatwg.org/multipage/forms.html#category-label
+exports.isLabelable = node => {
+  if (node.nodeType !== NODE_TYPE.ELEMENT_NODE) {
+    return false;
+  }
+  if (node.namespaceURI !== HTML_NS) {
+    return false;
+  }
+  switch (node.localName) {
+    case "button":
+    case "meter":
+    case "output":
+    case "progress":
+    case "select":
+    case "textarea":
       return true;
 
-    case "INPUT":
+    case "input":
       return node.type !== "hidden";
   }
 
--- a/lib/jsdom/living/nodes/HTMLLabelElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLLabelElement-impl.js
@@ -4,7 +4,8 @@ const HTMLElementImpl = require("./HTMLElement-impl").implementation;
 const MouseEvent = require("../generated/MouseEvent");
 const { domSymbolTree } = require("../helpers/internal-constants");
 const NODE_TYPE = require("../node-type");
-const { isLabelable, isDisabled } = require("../helpers/form-controls");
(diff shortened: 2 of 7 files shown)


lib/jsdom/living/helpers/form-controls.js
```diff
@@ -80,8 +80,45 @@ exports.normalizeToCRLF = string => {
...
+// https://html.spec.whatwg.org/multipage/dom.html#interactive-content-2
+exports.isInteractiveContent = node => {
+  if (node.nodeType !== NODE_TYPE.ELEMENT_NODE) {
+    return false;
+  }
+
+  if (node.hasAttributeNS(null, "tabindex")) {
+    return true;
+  }
+  switch (node.tagName) {
```
domenic: Stil nit: prefer localName, as the fundamental thing, to tagName, the derived thing.

More substantive issue: I think a HTML namespace check is needed for all of these.

I realize both of these follow the `isLabelable` implementation below. It would be _ideal_ to update both, and add tests (for non-HTML elements of those tag names) for both... but the tests at least are not required. I'd just rather us do the better-practice thing for new code, and preferably update the nearby old code at the same time.

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -60,9 +61,28 @@ class HTMLLabelElementImpl extends HTMLElementImpl {
...
+        if (isInteractiveContent(ancestor)) {
+          return;
+        }
+      }
+    }
+
     const node = this.control;
     if (node && !isDisabled(node)) {
+      // Check if the control is an inclusive ancestor of the event's target (and has already received this event).
+      // If so, do nothing.
```
domenic: Can you file a spec bug (or PR) for this case? I think the spec should include it.

A bug might be nice because we could discuss different possibilities. E.g. adding another "must do nothing" case, vs. adding this algorithm to the spec (perhaps all browsers follow it?), vs. adding this algorithm to the spec but with the "sendClickToAssociatedNode" step as optional...

test/web-platform-tests/to-upstream/html/semantics/forms/the-label-element/clicking-noninteractive-content.html
```diff
@@ -0,0 +1,110 @@
...
+
+<script>
+"use strict";
+
+const labelableNotInteractiveElements =
+  Array.from(document.getElementById("labelable-not-interactive-content").content.children);
+const label = document.getElementById("label");
+
+// This part is may be subject to platform-dependent operations in the spec, so we only check for obvious errors.
+// (Clicking once should register at least one click, but less than 30 clicks.)
```
domenic: I would be curious on the cross-browser results if we changed this to check for exactly one click. (Something to consider for the above-mentioned HTML spec issue, probably.)

TimothyGu:
> Is there any particular significance of this? Since you're adding the tests to to-upstream, I'm guessing this means you plan to add them to web-platform-tests/wpt via Chromium? (And maybe, eventually, remove them from jsdom when we do a WPT roll?)

Yes precisely. That's what I'm planning to do right now, when I fix https://github.com/whatwg/html/issues/5414 in https://crrev.com/2122369. But plans will change if we decide to change the spec instead, so we'll see.

TimothyGu:
> LGTM but expanding the tests by adding some non-HTML namespace elements would be even better.

Okay, I added some tests.

PR #2923, opened 2020-03-27
Run beforeParse hook in all windows (frames' windows included)

### LINKED ISSUE #2920: beforeParse option is not executed in iframe windows
### Basic info:

The option beforeParse is only executed in the main window currently. New frames create their own windows [here](https://github.com/jsdom/jsdom/blob/7242c4b238694a75b1a975572b767626dd550873/lib/jsdom/living/nodes/HTMLFrameElement-impl.js#L128). This does not allow us to add any new APIs or currently missing ones (my case with [window.performance.timing](https://developer.mozilla.org/en-US/docs/Web/API/Performance/timing)).

- **Node.js version:** 12.16.1
- **jsdom version:** 16.2.1

### Minimal reproduction case

```js
const {JSDOM, VirtualConsole} = require('jsdom');

virtualConsole.on('jsdomError', e=>console.log('JSDOM Error', e));
virtualConsole.on('log', console.log.bind(console));

new JSDOM(`
    <html>
        <script>console.log('MAIN WINDOW', window.performance.timing.responseStart)</script>
        <iframe src="https://www.youtube.com/embed/kGRX0BG-qms" width="300" height="300">
        </iframe>
    </html>`, {
        virtualConsole,
        beforeParse: w=>{
            w.performance.timing = {responseStart: Date.now()};
        },
        runScripts: 'dangerously',
        resources: 'usable',
    },
);
```

#### Expected
MAIN WINDOW ${Date.now()} log and no "Cannot read property 'responseStart' of undefined" jsomError Errors emitted.

#### What happens
MAIN WINDOW ${Date.now()} log and "Cannot read property 'responseStart' of undefined" error due to window.performance.timing default undefined value.

Browsers have windows.performance.timing support.

FYI, I would like to submit a PR on this one, if approved.

GabrielLomba:
Run beforeParse hook for all windows created and add parameter to allow to determine whether this is the main window or not.

Fixes #2920


README.md
```diff
@@ -247,13 +247,16 @@ Cookie jars are provided by the [tough-cookie](https://www.npmjs.com/package/tou
 
 ### Intervening before parsing
 
-jsdom allows you to intervene in the creation of a jsdom very early: after the `Window` and `Document` objects are created, but before any HTML is parsed to populate the document with nodes:
+jsdom allows you to intervene in the creation of a jsdom very early: after the `Window` and `Document` objects are created, but before any HTML is parsed to populate the document with nodes. The callback is also called with new `Window` objects created for each frame:
 
 ```js
 const dom = new JSDOM(`<p>Hello</p>`, {
-  beforeParse(window) {
+  beforeParse(window, main) {
```
domenic: Can we make the API `{ isMain }` instead of a boolean? See https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

lib/jsdom/browser/Window.js
```diff
@@ -34,8 +34,17 @@ const jsGlobals = require("./js-globals.json");
 const GlobalEventHandlersImpl = require("../living/nodes/GlobalEventHandlers-impl").implementation;
 const WindowEventHandlersImpl = require("../living/nodes/WindowEventHandlers-impl").implementation;
 
-exports.createWindow = function (options) {
-  return new Window(options);
+let beforeParse;
+const beforeParseDefaultOptions = { isMain: false };
+
+exports.setBeforeParse = function (cb) {
```
domenic: beforeParse should not be a module-level variable, and `setBeforeParse` should not be needed. It should be one of the window options, not separate.

lib/jsdom/browser/Window.js
```diff
@@ -34,8 +34,17 @@ const jsGlobals = require("./js-globals.json");
 const GlobalEventHandlersImpl = require("../living/nodes/GlobalEventHandlers-impl").implementation;
 const WindowEventHandlersImpl = require("../living/nodes/WindowEventHandlers-impl").implementation;
 
-exports.createWindow = function (options) {
-  return new Window(options);
+let beforeParse;
+const beforeParseDefaultOptions = { isMain: false };
```
domenic: A module-level `beforeParseDefaultOptions` is not necessary. Just update `createWindow` to:

```js
exports.createWindow = function (options, beforeParseOptions = { isMain: false }) {
```

lib/jsdom/browser/Window.js
```diff
@@ -34,8 +34,17 @@ const jsGlobals = require("./js-globals.json");
...
+
+exports.setBeforeParse = function (cb) {
+  beforeParse = cb;
+};
+
+exports.createWindow = function (options, beforeParseOptions) {
+  const window = new Window(options);
+  beforeParse(window, Object.assign({}, beforeParseDefaultOptions, beforeParseOptions));
```
domenic: There's no need for this `Object.assign()`; you can just pass `beforeParseOptions` directly.

GabrielLomba:
@domenic , the check is failing due to wrong node version in tests. Are these current changes acceptable?

marcschroeder:
Hey, is there any progress on this? We would need this as soon as possible?

zek:
Any progress?

PR #2926, opened 2020-03-28
SVG improvements

### LINKED ISSUE #2734: Implement `SVGGElement`
I guess SVGGElement ([MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/SVGGElement)) is not supported in jsdom ?

### Basic info:

- **Node.js version:** 10.150.
- **jsdom version:** 11.5.1

### Minimal reproduction case

```js

const options = {
  ... your options here ...
};
const dom = new JSDOM(`
  <svg>
    <g id="g"> 
  </svg>
`, options);

const g = dom.window.document.querySelector("#g");

if (g instanceof SVGGElement) { // throws "SVGGElement is not defined"
  console.log("ok");
}
```

[https://jsfiddle.net/p3ag4v85/](https://jsfiddle.net/p3ag4v85/)

(check in the console and look for `ok`)

### Notes

I have to say that I noticed the bug when I tested my code using Jest, that is using jsdom internally. Here is the error I got:

```javascript
● should throw an Error if g element is not found

    expect(received).toThrowError(expected)

    Expected message: "group element not found"
    Received message: "SVGGElement is not defined"

          46 |         if (!(svgElement instanceof SVGSVGElement))
          47 |             throw new Error("svg element not found");
        > 48 |         if (!(groupElement instanceof SVGGElement))
             |                                       ^
          49 |             throw new Error("g element not found");
          50 |         this.groupElement = groupElement;
          51 |         this.groupBBoxWidth = 0;

          at new Lib (lib/js/index.js:48:39)
          at test/MyLib/constructor.test.js:86:15
          at Object.<anonymous> (node_modules/expect/build/toThrowMatchers.js:81:11)
          at Object.toThrowError (node_modules/expect/build/index.js:342:33)
          at Object.<anonymous> (test/MyLib/constructor.test.js:87:7)

      85 |     expect(
      86 |         () => new Lib({ svgIdentifier: "svg", groupIdentifier: "g" })
    > 87 |     ).toThrowError(new TypeError("group element not found"));
         |       ^
      88 | });
      89 | 

      at Object.<anonymous> (test/MyLib/constructor.test.js:87:7)
```

TimothyGu:
New elements: `<defs>`, `<desc>`, `<g>`, `<metadata>`, `<switch>`, `<symbol>`, `<use>`.

New method: `SVGSVGElement.createSVGRect`

Auxiliary classes:

- DOMRect (new)
- DOMRectReadOnly (new)
- SVGAnimatedPreserveAspectRatio (new, reflection support)
- SVGAnimatedRect (new, reflection support)
- SVGAnimatedString (reflection support)
- SVGPreserveAspectRatio (new)
- SVGRect (new)

Tests are being written:

- [x] DOMRect, DOMRectReadOnly
- [x] SVGAnimatedPreserveAspectRatio, SVGPreserveAspectRatio
- [x] SVGAnimatedRect, SVGRect
- [x] SVGAnimatedString
- [x] `<defs>`
- [x] `<desc>`
- [x] `<g>`
- [ ] `<metadata>`
- [ ] `<switch>`
- [x] `<symbol>`
- [ ] `<use>`

Fixes #2734.

----

Some roadmap for future work:

- Implement `SVGAnimatedLength` + reflection (complicated, as it requires us to implement a CSS dimension parser and unit converter – with support for relative units like `em`, ideally)
- Grind out more elements
- Scripting and styling and other fanciness


lib/jsdom/living/geometry/DOMRect-impl.js
```diff
@@ -0,0 +1,83 @@
+"use strict";
+const DOMRectReadOnlyImpl = require("./DOMRectReadOnly-impl").implementation;
+
+class DOMRectImpl extends DOMRectReadOnlyImpl {
```
domenic: The handling of reflected elements in this file seems like a bit of a layering violation, and does not show up at all in the spec for https://drafts.fxtf.org/geometry-1/#domrect. Is this a spec omission? Or is there an alternate implementation strategy that's closer to the spec, but we're choosing this one? In either case I think we need a comment, and perhaps we should file a spec issue if it's the former.

lib/jsdom/living/geometry/DOMRectReadOnly-impl.js
```diff
@@ -0,0 +1,95 @@
...
+    if (this._reflectedElement) {
+      const attr = this._reflectedElement.getAttributeNS(null, this._reflectedAttribute);
+      return this._parser(attr).height;
+    }
+    return this._height;
+  }
+
+  get top() {
+    const { height, y } = this;
+    if (isNaN(height)) {
```
domenic: This NaN check is not in the spec, if I am reading correctly. Another spec bug to file? Or maybe the same one, if it can only happen for the reflection case?

lib/jsdom/living/geometry/DOMRectReadOnly-impl.js
```diff
@@ -0,0 +1,95 @@
...
+
+  get left() {
+    const { width, x } = this;
+    if (isNaN(width)) {
+      return x;
+    }
+    return Math.min(x, x + width);
+  }
+
+  toJSON() {
```
domenic: A link to https://github.com/jsdom/webidl2js/issues/185 here, as well as in the webidl file, would be helpful.

lib/jsdom/living/nodes/SVGUseElement.webidl
```diff
@@ -0,0 +1,12 @@
+// https://svgwg.org/svg2-draft/struct.html#InterfaceSVGUseElement
+[Exposed=Window]
+interface SVGUseElement : SVGGraphicsElement {
+  // [SameObject] readonly attribute SVGAnimatedLength x;
```
domenic: Any reason to omit these? At least some of them seem pretty easy.

lib/jsdom/living/svg/SVGAnimatedString-impl.js
```diff
@@ -4,14 +4,22 @@ class SVGAnimatedStringImpl {
   constructor(globalObject, args, privateData) {
     this._element = privateData.element;
     this._attribute = privateData.attribute;
-    this._attributeDeprecated = privateData.attributeDeprecated; // can be undefined
+
+    if (privateData.attributeDeprecated) { // can be undefined
```
domenic: This seems a bit unnecessary. Are there cases where `privateData.attributeDeprecatedNamespace` will be non-undefined, but `privateData.attributeDeprecated` will be undefined? If there are no such cases, then we could just keep this as two assignments, right?

lib/jsdom/living/svg/SVGPreserveAspectRatio-impl.js
```diff
@@ -0,0 +1,98 @@
+"use strict";
+
+const DOMException = require("domexception/webidl2js-wrapper");
+
+const alignmentToStringMap = [
```
domenic: Hmm, although I see upon reading the code that this is a "map" from the numbers 0 - 10, it's a bit too clever. Some options:

- Add a comment explaining
- Change the name, maybe `alignmentStringsByIndex` or something
- Compute the inverse of `alignmentToIntMap` at runtime, making this `const alignmentToStringMap = inverse(alignmentToIntMap)`.
- Just spell it out by using an object literal with keys 0-10.

lib/jsdom/living/svg/SVGPreserveAspectRatio.webidl
```diff
@@ -0,0 +1,25 @@
...
+  const unsigned short SVG_PRESERVEASPECTRATIO_NONE = 1;
+  const unsigned short SVG_PRESERVEASPECTRATIO_XMINYMIN = 2;
+  const unsigned short SVG_PRESERVEASPECTRATIO_XMIDYMIN = 3;
+  const unsigned short SVG_PRESERVEASPECTRATIO_XMAXYMIN = 4;
+  const unsigned short SVG_PRESERVEASPECTRATIO_XMINYMID = 5;
+  const unsigned short SVG_PRESERVEASPECTRATIO_XMIDYMID = 6;
+  const unsigned short SVG_PRESERVEASPECTRATIO_XMAXYMID = 7;
+  const unsigned short SVG_PRESERVEASPECTRATIO_XMINYMAX = 8;
+  const unsigned short SVG_PRESERVEASPECTRATIO_XMIDYMAX = 9;
+  const unsigned short SVG_PRESERVEASPECTRATIO_XMAXYMAX = 10;
```
domenic: This type is so stupid.

scripts/webidl/convert.js
```diff
@@ -134,7 +144,97 @@ const transformer = new Webidl2js({
...
+            element: ${implObj},
+            attribute: "${attrName}"
+          });
+        `
+      };
+    }
+
     throw new Error("Unrecognized reflection type " + idl.idlType.idlType);
+
+    function getExtAttrValue(extAttrName) {
```
domenic: It's not a big deal, but I think it would be nicer to move this to top-level of the file, and pass in the `idl` input as needed.

Also, the reflection handling is getting pretty big in this file; we might want to consider splitting it up.

lib/jsdom/living/geometry/DOMRect-impl.js
TimothyGu: The spec was written in a way that requires observing attribute changes (when attributes get changed, set DOMRect component values accordingly) as well as a setter hook (when DOMRect gets changed reserialize attribute). So I'd argue that there is a layering violation, but only for the setter part. (See https://svgwg.org/svg2-draft/types.html#InterfaceSVGAnimatedRect.)

In terms of the getter part, I chose a different implementation strategy that's closer to how the other reflection classes work: only get the values when the user asked for it / called the getter. DOMRect is only used in conjunction with SVGAnimatedRect, so even though this approach doesn't scale in terms of complexity of the DOMRect class in case there are other ways of reflecting a DOMRect, it should be fine in practice.

I will add a comment documenting this, if you are okay with this approach.

lib/jsdom/living/geometry/DOMRectReadOnly-impl.js
TimothyGu: No this is different from the reflection. In fact this is much worse. Essentially, the question is `new DOMRectReadOnly(0, 0, NaN, NaN).top/bottom/left/right` NaN, or 0?

The spec says one of them should be the result of min(0, 0 + NaN), but doesn't say how to compare NaN with actual numbers. Implementations don't have particular handling either, relying instead of platform primitives like `std::min`. The result, is that Chrome and Firefox on Linux (as well as Chrome on macOS) seem to prefer non-NaNs, but Safari/WebKit on Linux and macOS prefer NaNs. But theoretically the behavior could vary between browsers but also between different operating systems, compiler version, standard library, compiler flags in use, the function (min vs max), etc.

A spec bug seems definitely appropriate, but I'm not sure about what we should do here. Too many choices! ;)

lib/jsdom/living/nodes/SVGUseElement.webidl
TimothyGu: See above:

> - Implement SVGAnimatedLength + reflection (complicated, as it requires us to implement a CSS dimension parser and unit converter – with support for relative units like em, ideally)

lib/jsdom/living/svg/SVGAnimatedString-impl.js
TimothyGu: Ah yeah, this was a remnant from a previous iteration of the commit. The `if` isn't needed anymore.

lib/jsdom/living/geometry/DOMRectReadOnly-impl.js
TimothyGu: Well, guess there's already one from two years ago: https://github.com/w3c/fxtf-drafts/issues/222

lib/jsdom/living/geometry/DOMRect-impl.js
domenic: Can we file an issue to add a reserialization hook to the DOMRect spec, and point to it from here?

For lazily-getting the values instead of eagerly, I am in favor of your strategy, with appropriate documentation of the isomorphism.

lib/jsdom/living/geometry/DOMRectReadOnly-impl.js
domenic: Interesting. I'd say we should just let the NaNs propagate according to the Math.min/max behavior. Some of the later comments there are leaning in that direction too. And it seems like you've gone in that direction as well, per https://github.com/w3c/fxtf-drafts/pull/395 :).

lib/jsdom/living/geometry/DOMRect-impl.js
TimothyGu: Done: https://github.com/w3c/fxtf-drafts/issues/396.

julienw:
I was looking for DOMRect support in jsdom and found this PR. Maybe it's ready for another review ?

lib/jsdom/living/geometry/DOMRect.webidl
```diff
@@ -0,0 +1,15 @@
+[Exposed=(Window,Worker),
+ Serializable,
+ LegacyWindowAlias=SVGRect]
+interface DOMRect : DOMRectReadOnly {
+    constructor(optional unrestricted double x = 0, optional unrestricted double y = 0,
+            optional unrestricted double width = 0, optional unrestricted double height = 0);
+
+    // https://github.com/jsdom/webidl2js/issues/186
+    // [NewObject] static DOMRect fromRect(optional DOMRectInit other = {});
```
ExE-Boss, line 9:
```suggestion
    [NewObject, WebIDL2JSCallWithGlobal] static DOMRect fromRect(optional DOMRectInit other = {});
```

ExE-Boss (commented):
@julienw
That requires **webidl2js v16.2.0**, but **JSDOM** first has to be fixed to work with **webidl2js v16.1.0**, which is being done in <https://github.com/jsdom/jsdom/pull/2956>.

Kingwl:
Hey folks. Any progress?

We have a huge svg based app and cannot work fine. I think this PR may resolve the problem.

Zearin:
Same as @julienw for me; was looking for `DOMRect[ReadOnly]` support in jsdom and found this.

I second @Kingwl’s curiosity, too.

(Great work on JSDom in general, and @TimothyGu for this awesome work!)

julienw:
#2956 seems to be fixed now :-)

okry:
Any updates on this? Looks like this PR has been sitting around for a while and my team could really use the `DOMRect` support with our application that uses radix.

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+const numberRe = /^[+-]?(?:\d*\.)?\d+(?:[eE][+-]?\d+)?/;
+
+// https://svgwg.org/svg2-draft/coords.html#ViewBoxAttribute
+function parseViewBox(str) {
+  // Use a do-while "loop" as JavaScript `goto end`.
+  // eslint-disable-next-line no-unreachable-loop
+  do {
+    // When the attribute does not exist.
+    if (typeof str !== "string") {
+      break;
```
ExE-Boss, line 15:
```suggestion
      break block;
```

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    let i = 0;
+    skipSpace();
+    const xStr = matchNumber();
+    if (!xStr) {
+      break;
```
ExE-Boss, line 22:
```suggestion
      break block;
```

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+    if (!skipDelimiter()) {
+      break;
```
ExE-Boss, line 25:
```suggestion
      break block;
```

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+    const yStr = matchNumber();
+    if (!yStr) {
+      break;
```
ExE-Boss, line 29:
```suggestion
      break block;
```

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+    if (!skipDelimiter()) {
+      break;
```
ExE-Boss, line 32:
```suggestion
      break block;
```

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+    const widthStr = matchNumber();
+    if (!widthStr) {
+      break;
```
ExE-Boss, line 36:
```suggestion
      break block;
```

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+    if (!skipDelimiter()) {
+      break;
```
ExE-Boss, line 39:
```suggestion
      break block;
```

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+    const heightStr = matchNumber();
+    if (!heightStr) {
+      break;
```
ExE-Boss, line 43:
```suggestion
      break block;
```

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+    // Test for trailing junk.
+    skipSpace();
+    if (i < str.length) {
+      break;
```
ExE-Boss, line 48:
```suggestion
      break block;
```

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    // A negative value for <width> or <height> is an error and invalidates the 'viewBox' attribute.
+    const width = Number(widthStr);
+    if (width < 0) {
+      break;
```
ExE-Boss, line 54:
```suggestion
      break block;
```

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+    const height = Number(heightStr);
+    if (height < 0) {
+      break;
```
ExE-Boss, line 58:
```suggestion
      break block;
```

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+    function skipDelimiter() {
+      const start = i;
+      skipSpace();
+      if (i < str.length && str[i] === ",") {
+        i += 1;
+      }
+      skipSpace();
+      return i > start;
+    }
+  } while (false);
```
ExE-Boss, line 92:
```suggestion
  }
```

lib/jsdom/living/svg/SVGAnimatedRect-impl.js
```diff
@@ -0,0 +1,122 @@
...
+const SVGRect = require("../generated/SVGRect");
+
+// https://drafts.csswg.org/css-syntax/#number-token-diagram
+const numberRe = /^[+-]?(?:\d*\.)?\d+(?:[eE][+-]?\d+)?/;
+
+// https://svgwg.org/svg2-draft/coords.html#ViewBoxAttribute
+function parseViewBox(str) {
+  // Use a do-while "loop" as JavaScript `goto end`.
+  // eslint-disable-next-line no-unreachable-loop
+  do {
```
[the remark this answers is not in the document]
Zearin: Obscure language features FTW!

Zearin:
Almost a year after my last visit to this issue. It looks like a lot of work has been going on in this PR. 

Can I request a re-review?

Would love to use DOMRect and friends in my JSDOM tests…

Zearin:
@domenic Thank you for #3555! 🙏 

@TimothyGu Are you still working on this? 

If not, can someone mark this “HELP WANTED” (or similar), and find a steward to take over to completion?

charlieforward9:
Found this from [here](https://stackoverflow.com/questions/54382414/fixing-react-leaflet-testing-error-cannot-read-property-layeradd-of-null/54384719?noredirect=1#comment138380781_54384719). Would be awesome to see this merged! What is the status here. I see its been a few months.

domenic:
Rebased and merged in https://github.com/jsdom/jsdom/pull/3843.

PR #2928, opened 2020-03-28
feat: Add TextEncoder and TextDecoder 

### LINKED ISSUE #2524: Implement `TextEncoder` and `TextDecoder`
TextEncoder and TextDecoder are added as globals in node11.
Maybe it is possible to use these for jsdom?

https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder
https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder

https://nodejs.org/api/globals.html#globals_textdecoder
https://nodejs.org/api/globals.html#globals_textencoder

I made a change in jest to support these in jest-environment-node.  https://github.com/facebook/jest/pull/8022 I can make simmilar change in jsdom but I am not sure were it should be done.

fguitton:
This aims at addressing the lack of TextEncoder and TextDecoder.
This PR relies on the Node native object when present and utilises `text-encoding` package for Node 10. 

Closes #2524

### DIFF
--- a/lib/jsdom/living/encoding/TextDecoder-impl.js
+++ b/lib/jsdom/living/encoding/TextDecoder-impl.js
@@ -1,0 +1,26 @@
+"use strict";
+
+const hasNode10 = process.versions.node && Number(process.versions.node.split(".")[0]) === 10;
+const { TextDecoder } = !hasNode10 ? global : require("text-encoding");
+
+exports.implementation = class TextDecoderImpl {
+  constructor(globalObject, args) {
+    this._internal = new TextDecoder(...args);
+  }
+
+  get encoding() {
+    return this._internal._encoding;
+  }
+
+  get fatal() {
+    return this._internal._fatal;
+  }
+
+  get ignoreBOM() {
+    return this._internal._ignoreBOM;
+  }
+
+  decode(input, options) {
+    return this._internal.decode(input, options);
+  }
+};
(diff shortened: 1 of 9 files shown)


TimothyGu:
Apologies for the lack of comment on that bug, but this should be done through the https://github.com/jsdom/whatwg-encoding package. Right now, whatwg-encoding only exposes some lower-level primitive, but it could be augmented to support `TextEncoder` and `TextDecoder` as well, much like [whatwg-url](https://github.com/jsdom/whatwg-url) (which has `URL` and `URLSearchParams` in addition to lower-level primitives).

You could take a look at https://github.com/jsdom/whatwg-url/blob/master/webidl2js-wrapper.js and https://github.com/jsdom/whatwg-url/blob/master/scripts/convert-idl.js for a recommended pattern of adding these classes there.

fguitton:
@TimothyGu Thank you for letting me know. I have started jsdom/whatwg-encoding#11 targeted at moving from the current structure to one which is more resembling that of `whatwg-url`.

As the `whatwg-encoding` repository suggests `text-encoding` should be used, but as it would turn out is not spec compliant and causes the test to break for Node 10. Would you be able to offer suggestions as to how to proceed ?

https://github.com/jsdom/whatwg-encoding/blob/04c383e309e0d3f198e19b050295995028b6cc48/README.md#L50

TimothyGu:
Thanks for doing that! Let's continue this conversation over at https://github.com/jsdom/whatwg-encoding/pull/11 then. We could see about integrating that change in jsdom after that gets merged.

Issue #2932, opened 2020-03-31
ResourceLoader Not Fetching Relative URLs On 16.2.2


benjamts:
I'd like to report a regression from 16.2.1 to 16.2.2. We see the issue when fetching relative URLs with a custom `ResourceLoader`. 

### Basic info:

- **Node.js version:** 12.13.1
- **jsdom version:** 16.2.2

### Minimal reproduction case

```js
const jsdom = require('jsdom');

class CustomResourceLoader extends jsdom.ResourceLoader {
  fetch(url, options) {
    console.log('intercepted fetch', url)
    return null;
  }
}

const dom = new jsdom.JSDOM(`
    <html>
    <body>
      <script>console.log('inline');</script>
      <script src="http://localhost:8000/absolute.js"></script>
      <script src="/relative.js"></script>
    </body>
    </html>
  `, {
  runScripts: 'dangerously',
  resources: new CustomResourceLoader(),
});
```

### How does similar code behave in a previous version

With v16.2.1 the following is logged (as expected):
```
inline
intercepted fetch http://localhost:8000/absolute.js
intercepted fetch /relative.js
```

However, on v16.2.2 I see the following:
```
inline
intercepted fetch http://localhost:8000/absolute.js
```

(Side note: Thanks for all the great work on jsdom! It's an incredibly useful tool.)

benjamts:
I tried some other types of subresources this morning (with the canvas package installed as well):
- On both 16.2.1 and 16.2.2 absolute URLs are fetched for stylesheets, images, iframes, and scripts. Data and file URLs are fetched in both versions. 
- On 16.2.2 stylesheets, images, and iframes display the same behavior as scripts (absolute URLs are fetched but relative ones are not.)
- On 16.2.1 relative paths are fetched for images and scripts, but not for stylesheets or iframes.

I also tried using `resources: 'usable'` instead of a custom loader and saw the same results in each case.

benjamts:
Ah, that makes sense. 👍 

I probably should have thought to try that, since I did see that it "affects things like resolution of relative URLs" in the docs. I think this primarily tripped me up because it was a change in behavior from the previous version.

I don't know what the desired behavior is exactly, but the proposed documentation change in https://github.com/jsdom/jsdom/pull/2936 would have guided me past this.

domenic:
I've looked in to this a bit more. There was a slight regression, but I think not going through the resource loader is more correct.

[...]

PR #2936, opened 2020-04-06
Document need for URLs when loading subresources

overlookmotel:
This PR adds a line to README stating that when loading subresources, they must be fully qualified URLs, or `url` option in JS DOM constructor must be used.

This may seem obvious, but it wasn't the case in JSDOM 15, so it tripped me up when I upgraded.

In JSDOM 15 `<script src="/bundle.js"></script>` in a DOM instance with no `url` defined would cause a custom `ResourceLoader` to have its `.fetch()` method called with url `/bundle.js`.

In JSDOM 16 `.fetch()` is not called and there is no error or console output to indicate that/why it's failed.

I'm not sure if silent failure is the desired behavior or not, but since it's changed, I thought it'd be good to document it so others don't hit the same problem when upgrading.

But if you feel the docs are too crowded already, feel free to reject this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,8 @@ By default, jsdom will not load any subresources such as scripts, stylesheets, i
 * Scripts, via `<script>`, but only if `runScripts: "dangerously"` is also set
 * Images, via `<img>`, but only if the `canvas` npm package is also installed (see "[Canvas Support](#canvas-support)" below)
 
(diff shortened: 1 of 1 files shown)


domenic:
This seems related to #2932. We haven't looked into that yet, but the resolution might be a fix, or documentation, or a louder error. Thanks for preparing the docs PR, in case that's the route we take!

PR #2948, opened 2020-04-18
fix(canvas): support drawing an empty canvas into other canvas

zjffun:


domenic:
Thanks! Per the contributing.md, can you add a web platform test, instead of a test in the legacy to-port-to-wpts folder?

zjffun:
All right, I will add.

zjffun:
@dcousens 

Web platform test added.

https://github.com/web-platform-tests/wpt/pull/23091

zjffun:
Remove the test case from to-upstream, because wpt has merged it.

PR #2950, opened 2020-04-20
[WIP] add to-run config for to-upstream wpt.

zjffun:


domenic:
What is the motivation for this? The point of to-upstream is that we always run them. A to-run config is not desired.

zjffun:
Some to upstream tests may need special env, such as `needs-canvas`.

domenic:
I see. Well, we need to figure out a way to that with less code, and definitely with less duplicated code. Perhaps tests under the canvas subdirectory should only run when canvas is enabled, or something of that sort?

zjffun:
I plan to extract common methods to `to-run-utils.js`. 

May be we can do this after https://github.com/web-platform-tests/wpt/pull/23070 .

domenic:
What do you think of just not having such a complicated, infrastructure, and instead adding a simple two-line filter around here for canvas tests? https://github.com/jsdom/jsdom/blob/master/test/web-platform-tests/run-tuwpts.js#L35-L37

zjffun:
Indeed, filter is enough. It seems not all canvas needed test in 2dcontext dir. Should we need a yaml file like this. https://github.com/jsdom/jsdom/blob/master/test/web-platform-tests/to-run.yaml#L564-L575

zjffun:
https://github.com/jsdom/jsdom/pull/2948/commits/ba51d4fb79d473937899fad38f70772597833e86

PR #2951, opened 2020-04-20
update wpt

zjffun:
- [x] remove `assert_throws`
- [x] fix `connect ECONNREFUSED`
- [x] update `to-run.yaml`
- [x] fix or triage TempUnknown in `to-run.yaml`


zjffun:
I write a script to auto update to-run.yaml, but not sure if this project need. Array of yaml can fold and unfold, this's very convenient, and sort tests can be done automatically.

This update bring about one hundred fail, I only found out a few reasons of them.

domenic:
Please revert the dramatic changes to to-run.yml! The file is intended to be a flat list.

domenic:
I pushed a commit which undoes the dramatic reformatting of to-run.yaml and goes through and updates it for a few cases. It still needs more updates, but hopefully you, or another project contributor, can do that without destroying the readability and editability of the file.

lib/jsdom/living/traversal/NodeFilter-impl.js
```diff
@@ -0,0 +1,47 @@
+"use strict";
+const { addConstants } = require("../utils");
+
+module.exports = function (core) {
```
TimothyGu, line 4: This file should not be here. After https://github.com/jsdom/jsdom/commit/87fcce674447d0468285f3ef70f068f274d1b36a, the file will be automatically generated by webidl2js.

Make sure your local dependencies are up-to-date.

zjffun:
@TimothyGu I push this file by accident thank you for reminding.

test/web-platform-tests/to-run.yaml
```diff
@@ -973,6 +1015,10 @@ Document-prototype-currentScript.html: [timeout, Test not up to date next with u
 DocumentOrShadowRoot-prototype-elementFromPoint.html: [fail, offsetTop not implemented]
 MouseEvent-prototype-offsetX-offsetY.html: [fail, offsetTop not implemented]
 ShadowRoot-interface.html: [fail, shadowRoot.styleSheet is not yet implemented]
+declarative/declarative-after-attachshadow.tentative.html: [fail, Unknown]
+declarative/declarative-shadow-dom-attachment.tentative.html: [fail, Unknown]
+declarative/declarative-shadow-dom-basic.tentative.html: [fail, need nwsapi fix (browsers also fail now)]
+declarative/script-access.tentative.html: [fail, need nwsapi fix (browsers also fail now)]
```
domenic, line 1021: We can skip this whole directory as unimplemented.

domenic:
Thanks very much for all your hard work here @zjffun. It looks like you're almost done. Let us know if we can help.

test/web-platform-tests/to-upstream/html/semantics/forms/the-label-element/clicking-interactive-content.html
```diff
@@ -19,8 +19,6 @@
   <select></select>
   <textarea></textarea>
   <video controls></video>
-
-  <div tabindex=0></div>
 </template>
 
 <script>
```
TimothyGu, line 24: You should remove this file and some others in the directory, as they've been upstreamed in https://github.com/web-platform-tests/wpt/pull/22749.

zjffun:
My pleasure. There is a question need help :

I write a small script, it's not good and only work in current `to-run-yaml` construction. I wender if this script needed, or change to using yaml AST to generate `to-run-yaml` (current used js-yaml can't support this, may need import an other package).
https://github.com/zjffun/jsdom/blob/rm-assert-throw/test/web-platform-tests/tools/update-to-run.js

lib/jsdom/living/nodes/HTMLAnchorElement-impl.js
```diff
@@ -9,8 +9,10 @@ class HTMLAnchorElementImpl extends HTMLElementImpl {
     super(globalObject, args, privateData);
 
     this._htmlHyperlinkElementUtilsSetup();
+  }
 
-    this._hasActivationBehavior = true;
+  get _hasActivationBehavior() {
+    return this.hasAttributeNS(null, "href");
```
TimothyGu, line 15: You said

> I misunderstood before, not all elements has activation behavior are interactive content. like `HTMLAreaElement`

Should these changes be reverted then?

lib/jsdom/living/nodes/HTMLImageElement.webidl
```diff
@@ -6,7 +6,7 @@ interface HTMLImageElement : HTMLElement {
   [CEReactions, ReflectURL] attribute USVString src;
   [CEReactions, Reflect] attribute USVString srcset;
   [CEReactions, Reflect] attribute DOMString sizes;
-  [CEReactions, Reflect] attribute DOMString? crossOrigin;
+  [CEReactions, Reflect] attribute CORSSettings? crossOrigin;
```
TimothyGu, line 9: These do not appear to be part of the spec and should be reverted.

lib/jsdom/living/helpers/cors-settings.js
```diff
@@ -0,0 +1,47 @@
...
+
+Object.entries(states).forEach(stateKV => {
+  const state = states[stateKV[0]];
+  const { keywords } = stateKV[1];
+  keywords.forEach(keyword => {
+    if (keyword !== null) {
+      keywordsMap.set(keyword.toLowerCase(), state);
+    }
+  });
+});
```
TimothyGu, line 24: This all looks a bit over-engineered. We generally represent constants with just strings. What do you think of this:

```suggestion
const invalidValueDefault = "anonymous";
const missingValueDefault = "no-cors";

const keywordsMap = {
  __proto__: null,
  anonymous: "anonymous",
  "": "anonymous",
  "use-credentials": "use-credentials"
};
```

The `keywordsMap` could be a `Map` if you wish to use that instead. `getKeyword` will no longer be necessary after this since the state is represented already by the the canonical keyword. For the "No CORS" state representing it with `null` seems to suffice.

The `mode` and `credentialsMode` information should not exist in this file. Instead, they should be encoded as part of a function implementing https://html.spec.whatwg.org/multipage/urls-and-fetching.html#create-a-potential-cors-request, which probably goes into `resource-loader.js`.

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -332,11 +334,11 @@ class HTMLInputElementImpl extends HTMLElementImpl {
     return null;
   }
 
-  _isRadioGroupChecked() {
-    if (this.checked) {
+  _isRadioGroup(name) {
```
TimothyGu, line 337: Let's call this `_isAnyInRadioGroup` or `_someInRadioGroup`.

lib/jsdom/living/nodes/HTMLAnchorElement-impl.js
zjffun: If this revert, activation behavior of `<a></a>` will be used. This may not the desired result. https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/events/EventTarget-impl.js#L121-L123

lib/jsdom/living/nodes/HTMLLabelElement-impl.js
```diff
@@ -24,10 +24,8 @@ function sendClickToAssociatedNode(node) {
 }
 
 class HTMLLabelElementImpl extends HTMLElementImpl {
-  constructor(globalObject, args, privateData) {
-    super(globalObject, args, privateData);
-
-    this._hasActivationBehavior = true;
+  get _hasActivationBehavior() {
+    return true;
```
TimothyGu, line 28: This one could be reverted also.

lib/jsdom/living/nodes/HTMLElement-impl.js
```diff
@@ -37,10 +39,11 @@ class HTMLElementImpl extends ElementImpl {
   // https://html.spec.whatwg.org/multipage/dom.html#the-translate-attribute
   get translate() {
     const translateAttr = this.getAttributeNS(null, "translate");
+    const translateAttrString = asciiLowercase(translateAttr || "");
```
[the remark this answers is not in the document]
zjffun: This seems want to do the things in the table below. I can't think a better implementation. @domenic

| | | | | | |
| --- | --- | --- | --- | --- | --- |
| translateAttr | null | '' | 'yes' | 'no' | 'gibberish' |
| asciiLowercase | '' | '' | 'yes' | 'no' | 'gibberish' |
| return | the logic behind | true | true | false | the logic behind |

zjffun:
It's weird that `tabindex-focus-flag.html` was taged fail in `to-run.yaml` https://github.com/zjffun/jsdom/blob/rm-assert-throw/test/web-platform-tests/to-run.yaml#L488, but failing in tarvis `at window.add_completion_callback (test/web-platform-tests/run-single-wpt.js:197:22)`. And I can't reporduct it on my computer.

lib/jsdom/living/nodes/HTMLAnchorElement-impl.js
```diff
@@ -9,10 +9,16 @@ class HTMLAnchorElementImpl extends HTMLElementImpl {
...
+  }
 
-    this._hasActivationBehavior = true;
+  // An anchor element has an activation behavior only if it creates a hyperlink.
+  // An anchor element represents a hyperlink if it has an href attribute.
+  get _hasActivationBehavior() {
+    return this.hasAttributeNS(null, "href");
   }
 
+  set _hasActivationBehavior(v) {}
```
TimothyGu, line 20: Is the setter necessary?

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -121,10 +121,15 @@ class HTMLInputElementImpl extends HTMLElementImpl {
...
     this._labels = null;
+  }
 
-    this._hasActivationBehavior = true;
+  // An input element has an activation behavior only if it's not hidden.
+  get _hasActivationBehavior() {
+    return this.type !== "hidden";
   }
 
+  set _hasActivationBehavior(v) {}
```
TimothyGu, line 131: Is the setter necessary?

test/web-platform-tests/tools/update-to-run.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  newToRunStrings.push("");
+
+  fs.writeFileSync(resolveWpt("to-run.yaml"), newToRunStrings.join(os.EOL));
+}
+
+function resolveWpt(...paths) {
+  return path.resolve(__dirname, "../", ...paths);
+}
```
TimothyGu, line 97: This looks somewhat useful, but it doesn't look like the result is being used in this PR. Is my understanding correct?

lib/jsdom/living/helpers/cors-settings.js
```diff
@@ -0,0 +1,38 @@
+"use strict";
+const { asciiLowercase } = require("./strings");
+
+// https://html.spec.whatwg.org/multipage/urls-and-fetching.html#cors-settings-attributes
+const invalidValueDefault = "anonymous";
+const missingValueDefault = "no-cors";
```
TimothyGu, line 6: You could use `null` for this value to mean "No CORS". This way we don't need an additional `no-cors` string. We can also remove `getKeword`, which seems to be unused anyway.

scripts/webidl/convert.js
```diff
@@ -71,6 +71,19 @@ const transformer = new Webidl2js({
...
+          get: `
+            const value = ${implObj}.getAttributeNS(null, "${attrName}");
+            return ${getKeyword}(value);
+          `,
+          set: `
+            ${implObj}.setAttributeNS(null, "${attrName}", V);
+          `
+        };
+      }
+
```
TimothyGu, line 86: It looks like the `crossOrigin` property is currently marked as `[Reflect]` in the IDL files. Reflection of an enumerated attribute (like `crossorigin`) isn't currently implemented, so we shouldn't use reflection here.

The right fix would be to remove the `[Reflect]` and delete this part. Then, add `crossOrigin` getters and setters in the implementation classes that use `cors-settings`. To do so, you could also create a mixin class like this: https://github.com/jsdom/jsdom/blob/8921128ee1efe6737e3ea200959eb01fe75ff3aa/lib/jsdom/living/nodes/HTMLAnchorElement-impl.js#L11 https://github.com/jsdom/jsdom/blob/8921128ee1efe6737e3ea200959eb01fe75ff3aa/lib/jsdom/living/nodes/HTMLAnchorElement-impl.js#L46

TimothyGu:
> It's weird that `tabindex-focus-flag.html` was taged fail in `to-run.yaml` https://github.com/zjffun/jsdom/blob/rm-assert-throw/test/web-platform-tests/to-run.yaml#L488, but failing in tarvis `at window.add_completion_callback (test/web-platform-tests/run-single-wpt.js:197:22)`. And I can't reporduct it on my computer.

The `tabindex-focus-flag.html` test was fixed in be0059f908515c04154db7ebfe096e0f3c697965. Could you try rebasing your branch onto master?

lib/jsdom/living/nodes/HTMLAnchorElement-impl.js
zjffun: Remove those two setters will cause error, becauce the parent class `HTMLElement` need set `_hasActivationBehavior`. 
https://github.com/zjffun/jsdom/blob/rm-assert-throw/lib/jsdom/living/nodes/HTMLElement-impl.js#L23

test/web-platform-tests/tools/update-to-run.js
zjffun: Dealing with all the `TempUnknown` and delete a few lines of this script's result is the current `to-run.yaml`. 
Emm.. This script can get all fail tests that not in `to-run.yaml`, and add them in the right place of `to-run.yaml` with `TempUnknown` as reason.

lib/jsdom/living/nodes/HTMLAnchorElement-impl.js
zjffun: Seems a bug.
```js
class C1 {
    constructor() {
        this.foo = 'test';
    }
}

class C2 extends C1 {
    get foo(){}
}

let obj = new C2
/**
Uncaught TypeError: Cannot set property foo of #<C2> which has only a getter
    at new C1 (repl:3:18)
    at new C2 (repl:1:1)
*/
```

```js
class C3 {
    get foo(){}
}

let obj2 = new C3
obj2.foo = 'test' // ok
```

scripts/webidl/convert.js
zjffun: > The crossorigin attribute, valid on the `<audio>`, ``, `<link>`, `<script>`, and `<video>` elements...

test/web-platform-tests/to-run.yaml
```diff
@@ -476,12 +485,14 @@ sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-negative.h
 sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-order.html: [fail, scrollIntoView() not implemented]
 sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-positive.html: [fail, scrollIntoView() not implemented]
 sequential-focus-navigation-and-the-tabindex-attribute/focus-tabindex-zero.html: [fail, scrollIntoView() not implemented]
+tabindex-focus-flag.html: [fail, https://github.com/jsdom/jsdom/pull/2965]
```
zjffun, line 488: https://github.com/jsdom/jsdom/pull/2965

test/web-platform-tests/tools/update-to-run.js
```diff
@@ -0,0 +1,97 @@
...
+const fs = require("fs");
+const os = require("os");
+const jsYAML = require("js-yaml");
+const Mocha = require("mocha");
+
+const mocha = new Mocha({});
+
+const toRunFilename = resolveWpt("to-run.yaml");
+const toRunString = fs.readFileSync(toRunFilename, { encoding: "utf-8" });
+const toRunDocs = jsYAML.safeLoadAll(toRunString, { filename: toRunFilename });
```
ExE-Boss, line 16: The second argument to [`loadAll`](https://github.com/nodeca/js-yaml#loadall-string--iterator---options-) and [`safeLoadAll`](https://github.com/nodeca/js-yaml#safeloadall-string--iterator--options-) is an optional callback function.

The third is the options arguments.

```suggestion
const toRunDocs = jsYAML.safeLoadAll(toRunString, null, { filename: toRunFilename });
```

zjffun: I dive into jsyaml code find `safeLoadAll` seems has a bug caused this fix can't work. Change to the code below can fix.

```js
const toRunDocs = jsYAML.loadAll(toRunString, null, { filename: toRunFilename,
  schema: jsYAML.DEFAULT_SAFE_SCHEMA });
```

zjffun: I thought the commit suggestion is just to submit my suggestion and then click it...

zjffun:
Thanks everyone for the review. I reorganized commits of this pr, if there is anything else I need to do please @ me.

zjffun:
I self reviewed this PR and found that it was messy and had some errors, so I split them into separate branches. This PR only keeps the changes of update wpt.

In this PR the merged test cases are deleted, and the test under the html/canvas folder is set to require canvas.

Some failed tests cannot be expressed using wildcards, because some of them will pass even if they are not implemented. Like `elementFromPoint-parameters.html` and `image-loading-eager.html`.


Some fixes for the new test cases:

- translate-enumerated-ascii-case-insensitive.html: https://github.com/zjffun/jsdom/commits/fix/ascii-case-insensitivity
- tabindex-focus-flag.html: https://github.com/jsdom/jsdom/pull/2965
- radio-valueMissing.html: https://github.com/zjffun/jsdom/commits/fix/radio-value-missing
- script-crossorigin-network.sub.html: fix https://github.com/zjffun/jsdom/commits/feature/crossorigin-attr
- 2d.drawImage.canvas.html: https://github.com/jsdom/jsdom/pull/2948

domenic:
This is amazing; thank you very much. I appreciate the splitting; that definitely will make things easier. I'll get this merged now and then we can proceed with the other PRs and branches.

domenic:
I noticed a few of the deleted files are not landed in web platform tests upstream. I will un-delete them.

PR #2965, opened 2020-05-08
Treat SVG a element with the href attribute as focusable

zjffun:


zjffun:
Sorry I forgot that there are other namespaces. The test is already in wpt https://github.com/web-platform-tests/wpt/blob/master/html/interaction/focus/tabindex-focus-flag.html 

Thank you for review.

ExE-Boss:
The tests are failing because `to‑run.yaml` needs updating.

lib/jsdom/living/helpers/focusing.js
```diff
@@ -60,6 +56,16 @@ exports.isFocusableAreaElement = elImpl => {
     // attribute is not implemented.
   }
 
+  if (elImpl._namespaceURI === HTML_NS || elImpl._namespaceURI === SVG_NS) {
```
ExE-Boss, line 59: `_namespaceURI` and `_localName` should probably be destructured:
```js
const { _namespaceURI, _localName } = elImpl;
```

domenic: No, underscore-prefixed local variables are bad style. (Please remember to avoid unsolicited drive-by reviews on other people's projects.)

zjffun:
I moved using tabindex value check a focusable area element to HTML namespacea, becaue I find SVG and HTML use tabindex to check focusable areas element in different ways and no focusable area in other namespacea ( https://www.w3.org/TR/MathML/ , https://www.w3.org/TR/xlink/ ) .

> - SVG: any directly rendered element with a valid integer ‘tabindex’ attribute
> - HTML: the element's tabindex value is non-null, or the element is determined by the user agent to be focusable;

domenic:
Hmm, it looks like a good amount of tests are failing now...

lib/jsdom/living/helpers/focusing.js
```diff
@@ -60,6 +61,12 @@ exports.isFocusableAreaElement = elImpl => {
     // attribute is not implemented.
   }
 
+  if (elImpl._namespaceURI === SVG_NS) {
```
ExE-Boss, line 64: This should probably be an `else if`, since the namespace URI can’t be both the `HTML_NS` and the `SVG_NS`.

```suggestion
  } else if (elImpl._namespaceURI === SVG_NS) {
```

zjffun: Thanks for reminding.
I return the `false` in advance so that others don’t have to read the code behind to know the return value when `elImpl._namespaceURI` is `HTML_NS`.

lib/jsdom/living/helpers/svg/render.js
```diff
@@ -0,0 +1,44 @@
...
+  "symbol"
+]);
+
+// https://svgwg.org/svg2-draft/render.html#Rendered-vs-NonRendered
+exports.isRenderedElement = function isRenderedElement(elImpl) {
+  if (neverRenderedElements.has(elImpl._localName)) {
+    return false;
+  }
+
+  if (elImpl.getAttributeNS(null, "display") === "none") {
```
domenic: display is not an attribute; it is a CSS property. I'll remove this since we don't have the counterpart check for HTML.

lib/jsdom/living/helpers/svg/render.js
```diff
@@ -0,0 +1,44 @@
...
+    return false;
+  }
+
+  if (!elImpl._attached) {
```
domenic: To be consistent with the check introduced in https://github.com/jsdom/jsdom/pull/3005, I'll change this to use isConnected.

lib/jsdom/living/helpers/svg/render.js
```diff
@@ -0,0 +1,44 @@
...
+    return false;
+  }
+
+  if (!elImpl.parentNode || elImpl.parentNode._namespaceURI !== SVG_NS) {
```
domenic: I think elements with no parent should not be rendered, probably? This might be redundant with isConnected; I'll try and see what happens...

lib/jsdom/living/helpers/svg/render.js
zjffun: display is still an attribute in SVG, and you are right the spec says **computed style** value of none for the display **property**.

lib/jsdom/living/helpers/svg/render.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  // The spec is unclear about how to deal with non-SVG parents, so we only perform this check for SVG-namespace
+  // parents.
+  if (elImpl.parentElement && elImpl.parentElement._namespaceURI === SVG_NS &&
+                              !exports.isRenderedElement(elImpl.parentNode)) {
+    return false;
+  }
+
+  return true;
```
zjffun, line 45: Is tail recursion appropriate here?

PR #2967, opened 2020-05-08
Fix several HTMLImageElement#complete bugs

strager:
jsdom implements HTMLImageElement#complete poorly. I noticed this when I observed .complete giving false despite the onerror callback being called. To address this, perform the following changes:

* Make .complete true if src is unset or is empty.
* Make .complete true if an error occurs loading the image.
* Make .complete false if .src is reset to an unloaded URL.

This commit fixes most (but not all) test cases in html/semantics/embedded-content/the-img-element/img.complete.html. This commit does not address srcset or the image list.


lib/jsdom/living/nodes/HTMLImageElement-impl.js
```diff
@@ -5,7 +5,16 @@ const HTMLElementImpl = require("./HTMLElement-impl").implementation;
 const { Canvas } = require("../../utils");
 const { parseURLToResultingURLRecord } = require("../helpers/document-base-url");
 
+const REQUEST_STATE_BROKEN = "broken";
+const REQUEST_STATE_COMPLETELY_AVAILABLE = "completely available";
+const REQUEST_STATE_UNAVAILABLE = "unavailable";
```
domenic, line 10: Let's eliminate these and just use the strings directly.

test/web-platform-tests/to-upstream/html/semantics/embedded-content/the-img-element/img.complete-2.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
```
domenic, line 1: Can you name this `-dont-upstream.html` since it's a copy of an existing file and we need to remember not to upstream it?

domenic: Hmm I see it's not a copy of an existing file... I'll give this some thought.

domenic:
I'm trying to push fixes to this branch but I cannot do so. Can you ensure that "Allow edits from maintainers" is checked, per https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork ?

PR #2975, opened 2020-05-16
Use clean global when setting up Window

### LINKED ISSUE #2961: globalThis is not defined on Node 10
We're currently running into issues [upgrading Create React App](https://github.com/facebook/create-react-app/pull/8955) to use the latest versions of Jest and jsdom (26 and 16 respectively). When running our test suite on Node 10.20.1, we receive this error: 

```
    ReferenceError: globalThis is not defined

      at evalmachine.<anonymous>:1:1
      at setupWindow (node_modules/jsdom/lib/jsdom/browser/Window.js:51:55)
      at new Window (node_modules/jsdom/lib/jsdom/browser/Window.js:107:3)
      at exports.createWindow (node_modules/jsdom/lib/jsdom/browser/Window.js:38:10)
      at new JSDOM (node_modules/jsdom/lib/api.js:36:20)
```

After digging in with @SimenB, we discovered that it appears to potentially be caused by a core-js polyfill of `globalThis` (we're still investigating if we can work around in some way).

Here is a repro as copied from @SimenB's comment here: https://github.com/jsdom/jsdom/issues/2795#issuecomment-624460723 (thanks for digging into the issue with me Simen!)

> 
> FWIW, this reproduces the issue:
> 
> // polyfill globalThis for older nodes
> globalThis = global;
> const { JSDOM } = require('jsdom');
> new JSDOM('', {runScripts: 'dangerously'});
> Setting globalThis to anything at all will reproduce, I just picked global to be somewhat realistic. An even more realistic polyfill might be require('core-js'); which will also break, and might (I haven't verified) be inserted by @babel/preset-env and/or @babel/runtime.
> 
> Not sure it's necessarily something that needs to be fixed/changed in JSDOM, but it might be made more robust*? Anyways, hopes this helps folks hitting this issue to track it down. It's not Jest itself that does this.
> 
> *) e.g. filtering on vm.runInContext('this', windowInstance) rather than global in the outer context. This might not be feasible, I haven't played with it.

ianschmitz:
Fixes #2961.

Use a clean instance of the global object when looking for global entries to install into the jsdom environment. This avoids issues when the global object has had APIs polyfilled that aren't available within a new vm context.

Credit goes to @SimenB for helping troubleshoot!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,3 @@ npm-debug.log*
 benchmark/browser-bundle.js
 
 lib/jsdom/living/generated/**/*.js
-lib/jsdom/browser/js-globals.json
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -38,7 +38,12 @@ exports.createWindow = function (options) {
   return new Window(options);
 };
 
-const jsGlobalEntriesToInstall = Object.entries(jsGlobals).filter(([name]) => name in global);
+// Grab clean global to avoid issues when global APIS have been polyfilled.
+// https://github.com/jsdom/jsdom/issues/2961
+const vmGlobal = vm.runInNewContext("this");
+const jsGlobalEntriesToInstall = Object.entries(jsGlobals).filter(
+  ([name]) => name in vmGlobal
+);
 
 // https://html.spec.whatwg.org/#the-window-object
 function setupWindow(windowInstance, { runScripts }) {
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -29,7 +29,7 @@ const SessionHistory = require("../living/window/SessionHistory");
 const { forEachMatchingSheetRuleOfElement, getResolvedValue, propertiesWithResolvedValueImplemented } =
   require("../living/helpers/style-rules");
 const CustomElementRegistry = require("../living/generated/CustomElementRegistry");
-const jsGlobals = require("./js-globals.json");
+const jsGlobals = require("./js-globals");
 
 const GlobalEventHandlersImpl = require("../living/nodes/GlobalEventHandlers-impl").implementation;
 const WindowEventHandlersImpl = require("../living/nodes/WindowEventHandlers-impl").implementation;
@@ -38,10 +38,7 @@ exports.createWindow = function (options) {
   return new Window(options);
 };
 
-// Grab clean global to avoid issues when global APIS have been polyfilled.
-// https://github.com/jsdom/jsdom/issues/2961
-const vmGlobal = vm.runInNewContext("this");
-const jsGlobalEntriesToInstall = Object.entries(jsGlobals).filter(([name]) => name in vmGlobal);
+const jsGlobalEntriesToInstall = Object.entries(jsGlobals).filter(([name]) => name in global);
 
 // https://html.spec.whatwg.org/#the-window-object
 function setupWindow(windowInstance, { runScripts }) {
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -41,9 +41,7 @@ exports.createWindow = function (options) {
 // Grab clean global to avoid issues when global APIS have been polyfilled.
 // https://github.com/jsdom/jsdom/issues/2961
 const vmGlobal = vm.runInNewContext("this");
-const jsGlobalEntriesToInstall = Object.entries(jsGlobals).filter(
-  ([name]) => name in vmGlobal
-);
+const jsGlobalEntriesToInstall = Object.entries(jsGlobals).filter(([name]) => name in vmGlobal);
 
 // https://html.spec.whatwg.org/#the-window-object
 function setupWindow(windowInstance, { runScripts }) {
--- a/lib/jsdom/browser/js-globals.js
+++ b/lib/jsdom/browser/js-globals.js
@@ -1,19 +1,11 @@
 "use strict";
 const vm = require("vm");
-const fs = require("fs");
-const path = require("path");
 const assert = require("assert");
 
 // Creates a list of globals from the JS environment (not the web), so that Window.js can alias them when necessary.
 // The generated list should match https://tc39.es/ecma262/#sec-global-object, to the extent V8 implements the spec.
-// We generate this at build time instead of runtime because we want to avoid the performance and memory overhead of
-// creating a new context when scripting is disabled in the JSDOM.
 
-const dest = path.resolve(__dirname, "../lib/jsdom/browser/js-globals.json");
-
-const context = vm.createContext();
-
-const globals = vm.runInContext("Object.getOwnPropertyDescriptors(this)", context);
+const globals = vm.runInNewContext("Object.getOwnPropertyDescriptors(this)");
 
 // I guess VM contexts have a console, from V8? That isn't a JS global and JSDOM will install its own, so don't include
 // that.
@@ -28,4 +20,4 @@ for (const [key, global] of Object.entries(globals)) {
   delete global.value;
 }
 
-fs.writeFileSync(dest, JSON.stringify(globals, undefined, 2) + "\n");
+module.exports = globals;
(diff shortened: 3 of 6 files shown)


domenic:
Can you add a test that fails before this change, and work to pass the lint check?

TimothyGu:
Other than the linting issue, this looks good. I'll let @domenic and @pmdartus take a look though as they are more familiar with this.

ianschmitz:
Started writing a test today, but it started to get a little gross so wanted to run it by you first.

I wrote a describe block similar to https://github.com/jsdom/jsdom/blob/8921128ee1efe6737e3ea200959eb01fe75ff3aa/test/api/options-run-scripts.js#L127-L149

```js
  // `globalThis` is not supported by Node 10 which makes it useful for testing
  const jsSpecGlobalsPolyfillDescribe = hasNode10 ? describe : describe.skip;
  jsSpecGlobalsPolyfillDescribe("JS spec polyfilled globals", () => {
    afterEach(() => {
      delete global.globalThis;
    });

    // Allows us to modify global object prior to running module code
    function getFreshJsdomModule() {
      delete require.cache[require.resolve("../..")];
      // eslint-disable-next-line global-require
      return require("../..");
    }

    for (const optionValue of ["outside-only", "dangerously"]) {
      it(`should not include globals not available in fresh global when set to "${optionValue}"`, () => {
        global.globalThis = global;
        const { JSDOM: FreshJSDOM } = getFreshJsdomModule();
        new FreshJSDOM(undefined, { runScripts: optionValue })
        assert.doesNotThrow(() => new FreshJSDOM(undefined, { runScripts: optionValue }));
      });
    }
  });
```

However there are some issues with this approach:
- To be able to get a fresh instance of the module we'd have to remove the top level `require()` and rewrite the other references in the test suite
- This test won't actually fail on Node 10, because the `js-globals.json` that gets generated won't include `globalThis`, so we'd have to potentially also mutate that module?

Which lead me to thinking.. perhaps instead of running `generate-js-globals` as part of `prepare`, could it be done as part of `postinstall`? This would then generate it in the consumer's environment, instead of the environment at the time of publish (which appears to be Node 12, which is why `globalThis` is present). Thoughts?

SimenB:
This `vmGlobal` thing is essentially what the script does at build time

https://github.com/jsdom/jsdom/blob/8921128ee1efe6737e3ea200959eb01fe75ff3aa/scripts/generate-js-globals.js#L14-L16

So this PR partly removes the optimization introduced by having it in a script anyways.

Would it make sense to just do the `generate-js-globals` thing lazily in the code? Or even not lazily, since running the script (without the write to `fs`) takes about 35ms

```
Benchmark #1: node file.js
  Time (mean ± σ):      35.8 ms ±   1.9 ms    [User: 29.9 ms, System: 6.1 ms]
  Range (min … max):    33.3 ms …  42.6 ms    76 runs
```

Doing `runInNewContext` rather than creating a context manually and using that is a tiny bit faster, fwiw

```
Benchmark #1: node file.js
  Time (mean ± σ):      32.7 ms ±   1.6 ms    [User: 27.7 ms, System: 5.2 ms]
  Range (min … max):    30.2 ms …  39.7 ms    90 runs
```

I don't think those ms will make a huge difference compared to just loading `jsdom` itself

```
Benchmark #1: node -e "require(\"jsdom\")"
  Time (mean ± σ):     367.1 ms ±   4.9 ms    [User: 397.9 ms, System: 53.5 ms]
  Range (min … max):   358.6 ms … 375.6 ms    10 runs
```

This is with inlining the script into the runtime

```
Benchmark #1: node -e "require(\"jsdom\")"
  Time (mean ± σ):     372.8 ms ±   3.2 ms    [User: 398.7 ms, System: 56.2 ms]
  Range (min … max):   368.1 ms … 378.3 ms    10 runs
```

So about 5ms slower, but it fixes this bug where global env when `js-globals.json` is generated differs from the runtime globals

ianschmitz:
@domenic do you have any thoughts on which approach you prefer? I like @SimenB's suggestion but wanted to get your input before changing this PR.

ianschmitz:
I've updated the PR to generate the globals at runtime.

SimenB:
@domenic ping 🙂

domenic:
This currently fails the build and does not have a test.

ianschmitz:
@domenic i'm waiting for feedback on the approach.

domenic:
It's hard to evaluate the approach since I don't know if it works, or if it can be made to pass jsdom's tests. Sorry.

SimenB:
I'm not sure if it's possible to write a proper test for this since it'll involve switching node versions. It will only fail when running an old node against the json file generated on a newer version of node.

The bug is best boiled down to this, I think

```js
console.log(process.version);

global.globalThis = global;

new JSDOM('', {runScripts: 'outside-only'});

console.log('Great success!');
```

Running with node 10 prints

```sh-session
v10.20.1
evalmachine.<anonymous>:1
globalThis
^

ReferenceError: globalThis is not defined
    at evalmachine.<anonymous>:1:1
    at Script.runInContext (vm.js:133:20)
    at Object.runInContext (vm.js:311:6)
    at setupWindow (/Users/simen/repos/hhhhhu/node_modules/jsdom/lib/jsdom/browser/Window.js:51:55)
    at new Window (/Users/simen/repos/hhhhhu/node_modules/jsdom/lib/jsdom/browser/Window.js:107:3)
    at exports.createWindow (/Users/simen/repos/hhhhhu/node_modules/jsdom/lib/jsdom/browser/Window.js:38:10)
    at new JSDOM (/Users/simen/repos/hhhhhu/node_modules/jsdom/lib/api.js:36:20)
    at Object.<anonymous> (/Users/simen/repos/hhhhhu/file.js:7:1)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
```

Running with node 12 prints

```sh-session
v12.19.0
Great success!
```

domenic:
> The bug is best boiled down to this, I think

This seems like a great test to add.

SimenB:
I don't think it'd be a valid test since the issue is that the list of globals is generated at build time and might lead to a mismatch between what `jsdom` _thinks_ should exist and what _does_ exist, which explodes if some global it thinks exist is polyfilled. If the list is generated at runtime (like this PR implements) that's not an issue.

I guess the test might guard against a future optimization of again going with a static list? Not sure how valuable that's in practice?

---

CI is currently red due to attempting to run `vm.runInNewContext` in a browser. That raises the question - do you need the list of globals for some browser compatibility thing? If not just skipping the added test in the browser seems fine?

domenic:
> I don't think it'd be a valid test 

I'm pretty confused. You said that you have a test which fails on Node v10 on jsdom master, and would pass on Node v10 with this patch applied. If that's the case, it's a valid test. Or am I missing something?

> CI is currently red due to attempting to run vm.runInNewContext in a browser. That raises the question - do you need the list of globals for some browser compatibility thing? If not just skipping the added test in the browser seems fine?

Well, we need to make sure JSDOM continues to work in the browser, including all our existing tests. We shouldn't skip tests that pass today.

For the new test, i.e. "The bug is best boiled down to this, I think", that test should also be able to pass in the browser... I don't see a reason to skip it.

SimenB:
> > I don't think it'd be a valid test
> 
> I'm pretty confused. You said that you have a test which fails on Node v10 on jsdom master, and would pass on Node v10 with this patch applied. If that's the case, it's a valid test. Or am I missing something?

I have a script that illustrates how the current approach fails (which I pasted in the comment above), but I don't think it can be converted to a proper test. It would not fail on jsdom master unless the `prepare` script is first executed using node 12 followed by the script being ran with node 10.

>Well, we need to make sure JSDOM continues to work in the browser, including all our existing tests. We shouldn't skip tests that pass today.

Good point - it seems `vm` is shimmed today, I'm not sure why it fails now. Error is `TypeError: vm.runInNewContext is not a function`, but it's shimmed here: https://github.com/dfkaye/vm-shim/blob/b65eb2af267e944d908e7757daf139cbc559f919/vm-shim.js#L19

PR #2976, opened 2020-05-17
Prevent infinite loop when mutation observer calls observe within the callback

calebmer:
> Disclaimer: I haven’t carefully thought this change through. Probably consider this more like an issue with the change that fixed the issue for me.

While using jsdom I ran into a test that wouldn’t terminate. After debugging I discovered some code that fired in a mutation observer which was effectively:

```js
const observer = new MutationObserver(mutations => {
  // Many layers of abstraction...

  // In a totally different file:
  if (someCondition) {
    observer.disconnect();
    // Do some stuff...
    observer.observe();
  }

  // Many more layers of abstraction...
});
```

Since this happens during `notifyMutationObservers()` the new `observer.observe()` call added to the `_nodeList` which caused the for-loop to then execute on the new node, however that called the callback which added to `_nodeList` and so on forever.

### DIFF
--- a/lib/jsdom/living/helpers/mutation-observers.js
+++ b/lib/jsdom/living/helpers/mutation-observers.js
@@ -147,7 +147,7 @@ function notifyMutationObservers() {
   for (const mo of notifyList) {
     const records = [...mo._recordQueue];
     mo._recordQueue = [];
-    
+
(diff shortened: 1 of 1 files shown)


domenic:
Can you add tests for this?

crypt096 (approved):

Issue #2979, opened 2020-05-22
Form elements defined outside hierarchy with form='form-id' are not validated correctly


timrobinson33:
### Basic info:
reportValidity() and checkValidity() called on the form element return true (i.e. valid) if there is an invalid control that is outside the form element but linked to it using the form attribute. they correctly return false if the invalid control is nested inside the form

This might be caused by the same underlying problem as  https://github.com/jsdom/jsdom/issues/2628

- **Node.js version: v12.16.3
- **jsdom version:jsdom@16.2.1

### Minimal reproduction case
Sorry I'm not too familiar with directly using jsdom - I've only used it with React/Jest/react-testing-library, but I'm 95% sure the bug here is within jsdom and I'm sure it would be trivial for someone who knows jsdom to reproduce it directly. here is my React component (using TypeScript)
```js
export default function Foo() {
    const [foo, setFoo] = useState(".")
    function submitForm(event: FormEvent) {
        event.preventDefault()
        const target = event.target as HTMLFormElement
        console.log(target.reportValidity() + " " + target.checkValidity())
    }
    return (
        <div>
            <input type="text" form="theForm" value={foo} required={true} placeholder="foo" onChange={(e) => setFoo(e.target.value)} />
            <form id="theForm" onSubmit={submitForm}>
                <input type="submit" value="Submit"></input>
            </form>
        </div>
    )
}
```
and here is my test 
```
test("foo test", () => {
    const {getByPlaceholderText, getByText} = render(<Foo />)
    const textField = getByPlaceholderText("foo")
    fireEvent.change(textField, { target: { value : "" } })
    const submitButton = getByText("Submit")
    fireEvent.click(submitButton)
})
```
What's happening is that I'm setting the text field value to blank then submitting the form. The onSubmit handler reports true for reportValidity and checkValidity (i.e. the console log shows "true true" when running the test.

if you move the text input inside the form, the console log correctly shows "false false"

### How does similar code behave in browsers?
in Chrome and Firefox, the onSubmit handler is not called at all if the form is invalid. I'm not sure if you would consider that a separate bug

Issue #2628, opened 2019-07-05
Form elements defined outside hierarchy with form='form-id' are not captured by form.elements


FThompson:
Previous issues (#765 #2368) have addressed this issue resulting in PR #2369 but this fix does not add form elements defined outside form hierarchy with `form='form-id'` in the [`form.elements`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/elements) member.

### Basic info:

- **Node.js version:** 10.15.3
- **jsdom version:** 15.1.1

### Minimal reproduction case

```js
const { JSDOM } = require('jsdom');

let form = `
    <form id='test-form'></form>
    <input type='text' name='test' form='test-form' />
`;

let dom = new JSDOM(form);
console.log('expecting one form element');
console.log('actual: ' + dom.window.document.forms[0].elements.length); // 0
```

Repro: https://github.com/FThompson/jsdom-form-attribute-bug

### How does similar code behave in browsers?

`form.elements` detects elements defined outside form hierarchy with `form` attribute:
https://jsfiddle.net/fthompson/2njuLq3r/2/

chrisnicola:
Noticed that this also effects the submit button behaviour. A `type="submit"` button with `form=` set that is outside of the form element will not trigger submit on the form when clicked.

ccwebdesign:
This seems to fix the issue, if you (JSDOM contributors) want to add it. It needs proper testing (sorry -- low on time at the moment).

In master [lib/jsdom/living/nodes/HTMLFormElement-impl.js](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/nodes/HTMLFormElement-impl.js):50
```
get elements() {
    // TODO: Return a HTMLFormControlsCollection
    return HTMLCollection.createImpl([], {
      element: this,
      query: () => domSymbolTree.treeToArray(this.ownerDocument, {
        filter: node => isListed(node) && (node._localName !== "input" || node.type !== "image") && this == node.form
      })
    });
  }
```

FThompson:
@ccwebdesign Your solution looks like it would work, but one concern I have about that approach would be inefficient runtime when dealing with large web pages. I'd need to test it though to be sure. Otherwise, I think the best solution would be to live track which elements in the page correspond to each form in the page, but that would be far more involved to implement.

ccwebdesign:
@FThompson From what I can tell, the change (above) doesn't track mutations (or doesn't track mutations quickly enough -- re:performance). Below works.

[...]

PR #2980, opened 2020-05-22
#2979 : Form controls outside form hierarchy not validated

timrobinson33:
**WARNING:** This fix is dependent on the fix for https://github.com/jsdom/jsdom/issues/2628 which is at
https://github.com/ccwebdesign/jsdom but does not currently have a PR.

Sorry - I'm not sure what the protocol is for making a fix to an unreleased fix

As of the current released version, the elements collection does not include elements outside the form hierarchy. Once the fix for #2628 corrects it, this fix then allows the validation to reference the elements collection which will result in the correct validation being performed

### DIFF
--- a/lib/jsdom/living/nodes/HTMLFormElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLFormElement-impl.js
@@ -191,7 +191,7 @@ class HTMLFormElementImpl extends HTMLElementImpl {
   // https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#statically-validate-the-constraints
   _staticallyValidateConstraints() {
     const controls = [];
-    for (const el of domSymbolTree.treeIterator(this)) {
+    for (const el of this.elements) {
       if (el.form === this && isSubmittable(el)) {
         controls.push(el);
       }


domenic:
Can you add new tests, or find tests to enable, for this?

timrobinson33:
@domenic I can't write tests at the moment because my fix won't do anything until #2628 is merged. When there are tests for that, it will be easy to add tests for this. In fact since my fix is a single-line and this is is arguably the same problem as #2628, it might be easier to bundle this into it.

domenic:
Note that #2628 is not a pull request, so it's not something I can "merge".

timrobinson33:
@domenic Sorry I'm not that familiar with contributing to open source projects and I wasn't quite sure what to do so I apologise if I've broken the protocol. I just wanted to make it clear that there is a simple fix for this issue, even though I understand it can't be merged at the moment because it depends on #2628 which seems to be stalled. I'm happy for you to just delete the pull request if you think it's not appropriate to have it.

ccwebdesign:
@timrobinson33 Sorry -- didn't realize you were waiting on me. Created #3003

domenic:
#3003 is now merged!

crypt096 (approved):

PR #2981, opened 2020-05-22
Use `InnerHTML` mixin for `innerHTML` definition

ExE-Boss:
<https://github.com/w3c/DOM-Parsing/issues/21> has been resolved, so this can be updated.

I’ve kept the implementations in `Element‑impl.js` and `ShadowRoot‑impl.js`, as they have little actual overlap.

### DIFF
--- a/lib/jsdom/living/domparsing/InnerHTML-impl.js
+++ b/lib/jsdom/living/domparsing/InnerHTML-impl.js
@@ -1,0 +1,29 @@
+"use strict";
+
+const { parseFragment } = require("../../browser/parser");
+const { HTML_NS } = require("../helpers/namespaces.js");
+const { isShadowRoot } = require("../helpers/shadow-dom.js");
+const NODE_TYPE = require("../node-type.js");
+const { fragmentSerialization } = require("./serialization.js");
+
+// https://w3c.github.io/DOM-Parsing/#the-innerhtml-mixin
+exports.implementation = class InnerHTMLImpl {
+  // https://w3c.github.io/DOM-Parsing/#dom-innerhtml-innerhtml
+  get innerHTML() {
+    return fragmentSerialization(this, {
+      requireWellFormed: true,
+      globalObject: this._globalObject
+    });
+  }
+  set innerHTML(markup) {
+    const contextElement = isShadowRoot(this) ? this.host : this;
+    const fragment = parseFragment(markup, contextElement);
+
+    let contextObject = this;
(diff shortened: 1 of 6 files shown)


TimothyGu:
Hmm, the implementations are merged in the spec though. I think it would be a good idea to merge them.

lib/jsdom/living/domparsing/InnerHTML-impl.js
```diff
@@ -0,0 +1,6 @@
+"use strict";
+
+// https://w3c.github.io/DOM-Parsing/#the-innerhtml-mixin
+exports.implementation = class InnerHTMLImpl {
+  // innerHTML is implemented separately inside Element and ShadowRoot.
+};
```
ExE-Boss, line 6: This matches what’s done for `NonElementParentNodeImpl`: <https://github.com/jsdom/jsdom/blob/69b0205ed4cdafe94f9ffea519e00dcc67e17f1f/lib/jsdom/living/nodes/NonElementParentNode-impl.js#L3-L7>

TimothyGu (approved):

TimothyGu (changes requested):
Actually sorry, could you merge the implementations of innerHTML from Element and ShadowRoot in a manner similar to the [spec](https://w3c.github.io/DOM-Parsing/#dom-innerhtml-innerhtml)? I know the spec is not the cleanest but I'd rather us stick to it.

lib/jsdom/living/domparsing/InnerHTML-impl.js
```diff
@@ -0,0 +1,28 @@
...
+      globalObject: this._globalObject
+    });
+  }
+  set innerHTML(markup) {
+    let contextObject = isShadowRoot(this) ? this.host : this;
+    const fragment = parseFragment(markup, contextObject);
+
+    if (this.nodeType === NODE_TYPE.ELEMENT_NODE && this.localName === "template" && this.namespaceURI === HTML_NS) {
+      contextObject = contextObject._templateContents;
+    }
```
TimothyGu, line 24: This isn't quite right. The spec makes a distinction between "context element" and "context object".

```suggestion
    const contextElement = isShadowRoot(this) ? this.host : this;
    const fragment = parseFragment(markup, contextObject);

    let contextObject = this;
    if (this.nodeType === NODE_TYPE.ELEMENT_NODE && this.localName === "template" && this.namespaceURI === HTML_NS) {
      contextObject = contextObject._templateContents;
    }
```

TimothyGu (approved):

PR #2982, opened 2020-05-26
Implement `BeforeUnloadEvent`

ExE-Boss:
This implements the `BeforeUnloadEvent` interface so that we can perform brand checking in: <https://github.com/jsdom/jsdom/blob/bf5c6ce2e3a3ab6d9d79c923ee44330855f7a50d/lib/jsdom/living/helpers/create-event-accessor.js#L37>


domenic:
I don't think we want to implement this until we have something that actually fires it.

PR #2996, opened 2020-06-24
Ensure focusin/focusout fires with .focus()/.blur()

trueadm:
This PR alters how `HTMLOrSVGElement-impl` so that the `focus` and `blur` methods also dispatch the relevant `focusin` and `focusout` events in the bubble phase, as per how they occur on modern browsers. I wasn't sure where to add any tests for this change specifically, if anyone has any pointers, that would be great.

### DIFF
--- a/lib/jsdom/living/helpers/focusing.js
+++ b/lib/jsdom/living/helpers/focusing.js
@@ -65,14 +65,16 @@ exports.isFocusableAreaElement = elImpl => {
 
 // https://html.spec.whatwg.org/multipage/interaction.html#fire-a-focus-event plus the steps of
 // https://html.spec.whatwg.org/multipage/interaction.html#focus-update-steps that adjust Documents to Windows
-exports.fireFocusEventWithTargetAdjustment = (name, target, relatedTarget, bubbles) => {
+// It's extended with the bubbles option to also handle focusin/focusout, which are "defined" in
+// https://w3c.github.io/uievents/#event-type-focusin. See https://github.com/whatwg/html/issues/3514.
+exports.fireFocusEventWithTargetAdjustment = (name, target, relatedTarget, { bubbles = false } = {}) => {
   if (target === null) {
     // E.g. firing blur with nothing previously focused.
     return;
   }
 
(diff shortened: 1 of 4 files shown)


trueadm:
Thanks for the help @eps1lon. I found `focusing-focused-noop.html` and updated that to include the logic from this test as it seemed the most reasonable approach.

domenic:
Can we make sure that this doesn't cause infinite recursion, by enabling the test as @eps1lon suggests and then making it pass, instead of making it fail with `Maximum callstack size exceeded`?

trueadm:
@domenic Yep, I plan on looking into that later once I get some time to. :)

trueadm:
I've fixed the failing test. I ended up spending quite a bit longer on it, but using a build of Chrome and the test to cover the same sequence of how `activeElement` and `relatedTarget` get set, especially around iframes.

trueadm:
@domenic Are you happy with this PR or is there something you'd like me to look into still?

domenic:
Can you enable the tests @eps1lon suggests?

trueadm:
@domenic Ah sorry, I thought that mention @eps1lon was in reference to enabling the test for debugging purposes. I've commited that change to this PR now too.

trueadm:
The latest failures seem to be due to https://github.com/jsdom/jsdom/pull/2951 being merged on master.

Looking into why `focus-management-expectations.html` is failing, it seems to be with some other focus related logic, rather than something introduced in this PR.

trueadm:
@domenic Would it be possible to get a new release of JSDOM? We want to incorporate this change into the next version of React and thus ensure our Jest test suite is able to pass with the changes to React's event system. We can upgrade Jest once JSDOM releases a new version. :)

domenic:
For sure; I'll definitely do a release by next weekend, but hopefully can get it out sooner.

trueadm:
Thank you so much! I really appreciate it :)

domenic:
FYI I published v16.3.0 today.

rickhanlonii:
Thanks @domenic!

PR #3001, opened 2020-07-06
Including the need to set url for external scripts

evolverine:
I spent quite a bit of time trying to figure out why in my external scripts weren't loading, and this tidbit of information here could have saved me those hours. Maybe it can help others too.


domenic:
Thanks! I'll push a slightly-different commit that closes this and #2936.

Note also we have a bug where `error` events are not properly being fired, per https://github.com/jsdom/jsdom/issues/2932#issuecomment-619437074.

PR #3003, opened 2020-07-08
Form elements defined outside hierarchy with form='form-id' are not captured by form.elements

ccwebdesign:
Addresses #2628

Issues #2979 and related PR #2980 are based on this fix.

This fix hasn't been evaluated for performance.


domenic:
Can you add tests per https://github.com/jsdom/jsdom/blob/master/Contributing.md#web-platform-feature-tests ?

ccwebdesign:
@domenic Should the tests go in `test/web-platform-tests/to-upstream/html/semantics/forms/the-form-element/` or somewhere else?

domenic:
That seems reasonable if you indeed have to write them yourself, but I suspect tests for this already exist, and it's just a matter of enabling them in to-run.yaml...

ccwebdesign:
Looks like I can't edit the tests in the web-platform-tests, so here's a small modification in:
`/test/web-platform-tests/tests/html/semantics/forms/the-form-element/form-elements-filter.html`

```diff
test(() => {
  const elements = document.querySelector("#form").elements;
  const ids = Array.from(elements).map(el => el.id);

  const allCorrectIDs = Array.from(document.querySelectorAll("[data-in]")).map(el => el.id);

+  const notFound = ids.filter(id => !allCorrectIDs.includes(id));
-  assert_array_equals(ids, allCorrectIDs);
+  assert_array_equals(notFound, []);
});
```

Due to the way the HTMLFormElement-impl.js `get elements()` works, it doesn't return the nodes in document-order. It could be altered to, but could impact performance even more.

With the above alteration, 'form-elements-filter.html' can be re-enabled in `to-run.yaml` and should pass.

domenic:
I believe the spec requires document order, so we need to implement that.

ccwebdesign:
Yep -- you're right, made some modifications. Returning valid nodes early if there's no form.id. And, while it's a lot longer, filtering the document nodes should be more clear at first glance since the conditional was getting a little unwieldy.

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -47,14 +47,53 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
+          return true;
+        }
+
+        return false;
+      }
+    });
+  }
+
+  _filterElement(node) {
+    if (
```
domenic, line 81: This function can be reduced to `return isListed(node) && (node._localName !== "input" || node.type !== "image");`

At that point it might be simpler to inline into into the above.

Note also that line lengths can be up to 120 characters so there's no need for this kind of breaking.

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -47,14 +47,53 @@ class HTMLFormElementImpl extends HTMLElementImpl {
     super._descendantRemoved.apply(this, arguments);
   }
 
+  _getElementNodes() {
+    if (!this.hasAttributeNS(null, "id")) {
```
domenic, line 51: It looks like this optimization might be invalid. The spec says that the root of the collection is always the root element of the document: https://html.spec.whatwg.org/#dom-form-elements .

In particular, I think you want to consider what happens if you do something like

```js
const elements = form.elements;
form.id = "foo";
// elements should now contain things associated via form=""
```

Since it sounds like that case is not covered in the existing tests, would you mind adding a to-upstream test that fails before fixing that, and succeeds after?

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -47,14 +47,53 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
+      return domSymbolTree.treeToArray(this, {
+        filter: node => this._filterElement(node)
+      });
+    }
+
+    const id = this.getAttributeNS(null, "id");
```
domenic, line 57: Similarly to the above, caching this will not be valid, generally.

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -47,14 +47,53 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
+    return domSymbolTree.treeToArray(this.ownerDocument, {
+      filter: node => {
+        if (!this._filterElement(node)) {
+          return false;
+        }
+
+        if (this.contains(node)) {
```
domenic, line 64: I don't believe this works. In particular `form=""` should take precedent over `contains()` relationships. Another candidate for a test to write, I guess?

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -47,14 +47,53 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
 
+  _getElementNodes() {
+    if (!this.hasAttributeNS(null, "id")) {
+      return domSymbolTree.treeToArray(this, {
+        filter: node => this._filterElement(node)
+      });
+    }
+
+    const id = this.getAttributeNS(null, "id");
+    return domSymbolTree.treeToArray(this.ownerDocument, {
```
domenic, line 58: Since we want the _root_, which is not always the owner document, we should use `this.getRootNode({})`.

If you want to write a test for this, you can tell the difference using shadow DOM. However, that's pretty complicated, so I'm OK skipping that.

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -47,14 +47,53 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
+      filter: node => {
+        if (!this._filterElement(node)) {
+          return false;
+        }
+
+        if (this.contains(node)) {
+          return true;
+        }
+
+        if (
```
domenic, line 68: So, the trick here is that you can replace all of this code (after the initial filter) with a call to `formOwner` from `helpers/form-controls.js`, which does the full computation. It's probably slower, but we should first go for correctness, then optimize with caches or something later if necessary.

lib/jsdom/living/nodes/HTMLFormElement-impl.js
ccwebdesign: Just to double-check here, you want me to:
- test that the HTMLCollection root element is the root of the document
- test that `form.elements` fails to include new elements after changing `form.id`

From what I can find, the HTMLCollection root is never exposed. Not sure how to go about testing that,

Also from what I've found so far in tests, even after updating the HTMLCollection element to the root of the document, the query automatically includes / removes elements on changing `form.id`, as long as `this` in the query is scoped to the HTMLFormElement.

lib/jsdom/living/nodes/HTMLFormElement-impl.js
ccwebdesign: For this, added:  `test/web-platform-tests/to-upstream/html/semantics/forms/the-form-element/form-elements-form-owner.html`
test: "`form.elements` should not contain elements inside form with different form owner"

lib/jsdom/living/nodes/HTMLFormElement-impl.js
ccwebdesign: Thanks, that's much simpler. XD

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -47,14 +47,35 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
+        if (!this._filterElement(node)) {
+          return false;
+        }
+
+        const owner = formOwner(node);
+        if (owner && this === owner) {
+          return true;
+        }
+
+        return false;
```
domenic, line 62:
```suggestion
        return formOwner(node) === this;
```

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -47,14 +47,35 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
+      }
+    });
+  }
+
+  _filterElement(node) {
```
domenic, line 67: Since this function is used in only one place now, it would be simpler to inline it into the call site.

lib/jsdom/living/nodes/HTMLFormElement-impl.js
```diff
@@ -47,14 +47,35 @@ class HTMLFormElementImpl extends HTMLElementImpl {
...
     // TODO: Return a HTMLFormControlsCollection
     return HTMLCollection.createImpl(this._globalObject, [], {
-      element: this,
-      query: () => domSymbolTree.treeToArray(this, {
-        filter: node => isListed(node) && (node._localName !== "input" || node.type !== "image")
-      })
+      element: this.getRootNode({}),
+      query: function () {
+        return this._getElementNodes();
+      }.bind(this)
```
domenic, line 78:
```suggestion
      query: () => this._getElementNodes()
```

ccwebdesign:
@domenic That should address all of the issues so far. Didn't add the shadow DOM test you mentioned, since it seems like there are already tests for this within `form-elements-filter.html` ("form.elements only includes elements from the same shadow tree").

ccwebdesign:
@domenic Thanks for all of the suggestions and help on this. XD

PR #3008, opened 2020-07-16
feat: Cache Element.prototype.tagName

eps1lon:
`ByRole` from `@testing-library/dom` makes heavy use of window.getComputedStyle. This makes Element.prototype.tagName a hotpath in certain DOM trees. Where `asciiUppercase` can take up to 40% of the total time. Since all the conditions checked in `get tagName()` seem invariant we can cache the result and return that instead after the first computation.

For our specific benchmark we saw a drop of the median runtime down to 261.3774ms from 538.9752ms (which corresponds to the 40% of total time of `asciiUppercase`.

Benchmark repo: https://github.com/eps1lon/jsdom-tagName-cache

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -78,6 +78,8 @@ class ElementImpl extends NodeImpl {
     this._attributes = NamedNodeMap.createImpl(this._globalObject, [], {
       element: this
     });
+
+    this._cachedTagName = null;
   }
 
   _attach() {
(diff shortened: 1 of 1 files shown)


domenic:
Hmm. Can we eliminate the use of tagName in getComputedStyle instead? It's largely legacy and things shouldn't be using it...

eps1lon:
> Hmm. Can we eliminate the use of tagName in getComputedStyle instead? It's largely legacy and things shouldn't be using it...

I can take a look at it. Though it seems to me I'd have to touch `nwsapi`. That package might have to use `tagName`.

eps1lon:
> Hmm. Can we eliminate the use of tagName in getComputedStyle instead? It's largely legacy and things shouldn't be using it...

These are called from `nwsapi#matches`
Screenshot from 2020-07-16 02-53-07

I don't think we can get rid of them.

domenic:
I still think it would be best if this were fixed in https://github.com/dperini/nwsapi, and I encourage you to file an issue there. However, I'm happy to accept this for now; the extra memory cost isn't that great, and it's possible author code is also using the bad `tagName` property.

eps1lon:
I don't understand what the issue is with nwsapi. I think they're implementing the selector spec correctly.

domenic:
They're using `tagName`, which is not the fundamental primitive; an element's `localName` is the fundamental primitive.

eps1lon:
>  the extra memory cost isn't that great, and it's possible author code is also using the bad `tagName` property.

Note that caching tagName seems to be encouraged by the living standard:

> User agents could optimize qualified name and HTML-uppercased qualified name by storing them in internal slots.

-- https://dom.spec.whatwg.org/#concept-element-qualified-name

Depending on the targetted browser support of `nwsapi` https://github.com/dperini/nwsapi/pull/37 might not be acccepted. Regardless, thanks for the hint about `localName` and didn't know this existed an it seems like a better alternative in almost all cases (granted it's available).

Issue #3016, opened 2020-07-25
Bug: ResourceLoader needs decodeURI(file)


cata-code:
I have an html file that links to this css file `<link rel="stylesheet" href="css/page/hey 진 シーン/i 🥰 you 😘😘😘😘/yolo.css">` and when jsdom needs to load the css file with the `resources: 'usable'` option, it fails with this error:

```
Error: Could not load link: "file:///home/user/dev/test-folder/css/page/hey%20%EC%A7%84%20%E3%82%B7%E3%83%BC%E3%83%B3/i%20%F0%9F%A5%B0%20you%20%F0%9F%98%98%F0%9F%98%98%F0%9F%98%98%F0%9F%98%98/yolo.css"
    at onErrorWrapped (/home/user/dev/code/app/node_modules/jsdom/lib/jsdom/browser/resources/per-document-resource-loader.js:38:19)
    at Object.check (/home/user/dev/code/app/node_modules/jsdom/lib/jsdom/browser/resources/resource-queue.js:72:23)
    at /home/user/dev/code/app/node_modules/jsdom/lib/jsdom/browser/resources/resource-queue.js:83:27
    at processTicksAndRejections (internal/process/task_queues.js:94:5) [Error: ENOENT: no such file or directory, open '/home/user/dev/test-folder/css/page/hey %EC%A7%84 %E3%82%B7%E3%83%BC%E3%83%B3/i %F0%9F%A5%B0 you %F0%9F%98%98%F0%9F%98%98%F0%9F%98%98%F0%9F%98%98/yolo.css'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/home/user/dev/test-folder/css/page/hey %EC%A7%84 %E3%82%B7%E3%83%BC%E3%83%B3/i %F0%9F%A5%B0 you %F0%9F%98%98%F0%9F%98%98%F0%9F%98%98%F0%9F%98%98/yolo.css'
```
I think we need to add an `decodeURI(file)` when fetching the resource file

### Basic info:

- **Node.js version:** 14.4.0
- **jsdom version:** 16.3.0

domenic:
What URL does your browser try to load when you load that file into it? Can you provide the full example so that we can run it, per the issue template?

cata-code:
Probably here we need to add `encodeUri()`
https://github.com/jsdom/jsdom/blob/master/lib/jsdom/browser/resources/resource-loader.js#L46

[...]

PR #3027, opened 2020-08-16
Bug: ResourceLoader needs decodeURI(file) #3016

cata-code:
Fix for https://github.com/jsdom/jsdom/issues/3016

### DIFF
--- a/lib/jsdom/browser/resources/resource-loader.js
+++ b/lib/jsdom/browser/resources/resource-loader.js
@@ -43,7 +43,7 @@ module.exports = class ResourceLoader {
      *   2. We need to add an abort handler.
     */
     const promise = new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


domenic:
Can you add tests?

cata-code:
I tried to make a test but my system uses python 3 not 2 and I didn't want to mess that up again just for a small test

https://web-platform-tests.org/running-tests/from-local-system.html#system-setup
At least that's what this document says.

If it's a requirement for the pr to be accepted, then I will see what I can do this weekend, if I'm able to fix the python switching.

domenic:
Ah, right, sorry for missing your issue comment. Yeah, we require tests before accepting PRs. However, since we can't test file: URLs anyway in web platform tests, you probably can add these to the api/ tests folder, which doesn't have a python dependency. You can run that subset of tests using `yarn test-api`.

PR #3032, opened 2020-08-27
Fix for wrapper undefined

builtwith:
Needs to be checked before calling wrapper._ownerDocument.


domenic:
This needs tests before we could consider merging.

crypt096 (approved):

PR #3034, opened 2020-09-02
Call `validateTokens` just once with all tokens to validate

ankon:
This should be ever so slightly faster, but I doubt that that would be measurable. It's however more readable IMHO :)


domenic:
This is not correct, as validateTokens does not take an array; it takes rest arguments.

ankon:
Ugh, you're completely right -- sorry!

PR #3038, opened 2020-09-05
Support inputType on InputEvent

### LINKED ISSUE #3037: Can't create InputEvent with inputType property
### Basic info:

- **Node.js version:** 12.18.3
- **jsdom version:** 16.4.0

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom")
const { window } = new JSDOM("");
const event = new window.InputEvent("input", { inputType: "insertText" });
event.inputType;
```

It outputs `undefined`.

https://runkit.com/embed/rqw8sdcg6z3d

Copy and paste the code below in the browser console. It should output `"insertText"`
```js
const event = new window.InputEvent("input", { inputType: "insertText" });
event.inputType;
```

diegohaz:
Fixes #3037

### DIFF
--- a/lib/jsdom/living/events/InputEvent-impl.js
+++ b/lib/jsdom/living/events/InputEvent-impl.js
@@ -1,22 +1,9 @@
 "use strict";
-
 const UIEventImpl = require("./UIEvent-impl").implementation;
-
 const InputEventInit = require("../generated/InputEventInit");
 
 // https://w3c.github.io/uievents/#interface-inputevent
-class InputEventImpl extends UIEventImpl {
-  initInputEvent(type, bubbles, cancelable, data, isComposing, inputType) {
-    if (this._dispatchFlag) {
-      return;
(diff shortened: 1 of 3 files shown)


crypt096 (approved):

lib/jsdom/living/events/InputEvent-impl.js
```diff
@@ -6,14 +6,15 @@ const InputEventInit = require("../generated/InputEventInit");
 
 // https://w3c.github.io/uievents/#interface-inputevent
 class InputEventImpl extends UIEventImpl {
-  initInputEvent(type, bubbles, cancelable, data, isComposing) {
+  initInputEvent(type, bubbles, cancelable, data, isComposing, inputType) {
```
domenic, line 9: It looks like this function is never used or exposed (since it's not in the IDL), so we can delete it. I'll push a commit to do that.

Issue #2986, opened 2020-06-03
`getComputedStyle` doesn't returns correct style after I dynamically add `<style>` tag to the document whose content is set by `.innerHTML`.


StephenChips:
### Basic info:

- **Node.js version:** v12.13.0
- **jsdom version:** v16.2.2

I did some tests with Jest today, which use JSDOM to simulate the browser environment. I was going to dynamically inject some CSS codes after my component is loaded, so I created a `<style>` tag and set the style string to its `innerText` property. After the tag was inserted, surprisingly the test failed, and said that the style didn't get updated properly (I used the `window.getComputedStyle` function to get the current style). After some troubleshooting I didn't found any bugs, then I changed assignning style from `.innerText' to `.textContent`. The test passed  after such modification. So I though maybe it's the JSDOM's problem. I created following snippet and it proved that it was.

### Minimal reproduction case

```js
var jsdom = require('jsdom');
var jsdom = new jsdom.JSDOM('<html><head></head><body><p class="hello">hello, world.</p></body></html>');
var document = jsdom.window.document;
var window = jsdom.window;
var display = window.getComputedStyle(document.querySelector('.hello')).display;
console.log(display); // block

var style = document.createElement('style');
style.innerText = '.hello { display: flex; }';
console.log(style.innerText); // `.hello { display: flex; }
undefined
document.head.appendChild(style);

console.log(document.documentElement.outerHTML);
// <html><head><style></style></head><body><p class="hello">hello, world.</p></body></html>

display = **window.getComputedStyle(document.querySelector('.hello')).display;**
console.log(display);
// block (which should be 'flex')
```

### How does similar code behave in browsers?
The browser shows the value 'flex' as it should be.

milahu:
yes, <tt>[window.getComputedStyle](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle)(e [, p])</tt> should return a 'live' object

> The returned style is a **live** CSSStyleDeclaration object, which updates automatically when the element's styles are changed.

in
https://github.com/jsdom/jsdom/blob/5279cfda5fe4d52f04b2eb6a801c98d81f9b55da/lib/jsdom/browser/Window.js#L643
doing `return elt.style` is too simple
cos `elt.style` has less data than `getComputedStyle`

`elt.style` only contains data that was explicitly set
where GCS contains all data needed for layout rendering

jsdom already adds default values like `div{display: block}`
just the result is not live ....

milahu:
[...]

PR #3047, opened 2020-09-14
fix window.getComputedStyle(e) with proxy object

milahu:
fix issue #2986 

- [x] use proxy and custom class for computedStyle, to make it look like a style object
- [x] react to change of inline style (inlineStyle)
- [x] react to change of style sheets (sheetStyle)
- [x] after `element.remove()`, reset computedStyle to defaultStyle,  
remove cached inlineStyle and sheetStyle
- [x] make computedStyle read-only, for internal writes use secret write interface
- [x] add some tests
- [x] react to change of element.id - `web-platform-tests/tests/html/dom/elements/global-attributes/id-attribute.html`
  - [ ] id-style should take precedence over class-style, currently the 'last rule' wins  
`#a{color: a} .b{color: b}` and `<div id="a" class="b">` should eval to `color: a`
- [x] react to change of element.className and element.classList
- [ ] inhert style of parent elements
  - [ ] inlineStyle `web-platform-tests/to-upstream/cssom/getComputedStyle-visibility-inheritance-inline.html`
  - [ ] sheetStyle `web-platform-tests/to-upstream/cssom/getComputedStyle-visibility-inheritance.html`
- [ ] prettify interface of [DOMException](https://developer.mozilla.org/en-US/docs/Web/API/DOMException). props should be visible in `Object.keys(err)`: code, name, message - `console.dir(err)` should print these props
- [ ] move tests to mocha

13189df and 1d0e8dd should be ignored in merge


domenic:
I think this is pretty obsolete. Our getComputedStyle() is pretty spec-compliant these days.

PR #3031, opened 2020-08-25
2792 replace request with node fetch

tobyhinloopen:
Closes #2792

This is a work in progress.

PR #3054, opened 2020-09-23
2792 replace request with node fetch

### LINKED ISSUE #2792: Replace request with something better
We are currently using [both](https://github.com/jsdom/jsdom/blob/d240291edbe7d4180a5152993ced7950c834dc57/package.json#L35-L36) request and request-promise-native. We should replace them both with something better.

- **request** is in [maintenance mode](https://github.com/request/request/issues/3142).
- **request** is already quite [bloated](https://bundlephobia.com/result?p=request@2.88.0).
- **request-promise-native** uses [stealthy-require](https://www.npmjs.com/package/stealthy-require) which essentially results in two copies of request (see https://github.com/jsdom/jsdom/pull/2783#issuecomment-572415665)

Some alternatives are in https://github.com/request/request/issues/3143. Strongest candidate right now is probably **[node-fetch](https://www.npmjs.com/package/node-fetch)**.

andrewaylett:
This is a refinement of #3031, which I hope fixes the linter issues.

Should fix #2792, all being well.


andrewaylett:
I'm afraid I'd missed that @tobyhinloopen's branch wasn't _quite_ finished.  Will see if I can find time before he does :).

PR #3058, opened 2020-09-29
Fix event handler ObjectEnvironment instantiation

ExE-Boss:
This uses a method similar to what’s done by [the **Realms** shim](https://github.com/Agoric/realms-shim/blob/067bdd9404013e95b9603379cb1fb597f0bfdc50/src/evaluators.js#L60-L68).

It also works around the bug that prevents using a `Proxy` for the [named properties object](https://heycam.github.io/webidl/#named-properties-object): <https://github.com/jsdom/webidl2js/pull/167> and prevents a global `element` or `formOwner` variable from breaking the inner `with`s and also ensures that if there isn’t such a variable, it’s not added to the scope due to being passed as a named parameter.

## Depends on:
- [x] <https://github.com/web-platform-tests/wpt/pull/26091> (merged)
- [x] <https://github.com/jsdom/jsdom/pull/3156> (merged)

## See also:
- <https://github.com/whatwg/html/issues/6079>
- <https://github.com/whatwg/html/pull/6086>

### DIFF
--- a/lib/jsdom/living/helpers/create-event-accessor.js
+++ b/lib/jsdom/living/helpers/create-event-accessor.js
@@ -115,45 +115,49 @@ exports.getCurrentEventHandlerValue = (target, event) => {
 
     const createFunction = document.defaultView.Function;
     if (event === "error" && element === null) {
-      const wrapperBody = document ? body + `\n//# sourceURL=${document.URL}` : body;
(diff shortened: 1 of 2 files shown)


domenic:
This seems reasonable. Or, rather, that code is full of such hacks anyway, so it doesn't hurt much. Do you mind rebasing?

domenic (approved):

Issue #2160, opened 2018-02-26
Styles do not Inherit


Impally:
### Basic info:

- **Node.js version: 6.11.1** 
- **jsdom version: 11.6.2** 

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const dom = new JSDOM(`
<style>
  body {
    color: Red;
  }
</style>hello world
<div class="divgroup" id="d1" style="padding-right: 5px;">Content for div 1</div>
`);

 dom.window.getComputedStyle(dom.window.document.querySelector('#d1'))
dom.window.getComputedStyle(dom.window.document.querySelector('#d1')).getPropertyValue('color')
```

### How does similar code behave in browsers?

https://jsbin.com/cujumihabe/edit?html,js,console,output

The computed style for the div does not include the inherited color.  That's the major issue, the other being that there are a large amount of things left out, and the color would not be translated to an rgb value, but I believe both of those are issues with cssom.

domenic:
I think the issue is that we implement cascading, but not inheritance. Inheritance is harder :(.

TidyIQ:
Cascading isn't working properly if it is implemented. 

getComputedStyle(element).getAttribute("color") returns `#004ba0` instead of `#2e7d32` in jsdom for this element.

PR #3060, opened 2020-10-08
Inherit Styles: Color, Fill, Stroke

nickiaconis:
Related to #2160. It's been a frustration on my team that `getComputedStyles` doesn't work in JSDOM like it does in a browser. I'm hoping to move the status of this a little closer to feature parity, at least where it counts for us.

Not sure if this initial attempt is 100% how y'all would like to see inheritance implemented for these properties, but I needed to start the conversation somewhere. Happy to revise any portion, so just let me know.

### Basic info:

- **Node.js version:** 12.13.1 
- **jsdom version:** 16.4.0 

### Minimal reproduction case

```js

  <html><head><style>div { color: green; fill: blue; stroke: red; }</style></head>
  <body><div><span>hello</span><svg><circle cx=10 cy="10" r="10" /></svg></div></body></html>
`);

const spanStyle = window.getComputedStyle(window.document.querySelector("span"));
console.log(`span color is "${spanStyle.color}"`);

const svgStyle = window.getComputedStyle(window.document.querySelector("svg"));
console.log(`svg fill is "${svgStyle.fill}" and stroke is "${svgStyle.stroke}"`);
```

https://jsbin.com/saliyiz/3/edit?html,output

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -661,7 +661,7 @@ function Window(options) {
     const { forEach } = Array.prototype;
     const { style } = elt;
 
-    // Browsing-context connected, per https://github.com/whatwg/html/issues/4547
+    // Browsing-context connected, per https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle step 5
     if (!elt.isConnected || !elt._ownerDocument._defaultView) {
       return declaration;
     }
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -661,6 +661,11 @@ function Window(options) {
     const { forEach } = Array.prototype;
     const { style } = elt;
 
+    // Browsing-context connected, per https://github.com/whatwg/html/issues/4547
+    if (!elt.isConnected || !elt._ownerDocument._defaultView) {
+      return declaration;
+    }
+
     forEachMatchingSheetRuleOfElement(elt, rule => {
       forEach.call(rule.style, property => {
         declaration.setProperty(
--- a/lib/jsdom/living/helpers/colors.js
+++ b/lib/jsdom/living/helpers/colors.js
@@ -1,40 +1,8 @@
 "use strict";
+const notImplemented = require("../../browser/not-implemented");
 
-exports.computeColor = maybeColor => {
-  maybeColor = maybeColor.trim();
-
-  // sRGB?
-  const srgbRegex = /^rgb *\( *(\d{1,3}) *, *(\d{1,3}) *, *(\d{1,3}) *\)$/;
-  if (srgbRegex.test(maybeColor)) {
-    const [, r, g, b] = srgbRegex.exec(maybeColor);
-    return `rgb(${r}, ${g}, ${b})`;
-  }
-
-  // hex?
-  const shorthandHexRegex = /^#([a-f\d])([a-f\d])([a-f\d])$/;
-  maybeColor = maybeColor.replace(shorthandHexRegex, (_, r, g, b) => {
-    return r + r + g + g + b + b;
-  });
-  const hexRegex = /^#([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/;
-  if (hexRegex.test(maybeColor)) {
-    let [, r, g, b] = hexRegex.exec(maybeColor);
-    r = parseInt(r, 16);
-    g = parseInt(g, 16);
-    b = parseInt(b, 16);
-    return `rgb(${r}, ${g}, ${b})`;
-  }
-
-  // definded keyword?
-  if (exports.keywords[maybeColor]) {
-    const [r, g, b] = exports.keywords[maybeColor];
-    return `rgb(${r}, ${g}, ${b})`;
-  }
-
-  return "";
-};
-
-// https://www.w3.org/TR/css-color-3/#svg-color
-exports.keywords = {
+// https://drafts.csswg.org/css-color-3/#svg-color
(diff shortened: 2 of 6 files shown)


lib/jsdom/living/helpers/style-rules.js
```diff
@@ -18,6 +19,34 @@ exports.propertiesWithResolvedValueImplemented = {
...
+  color: {
+    inherited: true,
+    initial: {
+      "a:visited:active": "rgb(255, 0, 0)",
+      "a:active": "rgb(255, 0, 0)",
+      "a:visited": "rgb(85, 26, 139)",
+      "a:link": "rgb(0, 0, 238)",
+      mark: "rgb(0, 0, 0)",
+      body: "rgb(0, 0, 0)"
+    },
```
nickiaconis, line 34: I think this behavior of the color property's observed value might actually be due to the UA stylesheet. Does JSDOM implement a UA stylesheet? If so, I can transfer these over there (if they don't already exist).

The [color prop spec](https://drafts.csswg.org/css2/#propdef-color) states that the initial value "depends on user agent," but I've seen a few folks suggest that the true initial value behind the UA styles is `rgb(0, 0, 0)`. If we move these defaults to a UA stylesheet, then the initial value here probably becomes `rgb(0, 0, 0)`, and we can remove the special handling for object-type initial values from `getSpecifiedValue` too.

crypt096 (approved):

lib/jsdom/living/helpers/colors.js
```diff
@@ -0,0 +1,185 @@
...
+  // definded keyword?
+  if (exports.keywords[maybeColor]) {
+    const [r, g, b] = exports.keywords[maybeColor];
+    return `rgb(${r}, ${g}, ${b})`;
+  }
+
+  return "";
+};
+
+// https://www.w3.org/TR/css-color-3/#svg-color
```
domenic, line 36: Be sure to use Editor's drafts, so in this case https://drafts.csswg.org/css-color-3/. I suspect the list hasn't changed much, but at least the comment should change.

lib/jsdom/living/helpers/colors.js
```diff
@@ -0,0 +1,185 @@
+"use strict";
+
+exports.computeColor = maybeColor => {
```
domenic, line 3: Can you link this to a spec, and make sure it's following the spec algorithm?

lib/jsdom/living/helpers/colors.js
```diff
@@ -0,0 +1,185 @@
...
+  if (exports.keywords[maybeColor]) {
+    const [r, g, b] = exports.keywords[maybeColor];
+    return `rgb(${r}, ${g}, ${b})`;
+  }
+
+  return "";
+};
+
+// https://www.w3.org/TR/css-color-3/#svg-color
+exports.keywords = {
```
domenic, line 37: Why is this exported?

lib/jsdom/browser/Window.js
```diff
@@ -661,6 +661,11 @@ function Window(options) {
     const { forEach } = Array.prototype;
     const { style } = elt;
 
+    // Browsing-context connected, per https://github.com/whatwg/html/issues/4547
+    if (!elt.isConnected || !elt._ownerDocument._defaultView) {
+      return declaration;
+    }
```
domenic, line 667: Can you be sure to test this change? In particular the linked issue isn't about the element being connected; it's about the `<style>` element being connected. Probably the right link is https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle step 5?

domenic (changes requested):
Thanks for working on this @nickiaconis!

The biggest thing to be aware of is that the `to-port-to-wpts` test suite should not be expanded. Instead, follow the instructions in https://github.com/jsdom/jsdom/blob/master/Contributing.md#web-platform-feature-tests for finding an existing web platfrom test to enable, or writing a new one in `to-upstream`.

lib/jsdom/living/helpers/style-rules.js
domenic: We do implement the UA stylesheet. It is in https://github.com/jsdom/jsdom/blob/master/lib/jsdom/browser/default-stylesheet.js . Indeed, these should not be in the algorithms, but instead derived from there. But, several of these are already specified there...

lib/jsdom/living/helpers/colors.js
nickiaconis: I was originally going to handle the color computation elsewhere, but I decided against that and simply didn't remove the export. I'll do that now.

lib/jsdom/living/helpers/colors.js
nickiaconis: I can try. I based this off of the observation that colors returned by `getComputedStyles` are in the rgb format, so I'm not currently aware of what spec dictates this. I'm pretty new to reading/finding specs, so do you have any advice which might help me narrow the search?

domenic: Oh, nice! Also consider this approach:

Starting at https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle , it looks like most of the work is done in https://drafts.csswg.org/cssom/#resolved-value , so you'd want to implement some sort of `function resolvedValue(...) { ... }`, probably with lots of TODOs for the properties you're not currently working on. This would also ensure you get all color-related properties, not just color, fill, and stroke.

domenic:
Tests rolled into https://github.com/jsdom/jsdom/pull/4333. All but one pass in modern jsdom.

PR #3073, opened 2020-11-11
fix: support wildcard for access-control-allow-headers

silviot:
This PR improves on #2867 adding proper formatting. Solves #2408 

The tests fail for this PR on the latest node version [because they're failing on master](https://travis-ci.com/github/silviot/jsdom/builds/199962084).

I would love adding a test for this fix, but it looks like tests come from a different repository, and I'm not familiar at all with this codebase. If needed I can add a test, but I need some pointers about how to do it.

### DIFF
--- a/lib/jsdom/living/xhr/xhr-utils.js
+++ b/lib/jsdom/living/xhr/xhr-utils.js
@@ -90,7 +90,9 @@ function validCORSPreflightHeaders(xhr, response, flag, properties) {
   }
   const acahStr = response.headers["access-control-allow-headers"];
   const acah = new Set(acahStr ? acahStr.trim().toLowerCase().split(headerListSeparatorRegexp) : []);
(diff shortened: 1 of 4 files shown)


field123:
Any timeline on getting this merged? Blocking some testing I'm doing with the Firebase Auth Emulator which uses wildcard allow headers.

silviot:
I was able to have this fix in my project using a `postinstall` entry in the `scripts` part of my `package.json`.

@field123 and others might find this gist useful: https://gist.github.com/silviot/4b062421a2bc6e128ffe666873060cda

domenic:
Tests can be added in the https://github.com/jsdom/jsdom/tree/master/test/web-platform-tests/to-upstream directory. https://github.com/jsdom/jsdom/blob/master/test/web-platform-tests/to-upstream/xhr/access-control-preflight-request-header-returns-origin.html is a reasonable example to copy.

silviot:
@domenic Thanks for the pointer! I added a test and rebased against master.

PR #3076, opened 2020-11-16
fix(default-stylesheet): set svg title to display:inline

### LINKED ISSUE #3064: default display property of svg title is 'none' but should be 'inline'
### Basic info:

- **Node.js version:** v12.18.2
- **jsdom version:** v16.4.0

### Minimal reproduction case

```js

const { window } = new JSDOM(`<!DOCTYPE html>
<html lang="en">
  <head>
    <title>JSDOM Example</title>
  </head>
  <body>
    <svg><title>Hello World</title></svg>
  </body>
</html>`);

const title = window.document.querySelector('svg title');
const styles = window.getComputedStyle(title);

console.log(styles.getPropertyValue('display'));  // 'none'
```

In all major browsers, the svg title node is `display: inline` by default, however in jsdom the default display value is `none`. This caused an interesting problem when trying to detect if the svg element has a visible title node as the element is always not visible in jsdom.

Setting the `pretendToBeVisual` option does not fix the problem.

[Link to jsbin of working example](https://jsbin.com/soyisumixe/edit?html,output).

* Chrome: `display: inline`
* Firefox: `display: inline`
* Safari: `display: inline`
* IE11: `display: inline`
* Edge: `display: inline`

straker:
Closes #3064

### DIFF
--- a/lib/jsdom/browser/default-stylesheet.js
+++ b/lib/jsdom/browser/default-stylesheet.js
@@ -780,6 +780,10 @@ dialog {
     color: black
 }
 
+svg title {
+    display: inline;
+}
+
 /* noscript is handled internally, as it depends on settings. */
 
 `;


domenic:
This isn't the right fix, so I'll close this. As discussed in the issue, it's not clear this is right per spec. Even if it were right, the correct fix would be to change the default way `display` is computed, not add one-offs for individual elements.

Issue #3069, opened 2020-11-03
Jsdom includes a dependency on the deprecated library request-promise-native


yarhamjohn:
### Basic info:

`jsdom` has a dependency on `request-promise-native` which in turn has a peer-dependency on `request`. Both `request` and `request-promise-native` have been deprecated for some time now so I was wondering if there was a plan to migrate to a different library - some options were suggested here: https://github.com/request/request/issues/3143

guimard:
Hi,
looking at code, [axios](https://www.npmjs.com/package/axios) may be a code candidate:
 * provides native promise
 * has a "stream" mode that permits easily to fix [domexception/scripts/get-latest-platform-tests.js](domexception/scripts/get-latest-platform-tests.js)
 * there is a [axios-cookiejar-support](https://www.npmjs.com/package/axios-cookiejar-support) to fix [lib/jsdom/living/helpers/wrap-cookie-jar-for-request.js](lib/jsdom/living/helpers/wrap-cookie-jar-for-request.js)

guimard:
NB: duplicate of #3030

guimard:
See also #3031 and #3054

Issue #3030, opened 2020-08-24
Replace request-promise-native


guimard:
### Basic info:

- **Node.js version:** 12
- **jsdom version:** last

### Resume :

According to [upstream warning](https://github.com/request/request-promise-native#deprecated), request-promise-native is deprecated and should be replaced. It is used in

https://github.com/jsdom/jsdom/blob/5279cfda5fe4d52f04b2eb6a801c98d81f9b55da/lib/jsdom/browser/resources/resource-loader.js#L4-L6

SimenB:
Duplicate of #2792

zynyz:
Hi guys --

This has now become a breaking issue for me. My Google cloud function will no longer deploy. The build process is treating the deprecation warning as a fatal error. This seems to be new behavior since yesterday:

```
Build failed: npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142 
npm WARN deprecated request-promise-native@1.0.9: request-promise-native has been deprecated because it extends the now deprecated request package, see https://github.com/request/request/issues/3142 
npm WARN deprecated har-validator@5.1.5: this library is no longer supported 
npm ERR! Cannot read property 'failed' of undefined 
npm ERR! A complete log of this run can be found in: 
npm ERR! /builder/home/.npm/_logs/2020-10-20T23_16_10_668Z-debug.log; Error ID: 49341d49
```

* I apologize for any bad formatting. The preview pane does not seem to want to display today.

PR #3092, opened 2021-01-09
Replace request (as a direct dependency) with node-fetch

### LINKED ISSUE #2792: Replace request with something better

vegardbb:
# Relation to issues and other pull requests

Closes #2792, #3030, and #3069. Overrides #3031 and #3054.

# Description

This PR handles a few complaints from the linter aside from using the packages `node-fetch`  and `fetch-cookie` in the resource loader module, and implementing a new wrapper class for core http/https libraries when handling XHTTPRequests. The new Request class in the helper directory handles redirects, tunneling and basic "username:password" authentication using Base64 encoding. No other features of `request` have been reimplemented. I have also merged master into this branch.

# Note

One of our dependencies, `wd`, uses `request` as an HTTP utility, so even with this PR merged, `request` remains in the `yarn.lock` file as an indirect dependency. See [http-utils.js](https://github.com/admc/wd/blob/master/lib/http-utils.js).


domenic:
The API tests are much less important than the web platform tests. Please ping this PR when all of those pass and the build is green. Otherwise, we cannot merge this, as it breaks jsdom's actual web spec implementations.

oliverjessner:
Thanks we really need to get rid of request

vegardbb:
@domenic Running the test suite on my Ubuntu 20.04 system and Node v14.15 as shown in this [gist](https://gist.github.com/vegardbb/2b550ccfa83314a56ecfdff092e18234), 5064 of the tests pass, and the remaining 608 time out. The full test suite ran for 28 minutes. Does this suffice as evidence that this PR may be merged into master and published as a minor increment, or do we need to see the build tests on GitHub Actions pass first?

domenic:
Unfortunately, yeah, we need to have the tests passing on a CI system before we can merge them. I'm sorry the project is in such an unmaintained state that nobody has worked on the GitHub actions upgrade, but it does seem like that's a blocker at this point for merging any large changes to the codebase.

n0rbyt3:
@domenic In my system (Ubuntu 20.04), testing the wpts on ```origin/master``` also fails with timeouts.

@vegardbb Well done! I had a look at your changes and will push some updates next week, f.ex. you missed adding ```node-fetch``` to ```package.json``` and ```getProxyFromURI``` is never called. Anyway, that doesn't help to fix the timeouts.

domenic:
I see this is all passing; so great!

Can you open a separate PR to get all the tests passing on master (mostly through to-run.yml changes, I expect?) so that we can have a known good state, before merging a dramatic refactoring?

I'm in particular a bit worried about `abort-during-headers-received.window.html` and it'd be good to debug that in a dedicated PR.

domenic:
Oh... I just went to review the actual code for this. It seems that you've copied and pasted the source code of the deprecated request module into jsdom? That's not something we can accept... that's just hiding the warning for using deprecated code, while still using the deprecated code. I thought the goals was to implement on top of an actually-supported library. I'm sorry for not realizing this earlier.

jsg2021:
why `got`? over `node-fetch` or axios (both zero deps vs got's 12+)?

vegardbb:
@jsg2021 Pure experimentation, really. I've tried to use `node-fetch` in `xhr-utils` previously, but a horde of tests time out as a result of that.

jsg2021:
nodes builtin http module would probably give the most control. It would be a little verbose, but limit the dependencies. Just thinking out loud. Keep up the good work!

vegardbb:
@domenic I'll create a new PR with only the changes made in `to-run.yaml` shortly.

vegardbb:
@domenic I have implemented a new replacement prototype which handles request redirects. 43 of the tests in CI are now failing. Do you hav any pointers as to why and how I can fix them?

vegardbb:
When running the test status-preflight.htm, each of the preflight requests are receiving 500 responses. How do I get the test server to log out the stack trace if it throws an error?

allanbowe:
So close!!  Anything we can do to support?

vegardbb:
@allanbowe I had to dive into the HTTP client source code to figure out the discrepancy between v15 and the stable releases of Node. Turns out that the response object emits an error if the response is not yet finished when the socket object gets destroyed. No such event is fired in either v14, v12 or v10 of Node. I'm working on a version - specific hack as we speak.

@domenic I have noticed there are still a few other disabled tests that have been disabled because of internal behaviour in the request module, and will likely fail or hang if they were turned on. Should I investigate those tests too?

KenEucker:
Really looking forward to this update. Thanks for your hard work @vegardbb!

langyo:
Finally close to the merger, go for it!

domenic:
Very exciting to see the full test suite passing! And yes, any additional tests you can get to pass would be a big motivation for merging this. I can see there are already improvements from your to-rum.yaml updates!

I don't have time for a full review this weekend, but some things I noticed that will need fixing:

- Please don't disable ESLint rules. They are there for a reason. (E.g. in resource-loader.js.)
- Remove all jsdoc. (E.g. in resource-loader.js.)
- Don't create "old-style" classes; use the `class` keyword. (E.g. in http-request.js.)
- Some of this code looks like it might be copied from other modules. That isn't generally acceptable for this project, so please be careful. It also seems to lead to some over-complicated layers of indirection that we don't want. (E.g., the bottom of http-request.js from urlToOptions onward; and the custom error stuff earlier in the file.)

vegardbb:
@domenic I intend to refactor a little based on your initial feedback in the immediate future, could you hold off on the code review until I am done with that?

vegardbb:
I experimented with removing the lint disabling and using the "window" variable in the test/utils file, but could not find an alternative solution for the in-browser block without breaking either the browser build or the lint check. So I've left it as is. I have also shrunk the new module `http-request` by 80 loc, and exported the class directly.

The test suite should go green by now, so I'll be awaiting your PR review.

vegardbb:
@domenic @TimothyGu @Sebmaster Have any of the maintainers had time to review this pull request yet?

package.json
```diff
@@ -21,6 +21,7 @@
   "repository": "jsdom/jsdom",
   "dependencies": {
     "abab": "^2.0.5",
+    "abort-controller": "^3.0.0",
```
ls-andrew-goodale, line 24: This looks like just a `devDependency` since I only see it used in `test/util.js`

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+  getCookie() {
+    let cookies;
+    if (this._redirectCount === 0) {
+      this.originalCookieHeader = this.getHeader("Cookie");
+    }
+    const jar = this._options.cookieJar;
+    if (jar && typeof jar.getCookieStringSync === "function") {
+      cookies = jar.getCookieStringSync(this._currentUrl);
+    } else {
+      this._areCookiesDisabled = true;
```
domenic, line 51: It seems like this shouldn't be set every time you call getCookie, but instead should be determined in the constructor.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+    }
+    if (cookies && cookies.length) {
```
domenic, line 53: Why is a method called getCookie() setting headers?

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  write(data, encoding) {
+    if (this._ending) {
+      const writeAfterEndError = new Error("Writing data after ClientRequest.end() is called is not allowed");
+      writeAfterEndError.code = "ERR_STREAM_WRITE_AFTER_END";
+      throw writeAfterEndError;
+    }
+    if (!(typeof data === "string" || (typeof data === "object" && ("length" in data)))) {
+      throw new TypeError("data should be a string, Buffer or Uint8Array");
```
domenic, line 82: Since we are presumably the only people writing here, could we make this always use a single type (e.g. Uint8Array)?

Also, this kind of explicit type-checking is not desired, since again we control the code so we're just checking ourselves.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+        cookies.forEach(cookie => jar.setCookieSync(cookie, self._currentUrl, { ignoreError: true }));
+      } catch (e) {
+        this.emit("error", e);
+      }
+    }
+  }
+
+  write(data, encoding) {
+    if (this._ending) {
+      const writeAfterEndError = new Error("Writing data after ClientRequest.end() is called is not allowed");
```
domenic, line 77: This kind of self-checking doesn't seem necessary, since we're the ones writing. We should just not call this method in these cases.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+    this._currentRequest.setHeader(name, value);
+  }
+
+  removeHeader(name) {
+    delete this._options.headers[name];
+    this._currentRequest.removeHeader(name);
+  }
+
+  toJSON() {
+    const uri = new URL(this._currentUrl);
```
domenic, line 108: uri => url

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+    this._options.headers[name] = value;
+    this._currentRequest.setHeader(name, value);
+  }
+
+  removeHeader(name) {
+    delete this._options.headers[name];
+    this._currentRequest.removeHeader(name);
+  }
+
+  toJSON() {
```
domenic, line 107: What calls this?

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+    const uri = new URL(this._currentUrl);
+    const { method, headers } = this._options;
+    return { uri, method, headers };
+  }
+
+  _performRequest() {
+    const urlOptions = new URL(this._currentUrl);
+    const { protocol: scheme } = urlOptions;
```
domenic, line 115: Prefer `const scheme = urlOptions.protocol`

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+    this._options.agent = this._options.agents[scheme.substr(0, scheme.length - 1)];
```
domenic, line 116: Prefer `substring()` instead of the deprecated `substr()`

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+    const self = this;
+    function onNative(response) {
```
domenic, line 118: Use an arrow function instead.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+      self._processResponse(response);
+    }
+    const { request } = scheme === "https:" ? https : http;
+    const message = request(this._currentUrl, this._options, onNative);
```
domenic, line 122: Why is this file using http/https.request directly, and the other file using node-fetch? It'd be better if they were unified.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+    this._currentRequest = message;
+    this.getCookie();
```
domenic, line 124: Now I'm quite sure that getCookie() is a bad name for this function.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+    ["connect", "error", "socket", "timeout"].forEach(event => {
```
domenic, line 125: Use `for`-`of`

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+          }
+        }
+      })();
+    }
+  }
+
+  _processResponse(response) {
+    response.request = {};
+    response.request.headers = this._options.headers;
+    response.request.uri = new URL(this._currentUrl);
```
domenic, line 153: uri => url

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+    this.response = response;
+    this.setCookies(response);
+    const { statusCode } = response;
+    const { location } = response.headers;
+    const catchResErrors = err => {
+      if (!(majorNodeVersion === 15 && err.message === "aborted")) {
```
domenic, line 159: Add a comment explaining this sort of version-sniffing, if it is indeed necessary.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+      this._options.followRedirects &&
+      statusCode >= 300 &&
+      statusCode < 400) {
+      redirectAddress = location;
+    } else if (statusCode === 401 &&
+      /^Basic /i.test(response.headers["www-authenticate"] || "") &&
+      (this._options.user && this._options.user.length)) {
+      const { user, pass } = this._options;
+      this._options.auth = `${user}:${pass}`;
+      redirectAddress = 1;
```
domenic, line 176: 1??? It seems like this variable is serving two purposes and you should instead use two different variables.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+      if (((statusCode === 301 || statusCode === 302) && this._options.method === "POST") ||
+        (statusCode === 303 && !/^(?:GET|HEAD)$/.test(this._options.method))) {
+        this._options.method = "GET";
+        this._requestBodyBuffers = [];
+      }
+      let previousHostName = this.removeMatchingHeaders(/^host$/i);
+      if (!previousHostName) {
+        previousHostName = new URL(this._currentUrl).hostname;
+      }
+      const previousURI = this._currentUrl;
```
domenic, line 197: previousURI => previousURL

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+      this.emit("redirect");
+      try {
+        this._performRequest();
+      } catch (cause) {
+        this.emit("error", cause);
+      }
+    } else {
+      const isCompressed = this._options.headers["Accept-Encoding"] === "gzip, deflate";
+      const self = this;
+      function hasBody(code) {
```
domenic, line 218: This function is only called once and does the ugly `self = this` thing. It's best to just inline.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  getHeader(key, value) {
+    if (this._currentRequest) {
+      return this._currentRequest.getHeader(key, value);
+    }
+    return null;
+  }
+
+  removeMatchingHeaders(regex) {
```
domenic, line 259: This probably should be `_removeMatchingHeaders`

lib/jsdom/living/xhr/xhr-utils.js
```diff
@@ -269,26 +270,23 @@ function createClient(xhr) {
...
-    options.auth = {
-      user: flag.auth.user || "",
-      pass: flag.auth.pass || "",
-      sendImmediately: false
-    };
+    const { user, pass } = flag.auth;
+    const username = typeof user === "string" ? user : "";
+    const password = typeof pass === "string" ? pass : "";
+    options.user = username;
+    options.pass = password;
```
domenic, line 285: These lines could be much simpler as

```js
options.user = flag.auth.user || "";
options.pass = flag.auth.pass || "";
```

without four similarly-named intermediate variables.

lib/jsdom/living/xhr/xhr-utils.js
```diff
@@ -332,11 +352,11 @@ function createClient(xhr) {
   if (crossOrigin && (!simpleMethods.has(ucMethod) || nonSimpleHeaders.length > 0 || properties.uploadListener)) {
     client = new EventEmitter();
 
-    const preflightRequestHeaders = [];
+    const preflightRequestHeaders = {};
     for (const header in requestHeaders) {
       // the only existing request headers the cors spec allows on the preflight request are Origin and Referrer
```
domenic, line 357: Good catch on the spelling below. Update this comment too.

package.json
domenic: I see it used in resource-loader

test/util.js
```diff
@@ -125,9 +125,24 @@ function getTestFixtureUrl(relativePath) {
...
+        throw new Error(`Unexpected status ${response.status} fetching ${response.location}`);
+      }
+      clearTimeout(timeout);
+      return response.text();
+    }
+    function fail(e) {
+      clearTimeout(timeout);
+      throw e;
+    }
+    return fetch(exports.getTestFixtureUrl(relativePath), { method: "GET", signal }).then(res).catch(fail);
```
domenic, line 144: You can rewrite this using async/await to be simpler.

test/web-platform-tests/start-wpt-server.js
```diff
@@ -66,8 +67,12 @@ module.exports = ({ toUpstream = false } = {}) => {
 };
 
 function pollForServer(url, lastLogTime = Date.now()) {
-  return requestHead(url, { strictSSL: false })
-    .then(() => {
+  const agent = url.startsWith("https") ? new https.Agent({ rejectUnauthorized: false }) : null;
```
domenic, line 70: I'm surprised this is necessary in the testing context. Have you tried without?

package.json
ls-andrew-goodale: Indeed - I missed that. Sorry.

lib/jsdom/browser/resources/resource-loader.js
```diff
@@ -103,8 +115,91 @@ module.exports = class ResourceLoader {
...
+              args[2].ignoreError = true;
+            }
+            return cookieJar.__setCookie(...args);
+          };
+        }
+
+        let lastUrl = options.referrer;
+        let lastOpts = null;
+
+        const targetFetch = fetchCookie(
```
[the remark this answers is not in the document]
vegardbb: From what I gather, node-fetch is a bare-bones promise wrapper for http.ClientRequest, and ancillary features regarded as client-specific use case, such as handling cookies during redirect, are left outside the module's scope.

vegardbb:
Thanks for the feedback, I will be addressing all of the nitpicks over the weekend. I would also like to tinker with the setCookie monkeypatch and the Promise interface reimplementation in resource-loader, so as to ascertain why they are essential to the spec implementation, and then document them as inline comments.

lib/jsdom/living/helpers/http-request.js
vegardbb: I do not know why the testharness needs it, but if I remove that method in particular, a few of tests will throw "SyntaxError: Unexpected end of JSON input" and halts the test suit altogether.

test/web-platform-tests/start-wpt-server.js
vegardbb: I have tried leaving out the "agent" option in the fetch call, which caused the before-hook in Mocha to crash. Basically, without it we cannot poll the https address.

lib/jsdom/living/helpers/http-request.js
vegardbb: The `property `response.request.uri` is accessed by XMLHTTPRequestImpl.js in its redirect event listener, to mention one place.

```javascript
      client.on("redirect", () => {
        const { response } = client;
        const destUrlObj = new URL(response.request.headers.Referer);

        const urlObj = new URL(response.request.uri.href);

        if (destUrlObj.origin !== urlObj.origin && destUrlObj.origin !== flag.origin) {
          properties.origin = "null";
        }

        response.request.headers.Origin = properties.origin;

        if (flag.origin !== destUrlObj.origin &&
            destUrlObj.protocol !== "data:") {
          if (!xhrUtils.validCORSHeaders(this, response, flag, properties, flag.origin)) {
            return;
          }
          if (urlObj.username || urlObj.password) {
            properties.error = "Userinfo forbidden in cors redirect";
            xhrUtils.dispatchError(this);
          }
        }
      });
```

lib/jsdom/browser/resources/resource-loader.js
```diff
@@ -103,8 +115,91 @@ module.exports = class ResourceLoader {
 
       case "http":
       case "https": {
-        const requestOptions = this._getRequestOptions(options);
-        return request(urlString, requestOptions);
+        const { cookieJar } = options;
+        if (!cookieJar.__setCookie) {
+          cookieJar.__setCookie = cookieJar.setCookie;
```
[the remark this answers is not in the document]
vegardbb: All the API tests are passing with this block commented out.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,272 @@
...
+const https = require("https");
+const { Writable } = require("stream");
+const zlib = require("zlib");
+
+const ver = process.version.replace("v", "").split(".");
+const majorNodeVersion = Number.parseInt(ver[0]);
+const minorNodeVersion = Number.parseInt(ver[1]);
+
+function abortRequest(clientRequest) {
+  if (majorNodeVersion > 13 || (majorNodeVersion === 13 && minorNodeVersion > 13)) {
```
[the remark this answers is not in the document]
vegardbb: I tried to use only the `destroy` method, but that broke the test suite for versions 10 and 12 for Node. I've added a comment accordingly.

vegardbb:
@domenic Regarding your concerns with the different implementations written in `resource-loader` and `xhr-utils` have I come up with the following refactoring (the tests are all still passing, of course):

```javascript
      case "http":
      case "https": {
        const { accept, cookieJar, referrer } = options;
        const agents = agentFactory(this._proxy, this._strictSSL);
        const requestOptions = {
          followRedirects: true,
          cookieJar,
          agents,
          headers: {
            "User-Agent": this._userAgent,
            "Accept-Language": "en",
            "Accept-Encoding": "gzip",
            "Accept": accept || "*/*"
          }
        };
        if (referrer && !IS_BROWSER) {
          requestOptions.headers.Referer = referrer;
        }
        const requestClient = new Request(urlString, requestOptions);
        const promise = new Promise((resolve, reject) => {
          const accumulated = [];
          let accumulatedBytelength = 0;
          requestClient.once("response", ({ statusCode }) => {
            if (statusCode === 404) {
              requestClient.abort();
              reject(new Error("404 Not Found"));
            } else if (statusCode >= 500) {
              requestClient.abort();
              reject(new Error("Error encountered"));
            }
          });
          requestClient.on("data", chunk => {
            accumulatedBytelength += chunk.length;
            accumulated.push(chunk);
          });
          requestClient.on("end", () => {
            if (requestClient.isAborted) {
              resolve(null);
            } else {
              try {
                resolve(Buffer.concat(accumulated, accumulatedBytelength));
              } catch (error) {
                // This try-catch-loop was written to avoid the test suite halting
                reject(error);
              }
            }
          });
          requestClient.on("error", err => {
            if (!requestClient.isAborted) {
              reject(err);
            }
          });
        });
        requestClient.on("end", () => {
          promise.response = requestClient.response;
          promise.href = requestClient._currentURL;
        });
        promise.abort = requestClient.abort.bind(requestClient);
        promise.getHeader = name => requestOptions.headers[name] || requestClient.getHeader(name);
        requestClient.end();
        return promise;
      }
```

As far as I know, that would mean we could relegate `node-fetch` and `abort-controller` to dev-dependencies, no?

domenic:
Excellent, that would be perfect!

lib/jsdom/browser/resources/resource-loader.js
```diff
@@ -103,8 +82,56 @@ module.exports = class ResourceLoader {
...
+          requestClient.on("end", () => {
+            try {
+              resolve(Buffer.concat(accumulated, accumulatedBytelength));
+            } catch (error) {
+              reject(error);
+            }
+          });
+          requestClient.on("error", reject);
+        });
+        requestClient.on("end", () => {
```
domenic, line 127: Can you add comments explaining what parts of the code use response / href / getHeader? Eventually I'd like to eliminate those and use standard non-patched promises. (I know we need `abort()` pretty extensively, but I was surprised we need the others.)

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,261 @@
...
+    clientRequest.abort();
+  }
+  clientRequest.removeAllListeners();
+  clientRequest.on("error", () => {});
+}
+
+module.exports = class Request extends Writable {
+  constructor(url, options) {
+    super();
+    this._currentURL = url;
```
domenic, line 25: It looks like resource-loader uses this property, so it should not be underscore-prefixed.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,261 @@
...
+    this._options = options;
+    this.headers = this._options.headers;
+    this._ended = false;
+    this._redirectCount = 0;
+    this._requestBodyBuffers = [];
+    this._bufferIndex = 0;
+    this.response = Object.create(null);
```
domenic, line 32: Why initialize this to Object.create(null)? Seems like null would make more sense? Or just not assigning it?

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,261 @@
...
+  clientRequest.removeAllListeners();
+  clientRequest.on("error", () => {});
+}
+
+module.exports = class Request extends Writable {
+  constructor(url, options) {
+    super();
+    this._currentURL = url;
+    this._options = options;
+    this.headers = this._options.headers;
```
domenic, line 27: It looks like some of the XHR code accesses `realClient._options.headers`. Which is bad because it's accessing underscore-prefixed properties. Which makes me wonder why we have two copies of this property anyway.

Maybe the better pattern would be to have a public getter, e.g. `get headers() { return this._options.headers; }`?

Although even then there's a _third_ source of truth, right? `this._currentRequest.headers`? Hmm.

lib/jsdom/living/helpers/http-request.js
domenic: This is really surprising. If you have any time to look into it, that'd be much appreciated, but it's not a blocker for merging. If you can't figure it out then adding a comment like `// TODO: investigate this further and ideally remove` would be fine.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,261 @@
...
+      } else if (this._ended) {
+        this._currentRequest.end();
+      }
+    }
+  }
+
+  _performRequest() {
+    const urlOptions = new URL(this._currentURL);
+    const scheme = urlOptions.protocol;
+    this._options.agent = this._options.agents[scheme.substring(0, scheme.length - 1)];
```
domenic, line 93: I'm confused by this line. Nothing else in the PR uses `_options.agent` from what I can tell. So I guess this is used on the next line? But now it seems like `this._options` is serving multiple purposes: some of its properties (like `agents`) are inputs to this file, and some of the properties (like `agent`) are input to `http.request()`.

Is it possible to separate these uses out, so that `this._options` only serves a single purpose? Or maybe it goes away entirely, in favor of just underscore-prefixed properties for storing input/state + assembling a new options object in this function.

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,261 @@
...
+      });
+    }
+    if (this._isRedirect) {
+      this._bufferIndex = 0;
+      this._writeNext();
+    }
+  }
+
+  _processResponse(response) {
+    response.request = {};
```
domenic, line 127: Why is this stored as a property of `response`? It seems like it's only used in this function and so could be a normal variable?

lib/jsdom/living/xhr/xhr-utils.js
```diff
@@ -297,30 +292,52 @@ function createClient(xhr) {
...
-
-      if (hasBody && flag.formData) {
-        const form = client.form();
-        for (const entry of body) {
-          form.append(entry.name, entry.value, entry.options);
+      let requestBody = body;
+      let len = 0;
+      if (hasBody) {
+        if (flag.formData) {
+          requestBody = new FormData();
```
domenic, line 305: It would be nice not to depend on an external library for this, especially since the part of the library that we use is so small. Let's add something like

```js
// TODO: implement https://html.spec.whatwg.org/#multipart-form-data directly instead of using an external library
```

test/util.js
```diff
@@ -107,27 +108,40 @@ exports.inBrowserContext = () => {
...
+exports.readTestFixture = async relativePath => {
   if (exports.inBrowserContext()) {
-    return request(getTestFixtureUrl(relativePath), { timeout: 5000 });
+    const abortController = new AbortController();
+    const { signal } = abortController;
+    const timeout = setTimeout(() => {
+      abortController.abort();
+    }, 5000);
+    try {
+      const response = await fetch(exports.getTestFixtureUrl(relativePath), { method: "GET", signal });
```
domenic, line 134: It looks like we only use node-fetch in a browser context, right? In that case you could replace this with `await self.fetch(...)` and the above with `new self.AbortController()`, and remove the dependencies, I am pretty sure. Do you mind trying that?

lib/jsdom/living/helpers/http-request.js
```diff
@@ -0,0 +1,261 @@
...
+    clientRequest.destroy();
+  } else {
+    clientRequest.abort();
+  }
+  clientRequest.removeAllListeners();
+  clientRequest.on("error", () => {});
+}
+
+module.exports = class Request extends Writable {
+  constructor(url, options) {
```
domenic, line 23: It looks like some of the XHR code accesses `requestClient._currentRequest`, which implies that should not be underscore-prefixed.

domenic (commented):
This is looking really good. Going the extra mile to consolidate our two fetching paths is super-appreciated and I'm glad it worked out. We're getting to the point of mostly-stylistic nits, and if we get to that point completely I can just fix them up myself, so I'd suspect this might be the last or second-to-last real review before merging.

I think the following summarize the substantive issues I'd appreciate you looking in to:

- Since `http-request.js` is now going to be the core of HTTP in jsdom, we need to be very clear on what its public API is. Right now several parts of the code use its "private" (underscore-prefixed) properties, which is bad.
- Similarly, there's some confusion about what an "option" is. I think there's at least two kinds: options for constructing the `Request`, and options for calling `http.request()`/`https.request()`. It'd be ideal to separate those.
- It looks like we're only using `node-fetch` and `abort-controller` in a browser context now, where they're not actually needed. So maybe we can eliminate them entirely.
- I'm not very comfortable with the 404 + 5xx logic in resource-loader, and have suggested a couple potential paths for fixing it.

test/util.js
vegardbb: So, `self` is the stand-in for the `window` object here?

lib/jsdom/browser/resources/resource-loader.js
```diff
@@ -103,8 +82,56 @@ module.exports = class ResourceLoader {
...
+        };
+        if (referrer && !IS_BROWSER) {
+          requestOptions.headers.Referer = referrer;
+        }
+        const requestClient = new Request(urlString, requestOptions);
+        const promise = new Promise((resolve, reject) => {
+          const accumulated = [];
+          let accumulatedBytelength = 0;
+          requestClient.once("response", ({ statusCode }) => {
+            if (statusCode === 404) {
```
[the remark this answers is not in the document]
vegardbb: Failing tests: https://github.com/jsdom/jsdom/runs/2558580272?check_suite_focus=true

lib/jsdom/living/helpers/http-request.js
vegardbb: Required by XMLHTTPRequestImpl.js, but nowhere else (I think). There might yet be a way to refactor the XHR module such that this parameter patching can be circumvented.

lib/jsdom/browser/resources/resource-loader.js
vegardbb: All of these three properties are used in the method `fromURL` in `lib/api.js`, and the omission of either one of them leads to 2650 tests crashing in this very function.

```javascript
  static fromURL(url, options = {}) {
    return Promise.resolve().then(() => {
      // Remove the hash while sending this through the research loader fetch().
      // It gets added back a few lines down when constructing the JSDOM object.
      const parsedURL = new URL(url);
      const originalHash = parsedURL.hash;
      parsedURL.hash = "";
      url = parsedURL.href;

      options = normalizeFromURLOptions(options);

      const resourceLoader = resourcesToResourceLoader(options.resources);
      const resourceLoaderForInitialRequest = resourceLoader.constructor === NoOpResourceLoader ?
        new ResourceLoader() :
        resourceLoader;

      const req = resourceLoaderForInitialRequest.fetch(url, {
        accept: "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
        cookieJar: options.cookieJar,
        referrer: options.referrer
      });

      return req.then(body => {
        const res = req.response;

        options = Object.assign(options, {
          url: req.href + originalHash,
          contentType: res.headers["content-type"],
          referrer: req.getHeader("referer")
        });

        return new JSDOM(body, options);
      });
    });
  }
```

test/util.js
vegardbb: Actually, there is one spot within the Node execution context where node-fetch is required, the script `start-wpt-server.js`. Still makes it a dev-dependency only, so that should be acceptable, right?

domenic: Yeah, seems reasonable. Maybe that could be replaced too but it's not necessarily worth doing right now.

vegardbb:
@domenic I have addressed and implemented all of your suggestions. Now, we are down to three kinks.

 * The explicit promise rejection in resource-loader on non-2xx status remains, but in one condensed conditional loop instead of two covering two specific status codes, with a TODO signaling that this block deviates from the spec.
 * The odd **toJSON** method in http-request remains as well, but in this case I have triangulated which test halts if the method is removed, as well as where in the code it halts.
 * To my knowledge (should be looked into), three of the properties patched on the Promise instance in the HTTP switch case in `resource-loader`, `href`, `getHeader` and `response` is used only within one specific library file, `lib/api.js`.



```javascript
pipeRequest(form) {
  form.pipe(this._currentRequest);
}
```

I ultimately decided against making the ClientRequest instance within the Request class part of its public interface, in the interest of maintaining some level of abstraction. The cost was this awkward one-liner, which slightly harkens back to the old Request prototype which juggles too many responsibilities in the air. Outside of piping the ClientRequest instance into the FormData instance (a feature of the form-data module), the `_currentRequest` field is not referred to outside of the class. If this new method is unacceptable to you, the immediate alternative we have left is to define a *get* method on the `_clientRequest` field, and reverting the slight alteration in `xhr-utils`.

To further patch the abstraction leak, the XHR implementation was refactored a little such that a `response` field in the new Request class is no longer required. That's one more remnant of the deprecated Request prototype extinguished.

We may now proceed with one more iteration of reviewing this PR.

test/util.js
```diff
@@ -107,27 +106,40 @@ exports.inBrowserContext = () => {
...
+    const { signal } = abortController;
+    const timeout = setTimeout(() => {
+      abortController.abort();
+    }, 5000);
+    try {
+      const response = await self.fetch(exports.getTestFixtureUrl(relativePath), { method: "GET", signal });
+      if (!response.ok) {
+        throw new Error(`Unexpected status ${response.status} fetching ${response.location}`);
+      }
+      clearTimeout(timeout);
```
domenic, line 136: This could be done in a `finally` block for slightly nicer code.

test/util.js
```diff
@@ -107,27 +106,40 @@ exports.inBrowserContext = () => {
  * @param {string} relativePath Relative path within the test directory. For example "jsdom/files/test.html"
  * @returns {string} URL
  */
-function getTestFixtureUrl(relativePath) {
+exports.getTestFixtureUrl = relativePath => {
```
domenic, line 109: Not sure this needs to be exported?

lib/jsdom/browser/resources/resource-loader.js
```diff
@@ -103,8 +82,49 @@ module.exports = class ResourceLoader {
 
       case "http":
       case "https": {
-        const requestOptions = this._getRequestOptions(options);
-        return request(urlString, requestOptions);
+        const { accept, cookieJar, referrer } = options;
```
domenic, line 85: It'd be nice if we could put these into the function signature instead of destructuring them here. I need to look at the rest of the function to make sure that works though.

domenic (approved):
This looks really good. I found a couple nits and I want to try removing node-fetch from the test suite but I can fix those and try that myself tomorrow. Then we will merge and do a release. No further work needed from you, and thanks for refining this over time into a very nice architecture!

TimothyGu:
Sorry, fat fingered.

allanbowe:
Hurrah! Thanks!

Issue #2963, opened 2020-05-07
Unexpected token in webidl-conversions


su30007:
### Basic info:

- **Node.js version:** 10
- **jsdom version:** 16.22

### Minimal reproduction case

```js
import React from 'react';
import jsdom from 'jsdom';
import { Header, Footer, LinksContainer, Toolbox } from '../components';
import App from '../App';

const { JSDOM } = jsdom;
const { window } = new JSDOM(`...`);
describe('Landing page', () => {
  beforeEach(() => {
    jest.resetModules();
  });

  afterEach(() => {
    jest.clearAllMocks();
    jest.clearAllTimers();
  });

  it('should render without throwing an error', () => {
    window.chrome = {
      runtime: {}
    };

    const wrapper = shallow(<App />);

    expect(wrapper.find('h1').text()).toEqual(i18n('LANDING_PAGE_TITLE'));
    expect(wrapper.find(Header)).toBeDefined();
    expect(wrapper.find(Toolbox)).toBeDefined();
    expect(wrapper.find(LinksContainer).length).toEqual(2);
    expect(wrapper.find(Footer)).toBeDefined();

    window.chrome.webstore = {};
    wrapper.setProps({ test: 'test' });
  });

  it('handleCancel should hide the modal', () => {
    window.chrome = false;
    const wrapper = shallow(<App />);
    const instance = wrapper.instance();

    instance.handleCancel();

    expect(wrapper.state('visible')).toEqual(false);
  });
});

... your code that reproduces the problem here, probably using dom.window ...
```

### How does similar code behave in browsers?

Run a unit test by jest, then you will see

  ● Test suite failed to run

  /node_modules/jsdom/node_modules/webidl-conversions/lib/index.js: Unexpecte
d token, expected ( (357:12)

        355 | 
        356 |         return true;
      > 357 |     } catch {
            |             ^
        358 |         return false;
        359 |     }
        360 | }

domenic:
This does not follow the issue template, as it includes technologies besides jsdom (jest, react) and uses syntax (import, JSX) that is not supported in Node.js v10. Please add a comment after you fix the code sample to follow the issue template and verify that your new code sample still causes an issue.

partyka1:
I have this issue just from importing `jsdom` in new empty project, node 8.17.0 :(

eleanombre:
@partyka1 I just had this issue, it disappeared when using Node v10.16.1 (npm v6.9.0).

[...]

PR #3094, opened 2021-01-13
fix catch blocks missing identifier

erickmiller:
to comply with es5 javascript standard

### DIFF
--- a/lib/jsdom/living/custom-elements/CustomElementRegistry-impl.js
+++ b/lib/jsdom/living/custom-elements/CustomElementRegistry-impl.js
@@ -46,7 +46,7 @@ function isConstructor(value) {
     new P();
 
     return true;
-  } catch {
+  } catch (e) {
     return false;
   }
 }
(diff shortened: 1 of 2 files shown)


erickmiller:
> to comply with es5 javascript standard

As described in issue 2963 thread here: https://github.com/jsdom/jsdom/issues/2963

domenic:
This is not a change we're interested in.

erickmiller:
Why?  We are running a node version far past version 10 -- this has nothing to do with the Node version, this is an an incompatibility with the javascript standard and apparently an incompatibility with webpack and makes no sense to not fix it's a few lines of incorrect javascript syntax in your code.  99% of the places in the repo call try / catch correctly.  There are 3 files that contain a syntactically incorrect call to the catch statement.  I fixed it, took me 5 minutes.  This is a very useful library but is no good if it can't compile with other industry standard javscript tools. You should accept the change.  Why would you not want to?  Makes no sense.  Will have to keep patching on each build.  Explain any logical reason to keep incorrect syntax in these few files when the fix is so incredibly simple and makes no difference to the logical runtime of the code.

Issue #3107, opened 2021-02-05
Location href in iframe with "about:blank" src is not "about:blank"


SimonMueller:
Hey there.

It looks like iframes with a src attribute of "about:blank" inherit the location href from their parent which is not how most browsers behave. I already created a Pull Request to fix this and i will link the issue there.

### Basic info:

- Node.js version: 14.6.0
- jsdom version: 16.4.0
- Tested in: Edge (88.0.705.62), Firefox (85.0), Chrome (88.0.4324.146)

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const options = {
    url: "https://example.org/",
};
const dom = new JSDOM(`
  <iframe id="iframe" src="about:blank"></iframe>
`, options);

const document = dom.window.document;
const iframe = document.getElementById('iframe');

const iframeWindowHref = iframe.contentWindow.location.href;

console.log(iframeWindowHref === 'about:blank');  // false but should be true
console.log(iframeWindowHref === 'https://example.org/');  // true but should should be false
```

### How does similar code behave in browsers?

https://jsbin.com/ruhegifewa/edit?html,js,console

PR #3108, opened 2021-02-05
Location href in about blank src iframe

SimonMueller:
Fixes https://github.com/jsdom/jsdom/issues/3107

### DIFF
--- a/lib/jsdom/living/helpers/document-base-url.js
+++ b/lib/jsdom/living/helpers/document-base-url.js
@@ -1,5 +1,6 @@
 "use strict";
 const whatwgURL = require("whatwg-url");
+const { implForWrapper } = require("../generated/utils");
 
 exports.documentBaseURL = document => {
   // https://html.spec.whatwg.org/multipage/infrastructure.html#document-base-url
@@ -25,10 +26,11 @@ exports.fallbackBaseURL = document => {
 
   if (document.URL === "about:blank" && document._defaultView &&
       document._defaultView._parent !== document._defaultView) {
-    return exports.documentBaseURL(document._defaultView._parent._document);
+    const parentDocument = implForWrapper(document._defaultView._parent._document);
+    return exports.documentBaseURL(parentDocument);
   }
 
-  return whatwgURL.parseURL(document.URL);
+  return document._URL;
 };
 
 exports.parseURLToResultingURLRecord = (url, document) => {
--- a/lib/jsdom/living/helpers/document-base-url.js
+++ b/lib/jsdom/living/helpers/document-base-url.js
@@ -28,7 +28,7 @@ exports.fallbackBaseURL = document => {
     return exports.documentBaseURL(document._defaultView._parent._document);
   }
 
-  return document._URL;
+  return whatwgURL.parseURL(document.URL);
 };
 
 exports.parseURLToResultingURLRecord = (url, document) => {
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -158,7 +158,10 @@ class DocumentImpl extends NodeImpl {
     }
 
     this._URL = parsed;
-    this._origin = whatwgURL.serializeURLOrigin(parsed);
+    this._parentOrigin = privateData.options.parentOrigin;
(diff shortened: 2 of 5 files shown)


domenic (approved):
Thanks very much! I cleaned up the tests a bit. Interestingly while Firefox passes them all, Chrome appears to have a bug where it generates `about:blank#blocked` for the invalid URL case.

Let's see if we can get the CI passing (now that it's green on master) and then I'll merge this.

domenic:
Hmm, it looks like the code changes you made break several other tests:

- html/semantics/document-metadata/the-base-element/base_about_blank.html
- html/semantics/document-metadata/the-base-element/base_srcdoc.html
- html/semantics/scripting-1/the-script-element/execution-timing/101.html (timeout)
- webstorage/document-domain.html (timeout)
- webstorage/event_basic.html
- webstorage/event_case_sensitive.html
- webstorage/event_no_duplicates.html
- xhr/open-url-about-blank-window.htm

Maybe one thing to consider is that such iframes should have an origin equal to their parent, even if they have a URL of `about:blank`.

Let me know if you can get those tests passing, and then we can consider merging this.

domenic (changes requested):
Needs to pass the existing test suite.

SimonMueller:
@domenic Thanks for having a look at it. When i merged upstream i saw that those tests are failing but I wasn't sure why and couldn't make them pass right away. Currently I am really busy at work but I will have a look at it as soon as I have time. It looks like a side effect from using iframes in the tests...

If you have another hint of where i could start searching in the code i would really appreciate it.

SimonMueller:
@domenic ok nevermind i just couldn't leave it like that. the tests are  green now. 🚀 Your hint with the origin was pointing me in the right direction...

The only thing i wasn't able to figure out was why `document._URL` is undefined in `lib/jsdom/living/helpers/document-base-url.js` at this point.

lib/jsdom/living/helpers/document-base-url.js
```diff
@@ -28,7 +28,7 @@ exports.fallbackBaseURL = document => {
     return exports.documentBaseURL(document._defaultView._parent._document);
   }
 
-  return document._URL;
+  return whatwgURL.parseURL(document.URL);
```
domenic, line 31: Is this change necessary?

domenic (commented):
Thanks, but it looks like this no longer merges with the master branch? Can you get it rebased so there are no conflicts?

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -158,7 +158,9 @@ class DocumentImpl extends NodeImpl {
     }
 
     this._URL = parsed;
-    this._origin = whatwgURL.serializeURLOrigin(parsed);
+    this._origin = whatwgURL.serializeURL(parsed) === "about:blank" && privateData.options.referrer ?
```
domenic, line 161: Instead of using referrer, which is only supposed to impact document.referrer, can you just add a new origin option?

That would also avoid all this parsing and serializing. We could just pass `origin: parentDoc._origin`.

domenic, line 161: There's no need to re-serialize. Just use `urlOption`.

SimonMueller: good point. done.

lib/jsdom/living/helpers/document-base-url.js
SimonMueller: yes it is required to pass the tests for the base element.

i debugged this and because of my change in the iframe implementation the code now reaches  `return exports.documentBaseURL(document._defaultView._parent._document);` on line 28 and as far as i can tell we only have the public api `document.URL` to work with if we access the parent document this way. thats why i cannot access `document._URL` anymore and i think this code cannot work in other scenarios aswell because of this.

if you know a better way to do this then i would be interested to see it :)

domenic:
Hmm seeing repeated failures on v14 with the cloneNode test. I might have somehow broken it on a previous commit... I'll debug locally before merging this.

domenic:
I can't reproduce locally; the test doesn't even take that long. Hmm. I'll just re-run CI I guess...

PR #3113, opened 2021-02-12
Substitute GitHub Actions CI for Travis CI

vegardbb:
I have made the following suggestion for a Github Actions CI configuration which mimics the Travis config as closely as possible.

All reports for my repo are available [here](https://github.com/vegardbb/jsdom/actions).


domenic:
Oh, this is very exciting; thanks for tackling this!

I wonder why GitHub is not showing the action results in this repo? I looked through the settings page and couldn't find anything that needed to be enabled...

.github/workflows/jsdom-ci.yml
```diff
@@ -0,0 +1,82 @@
...
+      uses: actions/setup-node@v2
+      with:
+        node-version: ${{ matrix.node-version }}
+        architecture: ${{ matrix.architecture }}
+    - name: Setup HOSTS file for Web Platform Test server
+      run: ./test/web-platform-tests/tests/wpt make-hosts-file | sudo tee -a /etc/hosts
+    - name: Install dependencies
+      run: yarn --frozen-lockfile
+    - name: Run tests
+      run: yarn lint && yarn test --retries 1
```
Sebmaster, line 82: I think it might be need to put lint in its own job so we don't have to run it for every version in our matrix since it should produce the same results anyways? Also would allow for some more parallelism.

.github/workflows/jsdom-ci.yml
```diff
@@ -0,0 +1,82 @@
...
+          node-version: '15'
+      - name: Setup HOSTS file for Web Platform Test server
+        run: ./test/web-platform-tests/tests/wpt make-hosts-file | sudo tee -a /etc/hosts
+      - name: Install dependencies
+        run: yarn --frozen-lockfile
+      - name: Run tests
+        run: yarn test-browser
+  build:
+    runs-on: ubuntu-latest
+    continue-on-error: true
```
Sebmaster, line 57: Is my understanding correct that this'll mark tests as passing even if the job fails? If so we probably want to set this to false throughout.

.github/workflows/jsdom-ci.yml
vegardbb: On it.

.github/workflows/jsdom-ci.yml
vegardbb: I don't think that is the case, but I'm still running the CI for the commit with this option turned on. Keep in mind that with the latest master commit, two of the suites in the matrix build (Node 14 and 15) are supposed to go red.

vegardbb:
I can now confirm that the CI runs succesfully on pull request pushes to master, too. @domenic @Sebmaster Should I remove travis.yml, too?

domenic:
Yes, `.travis.yml` seems useless indeed, but it seems bad that GitHub Actions isn't running on this PR. Do you have any idea why or what we could do to fix it?

vegardbb:
@domenic Here is a similar case from [StackOverflow](https://stackoverflow.com/questions/61989951/github-action-workflow-not-running)
> Just wanted to add a silly case scenario that happened to me. Whenever you add a new workflow, sometimes it may happen that workflow may not get into the running phase even after everything is specified correctly. Just make sure to make some changes in whatever branch you are running the action in after adding the action itself. Make sure to commit these changes as well. This will trigger the action to run and show up on the dashboard.

TimothyGu:
It's possible that GitHub doesn't run workflows from outside contributors, just in case the workflow can leak secrets.

vegardbb:
I suggest we merge this PR, commit something on another pull request, and check if the workflow runs with the merged config. If not, we can always do a hard reset and go look for another CI.

domenic:
Alright, let's merge this. I suspect @TimothyGu is right.

Then we need to get the tests passing on master through some combination of disabling and debugging, then we can start merging new work.

vegardbb:
IT'S RUNNING!

PR #3114, opened 2021-02-13
ci: use bahmutov/npm-install

ph-fritsche:
That action uses the cache to speed up installations in CI.

### DIFF
--- a/.github/workflows/jsdom-ci.yml
+++ b/.github/workflows/jsdom-ci.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Setup HOSTS file for Web Platform Test server
         run: ./test/web-platform-tests/tests/wpt make-hosts-file | sudo tee -a /etc/hosts
       - name: Install dependencies
-        run: yarn --frozen-lockfile
+        uses: bahmutov/npm-install@v1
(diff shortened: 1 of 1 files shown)


domenic:
Installation is not a very large percentage of our CI time, and I'd rather not give third party actions access to the CI runner unless the gain is large. So I think it's best not to merge this.

Sebmaster:
Maybe we could consider using the github-native cache action like this https://github.com/actions/cache/blob/main/examples.md#node---yarn, maybe including node_modules? However, this currently takes <20s, I'm not sure there's gonna be much to be gained there.

ph-fritsche:
If we don't speed up the CI time, it does in fact not make much of a difference.

Just posted it here to make you aware of this action.

The `lint` and `build-with-browser` checks are fast - 20s install is one third of the runtime.
(With the action I guess it will be ~5s)

The other checks time out. And report possible leaks. So they need to be fixed.

domenic:
Indeed, some point release of Node.js 14 made backward incompatible changes (and Node v15 made more), and someone needs to work on debugging the tests to find the root cause and fixing them. This PR does not do that.

Issue #3106, opened 2021-02-02
Failing tests in newer versions of Node


vegardbb:
### Basic info:

- **Node.js version: 14 and up (to my knowledge)**
- **jsdom version: v16 (master branch)**

### Minimal reproduction case

`git checkout` to master and `yarn test-wpt` run the web platform tests.

### Description

The following six tests fail both locally and in one of the Travis builds (the one which runs Node v15, [the four other succeeded](https://github.com/jsdom/jsdom/runs/1482105649))

```
  5063 passing (23m)
  606 pending
  7 failing
  1) web-platform-tests
       cors
         preflight-failure.htm:
     Error: done() called multiple times
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:93:5)
  2) web-platform-tests
       cors
         preflight-failure.htm:
     Uncaught Error: Parse Error: Expected HTTP/
      at Socket.socketOnData (node:_http_client:530:22)
      at Socket.emit (node:events:376:20)
      at addChunk (node:internal/streams/readable:304:12)
      at readableAddChunk (node:internal/streams/readable:279:9)
      at Socket.Readable.push (node:internal/streams/readable:218:10)
      at TCP.onStreamRead (node:internal/stream_base_commons:192:23)
  3) web-platform-tests
       cors
         preflight-failure.htm:
     Error: done() called multiple times
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:93:5)
  4) web-platform-tests
       dom/events
         [expected fail] scrolling/input-text-scroll-event-when-using-arrow-keys.html:
     Uncaught UnhandledPromiseRejection: This error originated either by throwing inside of an async function
without a catch block, or by rejecting a promise which was not handled with .catch().
The promise rejected with the reason "TypeError: element.scrollIntoView is not a function".
  
  5) web-platform-tests
       dom/events
         [expected fail] scrolling/input-text-scroll-event-when-using-arrow-keys.html:
     Error: done() called multiple times
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:93:5)
  6) web-platform-tests
       dom/events
         [expected fail] scrolling/input-text-scroll-event-when-using-arrow-keys.html:
     Error: done() called multiple times
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:93:5)
```

[...]

PR #3116, opened 2021-02-16
Ignore tests that do not run properly in Node v15

vegardbb:
# Related issues
See #3106 for details on each failing test. The code changes are extracted from [this](https://github.com/vegardbb/jsdom/tree/replace-fetch) branch, and put on a separate PR.

# What
Ignores (by timeout) the following Web Platform Test files
 * preflight-failure.htm
 * scrolling/input-text-scroll-event-when-using-arrow-keys.html
 * abort-during-headers-received.window.html

# Why
For newer versions of Node.js, namely 14 and 15, These tests do not run as they should, and thereby fail, whereas in v10 and v12, they do not. The maintainers may choose to investigate this further at a later time, but for now this is the fast track to get all the tests to pass.


vegardbb:
I would also advice removing the Travis CI action, as we have no `.travis.yml` config file in master anymore (but the Travis check is still popping up in pull request, somehow).

domenic:
Done, thanks.

I'd like to investigate these a bit more before merging, as the fact that `done()` is called multiple times indicates some problem with the jsdom test runner that is only manifesting in new Node versions. I'm scared of hiding legitimate bugs.

domenic:
I spent some time trying to reproduce [the errors seen on CI](https://github.com/jsdom/jsdom/runs/1893798294) locally using Node v15.8.0 on Windows 10. The results were not great:

- If I run a test in isolation using `yarn test-wpt --grep preflight-failure` I get "Error: Error: test harness should not timeout: cors/preflight-failure.htm"

- If I run a test in isolation using `yarn test-wpt --fgrep preflight-failure` I get "Uncaught Error: Parse Error: Expected HTTP/". Note that fgrep and grep should be the same here!

- If I run a test in isolation using `npx mocha test/web-platform-tests/run-wpts.js --fgrep preflight-failure` I get "Error: Error: test harness should not timeout: cors/preflight-failure.htm"

All of the above were consistent each time I tried them (about three times each). In no cases was I able to get "done() called multiple times."

Further attempts to try to get "done() called multiple times":

- If I run a subset of tests, e.g. using `yarn test-wpt --fgrep preflight-`, I get ` Uncaught Error: Parse Error: Expected HTTP/` under failure `1)`. Mocha says `15 failing`. However there is a single line at the bottom of the output that says `16) preflight-failure.htm`, hinting that maybe it would exhibit the pattern seen on CI (of one failure for parse error, one for done() multiple times) but it got clipped??

- If I run all of the tests, *then* I get "done() called multiple times". Unfortunately I can't really afford to debug the entire test suite; it takes too long to run.

Finally, I'll note that it's hard to imagine how done() could be called multiple times given that our WPT runner uses promise-returns, instead of ever explicitly calling `done()`. (This error is from Mocha, by the way, not from the WPT harness.) I guess it would be a Mocha bug then: for example, maybe Mocha internally calls `done()` when an uncaught exception reaches `process.on('uncaughtException')`, and it also calls `done()` when the test's returned promise rejects, without accounting for the fact that both of these could happen?

So my conclusion is that something's deeply funky either in the innards of Mocha, or in how we're interfacing with it. My best guess is that Mocha does not deal well with tests that both throw uncaught global exceptions and return rejected promises, and that this manifests in strange flaky behavior that is inconsistent depending on how you run the tests. With this hypothesis, it's the upgrade to Node.js 15 that caused more global uncaught exceptions (e.g. HTTP parsing errors, or unhandled promise rejections).

I'd appreciate any further investigation that folks were able to do to confirm or deny this hypothesis. In particular, maybe if you are able to reproduce "done() called multiple times" using a single test, you'd be in a better position to debug than I am, where it only reproduces with the full test suite.

Otherwise, I'll plan to merge this PR this weekend.

test/web-platform-tests/to-run.yaml
```diff
@@ -33,6 +33,7 @@ cors-safelisted-request-header.any.html: [fail, Depends on fetch]
 image-tainting-in-cross-origin-iframe.sub.html: [timeout, Unknown]
 origin.htm: [timeout, https://github.com/jsdom/jsdom/issues/1833]
 preflight-cache.htm: [timeout, Cache should probably be implemented for simple requests before]
+preflight-failure.htm: [timeout, Node throws an error on code 304, done() is called multiple times]
```
ExE-Boss, line 36: This should probably support a value like: `timeout‑node15` or something.

test/web-platform-tests/to-run.yaml
```diff
@@ -163,7 +164,9 @@ document-level-touchmove-event-listener-passive-by-default.html: [timeout, Uses
 event-global-extra.window.html: [timeout, Not implemented]
 event-global.html: [fail, Not implemented]
 relatedTarget.window.html: [fail, Unknown]
-scrolling/**: [fail, Scroll methods not implemented]
+scrolling/input-text-scroll-event-when-using-arrow-keys.html: [timeout, Fires UnhandledPromiseRejectionWarning]
+scrolling/overscroll-**.html: [fail, Scroll methods not implemented]
+scrolling/scrollend-**.html: [fail, Scroll methods not implemented]
```
ExE-Boss, line 169: This should probably work with:
```suggestion
scrolling/**: [fail, Scroll methods not implemented]
```

since `%Array.prototype.find%` returns the first matching value.

test/web-platform-tests/to-run.yaml
vegardbb: Would that make the test run for Node v14 and older?

domenic: I think that's the implication. IMO it's fine as-is though. Getting improved test coverage on older node versions is not worth the burden I think.

vegardbb:
The 'stable' build was green as of the 8th of August. Back then it ran Node v14.7. In case somebody wanted to scroll through Node's changelogs, there's your version timeline.

@domenic The test suite is on version 7.1.1 of Mocha, right? I snooped around on their issues page, and [this search](https://github.com/mochajs/mocha/issues?q=uncaught+promise) led me to discover a PR that was merged in May last year for version 7.2.0, and another regarding "uncaught promise rejections" for v8.2, merged last November.

domenic:
Thanks @vegardbb for the suggestion! I upgraded us to the latest Mocha (and accidentally pushed to master instead of a PR, oops) and that fixes several of the issues: https://github.com/jsdom/jsdom/runs/1999352603?check_suite_focus=true .

I'll tweak and rebase this PR on the new master.

PR #3125, opened 2021-02-23
Update WPT

Sebmaster:
This requires upgrading to python3 which we get by just running ubuntu20 explicitly in CI.


Sebmaster:
Got the tests running. A significant number seem to be failing, we gotta go through em all and skip/mark them :(

domenic:
I've made a good amount of progress but something seems to be really slowing down the test runner, both locally and on CI, around the `html/semantics/scripting-1/the-script-element` section. It seems worse than before, and I'm not sure why.

I'm willing to use the fail-slow hammer on some of them, but even the successful ones seem to be very slow. I think maybe something about our timeout handling changed and WPT is no longer timing out after 10 seconds, but instead we're hitting Mocha's 70 second timeout?

domenic:
Yeah, it was a problem in recent changes to testharness.js. https://github.com/web-platform-tests/wpt/pull/27824

PR #3136, opened 2021-03-09
Use `IterableWeakSet` for `NodeIterator` tracking when supported

ExE-Boss:
When `WeakRef`s are supported (**Node ≥14.6.0**), then it’s possible to keep track of active `NodeIterator`s without leaking memory.

---

**Refs:** <https://github.com/tc39/proposal-weakrefs/issues/17#issue-292572565>

### DIFF
--- a/lib/jsdom/living/helpers/iterable-weak-set.js
+++ b/lib/jsdom/living/helpers/iterable-weak-set.js
@@ -1,0 +1,57 @@
+"use strict";
+
+// This is a static function to improve memory usage when multiple
+// `IterableWeakSet` instances are created, so that each `FinalizationRegistry`
+// uses the same shared function:
+function cleanup({ ref, set }) {
+  set.delete(ref);
+}
+
+// An iterable WeakSet implementation inspired by the iterable WeakMap
+// example code in the WeakRefs specification:
+//
+// @see https://github.com/tc39/proposal-weakrefs#iterable-weakmaps
+module.exports = class IterableWeakSet {
+  constructor() {
+    this._refSet = new Set();
+    this._refMap = new WeakMap();
+    this._finalizationRegistry = new FinalizationRegistry(cleanup);
+  }
+
+  add(value) {
+    if (!this._refMap.has(value)) {
+      const ref = new WeakRef(value);
+      this._refMap.set(value, ref);
+      this._refSet.add(ref);
+      this._finalizationRegistry.register(value, { ref, set: this._refSet }, ref);
+    }
+
+    return this;
+  }
+
+  delete(value) {
+    const ref = this._refMap.get(value);
+    if (!ref) {
+      return false;
+    }
+
+    this._refMap.delete(value);
+    this._refSet.delete(ref);
+    this._finalizationRegistry.unregister(ref);
+    return true;
+  }
+
+  has(value) {
+    return this._refMap.has(value);
+  }
+
+  * [Symbol.iterator]() {
+    for (const ref of this._refSet) {
+      const value = ref.deref();
+      if (value === undefined) {
+        continue;
+      }
+      yield value;
+    }
+  }
+};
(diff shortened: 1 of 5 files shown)


lib/jsdom/living/helpers/iterable-weak-set.js
```diff
@@ -0,0 +1,56 @@
+"use strict";
+
+function cleanup({ ref, set }) {
```
domenic, line 3: I recall it being important for these not to be closures for... reasons. Could you add a comment documenting these reasons?

lib/jsdom/living/helpers/iterable-weak-set.js
```diff
@@ -0,0 +1,56 @@
...
+  set.delete(ref);
+}
+
+/**
```
domenic, line 7: No jsdoc please.

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -169,17 +170,22 @@ class DocumentImpl extends NodeImpl {
...
-                                    Number(privateData.options.concurrentNodeIterators);
-
-    if (isNaN(this._workingNodeIteratorsMax)) {
-      throw new TypeError("The 'concurrentNodeIterators' option must be a Number");
-    }
+    if (hasWeakRefs) {
+      this._workingNodeIterators = new IterableWeakSet();
+    } else {
+      this._workingNodeIterators = [];
+      this._workingNodeIteratorsMax =
```
domenic, line 177: It looks like we never ported over the concurrentNodeIterators option to the "new" API: https://github.com/jsdom/jsdom/blob/69b0205ed4cdafe94f9ffea519e00dcc67e17f1f/lib/api.js#L285 . Since nobody has noticed, let's hard-code it to 10, and remove the validation code?

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -776,12 +782,19 @@ class DocumentImpl extends NodeImpl {
   }
 
   createNodeIterator(root, whatToShow, filter) {
-    const nodeIterator = NodeIterator.createImpl(this._globalObject, [], { root, whatToShow, filter });
+    const nodeIterator = NodeIterator.createImpl(this._globalObject, [], {
+      root, whatToShow, filter,
+      workingNodeIteratorsMax: hasWeakRefs ? null : this._workingNodeIteratorsMax
```
domenic, line 787: I don't think we need to check hasWeakRefs here since it's checked in NodeIterator-impl. (Alternately, you could remove the checks there, and check if `_workingNodeIteratorsMax` is null.)

domenic (commented):
This is great; thanks for working on this!

lib/jsdom/living/helpers/iterable-weak-set.js
ExE-Boss: I believe that the reason is memory usage, since the `cleanup` static method can be shared between all instances.

lib/jsdom/living/helpers/iterable-weak-set.js
ExE-Boss: Where was the decision to stop using **JSDoc** made?

domenic: We've asked several times before; this is not news to you.

ExE-Boss: I know, but I’m genuinely curious, especially because a good chunk of the **JSDoc** in this repository was contributed by _you_ according to `git blame`.

lib/jsdom/living/helpers/iterable-weak-set.js
```diff
@@ -0,0 +1,57 @@
+"use strict";
+
+// This is a static function to improve memory usage when multiple
+// `IterableWeakSet` instances are created, so that each `FinalizationRegistry`
+// uses the same shared function:
```
domenic, line 5: If this is the real reason then I'll just convert it to a closure as this is a fully general argument against any use of functions inside functions and it's not worth departing from the usual readable style.

PR #3137, opened 2021-03-09
Fix failing `window‑properties.html` tests

ExE-Boss:
Most of the failing tests in `test/web‑platform‑tests/to‑upstream/html/browsers/the‑window‑object/window‑properties‑dont‑upstream.html` have already been fixed at some point.

---

This fixes most of the remainder and adds `incorrectlyImplementedAsDataProperties` to the test to track properties that are incorrectly implemented as data properties instead of as accessors.

## Depends on:
- [ ] <https://github.com/jsdom/jsdom/pull/3139>


lib/jsdom/browser/Window.js
```diff
@@ -216,6 +216,22 @@ function setupWindow(windowInstance, { runScripts }) {
   windowInstance._globalObject = windowInstance;
 }
 
+function makeReplaceablePropertyDescriptor(property) {
+  return {
+    set: Object.getOwnPropertyDescriptor({
+      // eslint-disable-next-line accessor-pairs
+      set [property](value) {
```
ExE-Boss, line 223: This ensures that the setter has the correct `name` property.

test/web-platform-tests/to-run.yaml
```diff
@@ -848,7 +848,6 @@ execution-timing/074.html: [fail, Unknown]
 execution-timing/076.html: [fail, Unknown]
 execution-timing/077.html: [fail, Unknown]
 execution-timing/078.html: [fail, Unknown]
-execution-timing/079.html: [fail, Unknown]
```
ExE-Boss, line 851: This test was failing because it depends on setting `window.location`, which was previously unsupported.

test/web-platform-tests/to-upstream/html/browsers/the-window-object/window-properties-dont-upstream.html
```diff
@@ -21,23 +21,36 @@
...
-  assert_equals(typeof pd.set, hasSetter ? "function" : "undefined");
-  */
-  assert_equals(pd.enumerable, Enumerable);
-  assert_equals(pd.configurable, Configurable);
+  assert_equals(typeof pd.set, hasSetter ? "function" : "undefined", "typeof pd.set");
+  assert_equals(pd.enumerable, Enumerable, "pd.enumerable");
+  assert_equals(pd.configurable, Configurable, "pd.configurable");
 }
 
+var incorrectlyImplementedAsDataProperties = [
```
TimothyGu, line 38: Could you leave a comment here saying that this is jsdom-specific?

TimothyGu (commented):
The code looks good but can you take a look at the CI failures?

ExE-Boss:
> The code looks good but can you take a look at the CI failures?

Some of the failures (`dom/nodes/Node‑cloneNode.html`) are unrelated flaky tests.

---

Others (“Window: EventTarget methods” and “Window: Value Properties”) only work in **Node 12+**, which can’t be described in the current format, and depends on <https://github.com/jsdom/jsdom/pull/3139>.

test/web-platform-tests/to-upstream/html/browsers/the-window-object/window-properties-dont-upstream.html
ExE-Boss: I’d rather do: <https://github.com/jsdom/jsdom/pull/3139>

lib/jsdom/browser/Window.js
domenic: Can you instead define the function and then update its name with `Object.defineProperty`? This is too clever for future readers to understand.

lib/jsdom/browser/Window.js
```diff
@@ -375,6 +391,10 @@ function Window(options) {
     get location() {
       return idlUtils.wrapperForImpl(idlUtils.implForWrapper(window._document)._location);
     },
+    // [PutForwards=href]:
+    set location(value) {
+      Reflect.set((this || window).location, "href", value);
```
domenic, line 396: Why `this || window`? All other nearby methods use `window` and it seems best to stay consistent.

Why `Reflect.set()` instead of `window.location.href = value`?

domenic (changes requested):

lib/jsdom/browser/Window.js
ExE-Boss: Because that’s what matches what **WebIDL** does.

See: <https://heycam.github.io/webidl/#ref-for-PutForwards①⑥> and <https://github.com/jsdom/webidl2js/blob/e025c8487897c65240e0416805aec0183251c84d/lib/constructs/attribute.js#L157-L158>

domenic: Are there observable differences? If so, I'd expect a test. If not, I'd expect us to conform to the prevailing style.

domenic (approved):

romellem:
Sorry to resurrect an old thread, but if I previously had tests that _did_ change these properties, what would be the recommended way to go about doing that with JSDOM version 21 and beyond?

For example, if I had a jest test that checked to see that `window.location.reload()` was called when some button was clicked, previously I could

```js
jest.spyOn(window.location, 'reload').mockImplementation(jest.fn());
```

However, this now throws _"TypeError: Cannot assign to read only property 'reload' of object '[object Location]'"_

Any advice for what you recommend to overcome this new restriction?

domenic:
I'd suggest figuring out how to change those properties in a web browser, and then doing the same in jsdom.

Issue #3133, opened 2021-03-08
jsdom 16.5.0 break support for customElements.get


fdesforges:
### Basic info:

- Node.js version: v14.13.0 
- jsdom version: 16.5.0 

Issue has been reproduce using last version of jest.

### Minimal reproduction case

(minimal setup using `npm init -y` in an empty directory and just adding jsdom as dependency in the project)

```js
const {JSDOM} = require('jdsom')

const dom = new JSDOM(`<body>
<script>
class CustomElement extends HTMLElement {}
customElements.define('my-ce', CustomElement )
console.log(customElements.get('my-ce'))
</script>
</body>`, {runScripts: 'dangerously'})
```

with 16.4.0 ==> works as expected
```
$ node index.js
[class CustomElement extends HTMLElement]
```

with 16.5.0 ==>  Does not work as expected
```
$ node index.js
undefined
```

---

The origin come from this refactoring : https://github.com/jsdom/jsdom/commit/ff69a75706eac5ef90d9eb46aedd90a14e29984c

In the file [CustomElementRegistry.html line 184](https://github.com/jsdom/jsdom/blame/2d82763248296c5f118d6fd890317f5db5de1635/lib/jsdom/living/custom-elements/CustomElementRegistry-impl.js#184) the property `ctor` is removed :

```js
    const definition = {
      name,
      localName,
      // removed line  ctor,
      constructor,
      objectReference: ctor,
      observedAttributes,
      lifecycleCallbacks,
      disableShadow,
```
So the method get is not working anymore because it check the `ctor` property existence : (https://github.com/jsdom/jsdom/blame/2d82763248296c5f118d6fd890317f5db5de1635/lib/jsdom/living/custom-elements/CustomElementRegistry-impl.js#L217)

```js
  get(name) {
    const definition = this._customElementDefinitions.find(entry => entry.name === name);
    return definition && definition.ctor;
  }
```

domenic:
Fascinating... It's quite sad that our test suite didn't catch this.

ExE-Boss:
@domenic I’m updating the test suite to be able to catch this in <https://github.com/jsdom/jsdom/pull/2956>.

PR #3139, opened 2021-03-11
Handle failing subtests in web‑platform‑tests

ExE-Boss:
This makes it possible to migrate from using `dont‑upstream` tests, which tend to become outdated because the commented‑out code never runs, which leads to regressions like <https://github.com/jsdom/jsdom/issues/3133>.

Note that subtests don’t currently support match patterns.

---

This also migrates `window‑properties‑dont‑upstream.html` to use the new format.


test/web-platform-tests/utils.js
```diff
@@ -0,0 +1,22 @@
...
+
+const nodeMajor = Number(process.versions.node.split(".", 1)[0]);
+const hasNode10 = nodeMajor >= 10;
+const hasNode11 = nodeMajor >= 11;
+const hasNode12 = nodeMajor >= 12;
+const hasCanvas = Boolean(Canvas);
+
+exports.resolveReason = reason => {
+  return {
+    shouldSkip:
```
domenic, line 12: I like this function. But I think it reveals a chance for improvement. Since shouldSkip and expectFail are mutually exclusive, instead it can return one of "run", "skip", "expect-fail".

test/web-platform-tests/run-single-wpt.js
```diff
@@ -162,12 +181,8 @@ function createJSDOM(urlPrefix, testPath, expectFail) {
           };
 
           window.add_result_callback(test => {
-            if (test.status === 1) {
-              errors.push(`Failed in "${test.name}": \n${test.message}\n\n${test.stack}`);
-            } else if (test.status === 2) {
-              errors.push(`Timeout in "${test.name}": \n${test.message}\n\n${test.stack}`);
-            } else if (test.status === 3) {
-              errors.push(`Uncompleted test "${test.name}": \n${test.message}\n\n${test.stack}`);
+            if (test.status === test.FAIL || test.status === test.TIMEOUT || test.status === test.NOTRUN) {
```
domenic, line 184: I think we can just do `test.status !== test.PASS`

test/web-platform-tests/run-single-wpt.js
```diff
@@ -177,28 +192,58 @@ function createJSDOM(urlPrefix, testPath, expectFail) {
...
             } else if (errors.length && !expectFail) {
-              reject(new Error(`${errors.length} errors in test:\n\n${errors.join("\n")}`));
-            } else {
+              reject(new Error(`${errors.length}/${tests.length} errors in test:\n\n${errors.join("\n\n")}`));
+            } else if (typeof expectFail !== "object") {
               resolve();
+            } else {
+              const unexpectedErrors = [];
+              for (const test of tests) {
+                const testFailed =
```
domenic, line 221: As above

test/web-platform-tests/run-single-wpt.js
```diff
@@ -177,28 +192,58 @@ function createJSDOM(urlPrefix, testPath, expectFail) {
...
+                  test.status === test.FAIL ||
+                  test.status === test.TIMEOUT ||
+                  test.status === test.NOTRUN;
+                const data = expectFail[test.name];
+                let reason;
+
+                if (Array.isArray(data)) {
```
domenic, line 228: Won't it always be an array?

domenic (commented):
This is extremely exciting; thanks for working on this.

test/web-platform-tests/run-single-wpt.js
ExE-Boss: There is the `PRECONDITION_FAILED` status, which is reported when a normative optional feature is not implemented.

test/web-platform-tests/run-single-wpt.js
ExE-Boss: Not in the new format, where it will be a plain object.

ExE-Boss: Only if there’s a `to‑run.yaml` entry for this subtest, otherwise it will be `undefined`.

test/web-platform-tests/run-wpts.js
```diff
@@ -61,29 +63,28 @@ describe("web-platform-tests", () => {
...
+            case "skip":
+              specify.skip(`[${reason}] ${testFile}`);
+              break;
+            case "expect-fail": {
+              const failReason = reason !== "fail" ? `: ${reason}` : "";
+              runSingleWPT(testFilePath, `[expected fail${failReason}] ${testFile}`, true);
+              break;
+            }
+            default:
+              runSingleWPT(testFilePath, testFile, data || false);
```
domenic, line 87: Can we have a prefix like `[x/y] ` for ones where we expect some failures? If you look at the logs in https://github.com/jsdom/jsdom/pull/3139/checks?check_run_id=2102685649 they read `✓ window-properties.https.html` which is a bit misleading.

test/web-platform-tests/run-wpts.js
```diff
@@ -61,29 +63,28 @@ describe("web-platform-tests", () => {
...
-
-          if (shouldSkip) {
-            specify.skip(`[${reason}] ${testFile}`);
-          } else if (expectFail) {
-            const failReason = reason !== "fail" ? `: ${reason}` : "";
-            runSingleWPT(testFilePath, `[expected fail${failReason}] ${testFile}`, expectFail);
-          } else {
-            runSingleWPT(testFilePath, testFile, expectFail);
+          switch (resolveReason(reason)) {
+            case "skip":
```
domenic, line 78: Nit: braces on all `case`/`default`

test/web-platform-tests/run-wpts.js
```diff
@@ -61,29 +63,28 @@ describe("web-platform-tests", () => {
           });
 
           const testFile = testFilePath.slice((toRunDoc.DIR + "/").length);
-          let reason;
+          let reason, data;
 
           if (matchingPattern) {
             reason = toRunDoc[matchingPattern][0];
+            if (!Array.isArray(toRunDoc[matchingPattern])) {
```
domenic, line 70: Can you add a comment explaining what the array vs. non-array case is?

domenic (changes requested):
Just a few more small things

test/web-platform-tests/to-run.yaml
```diff
@@ -396,6 +396,120 @@ DIR: html/browsers/offline/browser-state
...
+named-access-on-the-window-object/window-named-properties.html: [timeout, Unknown]
+navigate-to-about-blank-while-initial-load-pending.html: [fail, Depends on window.open()]
+noopener-noreferrer-BarProp.window.html: [fail, Depends on window.open()]
+noopener-noreferrer-sizing.window.html: [fail, Depends on window.open()]
+proxy-getOwnPropertyDescriptor.html:
+  'Window target, no trap, "name" attribute': [fail, Incorrectly implemented as a data property]
+security-window/window-security.https.html: [fail, Exoticness of Window not implemented]
+self-et-al.window.html: [fail, Depends on window.open() and window.closed]
+window-indexed-properties-strict.html: [flaky, "Errors in `process.nextTick` callback:
+Uncaught TypeError: this[i].close is not a function
```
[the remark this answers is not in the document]
ExE-Boss: This is just a syntax highlighting error on **GitHub**.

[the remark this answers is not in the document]
domenic: Makes sense; it's still a bit hard to read though. Consider using YAML's multiline strings if that's an option, or different indenting, or various other possibilities.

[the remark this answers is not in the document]
ExE-Boss: **YAML**’s block multiline strings don’t work in flow sequences (`[foo, bar]`)

test/web-platform-tests/run-wpts.js
```diff
@@ -61,29 +63,48 @@ describe("web-platform-tests", () => {
...
+            case "expect-fail": {
+              const failReason = reason !== "fail" ? `: ${reason}` : "";
+              runSingleWPT(testFilePath, `[expected fail${failReason}] ${testFile}`, true);
+              break;
+            }
+
+            default: {
+              let failCount = 0;
+              if (data) {
+                failCount = Object.values(data).reduce((count, innerReason) => {
```
domenic, line 96: How about `Object.values(data).filter(innerReason => resolveReason(innerReason) === "expect-fail").length` ?

domenic (approved):
LGTM with nit

test/web-platform-tests/run-wpts.js
ExE-Boss: I guess that that reads nicer, but it requires allocating an extra intermediate array, which is one of the main reason I’m looking forward to when [the **Iterator Helpers** proposal](https://github.com/tc39/proposal-iterator-helpers) reaches stage 4 and gets implemented in all engines.

TimothyGu (approved):
Thanks for this! Would it be possible to go through the other `dont-upstream` tests we have and convert them to the new format? Doesn't have to be in the same PR of coursel

ExE-Boss:
@TimothyGu
I have began work on that in <https://github.com/jsdom/jsdom/pull/3137> and <https://github.com/jsdom/jsdom/pull/3140>.

---

After this is merged, I’ll open a PR to convert `html/dom/historical-dont-upstream.html`.

PR #3143, opened 2021-03-20
Fix radio/checkbox to not fire events when disconnected

domenic:


domenic:
CI failing because there are contradictory WPTs, so this is blocked on https://github.com/web-platform-tests/wpt/pull/28159 and a test roll.

PR #3144, opened 2021-03-20
Fix Text and Comment constructors

domenic:
They would not properly assign an owner document

### DIFF
--- a/lib/jsdom/living/nodes/Comment-impl.js
+++ b/lib/jsdom/living/nodes/Comment-impl.js
@@ -1,13 +1,13 @@
 "use strict";
-
 const CharacterDataImpl = require("./CharacterData-impl").implementation;
-
+const idlUtils = require("../generated/utils");
 const NODE_TYPE = require("../node-type");
 
 class CommentImpl extends CharacterDataImpl {
   constructor(globalObject, args, privateData) {
     super(globalObject, args, {
       data: args[0],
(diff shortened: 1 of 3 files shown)


lib/jsdom/living/nodes/Comment-impl.js
```diff
@@ -1,13 +1,13 @@
...
 const CharacterDataImpl = require("./CharacterData-impl").implementation;
-
+const idlUtils = require("../generated/utils");
 const NODE_TYPE = require("../node-type");
 
 class CommentImpl extends CharacterDataImpl {
   constructor(globalObject, args, privateData) {
     super(globalObject, args, {
       data: args[0],
+      ownerDocument: idlUtils.implForWrapper(globalObject._document),
```
TimothyGu, line 10: Would it be a good idea to move this to `CharacterDataImpl`'s constructor?

TimothyGu (approved):

lib/jsdom/living/nodes/Comment-impl.js
domenic: I'm not so sure. I think we have two possible approaches for setting ownerDocument:

- Pass it and set it explicitly when the spec says to do so. This is what this PR does.
- Set a default in NodeImpl, and then remember to override it where necessary (e.g. in createElement or friends)

Between these two, I tend to prefer the former. And something in between, where we set a default in CharacterDataImpl but not NodeImpl, seems not so good.

PR #3146, opened 2021-03-20
Fix checkbox indeterminate interaction with clicks

domenic:


lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -167,6 +167,8 @@ class HTMLInputElementImpl extends HTMLElementImpl {
   _legacyPreActivationBehavior() {
     if (this.type === "checkbox") {
       this.checked = !this.checked;
+      this._legacyActivationBehaviorPreviousIndeterminateState = this.indeterminate;
```
TimothyGu, line 170: Can we have this property "declared" in the constructor for better state tracking?

TimothyGu (approved):

PR #3155, opened 2021-03-23
Run all tests under `html/dom`

ExE-Boss:
This also migrates `html/dom/historical‑dont‑upstream.html` to use the new format that was introduced in <https://github.com/jsdom/jsdom/pull/3139>.


domenic:
Can you separate these changes? Getting rid of dont-upstream.html tests is valuable, but I want to evaluate that separately from whether enabling more tests in html/dom is a good idea, and especially whether we want to enable partially-failing tests since those create a higher maintenance burden.

domenic:
All tests under html/dom are now enabled.

Issue #3157, opened 2021-03-24
Convert `to‑run.yaml` to use the new subtest format


ExE-Boss:
Now that <https://github.com/jsdom/jsdom/pull/3139> is implemented, `to‑run.yaml` entries with more than 1 subtest that don’t have harness errors should be migrated to the new format.

---

The `test/web‑platform‑tests/run‑single‑wpt.js` diff that I’m using to detect these tests:
```diff
 if (errors.length === 0 && expectFail) {
   reject(new Error(unexpectedPassingTestMessage));
 } else if ((errors.length === 1 && (tests.length === 1 || harnessFail)) && !expectFail) {
   reject(new Error(errors[0]));
 } else if (errors.length && !expectFail) {
   reject(new Error(`${errors.length}/${tests.length} errors in test:\n\n${errors.join("\n\n")}`));
-} else if (typeof expectFail !== "object") {
+} else if (typeof expectFail !== "object" &&
+  (errors.length === 0 || harnessFail || tests.length === errors.length || tests.length === 1)) {
   resolve();
 } else {
```

## `dont‑upstream` tests:
- [ ] <code>[custom‑elements/CustomElementRegistry‑dont‑upstream.html](https://github.com/jsdom/jsdom/blob/03c3dd72536110b525a6616f117dee33b2e4673e/test/web-platform-tests/to-upstream/custom-elements/CustomElementRegistry-dont-upstream.html)</code> (<https://github.com/jsdom/jsdom/pull/3140>)
- [ ] <code>[domparsing/DOMParser‑dont‑upstream.html](https://github.com/jsdom/jsdom/blob/03c3dd72536110b525a6616f117dee33b2e4673e/test/web-platform-tests/to-upstream/domparsing/DOMParser-dont-upstream.html)</code>
- [ ] <code>[hr‑time/performance‑tojson‑dont‑upstream.html](https://github.com/jsdom/jsdom/blob/03c3dd72536110b525a6616f117dee33b2e4673e/test/web-platform-tests/to-upstream/hr-time/performance-tojson-dont-upstream.html)</code>
	- [ ] Depends on: <https://github.com/web-platform-tests/wpt/pull/28202>
- [ ] <code>[hr‑time/timeOrigin‑dont‑upstream.html](https://github.com/jsdom/jsdom/blob/03c3dd72536110b525a6616f117dee33b2e4673e/test/web-platform-tests/to-upstream/hr-time/timeOrigin-dont-upstream.html)</code>
[...]

PR #3163, opened 2021-03-28
Convert WPTs in `FileAPI` directory to use the new format

ExE-Boss:
Part of: <https://github.com/jsdom/jsdom/issues/3157>


domenic:
Let's discuss further in #3157.

domenic (approved):

PR #3173, opened 2021-04-09
Fix MutationObserver infinite loop bugs

### LINKED ISSUE #3096: MutationObserver enters infinite loop if new elements are observed in callback
### Basic info:

- **Node.js version:** 12.18.3
- **jsdom version:** 16.4.0

### Minimal reproduction case

```js

const dom = new JSDOM('<html><head></head><body></body></html>');
const { document, MutationObserver } = dom.window;

let callbackCount = 0;
const div1 = document.createElement('div');
document.body.appendChild(div1);
const observer = new MutationObserver(() => {
  callbackCount += 1;
  // prevents infinite loop
  if (callbackCount >= 10)
    return;
  const div2 = document.createElement('div');
  document.body.appendChild(div2);
  observer.observe(div2, { attributes: true });
});
observer.observe(div1, { attributes: true });
div1.setAttribute('data-test', 'test');

// flush queue
new Promise(resolve => resolve()).then(() => {
    console.log('callback calls: ' + callbackCount);
});
```

**Expected:** `callback calls: 1`
**Actual:** `callback calls: 10`

Outputs `callback calls: 1` as expected, see
https://jsbin.com/zaxuhifoye/edit?js,console

### LINKED ISSUE #3172: Mutation observer enters infinite loop if the same element is observed in callback after disconnect
### Basic info:

- **Node.js version:** 12.18.3
- **jsdom version:** 16.5.2

### Minimal reproduction case

```js
let callbackCount = 0;
const div1 = document.createElement('div');
document.body.appendChild(div1);
const observer = new MutationObserver(() => {
    callbackCount += 1;
    // prevents infinite loop
    if (callbackCount >= 10) {
        return;
    }
    observer.disconnect();
    observer.observe(div1, {attributes: true});
});
observer.observe(div1, { attributes: true });
div1.setAttribute('data-test', 'test');

// flush queue
new Promise(resolve => resolve()).then(() => {
    console.log('callback calls: ' + callbackCount);
});
```
Expected: `callback calls: 1`
Actual: `callback calls: 10`

Outputs `callback calls: 1`

https://jsbin.com/hepidugamo/edit?js,console

This is similar to #3096 however not quite the same, as this refers observing the same element and not a different element.

TimothyGu:
The code was misindented relative to the spec.

Fixes #3096. Fixes #3172.

### DIFF
--- a/lib/jsdom/living/helpers/mutation-observers.js
+++ b/lib/jsdom/living/helpers/mutation-observers.js
@@ -152,21 +152,21 @@ function notifyMutationObservers() {
       node._registeredObserverList = node._registeredObserverList.filter(registeredObserver => {
         return registeredObserver.source !== mo;
       });
+    }
 
-      if (records.length) {
-        try {
-          const moWrapper = idlUtils.wrapperForImpl(mo);
-          mo._callback.call(
-            moWrapper,
-            records.map(idlUtils.wrapperForImpl),
-            moWrapper
-          );
-        } catch (e) {
-          const { target } = records[0];
-          const window = target._ownerDocument._defaultView;
+    if (records.length > 0) {
+      try {
+        const moWrapper = idlUtils.wrapperForImpl(mo);
+        mo._callback.call(
+          moWrapper,
+          records.map(idlUtils.wrapperForImpl),
+          moWrapper
+        );
+      } catch (e) {
+        const { target } = records[0];
+        const window = target._ownerDocument._defaultView;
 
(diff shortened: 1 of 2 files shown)


Sebmaster (approved):

domenic:
Great catch, thank you. But we really should have a test :)

TimothyGu:
Working on it :)

domenic (approved):

Issue #2751, opened 2019-12-14
Guidance about mock implementation for layout related features


pmdartus:
## Observations

There is a lot of interest in providing mock implementations of layout related features in jsdom especially around scroll (#1422, #2626, #2332, ...). Based on the number of reactions on the different issues, 2 of the related scroll APIs are part of the top 10 most requested features.

## Goal

The goal of this issue is to capture jsdom contributors and consumers' points of view on this subject and reach a consensus on how to address such features in a coherent way.

pmdartus:
I think there are 2 types of layout related APIs: the APIs that can be safely mocked by either doing a no-op or returning a static value and the other ones (ok, it's really vague 🤣). jsdom already provides a `pretendToBeVisual` flag that can be turned ON to pretend that jsdom is rendering and displaying content. I think it would be coherent to use this flag to provide mock implementations for the simple APIs.

Scroll related APIs falls into the simple API bucket:
- when `pretendToBeVisual` is `false`: make the APIs throw an error with a message stating that `pretendToBeVisual` needs to be set to `true` to use such API.
- when `pretendToBeVisual` is `true`: do a no-op.

I am uncertain if we should use the same approach for APIs that jsdom already exposes (eg. window.innerWith, window.innerHeight, `Element.getBoundingClientRect()`, ...) because it would be a breaking change.

For other layout related APIs, that can't be simply replaced by a no-op or a static return value (eg. `HTMLElement.innerText`, `DocumentOrShadowRoot.elementFromPoint()`, `IntersectionObserver`), we don't expose them and require the consumer to mock such APIs.

mentierd:
@pmdartus I believe a lot of people are looking for a solution to mocking those things you mentioned above, as well as others.  I'm assuming that most people using jest are unknowingly introducing memory leaks into their tests by trying to mock such prototypes.  

One solution that I believe would be most beneficial is if there was a callback property, similar to beforeParse that would allow you to modify the window object (including prototypes) before the actual creation of the context. Perhaps the params would be the actual window object itself, as well as the utils that include the implSymbol and symbol un-wrapper function.

domenic:
> One solution that I believe would be most beneficial is if there was a callback property, similar to beforeParse that would allow you to modify the window object (including prototypes) before the actual creation of the context.

I'm not sure what the difference is between this and `beforeParse`, especially in terms of what they allow you to accomplish?

---

[...]

PR #3187, opened 2021-05-10
Add x, y parameters to getBoundingClientRect

### LINKED ISSUE #3188: getBoundingClientRect is missing x, y parameters
jsdom mocks `getBoundingClientRect` to return a DOMRect with all values set to 0:

```javascript
document.body.getBoundingClientRect()
// => { bottom: 0, height: 0, left: 0, right: 0, top: 0, width: 0 }
```

However, it's missing the `x` and `y` parameters, which are defined in the MDN for [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect) and supported in all modern browsers

I've made [a pull request](https://github.com/jsdom/jsdom/pull/3187) to add the parameters

eiko:
`getBoundingClientRect` returns a [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect) which is supposed to have an `x` and `y` property. This change makes `getBoundingClientRect` behave more reliably for code which expects its properties to have defined numerical values.

fixes #3188


domenic:
Although the general plan for these not-very-good mocks is still up in the air (see #2751), this seems like a reasonable improvement assuming CI passes, so I'll merge it pending further discussion on that point.

domenic:
Looks like this requires a few test updates.

eiko:
@domenic sorry for the delay on this- I had trouble getting tests to run locally.

I updated the "build" tests so they should pass now, but I couldn't figure out why "build-with-canvas" was failing with a segmentation fault. Let me know if this is expected, or if there is further action that needs to be done on my part.

TimothyGu (approved):

PR #3192, opened 2021-05-23
Start requiring Node.js v12 or higher

domenic:
When merged this will cause a major version bump. See individual commits for more details.


TimothyGu (approved):

TimothyGu:
Fixed the tests so this is ready to merge. However, ideally I'd like to release a non-major-bump version with #3197, #3203, and #3208 first. Up to you @domenic.

domenic:
Hmm gonna need to make some updates for https://github.com/websockets/ws/releases/tag/8.0.0

domenic:
I'm not sure why the wss variant of Create-blocked-port.any.js was marked as passing before, but with the upgrade it started failing like the normal (ws) variant, so I've marked it as such.

PR #3201, opened 2021-06-03
Implement ReadableStreams (from scratch)

ninevra:
This PR, if completed, would add an implementation of [`ReadableStream`s](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) to jsdom. The level of functionality is similar to present-day Firefox, and should be sufficient to implement the parts of `FileAPI`, `Blob`, and `fetch` that require streams.

This is an alternative to #3200.

## Status

- [x] Basic `ReadableStream` types
- [ ] `tee()`: TODO
- [ ] async iteration: TODO; `webidl2js` may need a patch for the `AsyncIterator` prototype
- [ ] `bytes` streams: non-goal; they're very complex, the spec is in flux, and full conformance is not possible in javascript
- [ ] `WritableStream`; the parts of `ReadableStream` that require it; `TransformStream`: non-goals
- [ ] transferable streams: non-goal


ninevra:
Closing in favor of continuing work of #3200.

PR #3208, opened 2021-06-16
Be robust against global polyfills

### LINKED ISSUE #3205: ReferenceError: AggregateError is not defined
### Basic info:

- **Node.js version:** 14.15.5
- **jsdom version:** 16.6.0

### Minimal reproduction case

I recently upgraded jest-environment-jsdom from 26.6.2 to 27.0.3 and I'm now seeing some issues that I believe are related to this package. If I run multiple test files at once, everything is fine, but if I run a single test file, then there's a crash in this package:

```
  ● Test suite failed to run

    evalmachine.<anonymous>:1
    AggregateError
    ^

    ReferenceError: AggregateError is not defined

      at evalmachine.<anonymous>:1:1
      at setupWindow (node_modules/jsdom/lib/jsdom/browser/Window.js:111:55)
      at new Window (node_modules/jsdom/lib/jsdom/browser/Window.js:223:3)
      at exports.createWindow (node_modules/jsdom/lib/jsdom/browser/Window.js:95:10)
      at new JSDOM (node_modules/jsdom/lib/api.js:36:20)
```

I'm not sure what's going wrong here. I've never seen this error before and I can't find any trace of it on the Internet, but after upgrading, everyone on my team is seeing this problem.

I had initially filed this issue here (https://github.com/simon360/jest-environment-jsdom-global/issues/47), but even after purging jest-environment-jsdom-global from the repo, I'm still hitting this crash.

TimothyGu:
If we are running on a JavaScript engine that does not natively support a global in lib/jsdom/browser/js-globals.json, but the user loaded a polyfill prior to loading jsdom, then we could be tricked into assuming that global is available in new VM contexts.

Do not assume such a thing, and always check for each global's availability in the VM context.

One could test this by running the following script on Node.js v14.x, which does not implement AggregateError natively:

```js
global.AggregateError = class AggregateError extends Error {};
new JSDOM("", { runScripts: "dangerously" });
```

Fixes #3205.

----

This is unfortunately hard to test on CI, since we regenerate js-globals.json as part of the `prepare` step after installing dependencies. We could switch to generating js-globals.json manually though, or as part of a `prepublishOnly` step.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -107,8 +107,14 @@ function setupWindow(windowInstance, { runScripts }) {
 
     // Without this, these globals will only appear to scripts running inside the context using vm.runScript; they will
     // not appear to scripts running from the outside, including to JSDOM implementation code.
-    for (const [globalName, globalPropDesc] of jsGlobalEntriesToInstall) {
-      const propDesc = { ...globalPropDesc, value: vm.runInContext(globalName, windowInstance) };
+    for (const [globalName] of jsGlobalEntriesToInstall) {
+      // jsGlobalEntriesToInstall may already be contaminated by polyfills and such, so we check whether windowInstance
+      // has the globalName first.
+      const propDesc =
+        vm.runInContext(`Object.getOwnPropertyDescriptor(this, ${JSON.stringify(globalName)})`, windowInstance);
+      if (!propDesc) {
+        continue;
+      }
       Object.defineProperty(windowInstance, globalName, propDesc);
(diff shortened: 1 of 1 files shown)


domenic:
This seems similar in spirit to https://github.com/jsdom/jsdom/pull/2975 (although not in implementation). In both cases I'd really like to have a test for this. If we don't have a test, then I would not count the scenario as supported, and so I'd prefer it to continue breaking.

TimothyGu:
Indeed. I prefer #2975.

PR #3219, opened 2021-07-11
Add tests for cleaning up global 

GalenWong:
Adding tests for #2975. The tests pass if the two scripts execute without error. The scripts exist outside of Mocha and just run in a plain node environment.

### DIFF
--- a/lib/jsdom/browser/js-globals.js
+++ b/lib/jsdom/browser/js-globals.js
@@ -1,19 +1,11 @@
 "use strict";
 const vm = require("vm");
-const fs = require("fs");
-const path = require("path");
 const assert = require("assert");
 
 // Creates a list of globals from the JS environment (not the web), so that Window.js can alias them when necessary.
 // The generated list should match https://tc39.es/ecma262/#sec-global-object, to the extent V8 implements the spec.
-// We generate this at build time instead of runtime because we want to avoid the performance and memory overhead of
-// creating a new context when scripting is disabled in the JSDOM.
 
-const dest = path.resolve(__dirname, "../lib/jsdom/browser/js-globals.json");
-
-const context = vm.createContext();
-
-const globals = vm.runInContext("Object.getOwnPropertyDescriptors(this)", context);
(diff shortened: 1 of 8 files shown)


lib/jsdom/browser/js-globals.js
```diff
@@ -1,19 +1,11 @@
...
 // The generated list should match https://tc39.es/ecma262/#sec-global-object, to the extent V8 implements the spec.
-// We generate this at build time instead of runtime because we want to avoid the performance and memory overhead of
-// creating a new context when scripting is disabled in the JSDOM.
 
-const dest = path.resolve(__dirname, "../lib/jsdom/browser/js-globals.json");
-
-const context = vm.createContext();
-
-const globals = vm.runInContext("Object.getOwnPropertyDescriptors(this)", context);
+const globals = vm.runInNewContext("Object.getOwnPropertyDescriptors(this)");
```
TimothyGu, line 8: The CI logs for build-with-browser say

```
TypeError: vm.runInNewContext is not a function
```

Indeed, our [browser shim](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/vm-shim.js) for the vm module does not support runInNewContext. Could you add that function there? It should be as simple as `return runInNewContext(code, createContext())`

test/api/js-globals-aggregateError.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+/**
+  * This test script is NOT included in index.js. It is ran from `yarn test`
+  * directly to prevent the contamination of gloablThis in the Mocha test
```
TimothyGu, line 5:
```suggestion
  * directly to prevent the contamination of globals in the Mocha test
```

test/api/js-globals-globalThis.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+/**
+  * This test script is NOT included in index.js. It is ran from `yarn test`
+  * directly to prevent the contamination of gloablThis in the Mocha test
```
TimothyGu, line 5:
```suggestion
  * directly to prevent the contamination of globals in the Mocha test
```

PR #3249, opened 2021-09-22
fix: added validation as a condition to form requestSubmit #3247

### LINKED ISSUE #3247: Constraint validation API does not prevent form submission when setCustomValidity is called
### Basic info:

When `setCustomValidity` is called on an input element within a form, the form can't be submitted thanks to the [Constraint validation API](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation#complex_constraints_using_the_constraint_validation_api).

Unlike a real browser, JSDOM does not prevent form submission when `setCustomValidity` is called.

- **Node.js version:** v14.17.3
- **jsdom version:** 17.0.0

### Minimal reproduction case

```js
const { JSDOM } = jsdom;

const dom = new JSDOM(`<!DOCTYPE html><html><body></body></html>`);

// Create form with submit handler
const formElement = document.createElement('FORM');
formElement.onsubmit = (event) => {
  event.preventDefault();
  console.log('SUBMITTED');
}
document.body.appendChild(formElement);

// Create text input element
const inputElement = document.createElement('INPUT');
inputElement.type = 'text';
formElement.appendChild(inputElement);

// Create submit button
const submitElement = document.createElement('INPUT');
submitElement.type = 'submit';
formElement.appendChild(submitElement);

// Submit invalid form
inputElement.setCustomValidity('invalid');
submitElement.click();
```

Form is not submitted and console does not display "SUBMITTED".
See it action here:
https://jsfiddle.net/yzdcsko1/20/

pozil:
This PR adds validation to form's requestSubmit method in accordance with spec.

Fixes #3247

### DIFF
--- a/lib/jsdom/living/nodes/HTMLFormElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLFormElement-impl.js
@@ -101,6 +101,12 @@ class HTMLFormElementImpl extends HTMLElementImpl {
       }
     }
 
+    // https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-submission-algorithm
+    // Step 6.3: if form doesn't have the 'novalidate' attribute, validate it and abort submission if form is invalid
+    if (!this.noValidate && !this.reportValidity()) {
+      return;
+    }
+
     if (!fireAnEvent("submit", this, undefined, { bubbles: true, cancelable: true })) {
       return;
     }


domenic:
We'll need tests before we can review this.

pozil:
Thanks @domenic, I agree and I need your input to provide tests.
All WPT tests for form submission (`html/semantics/forms/form-submission-0`) are disabled in jsdom. I've tried to include them but they're failing.

Where should I add those tests for this particular feature that's a mix of validation and form submission?

domenic:
A test under `to-upstream/html/semantics/forms/form-submission-0` seems good.

pozil:
@domenic it took a bit more time than expected because the form submit logic wasn't implemented as per specs (clicking a submit input/button was triggering form `submit()` instead of `requestSubmit(submitter)`) but I was able to fix this and provide the appropriate tests.

Let me know if you require anything else for the review.

domenic (approved):
Code LGTM. The styling of the tests is a bit weird (lots of spurious indents) but I can fix that and then merge, if CI passes.

pozil:
Thanks for the review @domenic and sorry about the style.
I must say that I'm a big fan of Prettier when it comes to this so I have a tendency to blindly trust my IDE to handle the formatting in the background.

SamuelWei:
@pozil I just updated to the latest version of jsdom and a lot of tests where failing, due to forms beeing submitted (with invalid data) to simulate server side validation errors. Sometimes it would greate for writing tests, if you can allow the form to be submitted even if invalid. Could this be a global config option of jsdom? Debbuing this error in my tests was quite a pain, it would be very helpfull if jsdom would print an error, or something similar to let the dev. know why the form was not submitted.

pozil:
Sorry to hear this. This contribution was meant to align the behavior of virtual forms in jsdom with the behavior of forms in a real browser (web standards).

I'm just a small contributor so don't have a say in the governance of jsdom but I believe that the general intent of this tool is to get as close as possible to the real thing. I don't think that the main contributors intend to add flags to disable standard behaviors. Perhaps you should open a new issue to discuss this and reference this PR?

SamuelWei:
@pozil thanks for your quick reply and the great contribution. I agree that jsdom should behaive as closely as possible to a regular web browser.

I will open an issue and PR regarding my proposal for a warning for form validation errors.

pozil:
@SamuelWei I just had an idea which should work for your use case: simply add a [`novalidate` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attributes) on your `form` element when you're in the context of your tests. This standard attribute disables HTML form validation.

PR #3251, opened 2021-09-23
upgrade webidl-conversions > 6.1.0

joriewong:
to fix https://github.com/jsdom/domexception/pull/11


domenic:
We'll be doing upgrades as part of a larger push to modernize webidl2js and its dependencies. Updating one dependency in isolation will break things, as you could see if you had run the tests.

PR #3256, opened 2021-10-05
fix: added tooShort/tooLong validation support. 

007tom:
Context: https://github.com/jsdom/jsdom/pull/2142#issuecomment-362946100


domenic:
That does not address my comment.

007tom:
Thank you, my real question is how to test `tooShort` validity using jest if `tooShort` alwsys return `false`.

007tom:
So what's the problem of my commit, it works for me.

domenic:
Your commit makes jsdom behave not like browsers behave, which is against our mission. You would realize this if you added tests (as is required by any PRs to this project), since those tests would not pass in browsers.

007tom:
That's also my concern, so it's in **Draft** state.

Whatever, maybe you're right!

Thank you!

PR #3258, opened 2021-10-16
Major performance improvements in some cases

GrantGryczan:
## Improve `_insert` performance from O(n^2) to O(n)

"n" here refers to the tree depth of the node being inserted, so lots of nested elements causes insertion to be extremely slow. Deep cloning, for example, calls `_insert` O(n) times, which makes cloning O(n^3) without this optimization and O(n^2) with it.

**This can make the difference between minutes and milliseconds** of synchronous blocking time for very deep node trees (which is my use case and what led me to need this optimization).

It works by only checking `isConnected` (which is O(n)) once on the node being inserted rather than checking it on every single descendant of the node being inserted. As far as I know, checking `isConnected` for every descendant isn't necessary since any node should have the same `isConnected` value as any other node in the same node tree.

## Improve `contains` performance

The current implementation of a node's `contains` method uses `compareDocumentPosition`, which runs a lot of extra features that aren't used by `contains`. I've run into a case where these extra features cost 560 ms of total synchronous blocking time, according to a CPU profile. The optimization I've implemented here has reduced that case down to 30 ms.

It works by simply checking each ancestor of the `other` node until it finds the `this` node (as in `this.contains(other)`), rather than checking a number of relationships between the two nodes but utilizing only one of them.

### DIFF
--- a/lib/jsdom/living/nodes/Node-impl.js
+++ b/lib/jsdom/living/nodes/Node-impl.js
@@ -11,7 +11,7 @@ const { setAnExistingAttributeValue } = require("../attributes");
 
 const NodeList = require("../generated/NodeList");
 
-const { nodeRoot, nodeLength } = require("../helpers/node");
+const { nodeRoot, nodeLength, isInclusiveAncestor } = require("../helpers/node");
 const { domSymbolTree } = require("../helpers/internal-constants");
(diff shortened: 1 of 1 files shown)


GrantGryczan:
Any updates?

domenic (approved):

GrantGryczan:
Awesome, thanks!

PR #3284, opened 2021-11-11
Add support for ports in postmessage

martinffx:
I need to be able to transfer a MessageChannel port via postmessage

https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API/Using_channel_messaging

### DIFF
--- a/lib/jsdom/living/post-message.js
+++ b/lib/jsdom/living/post-message.js
@@ -6,7 +6,7 @@ const { isValidTargetOrigin } = require("../utils");
 const { fireAnEvent } = require("./helpers/events");
 
 module.exports = function (globalObject) {
-  return function (message, targetOrigin) {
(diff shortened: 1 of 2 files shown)


martinffx:
Is JSDOM not accepting PR's?

benogle:
Is there a chance this will get merged?

martinffx:
@domenic @SimenB any chance of getting this added? Happy to start working on any feedback you may have?

domenic:
This is not a correct implementation. It appears to pass through any object, not just MessagePort objects. And jsdom doesn't implement MessagePort objects, so that would be the first thing to do.

In general, please follow the specifications when adding new code.

martinffx:
Cool, I'll have a look at updating it to only support [Transferable Objects](https://developer.mozilla.org/en-US/docs/Glossary/Transferable_objects)

PR #3285, opened 2021-11-11
Add AbortSignal.reason

MattiasBuelens:
Implements whatwg/dom#1027.

PR #3286, opened 2021-11-11
Update web platform tests

MattiasBuelens:
#3285 needs the updated WPTs for `AbortSignal.reason`, but it turns out that updating WPT is quite an undertaking. 😅

Hopefully it's a bit easier to review as a separate PR.


test/web-platform-tests/to-run.yaml
```diff
@@ -261,6 +272,7 @@ DIR: fetch/api/headers
 header-values-normalize.any.html: [fail, fetch is not defined]
 header-values.any.html: [fail, fetch is not defined]
 headers-no-cors.any.html: [fail, Request is not defined]
+headers-normalize.any.html: [fail, Unknown]
```
MattiasBuelens, line 275: This is caused by web-platform-tests/wpt#30979, where the Deno folks noticed that you shouldn't use `String.prototype.trim()` for normalizing header values. Of course, [we made the same mistake in jsdom](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/fetch/Headers-impl.js#L64)... 😅 

It's quite low-hanging fruit, so I'll fix it as part of this PR.

MattiasBuelens:
[CI fails on Node 12](https://github.com/jsdom/jsdom/runs/4184106901?check_suite_focus=true) with the following error:
```
INFO:root:Status of subprocess "http on port 9000": running

  5142 passing (36m)
  1184 pending
  1 failing

  1) web-platform-tests
       dom/events
         KeyEvent-initKeyEvent.html:
     Error: Error: test harness should not error: dom/events/KeyEvent-initKeyEvent.html
Unexpected token '.'
      at /home/runner/work/jsdom/jsdom/test/web-platform-tests/run-single-wpt.js:216:22
      at http://web-platform.test:9000/resources/testharness.js:3135:22
```

Looks like [this test](https://github.com/web-platform-tests/wpt/blob/57c5ccf5609520fcc25607df79e375c4f1747ebb/dom/events/KeyEvent-initKeyEvent.html#L12) uses optional chaining, which is only supported since Node 14... Sooo, now what? 😬

TimothyGu:
You can mark those tests with `needs-node14`. See test/web-platform-tests/run-wpts.js for the implementation for some older Node.js versions (which we should probably remove now that we require Node.js 12, heh).

MattiasBuelens:
It looks like we deleted the logic for minimum Node versions as part of e0e3e33d373dca80eb04377f100001d1011ea386, see [diff](https://github.com/jsdom/jsdom/commit/e0e3e33d373dca80eb04377f100001d1011ea386#diff-87b5a2e85858d5459c06637661176179339bb073e505e65bbbc7f1f3764c903b). I'll go ahead and resurrect that code.

lib/jsdom/living/events/EventTarget-impl.js
```diff
@@ -101,14 +101,14 @@ class EventTargetImpl {
...
-  _dispatch(eventImpl, targetOverride /* , legacyOutputDidListenersThrowFlag */) {
+  _dispatch(eventImpl, legacyTargetOverrideFlag /* , legacyOutputDidListenersThrowFlag */) {
     let targetImpl = this;
     let clearTargets = false;
     let activationTarget = null;
 
     eventImpl._dispatchFlag = true;
 
-    targetOverride = targetOverride || targetImpl;
+    const targetOverride = legacyTargetOverrideFlag ? idlUtils.implForWrapper(targetImpl._globalObject._document) : targetImpl;
```
MattiasBuelens, line 111: Thanks! That's much better. 😁

domenic: I had to break out the debugger to figure out why `targetImpl._document` wasn't working... darn global proxy vs. global object stuff.

PR #3289, opened 2021-11-18
add ESM usage

mariusa:


domenic:
This is not possible with jsdom.

mariusa:
Could you please detail? I'm importing jsdom in an ESM project and it works fine.

domenic:
If you can produce a script that you can run with `node test.js` that imports jsdom, that would be interesting information; please show us such a script. Otherwise I suspect you are using a third-party tool like TypeScript or babel.

mariusa:
:)
Using node 16

package.json
```
"type": "module",
```

test.js
```
import { JSDOM } from 'jsdom'
const dom = new JSDOM(`<!DOCTYPE html><p>Hello world</p>`)
console.log(dom.window.document.querySelector("p").textContent) // "Hello world"

```


```
$ node test.js 
Hello world
```

domenic:
Fascinating, that's good information!

Still, for those people using that setup, I assume they know how to mentally translate between CommonJS and import. I'd rather not clutter up the documentation with that.

Thanks for teaching me something!

mariusa:
> I'd rather not clutter up the documentation with that.

This isn't reasonable. We're talking about 2 lines, 1 of those being a comment.

Also:
1. It took me time to find out proper syntax. Alternatives are `import JSDOM from 'jsdom'` or `import * as JSDOM from 'jsdom'`
2. Others looking on README will conclude ESM isn't supported, if not mentioned.

Please consider adding these 2 lines. Thanks.

rayfoss:
This PR would have save me searching through 5 web pages for the info.  Node documentation is full of ESM and defaults to it... LTS has stable support for it. Once TypeScript finishes their Native ESM support, CommonJS will be deprecated.

Screenshot from 2021-12-17 00-28-35

https://stackblitz.com/edit/feathersjs-7kqlyt?file=server.js

P.S. I haven't been using CommonJS for anything in ~5 months... I started transitioning on Node 15 three years ago. I run local shebang scripts with mjs, or ts.

PR #3295, opened 2021-11-24
Fix crash when calling window.close() inside load event listener

MattiasBuelens:
When loading the following page, jsdom crashes:
```html
<script>
    window.close();
  });
</script>
```
Stack trace:
```
TypeError: Cannot read property '_pageShowingFlag' of null
    at Document.<anonymous> (C:\...\jsdom\lib\jsdom\browser\Window.js:915:27)
    at Document.callTheUserObjectsOperation (C:\...\jsdom\lib\jsdom\living\generated\EventListener.js:26:30)
    at innerInvokeEventListeners (C:\...\jsdom\lib\jsdom\living\events\EventTarget-impl.js:340:25)
    at invokeEventListeners (C:\...\jsdom\lib\jsdom\living\events\EventTarget-impl.js:276:3)
    at DocumentImpl._dispatch (C:\...\jsdom\lib\jsdom\living\events\EventTarget-impl.js:223:9)
    at fireAnEvent (C:\...\jsdom\lib\jsdom\living\helpers\events.js:18:36)
    at dispatchEvent (C:\...\jsdom\lib\jsdom\living\nodes\Document-impl.js:476:9)
    at C:\...\jsdom\lib\jsdom\living\nodes\Document-impl.js:481:11
    at new Promise (<anonymous>)
    at onLoad (C:\...\jsdom\lib\jsdom\living\nodes\Document-impl.js:479:14)
    at Object.check (C:\...\jsdom\lib\jsdom\browser\resources\resource-queue.js:76:23)
    at C:\...\jsdom\lib\jsdom\browser\resources\resource-queue.js:83:27
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```
This PR fixes this by checking if the window has been closed after firing the `load` event.

See https://github.com/whatwg/streams/pull/1189#issuecomment-978419971 for context.


MattiasBuelens:
I'm struggling with writing a proper test for this. I tried this:
```html
<!DOCTYPE html>
<title>Test window.close() inside load event listener.</title>
<script src=/resources/testharness.js></script>
<script src=/resources/testharnessreport.js></script>
<script>
  "use strict";
  setup({ single_test: true });

    window.close();
    done();
  });
</script>
```
But then it errors inside `testharness.js`, because it tries to build some new elements to show the tests results on the page... which we've just closed. 😅

domenic:
This might not be testable from inside jsdom, especially since we don't support `window.open()`. If so you can use https://github.com/jsdom/jsdom/blob/master/test/api/from-outside.js .

Another fix might be aligning with the spec for `window.close()`, which only closes the window from a queued task: https://html.spec.whatwg.org/#dom-window-close . I worry that might make some jsdom-using programs sad though, since we've so far been treating `window.close()` as a kind of "synchronously make stuff garbage-collectable" call, and waiting for a task to do so might cause programs to consume a lot more resources...

MattiasBuelens:
Okay, I managed to disable the test output from `testharness.js`, such that the test passes. Unfortunately, the test isn't perfect: if you remove the fix, it still passes - although it does print the call stack from the OP. So you'd have to review the logs to see if it actually worked... 😕

> This might not be testable from inside jsdom, especially since we don't support `window.open()`. If so you can use https://github.com/jsdom/jsdom/blob/master/test/api/from-outside.js .

I'll give that a try. 🙂

> Another fix might be aligning with the spec for `window.close()`, which only closes the window from a queued task: https://html.spec.whatwg.org/#dom-window-close . I worry that might make some jsdom-using programs sad though, since we've so far been treating `window.close()` as a kind of "synchronously make stuff garbage-collectable" call, and waiting for a task to do so might cause programs to consume a lot more resources...

Agreed, let's keep it synchronous.

PR #3296, opened 2021-11-29
Fix appending text to Text node's data in parse5 adaptor

### LINKED ISSUE #3261: Incorrect Mutation Records for document.write with whitespace
### Basic info:

- **Node.js version:** v12.16.3
- **jsdom version:** v18.0.0

### Minimal reproduction case

```js
const options = {};
const dom = new JSDOM("", options);
const window = dom.window;

document.open();

const observer = new window.MutationObserver(records =>
  records.forEach(record => console.log(formatMutationRecord(record)))
);

observer.observe(document, {
  attributes: true,
  attributeOldValue: true,
  characterData: true,
  characterDataOldValue: true,
  childList: true,
  subtree: true
});

document.write("a b c d");

document.close();

function formatMutationRecord(record) {
  const { target, oldValue } = record;

  switch (record.type) {
    case "characterData": {
      return target.nodeName + ": " + JSON.stringify(oldValue) + " => " + JSON.stringify(target.nodeValue);
    }

    case "childList": {
      const { addedNodes } = record;
      return "inserted " + Array.from(addedNodes).map(added => added.nodeName).join(", ")
    }
  }
}
```

Mutations records include a bunch of unexpected "CharacterData" updates:

```
inserted HTML
inserted HEAD
inserted BODY
inserted #text
#text: "a" => "a b c d"
#text: "a " => "a b c d"
#text: "a b" => "a b c d"
#text: "a b " => "a b c d"
#text: "a b c" => "a b c d"
#text: "a b c " => "a b c d"
```

Mutation record does not contain these updates:
```
inserted HTML
inserted HEAD
inserted BODY
inserted #text
```

https://cloudflareworkers.com/#8908b378ca311d7b21227af410fdaa19:https://tutorial.cloudflareworkers.com/
(open the browser console - displayed console is for the worker)

TimothyGu:
The ["insert a character"][1] algorithm in the HTML parser is clear on appending directly to the Text node's data rather than using the replace data algorithm (which creates mutation records).

[1]: https://html.spec.whatwg.org/multipage/parsing.html#insert-a-character

Fixes: #3261

### DIFF
--- a/lib/jsdom/browser/parser/html.js
+++ b/lib/jsdom/browser/parser/html.js
@@ -139,7 +139,7 @@ class JSDOMParse5Adapter {
   insertText(parentNode, text) {
     const { lastChild } = parentNode;
     if (lastChild && lastChild.nodeType === nodeTypes.TEXT_NODE) {
-      lastChild.data += text;
+      lastChild._replaceData(lastChild.length, 0, text, true);
     } else {
       const ownerDocument = this._ownerDocument();
       const textNode = Text.createImpl(this._globalObject, [], { data: text, ownerDocument });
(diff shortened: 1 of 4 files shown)


domenic:
Somewhat blocked on https://github.com/whatwg/html/issues/12744 since we've just added a not-per-spec suppressObservers flag.

PR #3298, opened 2021-11-29
Do not track location information for nodes created by innerHTML

### LINKED ISSUE #2968: Reporting `nodeLocation`s for nodes which were dynamically created is surprising
### Basic info:

- **Node.js version:** v12.16.2
 **jsdom version:** v16.2.2

### Minimal reproduction case

```js
let jsdom = require('jsdom');

let dom = new jsdom.JSDOM('<a><b></a></b>', { includeNodeLocations: true });

let b = dom.window.document.querySelector('b');
console.log(dom.nodeLocation(b).startOffset);

dom.window.document.querySelector('a').innerHTML = '<b></b>';
b = dom.window.document.querySelector('b');
console.log(dom.nodeLocation(b) && dom.nodeLocation(b).startOffset);
```

The first `console.log` prints `3`, which is good. The second prints `0`, which is surprising to me - at that point `b` does not exist in the actual source. Also, offset `0` in the original source has nothing to do with `b`.

This changed in 49353e224b6c59e48c3cc7da3700dbe93a2286ac (first released in 14.1.0); prior to that commit, the above would have printed `3` and `undefined`.

This breaks consumers which assume they can safely map location information, if it exists, back to the original source. E.g. this breaks ecmarkup, which is doing [exactly that](https://github.com/tc39/ecmarkup/blob/6aa114edb04c1e590f6d52760e7c9abd3e2545ce/src/Grammar.ts#L19).

N/A

bakkot:
Fixes #2968.

Specifically, this disables location information for nodes created by the HTML parser's `parseFragment`, which is used for `innerHTML`, `outerHTML`, and `createContextualFragment`. It does not change anything for nodes created by `parseIntoDocument`. Note that `nodeLocation` will return `undefined` rather than throwing for these nodes, which matches the behavior prior to https://github.com/jsdom/jsdom/commit/49353e224b6c59e48c3cc7da3700dbe93a2286ac.

In https://github.com/jsdom/jsdom/issues/2968#issuecomment-980497117 I mentioned three other APIs. Those all use `parseIntoDocument` and so are unaffected, but for reference:

- DOMParser's `parseFromString` is [not forwarding along `parseOptions`](https://github.com/jsdom/jsdom/blob/a61fdb886a30fce6a3aa68002e1af6e5bcb4c372/lib/jsdom/living/domparsing/DOMParser-impl.js#L42-L49), so it uses the default behavior of not tracking
-  XHR's `responseXML` [does forward `parseOptions`](https://github.com/jsdom/jsdom/blob/a61fdb886a30fce6a3aa68002e1af6e5bcb4c372/lib/jsdom/living/xhr/XMLHttpRequest-impl.js#L304), so it does track node locations. 
- I haven't checked frames because I couldn't figure out how to get them to work, but [looking at the source](https://github.com/jsdom/jsdom/blob/a61fdb886a30fce6a3aa68002e1af6e5bcb4c372/lib/jsdom/living/nodes/HTMLFrameElement-impl.js#L129-L139) I don't think `parseOptions` is forwarded, i.e. I think it will not track locations.

### DIFF
--- a/lib/jsdom/browser/parser/html.js
+++ b/lib/jsdom/browser/parser/html.js
@@ -202,6 +202,7 @@ function parseFragment(markup, contextElement) {
 
   const config = {
     ...ownerDocument._parseOptions,
+    sourceCodeLocationInfo: false,
     treeAdapter: new JSDOMParse5Adapter(ownerDocument, { fragment: true })
   };
 
--- a/test/api/methods.js
+++ b/test/api/methods.js
@@ -133,6 +133,39 @@ describe("API: JSDOM class's methods", () => {
         endOffset: 36
       });
     });
+
+    it("should return undefined for nodes created by innerHTML", () => {
+      const dom = new JSDOM(`<p>Hello</p>`, { includeNodeLocations: true });
+      const para = dom.window.document.querySelector("p");
+
+      para.innerHTML = `<div></div>`;
+
+      const div = dom.window.document.querySelector("div");
+
+      assert.deepEqual(dom.nodeLocation(div), undefined);
+    });
+
+    it("should return undefined for nodes created by outerHTML", () => {
+      const dom = new JSDOM(`<p>Hello</p>`, { includeNodeLocations: true });
+      const para = dom.window.document.querySelector("p");
+
+      para.outerHTML = `<div></div>`;
+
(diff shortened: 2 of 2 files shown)


domenic (approved):
Awesome, will find time to merge and do a release sometime soon.

Issue #1135, opened 2015-05-22
Intercepting Navigation Attempts


jpttrssn:
Is there a way to capture the request of [navigation](https://html.spec.whatwg.org/#navigate) attempts? If not, is it technically feasible and something that would fit within the scope of this project?

For example, if I'm crawling a website that happens to have a click event on links that perform navigation I would like to be able to capture the new URL. Same goes for form submissions.

Here's a simplified example:

```html
<!DOCTYPE html>
<html lang="en">
<head>
<title>Page 1</title>
<script src="http://code.jquery.com/jquery.js"></script>
</head>
<body>
    <h1>Page 1</h1>
    <a href="#">Index 2</a>
    <script>
        $().ready(function() {
            $('a').click(function(event) {
                window.location.href = 'index2.html';
            });
        });
    </script>
</body>
</html>
```

In the following test script I'm simply trying to get the changed window.location.href. Ideally you could get the entire request (by using something similar to the resourceLoader perhaps?). The window.location.href in this example doesn't change.

```javascript
'use strict';

var jsdom = require('jsdom');

jsdom.env({
    url: 'index.html',
    scripts: ['http://code.jquery.com/jquery.js'],
    done: function (errors, window) {
        window.$('a').each(function(index) {
            window.$(this).click();
            console.log(window.location.href);
        });
        window.close();
    },
    features: {
        FetchExternalResources : ['script'],
        ProcessExternalResources: ['script']
    }
});
```

domenic:
Great question. We do not support this currently. Technically this is within scope, but it is quite hard to implement, as the Window itself will change (as will all the other stuff within) when you navigate. I kind of doubt we'll get to it any time soon,

I know the zombie.js project, based on jsdom, has tackled a lot of these issues. I'd definitely suggest looking at them for a solution.

An easy thing we _could_ add to jsdom would be a kind of navigate hook. Something like `jsdom.jsdom(..., { navigate(newUrl) { ... } })`.

jpttrssn:
I'll take a look at zombie.js to see if it'll meet my needs. If not, the navigate hook would be what I'm looking for and I wouldn't mind helping make that happen, but I would need some help getting started. I'll let you know.

jpttrssn:
[...]

PR #3326, opened 2022-01-24
WIP: NotImplementedError and child exception classes

lezsakdomi:
Needs:

- [ ] Demo code
- [ ] Unit tests
- [ ] Ensuring that I've made the right decisions
  - Do we need such specific exceptions for nodes? Is the code repetition introduced worth it?
  - Is it OK to define "static methods" on the notImplemented function? Do we need to keep the backward compatibility in the internal API?
  - `window/notImplementedMethod()`'s argument changed kinda unnecessarily
  - `NotImplementedFormElementMethodError` vs `NotImplementedFetchNavigationError` for `HTMLFormElement.submit()`
  - Using commercial java style long class names - not really OK
- ~[ ] Replacing the most of errors I introduced with hooks~
  - WIth every call to notImplemented, hooks can still be easily introduced there - for ex. by emitting a cancelable NotImplementedError event on the window, which can prevent emitting a jsdom error in turn. (There is a lot to consider this way though.)

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -4,6 +4,7 @@ const webIDLConversions = require("webidl-conversions");
 const { CSSStyleDeclaration } = require("cssstyle");
 const { Performance: RawPerformance } = require("w3c-hr-time");
 const notImplemented = require("./not-implemented");
+const { NotImplementedWindowMethodError, NotImplementedMethodError } = require("./not-implemented");
 const { installInterfaces } = require("../living/interfaces");
 const { define, mixin } = require("../utils");
 const Element = require("../living/generated/Element");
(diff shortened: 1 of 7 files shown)


domenic:
Could you perhaps open an issue explaining what you're trying to do, and why jsdom doesn't support your work today? In general we prefer to avoid classes when strings or properties would suffice, and I think we have those already.

lezsakdomi:
The tests didn't run for me, but some were totally unrelated to my changes, so I'm not sure if I can rely on them.

[test-log.txt](https://github.com/jsdom/jsdom/files/7928576/test-log.txt)

lezsakdomi:
@domenic Sorry, I was trying to make my imagination from https://github.com/jsdom/jsdom/issues/1135#issuecomment-1019306705 to come into reality. I've linked this PR over there.

Actually I'm trying to replicate zombie.js's behaviour on a simple website which uses 2010-ish WebForms libraries. Of course it's way out of scope of this library, but I feel like detecting the event itself happening is still a needed behaviour.

My goal could be accomplished just by adding a `newURL? : string` property to the issue thrown, but I thought it would be too hacky, so I went the more complicated way.

I think that classes are the best way to represent different kind of exceptions, but I'm still not sure if separate `NotImplementedCanvasElementMethodError`, `NotImplementedFormElementMethodError` and `NotImplementedMediaElementMethodError` were needed.

lezsakdomi:
Ah, found the `Convert to draft` button. I also need to take a look into
- [ ] https://github.com/jsdom/jsdom/pull/1902
- [ ] https://github.com/jsdom/jsdom/pull/1905
before finalizing this MR.

Issue #2600, opened 2019-05-28
Lack of RadioNodeList


TechQuery:
## Basic info

- **Node.js version:** 8.0.0+
- **jsdom version:** 12.0.0+

## Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const { window: { document } } = new JSDOM(`
<form>
    <input type="radio" name="radio" value="r1">
    <input type="radio" name="radio" value="r2">

    <input type="checkbox" name="check" value="c1">
    <input type="checkbox" name="check" value="c1">
</form>
`);

const { radio, check } = document.form[0].elements;

console.assert(radio.constructor.name === 'RadioNodeList', 'List of Radio inputs');
console.assert(check.constructor.name === 'RadioNodeList', 'List of CheckBox inputs'); 
```

## Specification

https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#radionodelist

alex-cleo:
Hello there.

Any updates on this? My code depends on `if (element instanceof RadioNodeList) { ... }` which makes the test fail.

domenic:
https://mobile.twitter.com/slicknet/status/782274190451671040

alex-cleo:
> 

I'll take a look after work. If I'm able to do it myself I'll gladly create one

vlakoff:
This issue seems to have been fixed by #3337.

TechQuery:
> This issue seems to have been fixed by #3337.

@vlakoff Thanks, @UndefinedBehaviour referred this issue instead of `fix` it, so it keeps open for 3 years...

PR #3337, opened 2022-02-09
implement HTMLFormControlsCollection and RadioNodeList

UndefinedBehaviour:
https://github.com/jsdom/jsdom/issues/2600

### DIFF
--- a/lib/jsdom/living/nodes/HTMLFormControlsCollection-impl.js
+++ b/lib/jsdom/living/nodes/HTMLFormControlsCollection-impl.js
@@ -1,0 +1,33 @@
+"use strict";
+
+const HTMLCollectionImpl = require("./HTMLCollection-impl").implementation;
+const RadioNodeList = require("../generated/RadioNodeList");
+
+exports.implementation = class HTMLFormControlsCollectionImpl extends HTMLCollectionImpl {
+  namedItem(name) {
+    if (name === "") {
+      return null;
+    }
+
+    this._update();
+
+    const nodeList = RadioNodeList.createImpl(this._globalObject, [], {
+      element: this,
+      query: () => this._list.filter(
+        e => e.getAttributeNS(null, "id") === name || e.getAttributeNS(null, "name") === name
+      )
+    });
+
+    switch (nodeList.length) {
+      case 0: {
+        return null;
+      }
+      case 1: {
+        return nodeList.item(0);
+      }
+      default: {
+        return nodeList;
+      }
+    }
+  }
+};
--- a/lib/jsdom/living/nodes/RadioNodeList-impl.js
+++ b/lib/jsdom/living/nodes/RadioNodeList-impl.js
@@ -1,0 +1,49 @@
+"use strict";
+
+const NodeListImpl = require("./NodeList-impl").implementation;
+
+class RadioNodeListImpl extends NodeListImpl {
+  // https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#dom-radionodelist-value
+  // Note in general the spec says to manipulate/consult checkedness, but we use `checked` instead
+  // because the spec isn't very good here: https://github.com/whatwg/html/issues/7612.
+
+  get value() {
+    this._update();
+
+    const element = this._list.find(e => e._localName === "input" && e.type === "radio" && e.checked);
+    if (element === undefined) {
+      return "";
+    }
+
+    if (!element.hasAttributeNS(null, "value")) {
+      return "on";
+    }
+
+    return element.getAttributeNS(null, "value");
+  }
+
+  set value(value) {
+    let element;
+    if (value === "on") {
+      element = this._list.find(
+        e => e._localName === "input" &&
+             e.type === "radio" &&
+             (!e.hasAttributeNS(null, "value") || e.getAttributeNS(null, "value") === value)
+      );
+    } else {
+      element = this._list.find(
+        e => e._localName === "input" &&
+             e.type === "radio" &&
+             (e.hasAttributeNS(null, "value") && e.getAttributeNS(null, "value") === value)
+      );
+    }
+
+    if (element) {
+      element.checked = true;
+    }
+  }
+}
+
+module.exports = {
+  implementation: RadioNodeListImpl
+};
--- a/lib/jsdom/living/interfaces.js
+++ b/lib/jsdom/living/interfaces.js
@@ -31,6 +31,7 @@ const generatedInterfaces = {
   Comment: require("./generated/Comment"),
   DocumentType: require("./generated/DocumentType"),
   NodeList: require("./generated/NodeList"),
+  RadioNodeList: require("./generated/RadioNodeList"),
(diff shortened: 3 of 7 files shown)


domenic:
From a first glance, this looks very impressive! I won't have time to do a full review until the weekend, but I am optimistic. Nice work!

domenic:
Do fix the lint though :)

UndefinedBehaviour:
Thank you! lint fixed! 😛

lib/jsdom/living/nodes/HTMLFormControlsCollection-impl.js
```diff
@@ -0,0 +1,30 @@
...
+exports.implementation = class HTMLFormControlsCollectionImpl extends HTMLCollectionImpl {
+  namedItem(key) {
+    if (!key) {
+      return null;
+    }
+    this._update();
+    const nodeList = RadioNodeList.createImpl(this._globalObject, [], {
+      element: this,
+      query: () => this._list.reduce((a, b) => {
+        if (b.getAttributeNS && (b.getAttributeNS(null, "id") === key || b.getAttributeNS(null, "name") === key)) {
```
domenic, line 15: When would getAttributeNS not be truthy?

lib/jsdom/living/nodes/RadioNodeList-impl.js
```diff
@@ -0,0 +1,28 @@
+"use strict";
+
+const NodeListImpl = require("./NodeList-impl").implementation;
+
+class RadioNodeListImpl extends NodeListImpl {
+  get value() {
+    this._update();
+    const element = this._list.find(e => e._localName === "input" && e.type === "radio" && e.checked);
+    if (element) {
+      return element.value ? element.value : "on";
```
domenic, line 10: Why are you excluding elements with empty string value?

lib/jsdom/living/nodes/HTMLFormControlsCollection-impl.js
UndefinedBehaviour: just looking at the code in HTMLCollection-impl.js I didn't see any obvious check and wanted to play it safe... but yeah I guess if getAttributeNS is missing here there must a bigger problem somewhere else so this is not the place for that check...

lib/jsdom/living/nodes/RadioNodeList-impl.js
UndefinedBehaviour: oops... fixed!
(the wpt suite doesn't check this case 😮)

lib/jsdom/living/nodes/RadioNodeList-impl.js
```diff
@@ -0,0 +1,28 @@
+"use strict";
+
+const NodeListImpl = require("./NodeList-impl").implementation;
+
+class RadioNodeListImpl extends NodeListImpl {
+  get value() {
+    this._update();
+    const element = this._list.find(e => e._localName === "input" && e.type === "radio" && e.checked);
```
domenic, line 8: `e.checked` doesn't necessarily represent the checkedness that the spec asks about. As you can see from https://html.spec.whatwg.org/#attr-input-checked it's a pretty complicated relationship. Instead, check `e._checkedness`.

lib/jsdom/living/nodes/RadioNodeList-impl.js
```diff
@@ -0,0 +1,28 @@
...
+    if (element && element.value !== "") {
```
domenic, line 9: Let's follow the spec more exactly here. It has a step that says "if element is null, return the empty string". Please translate that, in this case becoming `if (element === undefined) { return ""; }`.

Then it says, "If element is an element with no [value](https://html.spec.whatwg.org/#attr-input-value) attribute, return the string "on"." This is not very well-written spec text, but again, let's try to translate it. In this case it'd be `if (!("value" in element)) { return "on"; }`.

And then there's a final step... that one should be clear.

lib/jsdom/living/nodes/RadioNodeList-impl.js
```diff
@@ -0,0 +1,28 @@
...
+      return element.value ? element.value : "on";
+    }
+    return "";
+  }
+  set value(value) {
+    this._update();
```
domenic, line 15: Again, this doesn't follow the spec very well. Please try to translate it more precisely.

lib/jsdom/living/nodes/RadioNodeList-impl.js
```diff
@@ -0,0 +1,28 @@
...
+    const element = this._list.find(
+      // https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#dom-radionodelist-value "On setting..."
+      e => e._localName === "input" && e.type === "radio" && (e.value ? value === e.value : value === "on")
+    );
+    if (element) {
+      element.checked = true;
```
domenic, line 21: As above, change the `_checkedness`, not `checked`.

domenic (changes requested):
OK, the full review is here!

In general the main advice is: follow the spec as precisely as possible. The closer of a translation it is, the easier maintenance is in the future, and the easier it is for reviewers or other code-readers to see that it's obviously correct. If you have any questions about how to translate, please feel free to ask them, and I or someone else should be able to chime in pretty fast.

Finally, I suspect a lot of the changes I'm requesting aren't covered by web platform tests. If you have the extra time, it'd be much appreciated to drop extra test cases into the to-upstream web platform tests directory, testing those edge cases. (E.g., cases where `checked` doesn't match the `_checkedness`.) It's not required but would be a nice bonus :).

lib/jsdom/living/nodes/RadioNodeList-impl.js
domenic: Hmm. Actually I suspect the spec is wrong here. Please leave as-is and add a comment linking to https://github.com/whatwg/html/issues/7612 .

If you change this to `_checkedness`, do any tests fail? If none do, then adding one would be super appreciated (but again not required). It would help us straighten out the spec situation.

UndefinedBehaviour: Yes, if you use `_checkedness` here then a couple tests in `radionodelist.html` will fail as multiple buttons will have their _checkedness set to true

lib/jsdom/living/nodes/HTMLFormControlsCollection-impl.js
```diff
@@ -0,0 +1,30 @@
...
+
+exports.implementation = class HTMLFormControlsCollectionImpl extends HTMLCollectionImpl {
+  namedItem(key) {
+    if (!key) {
+      return null;
+    }
+    this._update();
+    const nodeList = RadioNodeList.createImpl(this._globalObject, [], {
+      element: this,
+      query: () => this._list.reduce((a, b) => {
```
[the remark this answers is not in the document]
UndefinedBehaviour: Yeah `filter` works and is more straightforward... not sure why i thought i'd use `reduce` here...

domenic (approved):

PR #3342, opened 2022-02-25
Pass ports along with postMessage, allow non '*' target origins.

stevemolitor:
See #3284. This PR does that, an also lets you pass target origins besides just `*`. 

This is a bit of a hack though - for use in my own unit tests, not fit for merging.


stevemolitor:
Closing as I don't want this merged.

PR #3352, opened 2022-04-12
Implement crypto.getRandomValues(array).

sjrd:
This is the only member of `crypto` that is available even in non-secure contexts.

---

Motivation: Since we fixed a security in Scala.js related to `java.util.UUID.randomUUID()` (see https://github.com/scala-js/scala-js/security/advisories/GHSA-j2f9-w8wh-9ww4), the need for `crypto.getRandomValues` has become quite high. `jsdom` is one of the main testing environments used by the Scala.js community. The lack of `getRandomValues` is now causing several maintainers and project authors to use insecure variants because of `jsdom`. This in turn risks exposing them to security vulnerabilities of their own. I hope that by actually providing `crypto.getRandomValues` in `jsdom`, we will avoid this problematic situation.

---

This is my first contribution to `jsdom`. Don't hesitate to tell me if anything should be done differently. Also, if things should change, please tell me whether you prefer amending the commits and force-pushing on the branch, or whether you prefer to add fixup commits to the branch.

### DIFF
--- a/lib/jsdom/living/crypto/Crypto-impl.js
+++ b/lib/jsdom/living/crypto/Crypto-impl.js
@@ -1,0 +1,42 @@
+"use strict";
+
+const nodeCrypto = require("crypto");
+const DOMException = require("domexception/webidl2js-wrapper");
+
+// https://w3c.github.io/webcrypto/#crypto-interface
+class CryptoImpl {
+  constructor(globalObject) {
+    this._globalObject = globalObject;
+  }
+
+  // https://w3c.github.io/webcrypto/#Crypto-method-getRandomValues
+  getRandomValues(array) {
+    // Note: this rejects Float32Array and Float64Array.
+    if (!(array instanceof this._globalObject.Int8Array ||
+        array instanceof this._globalObject.Uint8Array ||
+        array instanceof this._globalObject.Uint8ClampedArray ||
+        array instanceof this._globalObject.Int16Array ||
+        array instanceof this._globalObject.Uint16Array ||
+        array instanceof this._globalObject.Int32Array ||
+        array instanceof this._globalObject.Uint32Array ||
+        array instanceof this._globalObject.BigInt64Array ||
+        array instanceof this._globalObject.BigUint64Array)) {
+      throw DOMException.create(this._globalObject, [
+        `The type of an object is incompatible with the expected type of the parameter associated to the object`,
+        "TypeMismatchError"
+      ]);
+    }
+
+    if (array.byteLength > 65536) {
+      throw DOMException.create(this._globalObject, [
+        `The quota has been exceeded.`,
+        "QuotaExceededError"
+      ]);
+    }
+
+    nodeCrypto.randomFillSync(array);
+    return array;
+  }
+}
+
+exports.implementation = CryptoImpl;
--- a/lib/jsdom/living/crypto/Crypto-impl.js
+++ b/lib/jsdom/living/crypto/Crypto-impl.js
@@ -11,25 +11,30 @@ class CryptoImpl {
 
   // https://w3c.github.io/webcrypto/#Crypto-method-getRandomValues
   getRandomValues(array) {
-    // Note: this rejects Float32Array and Float64Array.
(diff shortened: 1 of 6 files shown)


sjrd:
Are failures like
https://github.com/jsdom/jsdom/runs/5993755418?check_suite_focus=true#step:6:7255
and
https://github.com/jsdom/jsdom/runs/5993755195?check_suite_focus=true#step:6:7258
expected sometimes? The Node.js 14 build succeeded, but neither did 12 nor 16.

sjrd:
Rebased on top of master to get https://github.com/jsdom/jsdom/commit/4c7eed155e421c3b261667b6312d4c89d2a74c1b. The CI is green now. 🙂

lib/jsdom/living/crypto/Crypto-impl.js
```diff
@@ -0,0 +1,42 @@
...
+    // Note: this rejects Float32Array and Float64Array.
+    if (!(array instanceof this._globalObject.Int8Array ||
+        array instanceof this._globalObject.Uint8Array ||
+        array instanceof this._globalObject.Uint8ClampedArray ||
+        array instanceof this._globalObject.Int16Array ||
+        array instanceof this._globalObject.Uint16Array ||
+        array instanceof this._globalObject.Int32Array ||
+        array instanceof this._globalObject.Uint32Array ||
+        array instanceof this._globalObject.BigInt64Array ||
+        array instanceof this._globalObject.BigUint64Array)) {
```
domenic, line 23: This will fail on cross-realm instances, which is not really what the spec intends (though the spec is not precise).

The best way I can think of to do this is to check `array.constructor.name`.

lib/jsdom/living/crypto/Crypto-impl.js
```diff
@@ -0,0 +1,42 @@
...
+      throw DOMException.create(this._globalObject, [
+        `The type of an object is incompatible with the expected type of the parameter associated to the object`,
```
domenic, line 25: Let's do a more informative error message, e.g. "getRandomValues() only accepts integer typed arrays".

lib/jsdom/living/crypto/Crypto-impl.js
```diff
@@ -0,0 +1,42 @@
...
+        "TypeMismatchError"
+      ]);
+    }
+
+    if (array.byteLength > 65536) {
+      throw DOMException.create(this._globalObject, [
+        `The quota has been exceeded.`,
```
domenic, line 32: Again, a more informative error message. E.g. "getRandomValues() cannot generate more than 65536 bytes of random values; X bytes were requested"

lib/jsdom/living/crypto/Crypto-impl.js
sjrd: I must admit that I'm completely out of my depth when it comes to realms. I used `array.constructor.name` as suggested.

sjrd:
Thanks for the review. I added commit that should address your comments. Let me know if you'd like me to squash the commits instead.

domenic (approved):
LGTM. Will try and do a merge and release this weekend, although I'm worried about how apparently jsdom is really broken on Node v18 right now and might have trouble fitting this into my weekend open-source time... https://github.com/jsdom/jsdom/pull/3358#issuecomment-1107971243

lib/jsdom/living/crypto/Crypto-impl.js
```diff
@@ -0,0 +1,47 @@
...
+
+// https://w3c.github.io/webcrypto/#crypto-interface
+class CryptoImpl {
+  constructor(globalObject) {
+    this._globalObject = globalObject;
+  }
+
+  // https://w3c.github.io/webcrypto/#Crypto-method-getRandomValues
+  getRandomValues(array) {
+    // Note: this rejects Float32Array and Float64Array.
```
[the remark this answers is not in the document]
sjrd: Fixed. :)

SimenB:
Would love to see this in a minor release before dropping node 12 if possible (even though 12 is EOL later this week) 🙏

sjrd:
Thank you for merging and for the v20.0.0 release. :)

PR #3354, opened 2022-04-21
Update parse5 to 7.0.0

fb55:
A new version of parse5 is out, with fixes for many of the long-standing issues jsdom had. This removes a lot of the workarounds that were necessary before.


fb55:
There is currently an issue with optional chaining in Node 12. Please let me know if this is an issue — I've left it as jsdom has been fast with dropping Node versions that are End of Life, and there are only 9 days left before this applies to Node 12.

fb55:
Looks like there are a bunch of timeouts. I'd appreciate some pointers as to how to approach fixing them.

domenic:
Thanks so much for doing this!

No problem with dropping Node v12.

I don't see any timeouts (probably they are flaky... our test suite is annoying like that some times). `parsing/html_content_in_foreign_context.html` needs to get flipped from expected-fail to expected-pass (by deleting the line from `to-run.yml`).

The `frees up callback handles passed to setTimeout` is more mysterious. It seems unlikely to be your fault so we can try to figure it out and merge a fix that this can be rebased on... Any investigation is of course appreciated.

fb55:
Thanks for the help deciphering the results! I have removed the now-passing test from the expected failures. I also realised that I forgot to pass the (newly added) scripting flag to the `serialize*` methods.

The `setTimeout` issue is present in other current PRs as well, see eg. https://github.com/jsdom/jsdom/runs/6043730348/. I can reproduce it locally without any flakiness. My guess is that Node changed the way timers work in a minor update, which broke JSDOM (https://github.com/nodejs/node/pull/41231 looks like a good candidate).

PR #3393, opened 2022-07-03
I fixed some bugs.

LungZeno:
I wrote my project and then discovered some bugs. I fixed that bugs and then give my feedback.


domenic:
Sorry, we cannot accept changes like this without tests that fail before and pass after. Are you planning to add such tests?

domenic:
I cannot see any such tests in https://github.com/jsdom/jsdom/pull/3393/files ; did you forget to upload them?

LungZeno:
> I cannot see any such tests in https://github.com/jsdom/jsdom/pull/3393/files ; did you forget to upload them?

Sorry, I misunderstood that what you said is the original tests. I already wrote new tests.

test/api/from-url-in-browser.js
```diff
@@ -0,0 +1,58 @@
+/* globals location:false */
+"use strict";
+const { assert } = require("chai");
+const { describe, it } = require("mocha-sugar-free");
+
+const { JSDOM } = require("../..");
+
+require("chai").use(require("../chai-helpers.js"));
+
+describe("API: JSDOM.fromURL()", { skipUnlessBrowser: true, timeout: 5000 }, () => {
```
domenic, line 10: Instead of duplicating tests from from-url.js, please enable them in from-url.js instead. (By removing the top-level `{ skipIfBrowser: true }` and then adding it back for tests that don't pass in the browser.)

lib/jsdom/living/helpers/http-request.js
```diff
@@ -97,9 +97,18 @@ module.exports = class Request extends Writable {
...
-    this._currentRequest = request(this.currentURL, this._requestOptions, response => {
-      this._processResponse(response);
-    });
+    if (http.request.length === 2) {
+      const { href, origin, protocol, username, password, host, hostname, port, pathname, search } = urlOptions;
+      const urlOptionsClone =
+        { href, origin, protocol, username, password, host, hostname, port, pathname, search, path: pathname };
+      this._currentRequest = request({ ...urlOptionsClone, ...this._requestOptions }, response => {
+        this._processResponse(response);
+      });
```
domenic, line 106: The following version is simpler and also passes the tests:

```suggestion
      this._currentRequest = request({ ...urlOptions, ...this._requestOptions }, response => {
        this._processResponse(response);
      });
```

lib/jsdom/vm-shim.js
```diff
@@ -62,6 +153,52 @@ exports.runInContext = function (code, contextifiedSandbox, options) {
...
+          break;
+        case "LabeledStatement":
+          parent.body = getPatchBlockStatement(globalDeclaration);
+          break;
+        default:
+          throw new Error("don't know this parent type: " + parent.type);
+      }
+    }
+  }
+
```
domenic, line 201: The majority of what's added in this file is not tested. I was able to delete lines 180-197 with no tests failing, as well as lines 26-45 and 64-76.

It's also not clear what this code is trying to *do*. We already use https://github.com/ForbesLindesay/acorn-globals to detect globals and convert them into window assignments. I can see that it makes your new test pass, but why isn't acorn-globals good enough? Is there a bug in acorn-globals we should fix instead?

In fact, a lot of this code looks like it might be copied from acorn-globals?? What is going on??

test/api/from-url-in-browser.js
LungZeno: Yes, I merged them.

lib/jsdom/living/helpers/http-request.js
LungZeno: The spread operator only spreads own propertise, but href, origin, protocol, etc. all are setters and getters of prototype of urlOptions. { ...urlOptions } produces empty object. I fixed the bug of assert.isRejected and added test case that the simpler version should fail.

lib/jsdom/vm-shim.js
LungZeno: Yes, I addded test cases that the deletions should fail. 

findGlobals only finds all accessing of non-declared identifiers as accessing global variables, but "var" decarations and function declarations both out of all function scopes install new properties into global object. I don't ensure that it is a bug or feature of acorn-globals. The implemention of this vm-shim.js is to enclose the code into function scope, so all global declarations become local declarations, and then these new installation are dismissed. The patch is assignment being used to install these variables into global object.

I splited from my code in vm-shim.js that is modified from acorn-globals to create find-global-declarations.js and add license.

domenic: I don't quite understand. From what I can see of the existing vm-shim code, it runs `findGlobals` on `ast`, which is created by parsing `code`, which is the raw code itself. Nothing is enclosed in a function scope, that I can see. So it should work as-is.

Probably I am missing something, but I'd appreciate your help understanding before I merge this. Maybe there is a cleaner solution that doesn't involve so much duplicate work.

LungZeno: line 172:
`  return Function("window", rewrittenCode + suffix).bind(contextifiedSandbox)(contextifiedSandbox);`
The function `Function` create a function that body is `rewrittenCode + suffix`, so the code is in a function scope.

domenic: Right, but that is the rewritten code. findGlobals is run on the original code, so it should find all var declarations.

LungZeno: Right, but global declarations need differently rewriting. In addition, there is a bug that globally declared identifiers should also be found by findGlobals.

domenic: I don't understand your first sentence.

If there are bugs in the findGlobals npm package, can you submit a pull request to them to fix them, instead of duplicating the code with a bugfix into jsdom?

LungZeno: There are two bugs. I only partly fixed the bug about global declarations, but I don't fix the bug about globally declared identifiers yet.

LungZeno: > If there are bugs in the findGlobals npm package, can you submit a pull request to them to fix them, instead of duplicating the code with a bugfix into jsdom?
I removed the code about the bugfix of findGlobals npm package, and then I submitted a pull request to findGlobals package: https://github.com/ForbesLindesay/acorn-globals/pull/63

domenic:
I've merged in most of the changes here in 0c5aea76ec02dd87a915cb3c5e609df446fc27ea. What remains are the changes you've made to the globals-crawling. Those still seem best addressed by fixing acorn-globals, instead of jsdom. I saw you were working on some pull requests there, and encourage you to continue that.

If you think there's a reason why the changes really need to be in jsdom instead, please let me know and we can reopen and re-discuss. We'll also need some tests that fail before the change and pass after.

PR #3396, opened 2022-07-06
Allow subclasses in `crypto.getRandomValues()`

### LINKED ISSUE #3395: `crypto.getRandomValues()` throws on subclasses of `Uint8Array`
### Basic info:

- **Node.js version:** v16.13.0
- **jsdom version:** v20.0.0

### Minimal reproduction case

```js
const dom = new JSDOM();

class MyArray extends Uint8Array {}
const array = new MyArray(10);

// Throws
// TypeMismatchError: getRandomValues() only accepts integer typed arrays
dom.window.crypto.getRandomValues(array);
```

In browsers, the code
```js
class MyArray extends Uint8Array {}
const array = new MyArray(10);
crypto.getRandomValues(array);
```
works as expected.

### How can it be fixed?

I think that in `./living/crypto/Crypto-impl.js` the lines
```js
const typeName = array.constructor.name;
if (!(typeName === "Int8Array" ||
    typeName === "Uint8Array" ||
    typeName === "Uint8ClampedArray" ||
    typeName === "Int16Array" ||
    typeName === "Uint16Array" ||
    typeName === "Int32Array" ||
    typeName === "Uint32Array" ||
    typeName === "BigInt64Array" ||
    typeName === "BigUint64Array")) {
  throw DOMException.create(this._globalObject, [
    `getRandomValues() only accepts integer typed arrays`,
    "TypeMismatchError"
  ]);
}
```
should be replaced by
```js
if (!(array instanceof Uint8Array)) {
  throw DOMException.create(this._globalObject, [
    `getRandomValues() only accepts integer typed arrays`,
    "TypeMismatchError"
  ]);
}
```

sebamarynissen:
This PR fixes #3395, meaning that you can now pass subclasses of a `TypedArray` in `crypto.getRandomValues()`.

This PR also checks if the [webcrypto api](https://nodejs.org/api/webcrypto.html#web-crypto-api) is available in Node.js, and if so, uses that one. The webcrypto api is available in `>15`, so once jsdom drops support for Node 14, this can be the default implementation and no polyfilling is required anymore.


domenic (changes requested):
It seems like there are two changes here bundled together. One is to use the native webcrypto when it is available. The other is to change how type checks are done.

We should only do one of these in this PR. Probably the type checks.

Additionally, this is not a test of the JSDOM API, so it does not belong in test/api/. Instead, you need to add a web platform test, to the to-upstream directory. See https://github.com/jsdom/jsdom/blob/master/Contributing.md#tests for more information.

sebamarynissen:
@domenic About the web platform tests, I found that the repository has a file [WebCryptoAPI/getRandomValues.any.js](https://github.com/web-platform-tests/wpt/blob/master/WebCryptoAPI/getRandomValues.any.js). Does this means that I have to add a file with the same name in the `to-upstream` directory and then add my tests there?

domenic:
It's fine to have multiple files testing the same API. I'd suggest a filename like `getRandomValues-BufferSource-subclasses.any.js`

sebamarynissen:
@domenic As you suggested, I created a file for the wpt tests for subclasses and looks like they all pass. I also removed the check for using webcrypto if available, it can be added later on.

lib/jsdom/living/crypto/Crypto-impl.js
```diff
@@ -38,10 +37,30 @@ class CryptoImpl {
...
 }
 
 exports.implementation = CryptoImpl;
+
+// See #3395. Subclasses of TypedArrays should properly work, but we can't rely
+// on instanceof because Uint8Array may be different across different windows -
+// which can happen in JSDOM when running { runScripts: "dangerously" }. As a
+// solution, we imitate the behavior of instanceof by walking the proottype
+// chain.
+function getChain(array) {
```
domenic, line 52: It would be better for this function to have a clear name (I don't know what a "chain" is), like `getTypedArrayTypeName`, and for it to only return a string, not an array where its caller ignores the first value.

lib/jsdom/living/crypto/Crypto-impl.js
```diff
@@ -11,11 +11,10 @@ class CryptoImpl {
...
   // https://w3c.github.io/webcrypto/#Crypto-method-getRandomValues
   getRandomValues(array) {
-    // Note: this rejects Float32Array, Float64Array and DataView.
-    //
-    // We perform "instance tests" by comparing `array.constructor.name` so
-    // that the tests will be successful across realms.
-    const typeName = array.constructor.name;
+    // See #3395. Subclasses of TypedArrays should properly work, but we can't
+    // rely on instanceof because Uint8Array may be different across different
+    // windows.
```
domenic, line 16: This comment can be removed since it is redundant with the one 30 lines later.

test/web-platform-tests/to-upstream/WebCryptoAPI/getRandomValues-BufferSource-subclasses.any.js
```diff
@@ -0,0 +1,25 @@
...
+  "Uint16Array",
+  "Uint32Array",
+  "BigUint64Array"
+];
+
+for (const array of arrays) {
+  const ctor = globalThis[array];
+  test(() => {
+    class Buffer extends ctor {}
+    try {
```
domenic, line 19: This try/catch is unnecessary, and in fact would make the test harder to debug by hiding the thrown exception if it fails. Just add a comment like `// Must not throw for the test to pass`

domenic (changes requested):
Thanks for continuing to work on this!

test/web-platform-tests/to-upstream/WebCryptoAPI/getRandomValues-BufferSource-subclasses.any.js
sebamarynissen: Yeah, coming from a Mocha background, I was a bit surprised to not have something like `expect(fn).to.not.throw(Error)`, so that's what I went with. I simply removed the try/catch in 509564b.

domenic (approved):

Issue #2121, opened 2018-01-22
TypeError: Cannot read property '_location' of null


legraphista:
### Basic info:

- **Node.js version:** v8.9.1
- **jsdom version:** 11.6.0

### Minimal reproduction case

When there is a network request, if the window is closed after the message is sent but before the message is received, the code will continue to run and any window access will result in this kind of error: 
```js
(node:15703) TypeError: Cannot read property '_location' of null
    at Window.get location [as location] (/Users/Stefan/Desktop/Work/html-render/node_modules/jsdom/lib/jsdom/browser/Window.js:188:79)
    at fetch.then.then (./script.js:2:10)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)

```

In this example I polyfill fetch and make a request from an external script.
As soon as the window loads I close it. 

*main.js*
```js
const jsdom = require("jsdom");
const { JSDOM } = jsdom;
const window = (new JSDOM(`<html>
    <head></head>
    <body>
        <script src="./script.js"></script>
    </body>
</html>`, {
  runScripts: "dangerously",
  resources: 'usable',
})).window;

window.fetch = require('node-fetch');

window.onload = () => {
  window.close();
};
```
*script.js*
```js
fetch('https://api.ipify.org').then(x => x.text()).then((ip) => {
  window.location.href = ip;
});
```

### How does similar code behave in browsers?
Window closes and any network request is aborted.

Since fetch in this case is an external implementation and uses node's I/O callback, i'm not sure how that callback can be stopped.

domenic:
Note that this will fail to fetch ./script.js entirely, since you haven't given a base URL. So fetch is a red herring.

legraphista:
@domenic i was expecting that too, but for some reason, it fetches the script, runs it and the closes.
Here's the above example.
```bash
Stefans-MacBook:html-render Stefan$ node --trace-warnings main.js 
(node:16639) TypeError: Cannot read property '_location' of null
    at Window.get location [as location] (/Users/Stefan/Desktop/Work/html-render/node_modules/jsdom/lib/jsdom/browser/Window.js:188:79)
    at fetch.then.then (./script.js:2:10)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
(node:16639) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
[...]

PR #3397, opened 2022-07-07
Fixed some code prone to error

sinner:
The `idlUtils.implForWrapper` method can return null sometimes and it causes the error `Cannot read property '_location' of null`.

I'm not sure if you can use optional chaining feature, just let me know to make the changes in the PR. Another solution could be make that the `idlUtils.wrapperForImpl` method returns an empty object instead of a `null` value by default.

I think it could be related to this issue https://github.com/jsdom/jsdom/issues/2121
But, it happens when you run unit tests using jest and this library too.

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -372,10 +372,12 @@ function Window(options) {
       return external;
     },
     get location() {
-      return idlUtils.wrapperForImpl(idlUtils.implForWrapper(window._document)._location);
+      const _document = idlUtils.implForWrapper(window._document) || {};
+      return idlUtils.wrapperForImpl(_document._location);
     },
     get history() {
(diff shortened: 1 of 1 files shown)


mjb95:
We're noticing the same issue, which has been causing stability issues with our tests for the past few days.

domenic:
Sorry, this cannot be accepted without tests.

sinner:
Hi @domenic ! Sorry, could you guide to me where are the unit tests for this file in order to add the tests related to my suggested changes?

domenic:
https://github.com/jsdom/jsdom/blob/master/Contributing.md#tests should help!

mjb95:
Hi @sinner , did you ever get a chance to add these tests?

alann-sapone:
Sorry, but this issue still occurs very randomly ands breaks test suites :).
If a fix exists, it could be awesome to add it officially ^^'

Issue #3294, opened 2021-11-23
Any plans to add support for the HTMLDialogElement?


derekdon:
### Basic info:

- **Node.js version:** v16.13.0
- **jsdom version:** 18.1.0

### Minimal reproduction case

```js
/**
 * show and showModal don't exist (jsdom + HTMLDialogElement)
 */
// Suite('can have state toggled', async ({element}) => {
//   assert.is(element.open, false);
//   element.show();
//   assert.is(element.open, true);
//   element.close();
//   assert.is(element.open, false);
// });
```

https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement

zarahzachz:
Can anyone speak to the status of this? [`<dialog>`](https://caniuse.com/?search=dialog) and its show methods are now supported by all modern browsers, but implementation it is breaking my unit tests.

Receiving `TypeError: dialogNode.show is not a function`

Break seems to be related to [this change to `HTMLDialogElement`](https://github.com/jsdom/jsdom/commit/6d54df2404192257dfa9a4993621d93a226ebc80#diff-07afc48b207ab76e0fd1351ff5c7240e38efa417f2797b644f5b702b50713bee)

domenic:
The status is that we're waiting on you to submit a pull request implementing this feature :)

Asim-Tahir:
I think #3403 fixes this

ardunster:
It would be great if this could get some more visibility / priority. Using the `<dialog>` tag increases accessibility functionality (screen readers, keyboard control, automatic focusing on open, etc) over options like hiding/displaying a div to show a dialog and is supported in all the major browsers at this point. We used the workaround listed above to set the functions as prototypes but then also had to rewrite tests against implementation instead of behavior which feels terrible.

jtrein:
Adding to the workaround [posted](https://github.com/jsdom/jsdom/issues/3294#issuecomment-1196577616), if you are using Jest, this may help to do some assertions while we wait for [#3403](https://github.com/jsdom/jsdom/pull/3403) to be merged.

**Mocks**

```ts
  HTMLDialogElement.prototype.show = jest.fn(function mock(
    this: HTMLDialogElement
  ) {
    this.open = true;
  });

  HTMLDialogElement.prototype.showModal = jest.fn(function mock(
    this: HTMLDialogElement
  ) {
    this.open = true;
  });

  HTMLDialogElement.prototype.close = jest.fn(function mock(
[...]

PR #3403, opened 2022-07-22
Dialog Element Implementation

Hoekz:
This implements the Dialog Element according to the spec: https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element

There's a couple side effects of properly getting the web platform tests to pass:
 -  implemented a subset of the `autofocus` attribute functionality.
 - Computed styles was also only partially implemented in order to get the `inert-does-not-match-disabled-selector.html` test to pass.
 - The partial implementation of computed styles caused a few of the `pseudo-classes` tests to pass.


lib/jsdom/living/helpers/colors.js
```diff
@@ -0,0 +1,99 @@
+"use strict";
+
+const RGB_LOOKUP = {
```
domenic, line 3: Can you include everything from https://drafts.csswg.org/css-color/#named-colors ? Probably in a separate file.

I'm surprised this doesn't already exist on npm, but I couldn't find anything on https://www.npmjs.com/search?q=css%20named%20colors .

lib/jsdom/living/helpers/colors.js
```diff
@@ -0,0 +1,99 @@
...
+  olive: [0x80, 0x80, 0x00],
+  yellow: [0xFF, 0xFF, 0x00],
+  navy: [0x00, 0x00, 0x80],
+  blue: [0x00, 0x00, 0xFF],
+  teal: [0x00, 0x80, 0x80],
+  aqua: [0x00, 0xFF, 0xFF]
+};
+
+function colorKeywordToRGB(name) {
+  if (name.toLowerCase() in RGB_LOOKUP) {
```
domenic, line 23: This will give weird results if someone passes `name = "constructor"` or similar. You can use `Object.hasOwn(RGB_LOOKUP, name.toLowerCase())`.

lib/jsdom/living/helpers/focusing.js
```diff
@@ -102,3 +102,29 @@ exports.fireFocusEventWithTargetAdjustment = (name, target, relatedTarget, { bub
 
   target._dispatch(event);
 };
+
+exports.outsideUnopenedDialog = child => {
+  if (!child.parentNode) {
+    return true;
+  }
+
+  if (child.parentNode.tagName === "DIALOG" && !child.parentNode.open) {
```
domenic, line 111: Nit: Prefer `localName === "dialog"` instead of `tagName === "DIALOG"

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -77,14 +84,14 @@ function getCascadedPropertyValue(element, property) {
 
 // https://drafts.csswg.org/css-cascade-4/#specified-value
 function getSpecifiedValue(element, property) {
+  const { initial, inherited, coerce } = exports.propertiesWithResolvedValueImplemented[property];
   const cascade = getCascadedPropertyValue(element, property);
 
   if (cascade !== "") {
-    return cascade;
+    return coerce instanceof Function ? coerce(cascade) : cascade;
```
domenic, line 91: Let's just assume it's the right type and get a loud error if not:

```suggestion
    return coerce ? coerce(cascade) : cascade;
```

lib/jsdom/living/helpers/focusing.js
```diff
@@ -102,3 +102,29 @@ exports.fireFocusEventWithTargetAdjustment = (name, target, relatedTarget, { bub
...
+    return false;
+  }
+
+  return exports.outsideUnopenedDialog(child.parentNode);
+};
+
+exports.runAutoFocus = target => {
```
domenic, line 118: Hmm. The way this is done is very unlike the spec at https://html.spec.whatwg.org/#flush-autofocus-candidates , including when you call it. Do you think it'd be feasible to implement the spec here? I'm worried that this kind of implementation will cause people to raise more bugs about autofocus not working as it does in browsers, instead of our current status of it just not being implemented...

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -619,6 +632,12 @@ class DocumentImpl extends NodeImpl {
     for (const activeNodeIterator of this._workingNodeIterators) {
       activeNodeIterator._preRemovingSteps(oldNode);
     }
+    if (this._topLayer.includes(oldNode)) {
```
domenic, line 635: Can we run `_modalUnblock` here? it feels very duplicated.

lib/jsdom/living/nodes/HTMLDialogElement-impl.js
```diff
@@ -1,8 +1,123 @@
...
+
+class HTMLDialogElementImpl extends HTMLElementImpl {
+  constructor(globalObject, args, privateData) {
+    super(globalObject, args, privateData);
+    this._isModal = false;
+    this._returnValue = "";
+    this._previousFocusElement = null;
+  }
+
+  get open() {
```
domenic, line 21: This should be implemented using `[Reflect]` in the IDL file, so having it here is not necessary. Instead use `hasAttributeNS(null, "open")`.

lib/jsdom/living/nodes/HTMLDialogElement-impl.js
```diff
@@ -1,8 +1,123 @@
...
+    return this.getAttributeNS(null, "open") !== null;
+  }
+
+  get returnValue() {
```
domenic, line 25: The getters and setters for returnValue should not be necessary. Just rename the field from `_returnValue` to `returnValue`.

lib/jsdom/living/nodes/HTMLDialogElement-impl.js
```diff
@@ -1,8 +1,123 @@
...
+  show() {
+    if (this.open) {
+      return;
+    }
+
+    this.setAttributeNS(null, "open", "");
+
+    if (this._previousFocusElement) {
+      this._previousFocusElement.focus();
+    }
```
domenic, line 42: This seems incorrect. Step 3 of https://html.spec.whatwg.org/#dom-dialog-show stores the focused element, it doesn't focus it.

lib/jsdom/living/nodes/HTMLDialogElement-impl.js
```diff
@@ -1,8 +1,123 @@
...
+
+    this.setAttributeNS(null, "open", "");
+    this._isModal = true;
+
+    this._ownerDocument._modalBlock(this);
+
+    if (this._previousFocusElement) {
+      this._previousFocusElement.focus();
+    }
```
domenic, line 63: Same problem.

lib/jsdom/living/nodes/HTMLDialogElement-impl.js
```diff
@@ -1,8 +1,123 @@
...
+
+    const delegates = ["button", "input", "select", "textarea"]
+      .map(delegate => `${delegate}:not(:disabled):not([hidden])`);
+
+    const focusDelegates = this.querySelectorAll(
+      `${delegates.join(", ")}, [href], [tabindex]:not([tabindex="-1"]):not(dialog)`
+    );
+
+    const focusDelegate = [...focusDelegates].find(outsideUnopenedDialog);
```
domenic, line 84: These steps don't look like a good implementation of https://html.spec.whatwg.org/#focus-delegate

lib/jsdom/living/nodes/HTMLDialogElement-impl.js
```diff
@@ -1,8 +1,123 @@
...
+
+    this._dialogFocusSteps();
+  }
+
+  _dialogFocusSteps() {
+    if (this.inert) {
+      return;
+    }
+
+    if (runAutoFocus(this)) {
```
domenic, line 73: runAutoFocus doesn't look like a good implementation of https://html.spec.whatwg.org/#autofocus-delegate

lib/jsdom/living/nodes/HTMLDialogElement-impl.js
```diff
@@ -1,8 +1,123 @@
...
+    this._isModal = false;
+
+    if (result !== undefined) {
+      this._returnValue = result.toString();
+    }
+
+    this._ownerDocument._modalUnblock(this);
+
+    if (this._previousFocusElement) {
+      this._previousFocusElement.blur();
```
domenic, line 114: Step 6.3 of https://html.spec.whatwg.org/#close-the-dialog *focuses* the previous focused element; it doesn't blur it.

Also it's important to clear it before focusing it in case the focus event handler causes something to get messed up.

lib/jsdom/living/nodes/HTMLDialogElement-impl.js
```diff
@@ -1,8 +1,123 @@
...
+      this._previousFocusElement = null;
+    }
 
-class HTMLDialogElementImpl extends HTMLElementImpl { }
+    queueMicrotask(() => fireAnEvent("close", this, undefined, { returnValue: this.returnValue }));
```
domenic, line 118: This should be a task, not a microtask, so `setTimeout(, 0)` is better.

Also, I don't know why you're passing the return value here.

lib/jsdom/living/nodes/Node-impl.js
```diff
@@ -128,6 +128,15 @@ class NodeImpl extends EventTargetImpl {
     this._memoizedQueries = {};
     this._registeredObserverList = [];
     this._referencedRanges = new Set();
+    this._inert = false;
+  }
+
+  get inert() {
```
domenic, line 134: These getters and setters are not necessary; just remove the underscore from the declaration.

lib/jsdom/living/nodes/HTMLDialogElement-impl.js
```diff
@@ -1,8 +1,123 @@
...
+  close(result) {
+    if (!this.open) {
+      return;
+    }
+
+    this.removeAttributeNS(null, "open");
+    this._isModal = false;
+
+    if (result !== undefined) {
+      this._returnValue = result.toString();
```
domenic, line 108: The webidl2js wrapper layer will take care of ensuring this is always a string or undefined, so no need for toString().

domenic (commented):
Very exciting work! However the focusing stuff seems really mismatched from the spec. I'm hoping we can make it closer, to avoid problems in the future.

Hoekz:
Appreciate all the feedback on this! I've been looking at the focus algorithms you suggested implementing correctly (thank you for finding them, digging through the spec is a skill on its own apparently) and I think its doable, there's just a decent amount of requirements that get pulled in and life has gotten in the way of me working on it. I'm hoping to get some time to focus on it this week or at least find someone else who has some time to invest and help out.

Hoekz:
I think I've combed through all the autofocus tests and found which ones were passable and which ones were not, but probably best to double check that...I cannot get the [Script blocking style sheet](https://html.spec.whatwg.org/multipage/semantics.html#interactions-of-styling-and-scripting) implementation to work correctly and I'm not sure it can due to how some of the timing works right now, so it's been left to fail for now but is partially implemented.

Avi-Cohen-Nehemia:
Thank you so much for all the work done in this so far!
Would love to see this getting over the line in the near future! 🙏🏼

lib/jsdom/living/helpers/colors.js
```diff
@@ -1,26 +1,158 @@
...
+  wheat: [0xF5, 0xDE, 0xB3],
+  white: [0xFF, 0xFF, 0xFF],
+  whitesmoke: [0xF5, 0xF5, 0xF5],
+  yellow: [0xFF, 0xFF, 0x00],
+  yellowgreen: [0x9A, 0xCD, 0x32]
 };
 
 function colorKeywordToRGB(name) {
-  if (name.toLowerCase() in RGB_LOOKUP) {
+  if (Object.prototype.hasOwnProperty.call(RGB_LOOKUP, name.toLowerCase())) {
```
wesleycoder, line 155: Just a suggestion from a newbie here, but instead of a `Object.prototype.hasOwnProperty.call`, would a simple object access that gives `undefined` be less scarry and  in this case?
*I put a ternary here, but keeping the if can be more readable too.
```js
function colorKeywordToRGB(name) {
  const rgbKey = name.toLowerCase();

  return RGB_LOOKUP[rgbKey] ? `rgb(${RGB_LOOKUP[rgbKey].join(", ")})` : "";
}
```

domenic:
Unfortunately, I was unable to review this PR in a timely manner, since it's fairly large and I have less time for jsdom than I'd like. And now, a bunch of new web platform tests have been added, many of which fail. (Possibly due to changes in the dialog spec itself.) So we need someone to go through and figure out which ones we can pass, and update the implementation as appropriate.

I'm very sorry about this. I know it's a subpar contributor experience.

I'll try to help in a small way by pulling out the color-related changes into a separate PR, which I should be able to merge quickly on my own.

delijah:
Any progress here?

alexischarp:
Hey,
Thanks for this PR, any plans to merge it soon?

vcpablo:
Any news on this?

merlinz01:
Linking issue #3294.

anthonylex:
Is there any progress on getting this in?

MattyBalaam:
For the sake of your own sanity I would highly recommend testing behaviour like this in a real browser, nor a simulated envrionment.

PR #3404, opened 2022-07-23
Performance of appending option to select

TheHound:
Previous implementation of adding 'option' elements to 'select' elements caused an '(N*N/2)' issue where all previous elements would be added to an array to find either the last selected or first non disabled. Now this logic is run on just the 'option' element being appended.    

## Before

```
yarn benchmark -suites=html
yarn run v1.22.19
$ node ./benchmark/runner -suites=html
# html/select/text #
jsdom  x 1.56 ops/sec ±3.61% (8 runs sampled)
Done in 11.29s.
```

## After

```
yarn benchmark -suites=html
yarn run v1.22.19
$ node ./benchmark/runner -suites=html
# html/select/text #
jsdom  x 10.15 ops/sec ±6.53% (31 runs sampled)
Done in 7.23s.
```

Discovered this when I passed some html with a select with 20k option elements to JDSOM.

First pass at a fix and first time writing anything for nodejs so any and all feedback welcome including this not being the right way to solve the problem.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLSelectElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLSelectElement-impl.js
@@ -15,19 +15,6 @@ const { domSymbolTree } = require("../helpers/internal-constants");
 const { getLabelsForLabelable, formOwner, isDisabled } = require("../helpers/form-controls");
 const { parseNonNegativeInteger } = require("../helpers/strings");
 
-function toOptionArray(child) {
-  const array = [];
-  if (child._localName === "option" && !child.hasAttributeNS(null, "disabled")) {
-    array.push(child);
-  } else if (child._localName === "optgroup") {
-    for (const childOfGroup of domSymbolTree.childrenIterator(child)) {
-      if (childOfGroup._localName === "option" && !childOfGroup.hasAttributeNS(null, "disabled")) {
-        array.push(childOfGroup);
-      }
-    }
-  }
-  return array;
-}
 class HTMLSelectElementImpl extends HTMLElementImpl {
   constructor(globalObject, args, privateData) {
     super(globalObject, args, privateData);
@@ -56,8 +43,8 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
 
     this._labels = null;
 
-    // Essentially a cache for the current selected element for use while building a select by appending options
-    // See #_selectednessSettingAlgorithmWithKnownNewAddition
+    // Essentially a cache for the current selected element for use while building a select by appending options.
+    // See _selectednessSettingAlgorithmWithKnownNewAddition().
     this._lastSelectedOptionForFastPath = null;
   }
 
@@ -336,6 +323,20 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
 }
 
 mixin(HTMLSelectElementImpl.prototype, DefaultConstraintValidationImpl.prototype);
+
+function toOptionArray(child) {
+  const array = [];
+  if (child._localName === "option" && !child.hasAttributeNS(null, "disabled")) {
+    array.push(child);
+  } else if (child._localName === "optgroup") {
+    for (const childOfGroup of domSymbolTree.childrenIterator(child)) {
+      if (childOfGroup._localName === "option" && !childOfGroup.hasAttributeNS(null, "disabled")) {
+        array.push(childOfGroup);
+      }
+    }
+  }
+  return array;
+}
 
 module.exports = {
   implementation: HTMLSelectElementImpl
--- a/lib/jsdom/living/nodes/HTMLSelectElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLSelectElement-impl.js
@@ -15,6 +15,19 @@ const { domSymbolTree } = require("../helpers/internal-constants");
 const { getLabelsForLabelable, formOwner, isDisabled } = require("../helpers/form-controls");
 const { parseNonNegativeInteger } = require("../helpers/strings");
 
+function toOptionArray(child) {
+  const array = [];
+  if (child._localName === "option" && !child.hasAttributeNS(null, "disabled")) {
+    array.push(child);
+  } else if (child._localName === "optgroup") {
+    for (const childOfGroup of domSymbolTree.childrenIterator(child)) {
+      if (childOfGroup._localName === "option" && !childOfGroup.hasAttributeNS(null, "disabled")) {
+        array.push(childOfGroup);
+      }
+    }
+  }
+  return array;
+}
 class HTMLSelectElementImpl extends HTMLElementImpl {
   constructor(globalObject, args, privateData) {
     super(globalObject, args, privateData);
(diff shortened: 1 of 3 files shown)


lib/jsdom/living/nodes/HTMLSelectElement-impl.js
```diff
@@ -82,19 +83,69 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
       selected.forEach((option, index) => {
         option._selectedness = index === selected.length - 1;
       });
+      this._selectedOption = selected[selected.length - 1];
```
domenic, line 86: Shouldn't you also update this around line 77?

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
```diff
@@ -82,19 +83,69 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
...
+    }
+
+    // No selection so first option is selected
+    if ((!this._selectedOption || !this._selectedOption._selectedness) && this._displaySize === 1) {
+      this._selectedOption = optArr[0];
+      this._selectedOption._selectedness = true;
+    }
+  }
+
+  _toOptionArray(child) {
```
domenic, line 116: Let's factor this out into a standalone function that lives outside the class, instead of a member function, since it doesn't have any dependence on `this`.

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
```diff
@@ -82,19 +83,69 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
...
+    const array = [];
+    if (child._localName === "option") {
+      array.push(child);
```
domenic, line 119: It might be nicer to check for disabledness here, and then not need to check for it on line 94. Then the two branches of the toOptionArray function are symmetric.

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
```diff
@@ -82,19 +83,69 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
...
 
+  _updateSelected(child) {
+    if (this.hasAttributeNS(null, "multiple")) {
+      return;
+    }
+    if (child.hasAttributeNS(null, "disabled")) {
+      return;
+    }
+    const optArr = this._toOptionArray(child);
+    if (!optArr.length) {
```
domenic, line 98: Style nit: prefer numeric comparisons for `length`, so `if (optArray.length === 0)`, and below `if (selectedOptions.length > 0)`. (I realize there is some code elsewhere in this file that does not conform to that...)

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
```diff
@@ -82,19 +83,69 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
...
+
   _descendantAdded(parent, child) {
     if (child.nodeType === NODE_TYPE.ELEMENT_NODE) {
-      this._askedForAReset();
+      if (parent === this) {
+        this._updateSelected(child);
+      } else if (parent._localName === "optgroup") {
+        this._updateSelected(parent);
+      } else {
+        this._askedForAReset();
```
domenic, line 137: It's a bit weird to call `_askedForAReset()` when nothing actually asked for a form reset. Let's rename `_askedForAReset()` to `_selectednessSettingAlgorithm()` (see https://html.spec.whatwg.org/#selectedness-setting-algorithm), then add a quick wrapper `_askedForAReset() { this._selectednessSettingAlgortihm(); }`.

After you've done that, maybe it'd make sense to rename `_updateSelected()` to something related, like `_selectednessSettingAlgorithmWithKnownNewAddition(newChild)` or `_maybeRunSelectednessSettingAlgorithm(newChild)` or something? To make it clear it's an optimization related to minimizing how often we run the full selectedness setting algorithm.

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
```diff
@@ -42,6 +42,7 @@ class HTMLSelectElementImpl extends HTMLElementImpl {
     this._customValidityErrorMessage = "";
 
     this._labels = null;
+    this._selectedOption = null;
```
domenic, line 45: This doesn't seem like the right name, since sometimes `this._selectedOption._selectedness` will be false. And it would probably benefit from an explanatory comment. Maybe something like `_lastSelectedOptionForFastPath`?

domenic (commented):
Really nice work. A few small things to improve, mostly around making the code clearer for future readers.

TheHound:
Will get on that now - once done will amend commit unless I hear otherwise (i.e. to create a new commit)

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
TheHound: I didn't think was necessary due to mostly going via fast path when selectness is not set on incoming child but I think its safer to do so will add.

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
TheHound: Have made the change to check for disabledness in the `toOptionArray` here so the branches are symmetric and its consistent. Have left the check on line 94 as :    
1) Don't want to have to allocate a new array in here needlessly    
2) That also checks if the child is an optiongroup and the entire group is disabled which I am not currently doing here but could add.     
Not particularly passionate about any of that so if you also want it remove from the newly named `_selectednessSettingAlgorithmWithKnownNewAddition` just let me know.

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
TheHound: Done the rename and wrapper and the new name for the fast path method.

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
TheHound: Renamed to `_lastSelectedOptionForFastPath` and added a comment explaining reason for field.

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
TheHound: Done

lib/jsdom/living/nodes/HTMLSelectElement-impl.js
TheHound: Done

domenic (approved):
Really excellent work; I like the naming and structure a lot now. Thank you!

eps1lon:
This broke `HTMLSelectElement.selectedOptions` if `HTMLOptionElement.selected` is used: https://github.com/jsdom/jsdom/issues/3476

Revert in progress in https://github.com/jsdom/jsdom/pull/3475

Would be nice to reland with the new test passing though.

PR #3412, opened 2022-08-02
Add part as a global HTML Element attribute

### LINKED ISSUE #3366: 'part' property missing from Element prototype
### Basic info:

- **Node.js version:** 16.14.2
- **jsdom version:** 19.0.0

### Minimal reproduction case

```js
const { JSDOM } = jsdom;
const { window: { document }} = new JSDOM();

const el = document.createElement('div');
console.log('part' in el); // should be true
```

This will log `false`, when it should log `true`.

[Minimal browser repro](https://bl.ocks.org/nolanlawson/raw/a0ea8e885d4d49a5c667983d8f0fd01e/)

- Chrome 101: logs `true`
- Firefox 99: logs `true`
- Safari 15.4: logs `true`

For reference, please see [`Element.part` on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/part).

jmsjtu:
This PR resolves #3366 by adding the [part](https://developer.mozilla.org/en-US/docs/Web/API/Element/part) attribute as a global attribute to the HTML Element interface. 

While this PR does not include an implementation of the [css-shadow-parts spec](https://www.w3.org/TR/css-shadow-parts-1/#element-part-name-map), the [WPT's for css-shadow-parts](https://github.com/web-platform-tests/wpt/tree/master/css/css-shadow-parts) has been included as there is a test to check for the [part name in the idl](https://github.com/web-platform-tests/wpt/blob/master/css/css-shadow-parts/part-name-idl.html).

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -545,6 +545,16 @@ class ElementImpl extends NodeImpl {
     const matcher = addNwsapi(this);
     return matcher.closest(selectors, idlUtils.wrapperForImpl(this));
   }
+
+  get part() {
+    if (this._part === undefined) {
+      this._part = DOMTokenList.createImpl(this._globalObject, [], {
+        element: this,
+        attributeLocalName: "part"
+      });
+    }
+    return this._part;
+  }
 }
 
 mixin(ElementImpl.prototype, NonDocumentTypeChildNode.prototype);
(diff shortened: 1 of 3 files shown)


domenic:
Adding this without actually implementing CSS shadow parts breaks feature detection, so I don't think this is a great idea...

jmsjtu:
@domenic thanks for the feedback, I'll work on implementing CSS shadow parts based on the [spec](https://www.w3.org/TR/css-shadow-parts-1/) and update this PR.

domenic:
Closing for now; please feel free to ping when you have done such an implementation, and we'll reopen.

Issue #3419, opened 2022-08-22
CookieJar - Tough-Cookie breaking change 


arpowers:
ToughCookie just released a breaking change in a minor release and it's breaking all related tests using JSDom (when in localhost context)

Cause https://github.com/salesforce/tough-cookie/issues/246

Blown up tests look like this: 
<img width="1147" alt="Screen Shot 2022-08-22 at 3 44 12 PM" src="https://user-images.githubusercontent.com/308134/186032320-45e08f37-2d41-46fa-ac25-64ffcb424061.png">

Is it possible to provide the options required to the dependency? 

Otherwise, recommend pinning this particular dependency to version 4.0 until you have a chance to update

CSchulz:
You can configure the cookie jar via the jsdom options, it is called cookieJar.

Here the example we are using for our jsdom jest environment:

    testEnvironmentOptions: {
      url: 'http://localhost/',
      cookieJar: new toughCookie.CookieJar(undefined, {
        // first option copied from https://github.com/jsdom/jsdom/blob/master/lib/api.js
        // // jsdom cookie jars must be loose by default
        looseMode: true,
        allowSpecialUseDomain: true,
      }),
    },

arpowers:
Only problem for us is that we are using Vitest which doesn't provide an interface (that I've seen)

redbugz:
If tough-cookie is going to be picky about this now, it would seem that jsdom should configure their internal cookieJar to be loose and allow special domains by default, as that seems a common use case for testing is to use special use domains

andreschort:
Hi, this had me spinning a few hours today.
If you are using npm v8 you can use the overrides in package.json (https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides)
Like this:
`"overrides": {
    "tough-cookie": "4.0.0"
  },`

domenic:
I'm confident tough-cookie will revert their breaking change. I have no plans to address this on the jsdom side; please direct your attention to their repository instead, to urge them to fix their breaking change sooner.

CSchulz:
If anyone still has trouble with the new 4.1.1 release, check your used url for your testing environment. It needs to be a full qualified domain name in case of localhost you can use localhost.local

nagakiran:
> You can configure the cookie jar via the jsdom options, it is called cookieJar.
> 
> Here the example we are using for our jsdom jest environment:
> 
> ```
> testEnvironmentOptions: {
[...]

PR #3420, opened 2022-08-23
Removed caret from tough-cookie to mitigate v4.1.0 breaking change

archer56:
See issues: https://github.com/jsdom/jsdom/issues/3419 and https://github.com/salesforce/tough-cookie/issues/246 for more information.

v4.1.0 of tough-cookie is a breaking change,  and has broken many jest tests and builds.
An immediate solution is to fix the version of tough-cookie to v4.0.0 which is what this PR has done.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "parse5": "^7.0.0",
     "saxes": "^6.0.0",
     "symbol-tree": "^3.2.4",
-    "tough-cookie": "^4.0.0",
+    "tough-cookie": "4.0.0",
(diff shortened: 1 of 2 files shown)


elmarbeckmann (approved):

domenic:
I'm confident tough-cookie will revert their breaking change. I have no plans to address this on the jsdom side; please direct your attention to their repository instead, to urge them to fix their breaking change sooner.

AbhaySBhosale:
@domenic tough-cookie have release new version with the fix as per comment on respective issue.

Issue #3433, opened 2022-09-16
XmlHttpRequest timeouts are not cleaned up on network errors


rayvincent2:
### Basic info:

- **Node.js version:** 16.14.2
- **jsdom version:** 20.0.0

### Minimal reproduction case
When making more than one XMLHttpRequest in a test suite and one of them has a timeout applied and fails with a network error, the timeout is not cleared.

When running the test code below in Jest, you'll see that the timeout is not cleared and jest will not exit the test until the timeout naturally completes in 120 seconds

**`./xhr-timeout.test.js`**
```js
/**
 * @jest-environment jsdom
 */
 test("should create a long running timeout for 120 seconds", () => {
  const xhr1 = new XMLHttpRequest();
  xhr1.open("GET", "https://example.com");
  xhr1.send();
  
  const xhr2 = new XMLHttpRequest();
  xhr2.timeout = 120000;
  xhr2.open("GET", "https://a-domain-that-causes-a-network-error");
  xhr2.send();
});
```

**`./package.json`**
```json
{
  "name": "xhr-test",
  "version": "1.0.0",
  "scripts": {
    "test": "jest"
  },
  "dependencies": {
    "jest": "^29.0.3",
    "jest-environment-jsdom": "^29.0.3"
  }
}
```
**`Running the tests`**
```bash
npm install
npm test

// Optionally run this to see the open handles
npm test -- --detectOpenHandles
```

domenic:
This issue uses a third-party package (Jest), so we cannot help debug it. If others encounter it and can provide an example that uses jsdom only, we'll be happy to reopen.

PR #3434, opened 2022-09-16
Cleanup xhr timeouts on error

rayvincent2:
Fixes issue #3433 by ensuring that the timeouts are cleared if an error occurs


domenic:
Thanks, but we cannot accept this change without a test that fails before and passes afterward.

rayvincent2:
Very good.  I'll add a test.  I'm assuming that things look good otherwise?

domenic:
I can't really tell, because without a test I don't know if this fixes the issue :). But assuming it does, yes, this seems simple enough.

domenic:
Closing for now, but if you add a test we'll reopen.

PR #3436, opened 2022-09-22
add ResizeObserver polyfill to the window object. fixes #3368

### LINKED ISSUE #3368: ResizeObserver is not defined
### Basic info:

- **Node.js version:** v16.14.0 
- **jsdom version:** 19.0.0 

### Minimal reproduction case

```js

const dom = new JSDOM(`<!DOCTYPE html><p>Hello world</p>`);
console.log("RESIZE OBSERVER", dom.window.ResizeObserver);
```

In jsdom ResizeObserver is not defined on `window` object. In browsers it is defined -> https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

The lack of ResizeObserver in jsdom breaks unit tests in my project.

mreinstein:


domenic:
Sorry, but in jsdom we don't use polyfills; we implement the functionality ourselves.

PR #3440, opened 2022-10-03
Update readme document

zjffun:
* Fix the case of "MIME"
* Update the ResourceLoader example
* Update the link of jsdom-devtools-formatter


README.md
```diff
@@ -165,6 +165,12 @@ You can further customize resource fetching by subclassing `ResourceLoader` and
 ```js
 class CustomResourceLoader extends jsdom.ResourceLoader {
   fetch(url, options) {
+    const { origin } = new URL(url);
+
+    if (origin !== "https://example.com") {
+      return null;
+    }
```
domenic, line 172: What is the purpose of this change?

zjffun: Because I find this example is about "only returns results for requests to a trusted origin". https://github.com/jsdom/jsdom/blob/master/README.md?plain=1#L163

PR #3442, opened 2022-10-09
Feat add promiserejection event class

Smrtnyk:
In our repo we are missing this event constructor for testing some of the error capturing mechanisms we build in jest.

I am not sure actually about how to test this if it is needed?
Also not sure if this is all I have to do?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 .DS_Store
 .svn
 .*.swp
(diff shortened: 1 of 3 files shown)


domenic:
I don't think this is a good idea, since this doesn't actually add support for firing the event appropriately. So adding this without actually adding the event breaks feature detection.

But, to fire the event appropriately, we'd need to implement https://html.spec.whatwg.org/#the-hostpromiserejectiontracker-implementation . Which I believe is not possible using Node's vm module. So I don't think jsdom can ever add this.

I'll close this, but if you find some way to make all the tests in https://github.com/web-platform-tests/wpt/tree/master/html/webappapis/scripting/processing-model-2/unhandled-promise-rejections (which are included in jsdom, but disabled right now) then we can reconsider.

Smrtnyk:
thank you for the response, since it was just about one test suite, probably is not worth it anyway
we just stubbed it on our own

PR #3443, opened 2022-10-11
add missing abort method for sad path

whamtet:
Thanks @domenic and team for the great project,

Your XMLHttpRequest implementation wraps an internal client called `client`.  When we abort the request it aborts the internal client [here](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/xhr/XMLHttpRequest-impl.js#L364).  Now this internal client is created in `xhr-utils.createClient` which has a number of paths for data url, file url etc.  At the catch statement [here](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/xhr/xhr-utils.js#L326) we still assign client, but it is an `EventEmitter` which lacks an abort method.  If we then try to abort the wrapping XMLHttpRequest it throws a strange exception which doesn't tell the user about the true root cause.  I suggest adding an empty abort method to prevent confusion.

I want to add a unit test as well, but I'm not sure how to trigger the catch block.  If you agree with this change I can do so using e.g. Sinon or any other way that you suggest.

### DIFF
--- a/lib/jsdom/living/xhr/xhr-utils.js
+++ b/lib/jsdom/living/xhr/xhr-utils.js
@@ -327,6 +327,9 @@ function createClient(xhr) {
       const eventEmitterclient = new EventEmitter();
       process.nextTick(() => eventEmitterclient.emit("error", e));
       eventEmitterclient.end = () => {};
+      eventEmitterclient.abort = () => {
+        // do nothing
+      };
       return eventEmitterclient;
     }
   }
--- a/test/web-platform-tests/to-upstream/xhr/abort-with-error.any.js
+++ b/test/web-platform-tests/to-upstream/xhr/abort-with-error.any.js
@@ -1,0 +1,15 @@
+// META: title=XMLHttpRequest: abort() still works when error thrown internally
+
+      var test = async_test();
+
+      test.step(function() {
+        var client = new XMLHttpRequest();
+
+        client.open("GET", "invalid-protocol://example.com", true);
+        client.onabort = test.step_func(function() {
+          test.done();
+        });
+        
+        client.send(null);
+        client.abort();
+      });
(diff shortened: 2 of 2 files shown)


domenic:
Thanks for the fix!

To add a test, and to ensure this is fixing a real bug, we need to expose the bug in a web page that runs inside jsdom. See https://github.com/jsdom/jsdom/blob/master/Contributing.md#web-platform-feature-tests for how to write those.

whamtet:
According to [Contributing.md](https://github.com/jsdom/jsdom/blob/master/Contributing.md) 

All tests for web platform features (as opposed to features of jsdom itself, such as the JSDOM() constructor) should be in [web-platform-tests](https://github.com/web-platform-tests/wpt) format.

How do you decide whether a test should be an api test or web platform test?

domenic:
XHR is a web platform feature, so it needs a web platform test.

whamtet:
Sorry for the delay in updating @domenic , I do think this is quite an important bugfix however.  The issue is that if we make an invalid XMLHttpRequest and then abort it, the code triggers the internal abort method [here](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/xhr/XMLHttpRequest-impl.js#L365).  On the error path this will be the `EventEmitter` defined [here](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/xhr/xhr-utils.js#L327).  It needs its own abort method (no-op) to prevent user confusion.

In my use case the error was triggered by a third party library, but it was throwing the exception due to the missing abort method.  It looks like I need to create a new upstream web platform test for this one, please advice if you would like me to do it differently.

lib/jsdom/living/xhr/xhr-utils.js
```diff
@@ -327,6 +327,9 @@ function createClient(xhr) {
       const eventEmitterclient = new EventEmitter();
       process.nextTick(() => eventEmitterclient.emit("error", e));
       eventEmitterclient.end = () => {};
+      eventEmitterclient.abort = () => {
```
whamtet, line 330: this method has to exist to be consistent with the other code paths.

domenic:
Great, the test looks perfect. I'll try to find some jsdom time over the weekend to verify that the test fails without your patch, and then do a merge and release.

domenic (approved):
Thank you!

whamtet:
Thanks @domenic!!!

PR #3448, opened 2022-10-19
fixed a syntax error on line 170 & 175

Skad0sh:
removed the '.?' on line 170 and 175 which was causing a syntax error on import


domenic:
This is valid JavaScript syntax.

PR #3449, opened 2022-10-20
fix(getModifierState): query the correct property for alt, ctrl, meta, and shift keys

### LINKED ISSUE #3126: KeyboardEvent.getModifierState does not appropriately trigger in jsdom
### Basic info:

- **Node.js version:** 12.13.0
- **jsdom version:** 16.4.0

### Minimal reproduction case

```js
const dom = new JSDOM(`
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width">
    <title>JS Bin</title>
  </head>
  <body>
    <div id="text" />
  </body>
</html>
`);

dom.window.addEventListener('keydown', e => {
  if (e.getModifierState('Control')) {
    dom.window.document.getElementById('text').innerHTML = 'you pressed control too';
  } else {
    dom.window.document.getElementById('text').innerHTML = 'you did not press with control';
  }
});

const e = new dom.window.KeyboardEvent('keydown', { key: 'a', ctrlKey: true });
dom.window.dispatchEvent(e);
console.log(dom.window.document.getElementById('text').textContent);

const e2 = new dom.window.KeyboardEvent('keydown', { key: 'a', ctrlKey: true, modifierControl: true });
dom.window.dispatchEvent(e2);
console.log(dom.window.document.getElementById('text').textContent);
```

You can see in the output it is following the `else` clause because it does not think that ctrl key was also sent.

You can try the buttons as they manually fire the event similar to what happens in JSDOM, but also see it working as expected with actual keyboard events.
https://jsbin.com/ziqeyapuxo/edit?html,js

I also tried something similar to the test you have [here](https://github.com/jsdom/jsdom/blob/master/test/web-platform-tests/to-upstream/uievents/getModifierState.html) but that did not work for getting `getModifierState` to work, I'm unclear if I need to setup some options maybe to enable this?

juzerzarif:
This fixes #3126 

Chrome, Firefox, and Edge all return true for `getModifierState('Control')` when the `ctrlKey` property is true. Same for alt, meta, and shift modifiers as well. The spec doesn't seem to mention how the method should determine if a modifier is activated (https://www.w3.org/TR/uievents/#dom-keyboardevent-getmodifierstate) and since the `xKey` properties are the only ones on the instance that carry information about the state of these 4 modifiers, makes sense to look at those props for these 4 modifiers.

### DIFF
--- a/lib/jsdom/living/events/EventModifierMixin-impl.js
+++ b/lib/jsdom/living/events/EventModifierMixin-impl.js
@@ -11,6 +11,12 @@ class EventModifierMixinImpl {
   // these invalid modifiers will be undefined on `this` (thus `false` after casting it to boolean), we don't need to do
   // extra checking for validity.
   getModifierState(keyArg) {
+    if (keyArg === "Control") {
+      return Boolean(this.ctrlKey);
+    }
+    if (["Alt", "Meta", "Shift"].includes(keyArg)) {
+      return Boolean(this[`${keyArg.toLowerCase()}Key`]);
+    }
     return Boolean(this[`modifier${keyArg}`]);
   }
 }
--- a/test/web-platform-tests/to-upstream/uievents/getModifierState.html
+++ b/test/web-platform-tests/to-upstream/uievents/getModifierState.html
@@ -26,6 +26,34 @@ test(() => {
 }, "getModifierState false");
 
 test(() => {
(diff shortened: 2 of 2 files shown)


domenic (approved):
Great stuff; thank you!

PR #3453, opened 2022-10-24
Fix deferred scripts added after DOMContentloaded not being executed

### LINKED ISSUE #3451: Dynamically created defered scripts don't load
### Basic info:

- **Node.js version:** v18.10.0
- **jsdom version:** v20.0.1

### Minimal reproduction case

```js
const jsdom = require('jsdom')

const sleep = duration =>
  new Promise(resolve => {
    setTimeout(resolve, duration)
  })

const resources = new (class CustomResourceLoader extends jsdom.ResourceLoader {
  fetch(url) {
    // Override the contents of this script to do something unusual.
    if (url === 'http://localhost/script.js') {
      return Promise.resolve(
        Buffer.from(`
document.body.innerHTML = "it works!";
`)
      )
    }

    return null
  }
})()

;(async function () {
  const dom = new jsdom.JSDOM(
    /* HTML */ `<!DOCTYPE html>
      <html>
        <head>
          <meta charset="utf-8" />
          <title>Welcome</title>
          <script>
            console.log('hello')
            window.init = () => {
              const script = document.createElement('script')
              script.defer = true // <====================   comment out this line to make this work
              script.src = '/script.js'
              // firstScript.parentNode.insertBefore(script, firstScript)
              document.querySelector('head').appendChild(script)
            }
            ;['complete', 'interactive'].includes(document.readyState)
              ? window.init()
              : window.addEventListener('DOMContentLoaded', window.init)
          </script>
        </head>
        <body>
          [this should get replaced]
        </body>
      </html> `,
    {
      url: `http://localhost/`,
      runScripts: 'dangerously',
      resources,
    }
  )
  while (!dom.serialize().includes('it works')) {
    await sleep(100)
    console.log('.')
  }
  console.log('it works!')  // <=========== it never reaches this point
})()

```

The deferred script does run

kuba-orlik:
This fixes #3451

### DIFF
--- a/lib/jsdom/browser/resources/per-document-resource-loader.js
+++ b/lib/jsdom/browser/resources/per-document-resource-loader.js
@@ -68,7 +68,10 @@ module.exports = class PerDocumentResourceLoader {
 
     if (element.localName === "script" && element.hasAttributeNS(null, "async")) {
       this._asyncQueue.push(request, onLoadWrapped, onErrorWrapped, this._queue.getLastScript());
-    } else if (element.localName === "script" && element.hasAttributeNS(null, "defer")) {
+    } else if (
+      element.localName === "script" &&
+        element.hasAttributeNS(null, "defer") &&
+        this._document.readyState !== "interactive") {
       this._deferQueue.push(request, onLoadWrapped, onErrorWrapped, false, element);
     } else {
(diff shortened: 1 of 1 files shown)


domenic:
Oh, thanks, this looks nice! Can you add a test (in web platform tests format), per Contributing.md?

PR #3454, opened 2022-10-25
Don't look up the global custom elements registry through its sandbox-writable property.

bicknellr:
jsdom currently finds the global custom elements registry by reading the sandbox-writable `.customElements` property of the global object. This means that code running in the sandbox can break jsdom by overwriting or deleting that property. This PR updates jsdom to store and look up a global object's custom element registry through a `_`-prefixed property instead (similar to `._document`).


domenic (approved):
Thanks! One day we should really sort out our Window mess...

Issue #3363, opened 2022-05-02
[Bug]: structuredClone is not defined


tbertrand7:
### Basic info:

- **Node.js version:** 18.0.0
- **jsdom version:** 19.0.0

### Minimal reproduction case

Running a jest test that calls the structuredClone global using the jsdom environment

```js
structuredClone();
```

jacygao:
As an alternative before this is fixed, you could mock structuredClone in your test file:
```
global.structuredClone = jest.fn(val => {
    return JSON.parse(JSON.stringify(val));
});
```

TD-DO:
```
global.structuredClone = val => {
  return JSON.parse(JSON.stringify(val))
}
```
is what worked for me.
But is this going to be fixed?

AideTechBot:
I can vouch that this is still an issue using `jest-environment-jsdom`

amrutraj:
> As an alternative, you could mock structuredClone in your test file:
> 
> ```
> global.structuredClone = jest.fn(val => {
>     return JSON.parse(JSON.stringify(val));
> });
> ```

This won't work for copying functions as values of the keys. For the rest, no problem.

mftw:
> > As an alternative, you could mock structuredClone in your test file:
> > ```
> > global.structuredClone = jest.fn(val => {
> >     return JSON.parse(JSON.stringify(val));
> > });
> > ```
> 
> This won't work for copying functions as values of the keys. For the rest, no problem.

functions aren't supported by the structured clone algo anyway. https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types

Aleksandr-JS-Developer:
Any updates here?
---
> global.structuredClone = jest.fn(val => {
>    return JSON.parse(JSON.stringify(val));
> });

This method is good for objects that can be converted to JSON. This doesn't work for regular expressions, for example.

---

A smarter solution seems to be installing a polyfill.

1. Installing core-js (it is clear that if it is already installed, you do not need to install it again)
```bash
npm install core-js --save-dev
```

2. In the test file, add the line
```javascript
import 'core-js/stable';
// ...
// tests...
```

This is suitable if you are using the `jsdom` environment (```"testEnvironment": "jsdom"```). For other cases, the exact method may be slightly different, but the basic meaning is the same. This is the simplest one I've found. You can also try setting up Babel to automate this.

domenic:
[...]

PR #3459, opened 2022-11-03
Add the global structuredClone() method

diceride:
Adds the global [structuredClone()](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) method to the Window scope. The method was  introduced in Node.js version 17.0.0 ([release notes](https://nodejs.org/es/blog/release/v17.0.0/)). The available polyfills of the  structuredClone() method for Node.js are considered to be incomplete in comparison to the web browser JavaScript runtime implementations.

Example - Error message when the structuredClone() method is not defined in the global scope of Node.js (Node.js < 17)

```
ReferenceError: structuredClone is not defined
```

Users can provide their own polyfill of the structuredClone() method  (Node.js < 17)

```js
global.structuredClone = (value, options) => ...
```

References

- Specification: [HTML Standard](https://html.spec.whatwg.org/multipage/structured-data.html#dom-structuredclone)
- MDN: [structuredClone()](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone)
- Node.js Documentation: [structuredClone()](https://nodejs.org/docs/latest/api/globals.html#structuredclonevalue-options)

Resolves https://github.com/jsdom/jsdom/issues/3363

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -745,6 +745,10 @@ function Window(options) {
     return result;
   };
 
+  this.structuredClone = function (...args) {
+    return structuredClone(...args);
+  };
+
   this.stop = function () {
     const manager = idlUtils.implForWrapper(this._document)._requestManager;
     if (manager) {


diceride:
@domenic

domenic:
This does not implement the specification linked, or have any tests.

diceride:
Excuse me, but the structuredClone() method is a global object in the Node.js runtime, and the implementation simply reflects this method into the VM context. How does this "not implement the specification linked"?

domenic:
Because Node's structured clone operates on Node objects, not on jsdom objects.

diceride:
Alright, fair enough. So, according to the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#webapi_types), the following JSDOM Web/API types need to be covered differently, otherwise their non-enumerable properties, property accessors, and object prototypes are not preserved and they will be cloned as plain JavaScript objects.

- [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob)
- [DOMException](https://developer.mozilla.org/en-US/docs/Web/API/DOMException)
- [File](https://developer.mozilla.org/en-US/docs/Web/API/File)
- [FileList](https://developer.mozilla.org/en-US/docs/Web/API/FileList)

I am looking into this and will add some tests as requested.

diceride:
Please confirm whether this is what you have meant.

domenic:
You shouldn't need to add too many tests. They should already exist in web platform tests. See CONTRIBUTING.md. The main goal here is to enable https://github.com/web-platform-tests/wpt/blob/master/html/webappapis/structured-clone/structured-clone.any.js ; if you can get that passing, the PR is probably ready.

acywatson:
@domenic do you have any ideas on how I could move this forward? It seems like core-js has an implementation that follows the spec, AFAICT. Should we copy that code over, modifying it as necessary? It's unfortunate to have it in two places, but I'm not sure if you're interested in taking a dependency on core-js for this. Either way, I see no real reason to re-implement this from scratch.

domenic:
I think I gave pretty clear instructions above, about how to proceed. If you want to see someone following them, check out #3521 .

acywatson:
> I think I gave pretty clear instructions above, about how to proceed. If you want to see someone following them, check out https://github.com/jsdom/jsdom/pull/3521 .

I'll have to respectfully disagree - if it was clear, then I wouldn't be asking a clarifying question about it :). I think that's a reasonable thing to do before starting work on something like this.

Great to see someone else is working on this, though and I appreciate the response.

PR #3472, opened 2022-12-17
Fix: Add support to alias x and y to MouseEvent

### LINKED ISSUE #3470: MouseEvent: x and y support
### Basic info:

- **jsdom version:** 20.0.3

### Minimal reproduction case

jsdom

```js

const dom = new JSDOM(`<div></div>`, {});

// @ts-ignore

const div = document.querySelector("div");

const handleClick = (event) => {
  const { x, y, clientX, clientY } = event;
  console.log(`x: ${x}`);
  console.log(`y: ${y}`);
  console.log(`clientX: ${clientX}`);
  console.log(`clientY: ${clientY}`);
};
div.addEventListener("click", handleClick);

const event = new MouseEvent("click", {
  clientX: 10,
  clientY: 20,
});
div.dispatchEvent(event);
```

`x` and `y` properties are printed to the console

chrome

### Specification

Link: https://www.w3.org/TR/cssom-view-1/#extensions-to-the-mouseevent-interface

```
The x attribute must return the value of [clientX](https://www.w3.org/TR/cssom-view-1/#dom-mouseevent-clientx).

The y attribute must return the value of [clientY](https://www.w3.org/TR/cssom-view-1/#dom-mouseevent-clienty).
```

ViniciusFXavier:
# Add support to alias x and y to MouseEvent

Fix created to solve the problem where it does not return the x and y attributes of the MouseEvent interface.

When the following script runs, we get the attributes as a result.
```js
const event = new MouseEvent("click", {
  clientX: 10,
  clientY: 20,
});
const { x, y, clientX, clientY } = event;
console.log(`x: ${x}`);
console.log(`y: ${y}`);
console.log(`clientX: ${clientX}`);
console.log(`clientY: ${clientY}`);
```
image

image
Interface of reference: https://www.w3.org/TR/cssom-view-1/#extensions-to-the-mouseevent-interface

image
Reference: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent

This fixes #3470

### DIFF
--- a/lib/jsdom/living/events/MouseEvent-impl.js
+++ b/lib/jsdom/living/events/MouseEvent-impl.js
@@ -33,6 +33,8 @@ class MouseEventImpl extends UIEventImpl {
     this.screenY = screenY;
     this.clientX = clientX;
     this.clientY = clientY;
+    this.x = clientX;
+    this.y = clientY;
     this.ctrlKey = ctrlKey;
     this.altKey = altKey;
     this.shiftKey = shiftKey;
--- a/lib/jsdom/living/events/MouseEvent.webidl
+++ b/lib/jsdom/living/events/MouseEvent.webidl
@@ -7,6 +7,8 @@ interface MouseEvent : UIEvent {
   readonly attribute long screenY;
   readonly attribute long clientX;
   readonly attribute long clientY;
+  readonly attribute long x;
+  readonly attribute long y;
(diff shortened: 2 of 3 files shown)


domenic:
Looks like #3484 supersedes this and, unlike this PR, passes the build, so I'll close this.

Issue #3234, opened 2021-08-25
getComputedStyle performance


just-boris:
### Basic info:

- **Node.js version:** n/a
- **jsdom version:** any version is affected

### Minimal reproduction case

```js
const div = document.createElement('div');
document.body.appendChild(div);

console.time('getComputedStyle');
getComputedStyle(div);
console.timeEnd('getComputedStyle');
```

### How does similar code behave in browsers?

This code take a fraction of millisecond in real browsers, but 100 times slower in JSDOM. There are some libraries in the ecosystem heavily relying on this API and the tests become slow because of that.

Related issues: 

https://github.com/testing-library/dom-testing-library/issues/820 
https://github.com/theKashey/react-focus-lock/issues/151

just-boris:
As a possible solution, I think, it could cache the result of this call

https://github.com/jsdom/jsdom/blob/9dae17bf0ad09042cfccd82e6a9d06d3a615d9f4/lib/jsdom/browser/Window.js#L799-L807

External stylesheets usually do not change dynamically and this could give a significant improvement

donaldpipowitch:
I recently ran CPU profiling on our Jest tests (using testing-library) and indeed it looks like getComputedValue/getSpecifiedValue/getComputedStyle really eats up a lot of time.

nickroth:
We recently updated to use jsdom 16 in our jest tests and noticed a big slow down that I believe is related to this. Some components in React use react-focus-lock and the tests use RTL. 

Running `node --prof` with jest seems to indicate some native function called "__ZN2v88internal17Builtin_JsonParseEiPmPNS0_7IsolateE" as where it spends a lot of time and calls to that stem heavily from the functions that @donaldpipowitch listed. Node version is 14.18.3 but also tried on latest v16 and saw no improvement. Local OS is macOS 12.0.1 but similar slowdowns are seen across across the team and CI environment that all use varying OS's.

jsnajdr:
Some analysis about why `getComputedStyle` is so slow. JSDOM has a [default stylesheet](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/browser/default-stylesheet.js) with 120 rules. Your app can add additional rules, for example, one of our unit tests I've been debugging adds 90 additional rules from Emotion, the total becoming 210, but 120 rules is the baseline everyone starts with.

[...]

PR #3478, opened 2023-01-04
Add inheritance for pointer-events to getComputedStyle()

jsnajdr:
This PR is adding support for inheritance of the `pointer-events` CSS property to `getComputedStyle()`. If a parent element specifies `pointer-events: none`, then for a child element `getComputedStyle(childEl).pointerEvents` should be `none`, too.

This has been implemented for `visibility` for some time, and `pointer-events` is also an "Inherited: yes" ([#](https://svgwg.org/svg2-draft/interact.html#PointerEventsProperty)) property, with the same inheritance behavior, and deserves to be supported.

A real-world impact of this should be improved performance of Testing Library, namely the [`closestPointerEventsDeclaration` helper](https://github.com/testing-library/user-event/blob/1aa2027e5ec445ab413808556efa7763b65053d3/src/utils/pointer/cssPointerEvents.ts#L22-L32) in `@testing-library/user-event`. Every `userEvent.click(el)` call runs this to find out if `pointer-event`s are allowed. Its complexity should be reduced from _O(n^2)_ to just _O(n)_ (more [here](https://github.com/jsdom/jsdom/issues/3234#issuecomment-1359965734)).

### DIFF
--- a/lib/jsdom/living/helpers/style-rules.js
+++ b/lib/jsdom/living/helpers/style-rules.js
@@ -18,6 +18,12 @@ exports.propertiesWithResolvedValueImplemented = {
     inherited: true,
     initial: "visible",
     computedValue: "as-specified"
+  },
+  // https://svgwg.org/svg2-draft/interact.html#PointerEventsProperty
+  "pointer-events": {
+    inherited: true,
+    initial: "auto",
+    computedValue: "as-specified"
   }
 };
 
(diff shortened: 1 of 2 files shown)


jsnajdr:
> A real-world impact of this should be improved performance of Testing Library

Well, when I ran one of our Testing Library tests on this patched JSDOM, it's actually 50% slower 🙁 Most likely because for every `getComputedStyle` call, JSDOM used to traverse the parent elements, match styles against them, and figure out the inherited `visibility`. Now it does that traversal _twice_, newly also for `pointer-events`.

To fix this regression, we'd either have to optimize the `getResolvedValue` calls so that the `forEachMatchingSheetRuleOfElement` loop is done only once, or implement computed style caching.

jsnajdr:
> Well, when I ran one of our Testing Library tests on this patched JSDOM, it's actually 50% slower

I added a computed styles cache in #3482, and when this `pointer-events` patch is applied on top of that, it actually delivers the expected performance improvements for `@testing-library/user-event`. There's a 30% improvement. Described in detail in #3482 PR description.

domenic (approved):
Beautiful

PR #3479, opened 2023-01-04
Add support for unset CSS keyword to getComputedStyle()

jsnajdr:
Add support for the CSS-wide `unset` keyword when computing styles in `getComputedStyle()`. An `unset` value removes any previously set values for the element and forces it to be inherited or have initial value.

A companion PR to #3478, they should probably get merged together. The inspiration for this PR was seeing `unset` being handled in `@testing-library/user-event`'s [`closestPointerEventsDeclaration` helper](https://github.com/testing-library/user-event/blob/1aa2027e5ec445ab413808556efa7763b65053d3/src/utils/pointer/cssPointerEvents.ts#L22-L32).

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -801,26 +801,35 @@ function Window(options) {
 
     const declaration = new CSSStyleDeclaration();
     const { forEach } = Array.prototype;
-    const { style } = elt;
+
+    function handleProperty(style, property) {
+      const value = style.getPropertyValue(property);
+      // https://drafts.csswg.org/css-cascade-4/#valdef-all-unset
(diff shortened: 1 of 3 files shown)


domenic (approved):

domenic:
Looks great. I'm unclear why this should or should not be merged together with #3478; would something bad happen if we merged and released this before doing #3478 and #3482? I'll hold off for now, but let me know.

jsnajdr:
This is fully independent from the other two PRs (#3478 and #3482) and can be merged separately. It will only cause a merge conflict with #3482, because both PR touch the same code, the loop that goes through matching style rules and applies their properties, but that's fine.

PR #3480, opened 2023-01-04
Allow Image Buttons to submit forms

jenseng:
See https://html.spec.whatwg.org/multipage/input.html#image-button-state-(type=image)

I added a general input/button form submission test to assert which ones should/ should not submit a form, as I didn't see one in the existing wpt/tuwpt tests.


domenic (approved):
Nice!

This doesn't fully implement the spec, in particular steps 3-5 of https://html.spec.whatwg.org/#image-button-state-(type=image):input-activation-behavior . However, the coordinate stuff only feeds into constructing the form entry data, which is not yet implemented in jsdom. So we can omit it for now.

Issue #3435, opened 2022-09-20
form.submit() fires a submit event, but it shouldn't.


public:
According to the WHATWG specs:
* `form.submit()` [*should not* fire a 'submit' event](https://html.spec.whatwg.org/multipage/forms.html#dom-form-submit-dev)
* `form.requestSubmit()` [*should* fire a 'submit' event](https://html.spec.whatwg.org/multipage/forms.html#dom-form-requestsubmit-dev)

JSDOM's form.submit() always fires a 'submit' event: https://github.com/jsdom/jsdom/blob/28ed5877040f6e7eeddd01975b738c41d8f76329/lib/jsdom/living/nodes/HTMLFormElement-impl.js#L82-L83

It seems like really JSDOM should just log an error and not fire the event? [This is how it worked about 4 years ago](https://github.com/jsdom/jsdom/blame/d3facd595a28567443d39bc2c925f5366f7d0601/lib/jsdom/living/nodes/HTMLFormElement-impl.js#L76-L78) until https://github.com/jsdom/jsdom/pull/2388 made it start emitting submit events.

PR #3481, opened 2023-01-05
Add support for SubmitEvent.submitter

### LINKED ISSUE #3117: SubmitEvent.submitter is undefined when a click submits a form
### Basic info:

- **Node.js version:** 10.23.3
- **jsdom version:** 16.14.0

### Minimal reproduction case

Originally reported in https://github.com/testing-library/react-testing-library/issues/870

https://runkit.com/eps1lon/6024f498a573fe0019c227ca

```js

      <form>
        <p>
          <input type="firstname" value="Michael" readOnly>
        </p>
        <p>
          <input type="lastname" value="Scott" readOnly>
        </p>
        <button type="submit" value="next" name="action">
          Next
        </button>
        <button type="submit" value="save" name="action">
          Save
        </button>
      </form>
`);
const { document, MouseEvent } = window;

const form = document.querySelector("form");
form.addEventListener("submit", (event) => {
  event.preventDefault();
  console.log(String(event.submitter))
});

const submitNext = form.querySelector('button[type="submit"]');
submitNext.dispatchEvent(new MouseEvent('click'))
```
Logs:
```
"undefined"
```

OS: Linux 5.4 Ubuntu 20.04.2 LTS (Focal Fossa)
Chrome: 88.0.4324.150

https://jsbin.com/mamocowata/edit?html,js,console,output

Logs:
```
"[object HTMLButtonElement]"
```

### Click events and default behavior

In case it's unclear whether a untrusted click event should even submit the form:

By spec user agents must run the activation behavior on click events:
> If the event target has associated activation behavior, the default action MUST be to execute that activation behavior (see § 3.5 Activation triggers and behavior).
-- (https://w3c.github.io/uievents/#event-type-click). 

Generally, untrusted events don't run default behavior. Click events are the only exception here:

> Most untrusted events will not trigger default actions, with the exception of the click event. This event always triggers the default action, even if the isTrusted attribute is false (this behavior is retained for backward-compatibility). All other untrusted events behave as if the preventDefault() method had been called on that event.

-- https://w3c.github.io/uievents/#trusted-events

jenseng:
Fixes #3117

There are some equivalent disabled WPT tests (e.g. form-submission-algorithm.html), but they're intertwined with unsupported functionality like iframes/navigation, so it was just easier to add new tests that strictly validate the SubmitEvent 😅


test/web-platform-tests/to-upstream/html/semantics/forms/form-submission-0/submit-event.html
```diff
@@ -0,0 +1,49 @@
...
+    submitEvent = event;
+  });
+  document.body.appendChild(form);
+  return { form, submitButton };
+}
+
+test(() => {
+  const { submitButton } = prepareForm();
+  submitButton.click();
+  assert_equals(submitButton, submitEvent.submitter, "Submitter is set");
```
domenic, line 29: The order is (actual, expected). (I will push a commit to fix.)

domenic (approved):
LGTM with minor fixes I'll make.

test/web-platform-tests/to-upstream/html/semantics/forms/form-submission-0/submit-event.html
```diff
@@ -0,0 +1,49 @@
...
+test(() => {
+  const { form, submitButton } = prepareForm();
+  form.requestSubmit(submitButton);
+  assert_equals(submitButton, submitEvent.submitter, "Submitter is set");
+}, "Requesting submit with a submitter sets the submitter");
+
+test(() => {
+  const { form } = prepareForm();
+  form.requestSubmit();
+  assert_equals(undefined, submitEvent.submitter, "Submitter is not set");
```
domenic: This should be null, not undefined.

test/web-platform-tests/to-upstream/html/semantics/forms/form-submission-0/submit-event.html
```diff
@@ -0,0 +1,49 @@
...
+}, "Requesting submit without a submitter does not set the submitter");
+
+test(() => {
+  const { form } = prepareForm();
+  form.submit();
+  assert_equals(undefined, submitEvent.submitter, "Submitter is not set");
```
domenic, line 48: This does not fire a submit event in browsers, and so this test is invalid. So this means jsdom actually has a problem with its current implementation...

It's not easy to write a test for this since in browsers it will cause a navigation. Hmm.

domenic (changes requested):
Actually, found some problems with this when running the same tests against browsers, which we need to fix before merging...

test/web-platform-tests/to-upstream/html/semantics/forms/form-submission-0/submit-event.html
```diff
@@ -26,24 +26,25 @@
...
   form.requestSubmit(submitButton);
-  assert_equals(submitButton, submitEvent.submitter, "Submitter is set");
+  assert_equals(submitEvent.submitter, submitButton, "Submitter is set");
 }, "Requesting submit with a submitter sets the submitter");
 
 test(() => {
   const { form } = prepareForm();
   form.requestSubmit();
-  assert_equals(undefined, submitEvent.submitter, "Submitter is not set");
+  assert_equals(submitEvent.submitter, null, "Submitter is not set");
```
jenseng, line 41: this test now fails, i'll tweak the implementation so that this property is null rather than undefined

test/web-platform-tests/to-upstream/html/semantics/forms/form-submission-0/submit-event.html
jenseng: i guess ideally once that bug is fixed, this test should just be removed? i'd be happy to take a stab at fixing that issue in another PR

test/web-platform-tests/to-upstream/html/semantics/forms/form-submission-0/submit-event.html
jenseng: Done

test/web-platform-tests/to-upstream/html/semantics/forms/form-submission-0/submit-event.html
domenic: Well, we should not add tests that fail in browsers. And we shouldn't be passing tests that fail in browsers.

Ideally, we should write a test that shows that the event does *not* happen. However, writing such a test is difficult, because in a browser it would cause a navigation.

jenseng: yeah that makes sense, testing this gets a bit tricky 😕

i took a stab at just removing [these lines](https://github.com/jsdom/jsdom/blob/a7c85452d0c4b1947b0410baf11b040f0ed076a9/lib/jsdom/living/nodes/HTMLFormElement-impl.js#L83-L85) on master, and the wpt/tuwpt tests all passed locally. so that's promising, though perhaps there are factors i'm not considering 🤔

jenseng: relevant existing issue: #3435

jenseng: @domenic wdyt of removing the test altogether from this PR (since here we're mainly focused on SubmitEvent.submitter, not whether or not the `submit` event fires)? then we can figure out the best way to test the behavior of `submit()` on #3493

domenic: Yeah, that sounds perfect. Thanks for working on it!

jenseng: Sure thing, done!

domenic (approved):

PR #3484, opened 2023-01-05
Add CSSOM View extensions to MouseEvent

### LINKED ISSUE #3483: MouseEvent: offsetX and offsetY return undefined
### Basic info:

- **jsdom version:** 20.0.3

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom")
const { document } = new JSDOM().window;

div.addEventListener("click", ({ offsetX, offsetY }) => console.log({ offsetX, offsetY }));
div.click();
// { offsetX: undefined, offsetY: undefined }
```

```js
div.addEventListener("click", ({ offsetX, offsetY }) => console.log({ offsetX, offsetY }));
div.click();
// { offsetX: 0, offsetY: 0 }
```

### Details

The issue appears to be that jsdom doesn't support the [CSSOM View extensions to MouseEvent](https://www.w3.org/TR/cssom-view/#extensions-to-the-mouseevent-interface) (and possibly others). For parity with browsers and to make jsdom more useful, `offsetX` and `offsetY` attributes should be set and not return null/undefined. Since layout is out of scope for jsdom, `0` seems like the appropriate value to return (just as browsers do with synthetic click events).

### Related Issues
* #3470

jenseng:
Ensure that the [CSSOM View MouseEvent attributes](https://www.w3.org/TR/cssom-view/#extensions-to-the-mouseevent-interface) are set. Specifically this ensures that `x`/`y` mirror `clientX`/`clientY` and that `offsetX`/`offsetY` are set (always `0` since there is no layout).

Fixes #3483, #3470
Builds on and supersedes #3472 (reworks it to fix the default case and the tests)

### DIFF
--- a/lib/jsdom/living/events/MouseEvent-impl.js
+++ b/lib/jsdom/living/events/MouseEvent-impl.js
@@ -7,6 +7,13 @@ const UIEventImpl = require("./UIEvent-impl").implementation;
 const MouseEventInit = require("../generated/MouseEventInit");
 
 class MouseEventImpl extends UIEventImpl {
+  get x() {
+    return this.clientX;
+  }
+  get y() {
+    return this.clientY;
+  }
+
   initMouseEvent(
     type,
     bubbles,
@@ -33,8 +40,6 @@ class MouseEventImpl extends UIEventImpl {
     this.screenY = screenY;
     this.clientX = clientX;
     this.clientY = clientY;
-    this.x = clientX;
-    this.y = clientY;
     this.ctrlKey = ctrlKey;
     this.altKey = altKey;
     this.shiftKey = shiftKey;
--- a/lib/jsdom/living/events/MouseEvent-impl.js
+++ b/lib/jsdom/living/events/MouseEvent-impl.js
@@ -33,6 +33,8 @@ class MouseEventImpl extends UIEventImpl {
     this.screenY = screenY;
     this.clientX = clientX;
     this.clientY = clientY;
+    this.x = clientX;
+    this.y = clientY;
     this.ctrlKey = ctrlKey;
     this.altKey = altKey;
     this.shiftKey = shiftKey;
--- a/lib/jsdom/living/events/MouseEvent-impl.js
+++ b/lib/jsdom/living/events/MouseEvent-impl.js
@@ -5,6 +5,7 @@ const EventModifierMixinImpl = require("./EventModifierMixin-impl").implementati
 const UIEventImpl = require("./UIEvent-impl").implementation;
 
 const MouseEventInit = require("../generated/MouseEventInit");
+const { wrapperForImpl } = require("../generated/utils");
 
 class MouseEventImpl extends UIEventImpl {
   get x() {
@@ -13,11 +14,31 @@ class MouseEventImpl extends UIEventImpl {
   get y() {
     return this.clientY;
   }
+  get pageX() {
+    if (this._dispatchFlag) {
+      return 0;
+    }
+    const offset = wrapperForImpl(this.view)?.scrollX || 0;
+    return offset + this.clientX;
+  }
+  get pageY() {
+    if (this._dispatchFlag) {
+      return 0;
+    }
+    const offset = wrapperForImpl(this.view)?.scrollY || 0;
+    return offset + this.clientY;
+  }
   get offsetX() {
-    return 0;
+    if (this._dispatchFlag) {
+      return 0;
+    }
+    return this.pageX;
   }
   get offsetY() {
-    return 0;
+    if (this._dispatchFlag) {
+      return 0;
+    }
+    return this.pageY;
   }
 
   initMouseEvent(
--- a/lib/jsdom/living/events/MouseEvent-impl.js
+++ b/lib/jsdom/living/events/MouseEvent-impl.js
@@ -13,6 +13,12 @@ class MouseEventImpl extends UIEventImpl {
   get y() {
     return this.clientY;
   }
+  get offsetX() {
+    return 0;
+  }
+  get offsetY() {
+    return 0;
+  }
 
   initMouseEvent(
     type,
--- a/lib/jsdom/living/events/MouseEvent.webidl
+++ b/lib/jsdom/living/events/MouseEvent.webidl
@@ -7,9 +7,6 @@ interface MouseEvent : UIEvent {
   readonly attribute long screenY;
   readonly attribute long clientX;
   readonly attribute long clientY;
-  readonly attribute long x;
-  readonly attribute long y;
-
   readonly attribute boolean ctrlKey;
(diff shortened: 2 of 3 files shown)


lib/jsdom/living/events/MouseEvent.webidl
```diff
@@ -52,3 +51,27 @@ partial interface MouseEvent {
                         optional short buttonArg = 0,
                         optional EventTarget? relatedTargetArg = null);
 };
+
+// https://www.w3.org/TR/cssom-view/#extensions-to-the-mouseevent-interface
```
domenic, line 55: Use https://drafts.csswg.org/cssom-view/ here and below; we don't use TR URLs.

lib/jsdom/living/events/MouseEvent.webidl
```diff
@@ -52,3 +51,27 @@ partial interface MouseEvent {
...
+  readonly attribute double clientX;
+  readonly attribute double clientY;
+  readonly attribute double x;
+  readonly attribute double y;
+  readonly attribute double offsetX;
+  readonly attribute double offsetY;
+};
+
+// https://www.w3.org/TR/cssom-view/#extensions-to-the-mouseevent-interface
+// Changes existing coordinate entries to doubles
```
domenic, line 71: I'm not sure how well this works in webidl2js. I guess it didn't break the build, which is good, but does it properly "override" the existing ones?

I think before merging, I need a test to ensure that these correctly come out as doubles instead of longs. So, construct a new MouseEvent with its screenX/screenY/clientX/clientY constructor options set to something fractional (e.g. `1.5`), and then read off the results from the object and ensure they're still fractional.

Bonus points if you add a test that using initMouseEvent does *not* preserve the fractionalness, since the arguments are still defined to be longs.

lib/jsdom/living/events/MouseEvent-impl.js
```diff
@@ -7,6 +7,19 @@ const UIEventImpl = require("./UIEvent-impl").implementation;
 const MouseEventInit = require("../generated/MouseEventInit");
 
 class MouseEventImpl extends UIEventImpl {
+  get x() {
+    return this.clientX;
+  }
+  get y() {
+    return this.clientY;
+  }
+  get offsetX() {
```
domenic, line 16: Per spec these should be the same as pageX/pageY. A test, or enabling of an existing test, would be good.

lib/jsdom/living/events/MouseEvent-impl.js
```diff
@@ -7,6 +7,19 @@ const UIEventImpl = require("./UIEvent-impl").implementation;
...
+    return 0;
+  }
+  get offsetY() {
+    return 0;
+  }
+
```
domenic, line 22: You're missing getters for pageX and pageY. See how they are defined in steps 2-3 of https://drafts.csswg.org/cssom-view/#dom-mouseevent-pagex .

domenic (changes requested):

lib/jsdom/living/events/MouseEvent-impl.js
jenseng: i'm not sure if the getters are actually needed, since (i think) by default properties are readable/writable (e.g. see clientX usage above, which works and doesn't have an explicit getter)

lib/jsdom/living/events/MouseEvent.webidl
jenseng: Interestingly, it seems that although major browsers (chromium/firefox/webkit) have implemented these MouseEvent properties, they still convert doubles to longs (possibly for [legacy compatibility reasons](https://w3c.github.io/pointerevents/#the-click-auxclick-and-contextmenu-events)?) I don't know if that means jsdom **_shouldn't_** support doubles, but perhaps something to think about.

lib/jsdom/living/events/MouseEvent.webidl
jenseng: Done

lib/jsdom/living/events/MouseEvent.webidl
jenseng: Added tests around this

lib/jsdom/living/events/MouseEvent-impl.js
jenseng: Done, and added a test

lib/jsdom/living/events/MouseEvent-impl.js
jenseng: Done

domenic (approved):

lib/jsdom/living/events/MouseEvent-impl.js
```diff
@@ -5,8 +5,42 @@ const EventModifierMixinImpl = require("./EventModifierMixin-impl").implementati
...
+  get x() {
+    return this.clientX;
+  }
+  get y() {
+    return this.clientY;
+  }
+  get pageX() {
+    if (this._dispatchFlag) {
+      return 0;
+    }
```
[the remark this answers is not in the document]
jenseng: @jsnajdr when implementing this, my goal was to:
1. Follow the spec as closely as possible, and
2. Return something reasonable where we can't fully implement it, i.e. since jsdom doesn't implement layout/rendering, some things are unknowable so perhaps 0 is the least bad option

A (minor) risk with returning the current `clientX + scrollX` is that it could be inconsistent with browsers, since `scrollX/Y` could change between the time of event initialization and when `pageX`/`pageY` are read during dispatch (e.g. if explicitly set or `scrollTo` is called during an earlier click handler).

That said, it's probably still a better default than zero, and assuming `scrollX/Y` hasn't changed, it should be synonymous with `the position where the event occurred relative to the origin of the initial containing block`.

@domenic wdyt? Seems like a reasonable change to me, though perhaps it's still a good idea to keep the dispatch branch even if the return value is the same. That way it's a good placeholder/indicator for when layout support is added.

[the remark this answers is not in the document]
jenseng: I do have an [existing PR](https://github.com/jsdom/jsdom/pull/3496) that is closely related to this, and I think we can solve the minor risk outlined above by setting the new [`_initialContainingBlockRelativeX/Y`](https://github.com/jsdom/jsdom/pull/3496/files#diff-10b0a4b9fcc799389fd0a69fdce632ca6b9b82319f1ce74fa005913a9755a6f8R49-R50) to `clientX/Y + scrollX/Y`.

That way the values will be persisted at init time, and not risk changing if `scrollX/Y` changes later before/during dispatch.

Maybe I'll just split it out into its own PR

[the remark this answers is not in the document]
jenseng: #3514 should resolve this

PR #3493, opened 2023-01-10
Don't fire submit event from form.submit()

### LINKED ISSUE #3435: form.submit() fires a submit event, but it shouldn't.

jenseng:
Fixes #3435

Open to suggestions on how to test this in a way that works for both jsdom and browsers, as the navigation aspect makes it difficult. See also discussion on #3481


jenseng:
I wonder if a mocha/API test would be the best way to test this, instead of WPT? Really we only care to test that jsdom isn't firing the event; there's not a reliable way to cancel the form submission in browsers, and we can generally trust that they're already following the spec in this regard

domenic:
Yeah, I think we can't write a test here that will work in both jsdom and browsers. So I think we have two options:

- Write a "to-upstream" web platform test but put "dont-upstream" in the name. We've done that a couple times, although usually for slightly different reasons. Maybe this one, being quite special, would be in a different folder, e.g. a top-level jsdom-limitations/ or something.

- Write a Mocha API test. Probably in https://github.com/jsdom/jsdom/blob/master/test/api/from-outside.js , or maybe from a separate file like `jsdom-limitations.js`.

I think I'd be OK with either option. Both are kind of unpleasant in that the test we're talking about doesn't seem to fit into either existing format, but maybe it's fine. In both cases, we could be setting a precedent for future tests, e.g. tests of the various layout-ish methods that always return 0.

jenseng:
> * Write a "to-upstream" web platform test but put "dont-upstream" in the name. We've done that a couple times, although usually for slightly different reasons. Maybe this one, being quite special, would be in a different folder, e.g. a top-level jsdom-limitations/ or something.

Rebased and added a test along these lines, let me know if you'd like me to change the naming or anything

domenic (approved):

Issue #3494, opened 2023-01-11
Wrong behavior when location.hash is set to empty string


matheusmb:
### Minimal reproduction case

When manipulating the jsdom `window.location.hash`  to an empty string, its behavior differs from the browser. On browsers when the hash is cleared by setting it to an empty string, its return an empty string but the hash symbol is kept in the URL. Please read the steps bellow to understand it better:

1. Open http://example.com/
> window.location.hash // -> ""
> window.location.href // -> "http://example.com/"

2. Update the hash value
> window.location.hash = "test"
> window.location.hash // -> "#test"
> window.location.href // -> "http://example.com/#test"

3. Set hash as an empty string
> window.location.hash = ""
> window.location.hash // -> ""
> window.location.href // -> "http://example.com/#"

I could reproduce this behavior on Chrome, Edge and Firefox running on a Windows machine.

The jsdom implementation wipes out the hash symbol when the hash is set as an empty string, hence on jsdom implementation, the step 3 would return:
4. Set hash as an empty string on jsdom environment
> window.location.hash = ""
> window.location.hash // -> ""
> window.location.href // -> "http://example.com/"

The PR #3495 addresses this bug.

PR #3495, opened 2023-01-11
Fix location hash set empty behavior

matheusmb:
This PR fixes the issue described by #3494

### DIFF
--- a/lib/jsdom/living/window/Location-impl.js
+++ b/lib/jsdom/living/window/Location-impl.js
@@ -173,7 +173,7 @@ exports.implementation = class LocationImpl {
   }
 
   get hash() {
-    if (this._url.fragment === null) {
+    if (this._url.fragment === null || this._url.fragment === "") {
       return "";
     }
 
--- a/lib/jsdom/living/window/Location-impl.js
+++ b/lib/jsdom/living/window/Location-impl.js
@@ -173,7 +173,7 @@ exports.implementation = class LocationImpl {
   }
 
   get hash() {
-    if (this._url.fragment === null || this._url.fragment === "") {
+    if (this._url.fragment === null) {
       return "";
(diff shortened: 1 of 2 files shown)


domenic:
Thanks! However we do not accept new tests in the to-port-to-wpts format. Can you add them in the WPT to-upstream format, per CONTRIBUTING.md?

matheusmb:
Sure @domenic, I'll work on this tomorrow. I will take the opportunity and migrate this test suit to the new format. Thanks.

matheusmb:
Hey @domenic, I took a look at the WPT tests and I could not find any case that would cover this behavior under [the location interface test cases](https://github.com/web-platform-tests/wpt/tree/master/html/browsers/history/the-location-interface). I'm wondering if this is a bug in browsers implementation 🤔. jsdom was implementing the hash setter according to the [URL Living Standard](https://url.spec.whatwg.org/#dom-url-hash), although the spec does not seem clear to me if the # will be removed from the URL.

What do you think about this? What is the approach for this case?

domenic:
> I took a look at the WPT tests and I could not find any case that would cover this behavior under [the location interface test cases](https://github.com/web-platform-tests/wpt/tree/master/html/browsers/history/the-location-interface).

I agree. I think you'll need to add a new test, in the to-upstream folder.

> I'm wondering if this is a bug in browsers implementation 🤔.

I don't think so. What leads you to that conclusion?

> . jsdom was implementing the hash setter according to the [URL Living Standard](https://url.spec.whatwg.org/#dom-url-hash), although the spec does not seem clear to me if the # will be removed from the URL.

That's the bug. It should be implementing the spec for the Location object, under "the hash setter steps" at https://html.spec.whatwg.org/#dom-location-hash.

PR #3496, opened 2023-01-12
FormData.constructor: implement optional submitter parameter 

jenseng:
Add support for the new optional `submitter` parameter to the `FormData` constructor. This feature is [implemented in the latest versions](https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData#browser_compatibility) of Chrome/Firefox/Safari.

Spec: https://xhr.spec.whatwg.org/#interface-formdata
Spec PR: https://github.com/whatwg/xhr/pull/366

In order to fully implement it in jsdom, this requires more robust Image Button support (i.e. track selected coordinate, use when constructing form data set).

Testing notes:
* The newly enabled WPTs all pass in jsdom and in the big three browsers.
* The newly added to-upstream test passes in jsdom and Chrome/Safari (it fails in Firefox, but that's a Firefox bug imo, I'm planning to file a FF bug and upstream the test out of band)

### DIFF
--- a/lib/jsdom/living/nodes/HTMLInputElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLInputElement-impl.js
@@ -188,7 +188,7 @@ class HTMLInputElementImpl extends HTMLElementImpl {
     }
   }
 
-  _activationBehavior() {
+  _activationBehavior(event) {
     if (!this._mutable && this.type !== "checkbox" && this.type !== "radio") {
       return;
     }
@@ -200,7 +200,14 @@ class HTMLInputElementImpl extends HTMLElementImpl {
         fireAnEvent("input", this, undefined, { bubbles: true });
         fireAnEvent("change", this, undefined, { bubbles: true });
       }
-    } else if (form && (this.type === "submit" || this.type === "image")) {
+    } else if (form && this.type === "image") {
+      // https://html.spec.whatwg.org/multipage/input.html#image-button-state-(type=image):input-activation-behavior
+
+      // TODO: if/when layout is implemented, record the selected coordinate at the start of dispatch and use it here,
+      // rather than relying on these getters that just mirror pageX/Y outside of dispatch
+      this._selectedCoordinate = { x: event.offsetX, y: event.offsetY };
+      form._doRequestSubmit(this);
+    } else if (form && this.type === "submit") {
       form._doRequestSubmit(this);
     } else if (form && this.type === "reset") {
       form._doReset();
--- a/lib/jsdom/living/xhr/FormData-impl.js
+++ b/lib/jsdom/living/xhr/FormData-impl.js
@@ -1,7 +1,8 @@
 "use strict";
+const DOMException = require("domexception/webidl2js-wrapper");
 const idlUtils = require("../generated/utils");
 const { closest } = require("../helpers/traversal");
-const { isDisabled, isSubmittable, isButton } = require("../helpers/form-controls");
+const { isDisabled, isButton, isSubmitButton } = require("../helpers/form-controls");
 const Blob = require("../generated/Blob.js");
 const File = require("../generated/File.js");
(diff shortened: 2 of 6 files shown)


lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -200,7 +200,14 @@ class HTMLInputElementImpl extends HTMLElementImpl {
         fireAnEvent("input", this, undefined, { bubbles: true });
         fireAnEvent("change", this, undefined, { bubbles: true });
       }
-    } else if (form && (this.type === "submit" || this.type === "image")) {
+    } else if (form && this.type === "image") {
+      // https://html.spec.whatwg.org/multipage/input.html#image-button-state-(type=image):input-activation-behavior
+
+      // At this point the dispatch flag is false, so offsetX/Y reflect pageX/Y, which is not what we want.
+      // Instead we use _paddingEdgeRelativeX/Y, since they reflect the actual selected coordinates.
+      this._selectedCoordinate = { x: event._paddingEdgeRelativeX, y: event._paddingEdgeRelativeY };
```
jenseng, line 208: Perhaps there's a better way? The challenge here is that `_activationBehavior` is our best hook to know a click happened, but it happens after the dispatch flag is unset.

This allows us to follow the spec, but perhaps this is all overkill, since we don't implement layout and it would probably be ok just to return `(0, 0)` 🤷  ... regarding CSSOM and offsetX/offsetY it looks like browsers [don't even consistently implement](https://github.com/w3c/csswg-drafts/issues/1297) the non-dispatch branch ([step 2](https://drafts.csswg.org/cssom-view/#dom-mouseevent-offsetx))

lib/jsdom/living/xhr/FormData-impl.js
```diff
@@ -96,7 +109,7 @@ function constructTheEntryList(form, submitter) {
   // TODO: handle encoding
   // TODO: handling "constructing entry list"
 
-  const controls = form.elements.filter(isSubmittable); // submittable is a subset of listed
```
jenseng, line 99: This was not compliant with the spec, as it excludes Image Buttons

lib/jsdom/living/events/MouseEvent-impl.js
```diff
@@ -16,31 +16,42 @@ class MouseEventImpl extends UIEventImpl {
   }
   get pageX() {
     if (this._dispatchFlag) {
-      return 0;
+      return this._initialContainingBlockRelativeY;
```
jenseng, line 19: should be `_initialContainingBlockRelativeX`, but the point is moot as I'm splitting this out into a separate PR

jenseng:
Going to pull some of this out into a separate PR to improve CSSOM coordinates

lib/jsdom/living/nodes/HTMLInputElement-impl.js
```diff
@@ -200,7 +200,14 @@ class HTMLInputElementImpl extends HTMLElementImpl {
         fireAnEvent("input", this, undefined, { bubbles: true });
         fireAnEvent("change", this, undefined, { bubbles: true });
       }
-    } else if (form && (this.type === "submit" || this.type === "image")) {
+    } else if (form && this.type === "image") {
+      // https://html.spec.whatwg.org/multipage/input.html#image-button-state-(type=image):input-activation-behavior
+
+      // TODO: if/when layout is implemented, record the selected coordinate at the start of dispatch and use it here,
+      // rather than relying on these getters that just mirror pageX/Y outside of dispatch
+      this._selectedCoordinate = { x: event.offsetX, y: event.offsetY };
```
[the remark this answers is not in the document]
jenseng: Added a test

domenic (approved):
Sorry for taking so long to review this, especially given all the prerequisite work you did. It's great!

PR #3507, opened 2023-02-02
add support for ports

iammerrick:
Is there a reason this parameter isn't forwarded along?

### DIFF
--- a/lib/jsdom/living/post-message.js
+++ b/lib/jsdom/living/post-message.js
@@ -6,7 +6,7 @@ const { isValidTargetOrigin } = require("../utils");
 const { fireAnEvent } = require("./helpers/events");
 
 module.exports = function (globalObject) {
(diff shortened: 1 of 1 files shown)


domenic:
Because MessagePorts are not implemented in jsdom.

iammerrick:
@domenic Thank you! That makes sense, when using this PR I was creating new instances from node's built in support. https://nodejs.org/api/worker_threads.html#class-messageport

For cases like this, where the port is provided by the API user. What would you like JSDOM support to look like? Polyfilling it always to support all the environments? 

I'd be interested in contributing support if I understood what you'd like that to look likes

domenic:
The idea would be to implement https://html.spec.whatwg.org/#message-ports , and in particular to pass various tests in https://github.com/web-platform-tests/wpt/tree/master/webmessaging .

PR #3488, opened 2023-01-07
Roll web platform tests

domenic:
Let's see what CI spits out in terms of failures...

PR #3508, opened 2023-02-07
Update web-platform-tests

jenseng:
Update to the latest web-platform-tests. I see #3488 overlaps quite a bit -- I didn't notice it until after I posted this PR 🤦 -- but a difference here is I'm solely updating WPTs and the run configuration, there are no behavior changes.

## Motivation

Keeps things up to date, which will let me land [FormData submitter support](https://github.com/whatwg/xhr/pull/366) ([WIP](https://github.com/jsdom/jsdom/pull/3496)) without duplicating [existing WPT tests](https://github.com/web-platform-tests/wpt/blob/master/xhr/formdata/constructor-submitter.html) :)

## Notes

* I tried to categorize new failures where the cause was obvious, though there are a few unknowns
* For skips I generally went with `fail-slow` over `timeout`, since (I think?) that more clearly communicates it's expected to fail, versus an unknown/harness issue.
* Some of the `to-run.yml` changes are just things being moved around, due to WPT renames
* I changed a dozen or so existing `fail` to `fail-slow`; this reduces the overall local run time by almost 4 min (~20:00 -> ~16:20)
* A handful of formerly failing tests now pass, so I removed the expected failures from `to-run.yml`

### DIFF
--- a/test/web-platform-tests/to-run.yaml
+++ b/test/web-platform-tests/to-run.yaml
@@ -1,7 +1,7 @@
 DIR: FileAPI
 
-Blob-methods-from-detached-frame.html: [fail-slow, iframe onload timing]
-BlobURL/cross-partition.tentative.https.html: [fail-slow, iframe onload timing]
+Blob-methods-from-detached-frame.html: [timeout, Unknown]
+BlobURL/cross-partition.tentative.https.html: [timeout, Unknown]
 FileReader/workers.html: [fail, Needs Worker implementation]
 blob/Blob-array-buffer.any.html: [fail, Depends on TextEncoder]
 blob/Blob-constructor.any.html:
@@ -34,7 +34,7 @@ url/**: [timeout, blob URLs not implemented]
 DIR: WebCryptoAPI
 
 "*/**": [fail, Not implemented]
-algorithm-discards-context.https.window.html: [fail-slow, Not implemented]
+algorithm-discards-context.https.window.html: [fail-slow, Tests non-implemented methods]
 idlharness.https.any.html: [fail, Unknown]
 randomUUID.https.any.html: [fail, Not implemented]
 
@@ -59,7 +59,7 @@ preflight-failure.htm: [timeout, Node throws an error on code 304, done() is cal
 redirect-preflight-2.htm: [fail, Preflight should also be done before redirected requests but request module redirects cannot be paused while doing preflight]
 remote-origin.htm: [timeout, postMessage event does not contain source]
 response-headers.htm: [timeout, I don't find a spec about combining same value response headers; also https://github.com/jsdom/jsdom/issues/1833]
-simple-requests-ch.tentative.htm: [fail, Client Hints not implemented]
+simple-requests-ch.tentative.htm: [fail-slow, Client Hints not implemented]
 simple-requests.htm: [timeout, Maybe https://github.com/jsdom/jsdom/issues/1833 but it fails locally too]
 
 ---
@@ -76,8 +76,7 @@ checkVisibility.html: [fail, Not implemented]
 client-props-input.html: [fail, Not implemented]
 client-props-root.html: [fail, Not implemented]
 cssom-getBoundingClientRect-003.html: [fail, Not implemented]
-cssom-getBoxQuads-001.html: [fail, Not implemented]
-cssom-getBoxQuads-002.html: [fail, Not implemented]
+cssom-getBoxQuads-**.html: [fail, Not implemented]
 cssom-getClientRects-002.html: [fail, Unknown]
 cssom-view-img-attributes-001.html: [fail, Unknown]
 dom-element-scroll.html: [fail, Scroll methods not implemented]
@@ -161,7 +160,7 @@ HTMLElement-constructor.html:
 adopted-callback.html: [fail, "Failing test due to https://github.com/whatwg/dom/issues/813"]
 attribute-changed-callback.html:
   "attributedChangedCallback must be enqueued for style attribute change by mutating inline style declaration": [fail, attributeChangedCallback doesn't work with CSSStyleDeclaration, https://github.com/jsdom/cssstyle/issues/113]
-cross-realm-callback-report-exception.html: [fail, unknown iframe issue]
+cross-realm-callback-report-exception.html: [fail, No relevant realm support for callbacks in webidl2js]
 custom-element-reaction-queue.html:
   "Upgrading a custom element must invoke attributeChangedCallback and connectedCallback before start upgrading another element": [fail, document.write() implementation is not spec compliant]
   "Mutating a undefined custom element while upgrading a custom element must not enqueue or invoke reactions on the mutated element": [fail, document.write() implementation is not spec compliant]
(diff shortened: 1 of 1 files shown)


test/web-platform-tests/to-run.yaml
```diff
@@ -192,11 +226,49 @@ DIR: dom/events
...
 EventListener-incumbent-global-1.sub.html: [timeout, Multi-globals]
 EventListener-incumbent-global-2.sub.html: [timeout, Multi-globals]
 EventListener-invoke-legacy.html: [timeout, Animation stuff not implemented]
 EventTarget-dispatchEvent.html: [fail, We don't support every event interface yet]
 event-global-extra.window.html: [fail, We're supposed to check the event listener's global (not the Event's global)]
+event-global-is-still-set-when-coercing-beforeunload-result.html: [fail-slow, unknown iframe issue]
+event-global-is-still-set-when-reporting-exception-onerror.html: [fail, unknown iframe issue]
+event-global-set-before-handleEvent-lookup.window.html: [fail, unknown iframe issue]
+mouse-event-retarget.html: [fail, Requires a layout engine]
+passive-by-default.html:
```
jenseng, line 239: rather than just `fail` the whole file, i enumerated the failures, as the `touchend` variants do all pass under jsdom. i haven't looked into the specifics though, so maybe it'd be better to just `fail` the whole file, due to the lack of Pointer events support?

test/web-platform-tests/to-run.yaml
```diff
@@ -216,22 +288,23 @@ DIR: dom/nodes
...
 Element-closest.html: [fail, :has is not supported (by all major browsers as well)]
 Element-firstElementChild-entity-xhtml.xhtml: [fail, Unknown]
 Element-firstElementChild-entity.svg: [fail, Unknown]
 Element-getElementsByTagName-change-document-HTMLNess.html: [fail, Unknown]
+MutationObserver-cross-realm-callback-report-exception.html: [fail, unknown iframe issue]
 MutationObserver-document.html: [fail, Usage of external scripts doesn't block HTML parsing, https://github.com/jsdom/jsdom/issues/2413]
 Node-appendChild-cereactions-vs-script.window.html: [fail, customElements not implemented]
 Node-cloneNode.html: [timeout, Often takes more than 70 seconds to pass, https://github.com/jsdom/jsdom/issues/3154]
 Node-isConnected.html: [fail, Last two tests with <iframe> are failing]
+NodeList-static-length-getter-tampered*: [flaky, Nested for-loops are really slow and these often time out]
```
jenseng, line 303: these tests are weird, looks like they were added to validate the behavior/perf of NodeList in WebKit. they usually pass, but take about a minute each in jsdom 😬

test/web-platform-tests/to-run.yaml
```diff
@@ -577,48 +686,49 @@ DIR: html/canvas/element/drawing-images-to-the-canvas
...
-the-lang-attribute-002.html: [fail, Unknown]
-the-lang-attribute-003.html: [fail, Unknown]
-the-lang-attribute-004.html: [fail, Unknown]
-the-lang-attribute-005.html: [fail, Unknown]
-the-lang-attribute-006.html: [fail, Unknown]
-the-lang-attribute-007.html: [fail, Unknown]
-the-lang-attribute-008.html: [fail, Unknown]
-the-lang-attribute-009.html: [fail, Unknown]
-the-lang-attribute-010.html: [fail, Unknown]
+DIR: html/dom
```
jenseng, line 689: this whole section (and subdirs) got refactored due to the `aria-` files getting moved over from `dom/nodes`; to keep testing those two I had reorganize things this way (maybe there's a better way?)

jenseng:
@domenic as mentioned in the summary, I didn't realize you already had a PR going when I first uploaded this. LMK how you'd like to proceed, I'm happy to incorporate reasons/descriptions from yours ... overall they are pretty similar, but some of yours seem more accurate/precise 😆

test/web-platform-tests/to-run.yaml
```diff
@@ -1,5 +1,7 @@
 DIR: FileAPI
 
+Blob-methods-from-detached-frame.html: [fail-slow, iframe onload timing]
+BlobURL/cross-partition.tentative.https.html: [fail-slow, iframe onload timing]
```
domenic, line 4: The intent of fail-slow is that, if we ran this test, it would fail with a non-timeout error, but it would take so long to fail that we shouldn't bother.

Whereas the intent of timeout is, if we ran this test, it would fail with a timeout error.

It's possible we haven't applied these consistently so far, but you could imagine a CI run which runs all the skipped tests and checks that we're using the right status...

(Although, I'm not sure if there's a lot of value in separating these anymore, hmm.)

Anyway, don't stress about getting it exactly right, but if there's clear cases where timeout is more correct, switching would probably be good. Or maybe we should just work on a separate PR to get rid of "timeout" and replace it with "fail-slow" everywhere, after this merges...

test/web-platform-tests/to-run.yaml
```diff
@@ -137,50 +146,75 @@ CustomElementRegistry.html:
...
 parser/parser-fallsback-to-unknown-element.html: [fail, Usage of external scripts doesn't block HTML parsing, https://github.com/jsdom/jsdom/issues/2413]
 parser/parser-sets-attributes-and-children.html: [fail, Usage of external scripts doesn't block HTML parsing, https://github.com/jsdom/jsdom/issues/2413]
 parser/parser-uses-constructed-element.html: [fail, Usage of external scripts doesn't block HTML parsing, https://github.com/jsdom/jsdom/issues/2413]
+perform-microtask-checkpoint-before-construction-xml-parser.xhtml: [fail, needs URL.createObjectURL]
 perform-microtask-checkpoint-before-construction.html: [fail, impossible to implement microtask checkpoint without patching Promise]
+pseudo-class-defined-customized-builtins.html: [fail-slow, :defined is not defined and throws]
 pseudo-class-defined.html: [timeout, :defined is not defined and throws]
+reactions/Animation.html: [fail, animate is not implemented]
+reactions/AriaMixin-element-attributes.html: [fail, attributeChanged is not implemented]
+reactions/AriaMixin-string-attributes.html : [fail, attributeChanged is not implemented]
```
domenic, line 184: What do you mean by this? I think this should be something like "ARIAMixin is not implemented".

test/web-platform-tests/to-run.yaml
domenic: Yeah, I think it'd be best to fail this entire file, as it's about a spec change we have not worked on: https://github.com/whatwg/dom/commit/b29449709a7b797cdb9721600bb552f8f140e64a

test/web-platform-tests/to-run.yaml
domenic: I think this is a weakness of our current setup, so there's not much we can do. Thanks for doing all that extra work.

domenic (approved):
Wow, this is some truly amazing work. Thank you so much for all the time you put into this! Indeed, I made a start on this in #3488, but very slowly; whereas you just went and did it all!!

Feel free to copy over any reasons from my PR; I'll trust your judgement.

So yeah, I'll let you push another commit or so to address minor comments, but I'm excited to merge this whenever you feel it's ready.

jenseng:
Thanks @domenic! I'll address the PR feedback and incorporate some descriptions/reasons from yours

jenseng:
Ok should be good to go... rebased, addressed feedback, tweaked some descriptions, and bumped the wpt submodule again 🤞

TimothyGu:
Thanks a lot!

PR #3514, opened 2023-03-02
MouseEvent: fix `pageX/Y` and `offsetX/Y` during dispatch

jenseng:
Ensure that `pageX/Y` reflect the coordinate relative to the origin of the document (i.e. `clientX/Y` in the absence of scrolling support) during dispatch.

Also fix the target-relative coordinate (`offsetX/Y`) during dispatch. This will typically be `(0, 0)` (e.g. in a synthetic click or a dispatched `MouseEvent` with default coordinates), but it will mirror `clientX/Y` if those are set.

Lastly, remove some `scrollX/Y` logic for values computed outside of dispatch. While it logically followed the spec, scrolling is not implemented and it is unsafe to base the calculation on those replaceable properties.

cc @jsnajdr

See also:
* https://github.com/jsdom/jsdom/pull/3484#discussion_r1123237268
* https://github.com/facebook/jest/pull/13825


lib/jsdom/living/events/MouseEvent-impl.js
```diff
@@ -16,31 +16,43 @@ class MouseEventImpl extends UIEventImpl {
...
-      return 0;
+      return this._initialContainingBlockRelativeX;
     }
     const offset = wrapperForImpl(this.view)?.scrollX || 0;
     return offset + this.clientX;
   }
   get pageY() {
     if (this._dispatchFlag) {
-      return 0;
+      return this._initialContainingBlockRelativeY;
```
jsnajdr, line 26: This way, we're returning `clientY` adjusted to the `scrollY` value at the time when the event was created. Which could be a long time ago. If the viewport has scrolled in the meantime, the `pageY` value returning during dispatch won't reflect that.

That seems wrong and testing this in Chrome, I see different behavior:
1. After creating the event with `new MouseEvent`, I can inspect its `.pageY` field and it will always reflect the currrent scroll position. We are outside dispatch here. Your code implements this correctly.
2. If I scroll to another position and then dispatch the event, event listener (inside dispatch) will report `.pageY` value that reflects the current scroll position. Your code doesn't do that. It will use the scroll position from the time when the event was created. This value is even older than the one returned in step 1 when reading `.pageY` outside dispatch.
3. The `.pageY` value will probably be frozen during the entire dispatch. Setting `window.scrollY` inside a listener, or scrolling during the dispatch (is that even possible? event dispatch is completely synchronous) won't change the `.pageY` value.

lib/jsdom/living/events/MouseEvent-impl.js
```diff
@@ -16,31 +16,43 @@ class MouseEventImpl extends UIEventImpl {
...
     return this.pageY;
   }
 
+  constructor(globalObject, args, privateData) {
+    super(globalObject, args, privateData);
+
+    // We cache this during init so that we have a stable value during dispatch
+    const { scrollX = 0, scrollY = 0 } = wrapperForImpl(this.view) ?? {};
+    this._initialContainingBlockRelativeX = this.clientX + scrollX;
+    this._initialContainingBlockRelativeY = this.clientY + scrollY;
```
jsnajdr, line 50: I would call these fields simply `this._pageX`, `this._pageY`, `this._offsetX`, `this._offsetY`. These names are shorter, and yet more comprehensible and more to the point.

test/web-platform-tests/to-upstream/html/editing/activation/click-event-properties.html
```diff
@@ -104,6 +104,24 @@
...
+    assert_equals(e.pageX, 110);
+    assert_equals(e.pageY, 220);
+    assert_equals(e.offsetX, 0);
+    assert_equals(e.offsetY, 0);
+  }));
+  window.scrollX = 10;
+  window.scrollY = 20;
+  element.dispatchEvent(new MouseEvent("click", {
+    clientX: 100, clientY: 200, view: window
+  }));
```
jsnajdr, line 122: In this test, we should check also the case with:
```js
const ev = new MouseEvent();
window.scrollX = 10;
window.scrollY = 20;
el.dispatchEvent(ev);
```
and expect the updated `window.scroll*` values to be used inside the dispatch.

Maybe there are also WPTs for this case?

lib/jsdom/living/events/MouseEvent-impl.js
jenseng: While this PR might seem wrong in terms of Chrome compatibility, I believe it's right in terms of following the spec. Given the browsers are all over the place and internally inconsistent (see below), wouldn't that be the best course?

Here's what my testing shows for `pageY` and `offsetY`... I got slightly different results that what you described, in particular around scrolling, though maybe I overlooked something? I should point out that you can synchronously scroll during dispatch (e.g. via `window.scroll(...)`), see the novel below and this note[^1] for more info 😄 

**Legend:**
* ✅ = implementation matches the spec and/or everybody agrees
* ❌ = doesn't match the spec but another implementation does
* 🤷 = spec is unclear and implementations disagree

Given:
```html
<div style="height:1000px; margin-top: 500px">
  <button>Go!</button>
</body>
<script>
  const evt = new MouseEvent("click", { view: window, clientY: 123 });
  const btn = document.querySelector("button");
  btn.addEventListener("click", (e) => {
    // window.scroll(0, 100); // optionally test browser scrolling during dispatch
    // window.scrollY = 100; // optionally test jsdom "scrolling" during dispatch
    console.log({pageY: e.pageY, offsetY: e.offsetY});
  });
</script>
```

1. If I scroll[^1] (outside of dispatch):
   * **`pageY`**
       * Chrome: reflects the event's `clientY` value but never considers `scrollY` (initial or current), i.e. `123` ❌
       * WebKit: reflects the event's `clientY` value but never considers `scrollY` (initial or current), i.e. `123` ❌
       * Gecko: reflects the event's `clientY` value plus the current `scrollY` value ✅ 
       * jsdom (prior to this PR): reflects the event's `clientY` value plus the current `scrollY` value ✅ 
       * jsdom (with this PR): reflects the event's `clientY` value plus the current `scrollY` value ✅ 
   * **`offsetY`**
       * Chrome: reflects the event's incorrect `pageY` value ❌
       * WebKit: reflects the event's incorrect `pageY` value ❌
       * Gecko: reflects the event's correct `pageY` value ✅ 
       * jsdom (prior to this PR): reflects the event's correct `pageY` value ✅ 
       * jsdom (with this PR): reflects the event's correct `pageY` value ✅ 
2. If I scroll to another position and then `btn.dispatchEvent(evt)`, within the event listener: 
   * **`pageY`**
       * Chrome: reflects the event's `clientY` value and never considers `scrollY` (initial or current), i.e. `123` ❌
       * WebKit: reflects the event's `clientY` value and never considers `scrollY` (initial or current), i.e. `123` ❌
       * Gecko: reflects the event's `clientY` value plus the current `scrollY` value ❌
         If the event listener scrolls (see commented-out code), `pageY` reflects the event's `clientY` value plus the newly changed `scrollY` value, i.e. `223` ❌
       * jsdom (prior to this PR): returns `0` ❌
       * jsdom (with this PR): reflects where the synthetic click happened relative to the document origin (i.e. the `scrollY` value at the time the event was created, plus `evt.clientY` which defaults to `0` absent a layout implementation) ✅
   * **`offsetY`**
       This is a weird one IMO. In order to get a reasonable value back, the event needs a reasonable `clientX/Y`. But even then browsers are inconsistent, and since jsdom doesn't implement `offsetTop` and friends, we can't return the same value as any browser. Additionally the browser `offsetX/Y`s are inconsistent with their own `pageX/Y`s, so something has to give (i.e. `pageX/Y` should be equal to `offsetX/Y` plus the target's `offsetLeft/offsetTop`, and that is not the case).
       * Chrome: reflects `evt.clientY + window.scrollY - btn.offsetTop`, i.e. `123 + ??? - 500` 🤷
         If the event listener scrolls (see commented-out code), `offsetY` changes accordingly; it's cached on the first access
       * WebKit: reflects (as of event init) `evt.clientY - btn.offsetTop`, i.e. `123 - 500` 🤷
       * Gecko: reflects (as of event init) `evt.clientY + window.scrollY - btn.offsetTop`, i.e. `123 + ??? - 500` 🤷
         If the event listener scrolls (see commented-out code), `offsetY` changes accordingly; it's recomputed each time
       * jsdom (prior to this PR): returns `0` ... at least it's consistent with scenarios 3 and 4 🤷
       * jsdom (with this PR): returns `0` ... at least it's consistent with scenarios 3 and 4 🤷
3. If I scroll to another position and then `btn.click()`, within the event listener:
   * **`pageY`**
      * Chrome: reflects the current `scrollY` value ❌
        If the event listener scrolls (see commented-out code), `pageY` reflects the value before the change ❌
      * WebKit: returns `0` ❌
      * Gecko: returns `0` ❌
      * jsdom (prior to this PR): returns `0` ❌
      * jsdom (with this PR): reflects where the synthetic click happened relative to the document origin (i.e. the `scrollY` value at the time the event was created, plus `evt.clientY` which defaults to `0` absent a layout implementation) ✅
   * **`offsetY`**
       I don't know that the spec spells out that it should be zero, but it's a reasonable default and all implementations agree
       * Chrome: returns `0` ✅
       * WebKit: returns `0` ✅
       * Gecko: returns `0` ✅
       * jsdom (prior to this PR): returns `0` ✅
       * jsdom (with this PR): returns `0` ✅
4. If I scroll to another position and then actually click, within the event listener:
   * **`pageY`**
      * Chrome: reflects where the click happened relative to the document origin (equivalent to `evt.clientY` plus the `scrollY` value at the time the event was created) ✅
      * WebKit: reflects where the click happened relative to the document origin (equivalent to `evt.clientY` plus the `scrollY` value at the time the event was created) ✅
      * Gecko: reflects where the click happened relative to the document origin (equivalent to `evt.clientY` plus the `scrollY` value at the time the event was created) ✅
      * jsdom (prior to this PR): n/a, can't actually click
      * jsdom (with this PR): n/a, can't actually click
   * **`offsetY`**
       * Chrome: returns relative offset from the target padding edge ✅
       * WebKit: returns relative offset from the target padding edge ✅
       * Gecko: returns relative offset from the target padding edge ✅
       * jsdom (prior to this PR): n/a, can't actually click
       * jsdom (with this PR): n/a, can't actually click


[^1]: Scrolling in a browser should be done by actually scrolling, or by doing something like `window.scroll(...)`. Setting `scrollX/Y` won't actually cause it to be used in calculations, nor will it actually scroll. In jsdom, OTOH, `scroll` and friends are not implemented, and thus "scrolling" must be done by setting `scrollX/Y`

lib/jsdom/living/events/MouseEvent-impl.js
```diff
@@ -16,31 +16,43 @@ class MouseEventImpl extends UIEventImpl {
...
+    // TODO: support non-zero values somehow if/when jsdom adds layout support
+    this._paddingEdgeRelativeX = 0;
+    this._paddingEdgeRelativeY = 0;
```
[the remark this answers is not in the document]
jenseng: If the aim is to mimic Chrome's implementation details, then maybe? But I'm not sure how how that changes the end result, and IMO it would be less clear at a glance that the code is following the spec. There are some other things happening in the Chrome implementation which I'm not sure we want to follow, e.g. lazily resolving coordinates when first access, which is [not without problems](https://github.com/w3c/csswg-drafts/issues/1070#issuecomment-715590900).

It seems that "having a target" is Chrome's proxy for "is the dispatch flag set?", so it appears the only time the `pageX/Y -> offsetX/Y` defaults get used is outside of dispatch, but the implementation appears inconsistent with the spec and others (see my other comment).

Considering all that, `0` feels like a reasonable initial value for `_offsetX/Y` IMO. If/when jsdom adds layout support, we could then of course implement similar logic to Chrome/others.

test/web-platform-tests/to-upstream/html/editing/activation/click-event-properties.html
jenseng: I'm not sure if that's a valid test in terms of the spec, but I suppose that depends on how you interpret the "when" behind `the position where the event occurred`. Does the event occur when you construct it, or when you dispatch it? Most of the relevant coordinates are set during construction via MouseEventInit, so it seems reasonable to conclude that that's when the coordinate was selected.

That said, I'll take a closer look at the existing WPTs and see what can be done, in particular I'll look to add/enhance tests to coverage what I outlined in my other comment.

One problem I've noticed with my tests is that setting `scrollX/Y` can't be upstreamed, since those properties should be readonly and it would be better to call `scroll(...)`. But jsdom doesn't implement scroll yet 🙃

jenseng: Looks like there aren't many CSSOM View MouseEvent tests, but what I've found bears out my testing, e.g. https://github.com/web-platform-tests/wpt/blob/master/css/cssom-view/mouseEvent.html has a failure for both [WebKit](https://github.com/WebKit/WebKit/blob/136a23f12bae93a6dd2c9f93565aa6da6efdfef5/LayoutTests/imported/w3c/web-platform-tests/css/cssom-view/mouseEvent-expected.txt#L4) and [Chrome](https://github.com/chromium/chromium/blob/main/third_party/blink/web_tests/external/wpt/css/cssom-view/mouseEvent-expected.txt#L3) since they don't consider the scroll offset.

Doesn't look like there's any testing around what happens during dispatch :-/

lib/jsdom/living/events/MouseEvent-impl.js
jenseng: I should point out though, several of the ❌ could be thought of as ✅ if we’re not concerned about any scrolling (user or programmatic) that happens between event creation and retrieval of coordinates during dispatch. It probably never actually matters in the wild, except perhaps in weird synthetic dispatch tests like the ones I wrote 😆. 

Just trying to be thorough and ensure we don’t deviate from the spec unless 1. there’s a compelling reason to, and 2. major implementations are already in agreement (in which case the spec should change)

lib/jsdom/living/events/MouseEvent-impl.js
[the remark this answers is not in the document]
jsnajdr: > Considering all that, `0` feels like a reasonable initial value for `offsetX/Y` IMO.

Seems we both have different criteria about what "reasonable" means, so let me explain how I'm thinking about this 🙂

The `0` value maybe satisfies the spec, but in my view that's not enough. The JSDOM's `MouseEvent` implementation must also serve the practical purposes for which people are using it. And one of the main practical purposes of JSDOM is unit testing of web code, like React components. The reason why I came here to discuss `MouseEvent` is that after upgrading JSDOM to the latest version, one of our unit tests started failing. The test uses the [`react-colorful`](https://omgovich.github.io/react-colorful/) library for implementing color sliders like this:

Screenshot 2023-03-06 at 8 18 10

And the test creates and dispatches `mousedown` and `mousemove` events to simulate moving the slider. On `mousemove`, the tested component calculates relative position in the `getRelativePosition` [method](https://github.com/omgovich/react-colorful/blob/master/src/components/common/Interactive.tsx#L39-L40), reading the `.pageX` and `.pageY` fields. It would be probably better off reading `.offsetX` and `.offsetY`, but anyway. Because JSDOM doesn't support layout, and `getBoundingClientRect` always returns zero height and width, the test is very fragile, but it can at least detect that the slider has moved. If `.pageX` starts always returning `0`, the move will not be detected: the mouse position is always `(0, 0)`!

That's why I ask that `.pageX` and `.offsetX` don't completely ignore the user-specified `.clientX`, but that they should be derived from it. One day, when JSDOM implements even the most minimal layout support, like `getBoundingClientRect` looking at the `width` and `height` CSS styles, the test will become much closer to the browser environment.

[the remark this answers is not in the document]
jenseng: This is a really illustrative example, thank you!

I think we we were already basically in agreement about `pageX/Y` (thus this PR 😉), so it seems like the problem you describe would mostly be covered by this fix, no? That said, you do make a compelling case for `offsetX/Y` simply mirroring `pageX/Y` in the dispatch case too, in the absence of layout support.

I had been viewing this behavior more from the use case of `click` events on things like image buttons, rather than for scenarios like yours (e.g. detecting mouse movement over time). My concern was primarily around `offsetX/Y` returning surprising values during the former, but upon further reflection that is unlikely to be an issue unless `clientX/Y` or `scrollX/Y` are also explicitly set; typically they wouldn't be for these use cases, so the values will continue to be zero.

Although neither approach fully follows the spec, yours does get us marginally closer and helps solve the problems you've outlined, so I'm on board, sorry it took me a bit to get there 😆

lib/jsdom/living/events/MouseEvent-impl.js
jenseng: Thanks @jsnajdr! I think I'm generally in agreement now, just a couple notes:
1. I think it's still a good idea to follow the spec and use `scrollX/Y` in the calculation, even if scrolling isn't fully supported yet. This way we don't need to revisit it when scrolling is implemented; at most some tests would just need to change to call `scroll/scrollBy` instead of setting those properties.
2. I'm going to expand the tests a little more around `offsetX/Y`, specifically I want to make sure we're well covered both in terms of 1. what we do implement spec-wise and 2. being clear around where we deviate due to not implementing layout support (e.g. either find and mark some existing WPTs as failing, or add some `dont-upstream` tests)

jenseng:
Filed [this CSSOM View issue](https://github.com/w3c/csswg-drafts/issues/8531) to address the issues around synthetic pointer events

lib/jsdom/living/events/MouseEvent-impl.js
jsnajdr: > I think it's still a good idea to follow the spec and use `scrollX/Y` in the calculation, even if scrolling isn't fully supported yet.

I agree: returning `pageX: clientX + window.scrollX` and `offsetX: clientX`, both values look solid and shouldn't lead to any suprises :+1:

jenseng:
The second commit:
1. Defaults `offsetX/Y` to `clientX/Y - domRect.x/y` rather than `0`. Although the values will still be zero in many cases (e.g. synthetic click, or a dispatched event w/ zeroed `clientX/Y`), this makes the values more useful as discussed earlier in the PR (e.g. when testing mouse movement across events). By including the target's domRect in the calculations, this also means it will also return the right values if/when jsdom adds full layout support.
2. Splits out some of the new and existing tests into the `dontupstream` variant, as explained in a comment in that test
3. Reworks the timing around when we finalize the event position for `pageX/Y` and `offsetX/Y` during dispatch. Rather than do it at init when we don't know what the target will be, we (re)calculate them right when the target is set (just before the bubble and cancel phases)

On a related note, I opened a [wpt PR](https://github.com/web-platform-tests/wpt/pull/38912) to expand the CSSOM View MouseEvent tests. There's a bit of overlap with the tests in this PR, but we can't use the wpt ones until JSDOM implements `scroll`, layout, etc., so for now it's probably good to have them in both places.

test/web-platform-tests/to-upstream/html/editing/activation/click-event-properties-dontupstream.html
```diff
@@ -0,0 +1,55 @@
...
+// these tests should not be upstreamed since:
+// 1. we're testing pageX/Y and scroll offsets by setting scrollX/Y which should be readonly; jsdom doesn't implement
+//    scroll() and friends (yet)
+// 2. getBoundingClientRect() doesn't know about layout/scrolling, so jsdom can return some unusual offsetX/Y values
+//    during dispatch despite the calculation logic being correct (see below)
+// 3. equivalent tests are being added upstream in https://github.com/web-platform-tests/wpt/pull/38912 that use scroll()
+//    and expect correct offsetX/Y values during dispatch
+test(() => {
+  window.scrollX = 10;
+  window.scrollY = 20;
```
domenic, line 23: Oof, I didn't realize we were setting these in tests. I'd like to avoid writing tests that rely on invalid behavior like this. Ideally we should fix scrollX/scrollY to be readonly.

lib/jsdom/living/events/MouseEvent-impl.js
```diff
@@ -16,31 +16,52 @@ class MouseEventImpl extends UIEventImpl {
...
   }
   get offsetY() {
     if (this._dispatchFlag) {
-      return 0;
+      return this._offsetY;
     }
     return this.pageY;
   }
 
+  /** @override */
```
domenic, line 44: Please avoid jsdoc here and elsewhere

domenic (commented):
Thanks for all the iteration here, and sorry I haven't had time to participate in the discussion until now!

My issue is that I'm really uncomfortable with how window.scrollX and scrollY setters appear to be a key part of the story here. I'd like us to make those readonly, or at the very least assume they never change and not write tests which rely on changing them.

Where does that leave us?

jenseng:
Yeah that sounds reasonable, just two comments:
* Even if the tests don't do anything w/ `scrollX/Y`, I think it's important for the implementation to still read them (as well as `getBoundingClientRect`). That way if/when scrolling and layout are implemented, the code is already correct and it's just a matter of enabling some upstream WPTs.
* Any concerns around making `scrollX/Y` without also implementing `scroll/scrollBy/etc` at the same time? Seems like there could be jsdom users in the wild that are relying on being able to set those in the absence of scrolling support.

lib/jsdom/living/events/MouseEvent-impl.js
jenseng: Would it be worth spelling this out in Contributing.md or elsewhere? Seems like there's some existing jsdoc comments/tags in the codebase

lib/jsdom/living/events/MouseEvent-impl.js
```diff
@@ -16,31 +16,48 @@ class MouseEventImpl extends UIEventImpl {
...
     }
     return this.pageY;
   }
 
+  get target() {
+    return super.target;
+  }
+
+  set target(target) {
+    super.target = target;
```
jsnajdr, line 49: Another way to implement this, maybe more elegant, would be a `_receivedTarget` method on the parent `EventImpl` class:
```js
class EventImpl {
  set target(target) {
    this._target = target;
    this._receivedTarget();
  }

  _receivedTarget() {}
}

class MouseEventImpl {
  _receivedTarget() {
    // calculate offsets from this.target
  }
}
```
Chrome's `Event` implementation also has a `::ReceivedTarget` method, and JSDOM also has similar prior art in the `_activationBehavior()` method on elements.
```

**Review by jsnajdr — APPROVED** [review]

**Comment by jsnajdr** [comment]
> Any concerns around making `scrollX/Y` without also implementing `scroll/scrollBy/etc` at the same time?

In a real browser, setting `window.scrollX` is possible, but it won't affect the scroll position at all. The `scrollX` position is `[[Replaceable]`](https://webidl.spec.whatwg.org/#Replaceable) which means that you can set it, but then it will be just another custom property of `window` that will shadow the internal one.

Compliant code would have access to the `window`'s internal `_scrollX` value (which `window` exposes through a replaceable `scrollX` getter), and there would be `scroll/scrollBy` methods that set that internal value. `scroll/scrollBy` implementation would also have to fire the `scroll` events on all expected targets (`document`, `window`, ...).

If we want to avoid snowballing this PR too much, I think we'll either need to accept the incorrect `window.scrollX =` setting for now, or to give up reading the scroll position at all, returning raw `clientX` as `pageX`, without any adjustment.

**Comment by jsnajdr** [comment]
Hi @domenic :wave: what are you thoughts about what needs to be done to move this PR forward? There were concerns about setting `window.scrollX` in tests, but if we don't want to implement full `window.scrollBy` including `scroll` and `scrollend` events, at least not right now, does that leave us with any other options?

**Comment by domenic** [comment]
I admit I've lost the plot on what we're trying to accomplish here. This thread is too long and this PR contains many changes which seem to be about future-proofing for some hypothetical future where we support layout or scrolling, which makes it hard to understand. The WPT added here does not pass in Firefox or Chrome: https://jsbin.com/sexirixixo/edit?html,output which maybe is a spec problem, maybe is a test problem.

I tried setting out guidelines I'd like us to follow about, but perhaps they were too implicit. The guidelines I'd like are:

- Web platform tests should only use standardized functionality, not anything nonstandard like modifiable `scrollX` properties.
- Web platform tests should be written such that:
  - If all browsers agree, they should match that behavior. (If the spec disagrees with browser consensus, file a spec bug, but keep the tests aligned with browser behavior.)
  - If not all browsers agree, and at least one of them matches the spec, then follow the spec behavior.
  - If no browsers agree with the spec and at least two agree with each other, follow the two-browser behavior and file a spec bug.
  - If no browsers agree with each other and there's a fourth behavior in the spec, we can't handle this on the jsdom level, and need to escalate this further in the standards process before proceeding.
- A change will be accepted to jsdom if it improves our WPT pass rate, including newly-added WPTs that follow the above guidelines.
- A change will be much easier for me to review if there's a 1:1 correspondence with the spec.
- However, putting in dead code that only matters in some future where we support scrolling or layout, is distracting and problematic, and I'd rather we just add a comment saying why a certain spec line is being omitted.

I hope that helps... I think the current PR does not meet these criteria because the WPT that was added fails in two browsers, but admittedly I haven't tested Safari, so maybe it matches the spec and Safari. So that's the immediate problem. After that, I'd encourage getting the PR into shape such that it's easy to review, per the above guidelines.

**Comment by jenseng** [comment]
All fair points, I'll clean this up and remove scrollX/Y stuff in the code in favor of comments. [Upstream wpts](https://github.com/web-platform-tests/wpt/pull/38912) should have this covered for if/when jsdom ever implements scrolling/layout, at which point the implementation could be fleshed out and comments removed.

As for the test, there's a dumb oversight on my part, as it can be made to pass in browsers and jsdom. I wrote it with jsdom in mind, so I didn't consider the body margin and button border, which makes it fail in browsers 🙃  ... the test passes if the event is dispatched against window or some other border-less element at the page origin.

**Comment by jenseng** [comment]
Ok latest commit hopefully addresses everything. I also updated the PR summary to reflect the current state of things 🤞

File: lib/jsdom/living/events/MouseEvent-impl.js [hunk 8]
```diff
@@ -15,44 +14,22 @@ class MouseEventImpl extends UIEventImpl {
...
+    return this.clientX; // TODO: add horizontal scroll offset once jsdom implements scrolling support
   }
   get pageY() {
-    if (this._dispatchFlag) {
-      return this._pageY;
-    }
-    const offset = wrapperForImpl(this.view)?.scrollY || 0;
-    return offset + this.clientY;
+    // TODO: consider dispatch flag and return page-relative event position once layout is supported
+    return this.clientY; // TODO: add horizontal scroll offset once jsdom implements scrolling support
```
jenseng, line 22: nit: vertical

jsnajdr:
Thanks @jenseng for greatly simplifying the PR. I'm confirming that it addresses the main concern I had: when I create a `MouseEvent` by calling a constructor:
```js
e = new MouseEvent('mousedown', { clientX: 10, clientY: 20 });
```
the `clientX` and `clientY` values are propagated to the return values of the `e.pageX` and `e.offsetX` getters, even during dispatch. Before this PR, the getters returned `0` during dispatch, ignoring the constructor parameters.

domenic (approved):
Thanks for your patience here!

uladzimirdev:
@domenic is there any blocker or do you require any help with additional testing to make a release that will contain a fix from current PR?

PR #3529, opened 2023-04-13
fix: proper end index in setRangeText() for input and textarea

### LINKED ISSUE #2601: Textarea setRangeText(text, start, end, 'end') creates the wrong selection
JSDOM's implementation and test of the `textarea`'s `setRangeText` method is incorrect with `'end'` as the 4th argument (maybe other cases too).

The offending test: https://github.com/jsdom/jsdom/blob/03712154a4ab3a3aebf33fdeb8e0ae436b0c41e3/test/web-platform-tests/to-upstream/html/semantics/forms/the-textarea-element/setRangeText.html#L56

## Expected behaviour
`setRangeText(..., 'end')` sets the textarea's selection to the end of the inserted text.

## Observed behaviour
`setRangeText(..., 'end')` sets the textarea's selection to the end of the textarea.

## Relevant code and docs
In the aforementioned test, the lines:

    assert_equals(element.selectionStart, 6);
    assert_equals(element.selectionEnd, 6);

should be:

    assert_equals(element.selectionStart, 3);
    assert_equals(element.selectionEnd, 3);

This [fiddle](https://jsfiddle.net/m9znght5/2/), based on the incorrect test, shows the correct behaviour in Firefox and Safari on OSX.

The [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setRangeText) and [official HTML5 spec](https://www.w3.org/TR/html50/forms.html#dom-textarea/input-setrangetext) also disagree with the current implementation.

pmstss:
fixes #2601

### DIFF
--- a/lib/jsdom/living/nodes/HTMLInputElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLInputElement-impl.js
@@ -596,7 +596,7 @@ class HTMLInputElementImpl extends HTMLElementImpl {
   }
 
   _dispatchSelectEvent() {
-    fireAnEvent("select", this, undefined, { bubbles: true, cancelable: true });
+    setTimeout(() => fireAnEvent("select", this, undefined, { bubbles: true, cancelable: false }), 0);
   }
 
   _getValueLength() {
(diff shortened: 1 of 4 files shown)


domenic (approved):
Nice, thank you!

Issue #3002, opened 2020-07-08
Fix for TypeError: range(...).getBoundingClientRect is not a function


D-to-the-K:
### Basic info:

- **Node.js version:** 12.16.3
- **jsdom version:** 16.2.2

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const options = {
  runScripts: 'dangerously',
};

const dom = new JSDOM(`
<!DOCTYPE html>
<html>
<body>
  <p id="p0">Range: </p>
  <p id="p1">Bounds: </p>
  <p id="p2">Rects: </p>
  <script>
    let range = document.createRange();
    let p0 = document.getElementById("p0");
    p0.innerHTML += JSON.stringify(range);

    // prevent crash
    try {
      let bounds = range.getBoundingClientRect();
      let p1 = document.getElementById("p1");
      p1.innerHTML += JSON.stringify(bounds);
    }
    catch(e) {
      console.log(e.message);
    }
    
    // prevent crash
    try {
      let rects = range.getClientRects();
      let p2 = document.getElementById("p2");
      p2.innerHTML += JSON.stringify(rects);
    }
    catch(e) {
      console.log(e.message);
    }
  </script>
</body>
</html>
`, options);

console.log(dom.window.document.getElementById("p0").innerHTML);
console.log(dom.window.document.getElementById("p1").innerHTML);
console.log(dom.window.document.getElementById("p2").innerHTML);
```

Outputs:
```
range.getBoundingClientRect is not a function
range.getClientRects is not a function
Range: {}
Bounds:
Rects:
```

### How does similar code behave in browsers?

Shouldn't crash. Should instead show the output:
```
Range: {}
Bounds: {"x":0,"y":0,"width":0,"height":0,"top":0,"right":0,"bottom":0,"left":0}
Rects: {}
```

JS Bin: https://jsbin.com/keconidona/edit?html,output

### Suggestions for a fix:

Based on my own tests, simply adding these functions to the Range class should already fix the problem:

```js
getBoundingClientRect() {
  return {
    bottom: 0,
    height: 0,
    left: 0,
    right: 0,
    top: 0,
    width: 0
  };
}

getClientRects() {
  return [];
}
```

[...]

PR #3533, opened 2023-04-20
Add getClientRects and getBoundingClientRect stubs to Range

jeffrifwald:
Adds Rect methods to Range following the current implementation of the `Element.prototype.getBoundingClientRect` and `Element.prototype.getClientRects` stubs.

Should address #3002


domenic:
As discussed in https://github.com/jsdom/jsdom/issues/2751, we're not clear on the right path for these sorts of stubs, and it'd be better to address them holistically.

Additionally, we'd want to introduce proper implementations of `DOMRect`, instead of using POJSOs.

So, if you're interested in pursuing this, I'd suggest:

- Doing a separate PR which implements DOMRect support. (I might try to do this today by extracting the relevant pieces of the mega-PR https://github.com/jsdom/jsdom/pull/2926.)
- Helping drive the discussion in https://github.com/jsdom/jsdom/issues/3002 by identifying an exhaustive list of all stub methods that are desired, and whether or not jsdom is currently stubbing them.

I'm sorry that this is more work than you might have anticipated.

Issue #3432, opened 2022-09-16
Nested `:is` selectors are treated as invalid


enjoythelive1:
### Basic info:

- **Node.js version:** 18.8.0
- **jsdom version:** 20.0.0

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const dom = new JSDOM(`
<button>hi</button>
<input type="submit" value="weee" />
`);

dom.window.document.querySelector(':is(:is(input), button)');
```

This throws the following exception:

```
/node_modules/nwsapi/src/nwsapi.js:572
        throw err;
        ^
DOMException [SyntaxError]: ':is(:is(input), button)' is not a valid selector
    at emit (/node_modules/nwsapi/src/nwsapi.js:570:17)
    at compileSelector (/node_modules/nwsapi/src/nwsapi.js:1310:11)
    at compile (/node_modules/nwsapi/src/nwsapi.js:758:16)
    at collect (/node_modules/nwsapi/src/nwsapi.js:1563:22)
    at _querySelectorAll (/node_modules/nwsapi/src/nwsapi.js:1523:36)
    at Object._querySelector [as first] (/node_modules/nwsapi/src/nwsapi.js:1429:14)
    at DocumentImpl.querySelector (/node_modules/jsdom/lib/jsdom/living/nodes/ParentNode-impl.js:69:44)
    at Document.querySelector (/node_modules/jsdom/lib/jsdom/living/generated/Document.js:1026:58)
    at Object.<anonymous> (/test.js:8:21)
    at Module._compile (node:internal/modules/cjs/loader:1119:14)

Node.js v18.8.0

Process finished with exit code 1

```

The selector that i was using that triggered the fail was: `:is(:is(button, input)[type=submit], button:not([type])):not([disabled])`

### How does similar code behave in browsers?

This selector works perfectly in Chrome 105.0.5195.102 and Firefox 104.0.1

Issue #3427, opened 2022-09-07
Query selector fails with nested namespaced tag names


sasensi:
### Basic info:

- **Node.js version:** v16.13.0
- **jsdom version:** 20.0.0

### Minimal reproduction case

```js
const dom = new JSDOM(`<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<a xmlns:ns="http://schemas.openxmlformats.org/drawingml/2006/main">
    <ns:b>
        <ns:c></ns:c>
    </ns:b>
    <b>
        <c></c>
    </b>
</a>`, {contentType: 'application/xml'});
const a = dom.window.document.querySelector('a');
const nsB = a.querySelector('ns\\:b');
const nsC = a.querySelector('ns\\:c');
const nsBC = a.querySelector('ns\\:b ns\\:c');
const b = a.querySelector('b');
const c = a.querySelector('c');
const bc = a.querySelector('b c');

// The nsBC variable is null while it should not
```

### How does similar code behave in browsers?

The `namespace\\:tagname` syntax doesn't seem to work in browser.
I found it [here](https://github.com/jsdom/jsdom/issues/2159#issuecomment-368318385) but I don't really understand why it work for single level selectors to be honest.

asamuzaK:
You should set `ns|b` for the namespaced elements, attributes, not `ns:b`.
See [Type selectors - CSS: Cascading Style Sheets | MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Type_selectors#namespaces)

domenic:
I created a test page at http://boom-bath.glitch.me/selector.xhtml and it seems browsers don't support the `a|b` syntax in querySelector in XML documents.

And as @asamuzaK points out, the `a\\:b` syntax is meaningless in such XHTML documents.

So, I'll close this, as I don't think there's any work for us to do here.

Issue #3416, opened 2022-08-06
[XML] querySelector with two elements, one capitalized


regseb:
### Basic info:

- **Node.js version:** v18.7.0
- **jsdom version:** 20.0.0

### Description

I found a difference between jsdom and browsers (Firefox and Chromium). For an XML document with a capitalized element, the `querySelector()` method doesn't work when searching by putting two elements in the selector.

- :heavy_check_mark: `querySelector("qux")`
- :heavy_check_mark: `querySelector("Qux")`
- :heavy_check_mark: `querySelector("qux corge")`
- :x: `querySelector("Qux corge")`

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");
const DOMParser = new JSDOM().window.DOMParser;

const xml = `<Foo><bar>baz</bar></Foo>`;
const root = "Foo";
const child = "bar";

const doc = new DOMParser().parseFromString(xml, "application/xml");
console.log("root: " + doc.querySelector(root).textContent);
console.log("child: " + doc.querySelector(child).textContent);
console.log("root child: " + doc.querySelector(`${root} ${child}`).textContent);
```

```
root: baz
child: baz
file:///home/regseb/testcase/index.js:11
console.log("root child: " + doc.querySelector(`${root} ${child}`).textContent);
                                                                  ^

TypeError: Cannot read properties of null (reading 'textContent')
    at file:///home/regseb/testcase/index.js:11:67
    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:541:24)
    at async loadESM (node:internal/process/esm_loader:91:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)

Node.js v18.7.0
```

### How does similar code behave in browsers?

https://jsbin.com/xotuzuzage/edit?js,console

regseb:
I opened an issue on https://github.com/dperini/nwsapi/issues/62.

tinco:
We ran into this issue as well, temporary workaround that fixed it for us was adding:

```
    "resolutions": {
        "nwsapi": "2.2.0"
    },
```

To the top of our `package.json`. A fix seems to already been in the repo, but nwsapi is on a slow release schedule.

regseb:
The bug in _nwsapi_ has been fixed. I created a PR to update the dependency.

- https://github.com/dperini/nwsapi/issues/62
[...]

Issue #3362, opened 2022-04-29
Issue with nesting Direct Child Selector


alopix:
### Basic info:

- **Node.js version:** v14.19.1
- **jsdom version:** 19.0.0

### Minimal reproduction case

```js
const { JSDOM } = require('jsdom');

const HTML = `<!DOCTYPE html><html><body><main id="main-content"><div><div><div><h1>Hello, World!</h1><p>Paragraph</p></div></div></div></main></body></html>`;

test('tripple direct child selector', () => {
  const {
    window: { document }
  } = new JSDOM(HTML);

  const header = document.querySelector('#main-content > div > div > div h1');
  expect(header).toBeTruthy();
  expect(header.textContent).toBe('Hello, World!');
});

test('double direct child selector', () => {
  const {
    window: { document }
  } = new JSDOM(HTML);

  const header = document.querySelector('#main-content > div > div h1');
  expect(header).toBeTruthy();
  expect(header.textContent).toBe('Hello, World!');
});

test('double direct child with nth-child selector', () => {
  const {
    window: { document }
  } = new JSDOM(HTML);

  const header = document.querySelector('#main-content > div > div:nth-child(1) h1');
  expect(header).toBeTruthy();
  expect(header.textContent).toBe('Hello, World!');
});

test('descendant selector', () => {
  const {
    window: { document }
  } = new JSDOM(HTML);

  const header = document.querySelector('#main-content div div h1');
  expect(header).toBeTruthy();
  expect(header.textContent).toBe('Hello, World!');
});

test('descendant with nth-child selector', () => {
  const {
    window: { document }
  } = new JSDOM(HTML);

  const header = document.querySelector('#main-content div div:nth-child(1) h1');
  expect(header).toBeTruthy();
  expect(header.textContent).toBe('Hello, World!');
});
```

### How does similar code behave in browsers?
All three tests should find the `h1` element, but only the first one does. See comparison in [jsbin](https://jsbin.com/tiqikepoxo/edit?html,js,console)

I’m also attaching a minimal demo in Node.
[minimal-demo.zip](https://github.com/jsdom/jsdom/files/8589298/minimal-demo.zip)

alopix:
Still an issue in `20.0.0`.

domenic:
When I tried to run this minimal demo using `node index.test.js`, I got this error: 

```
test('tripple direct child selector', () => {
^

ReferenceError: test is not defined
[...]

PR #3535, opened 2023-04-22
Replace nwsapi with selery

### LINKED ISSUE #3015: querySelector has problems with nodes starting with underscores
### Basic info:

- **Node.js version:** 12.16.1
- **jsdom version:** 16.3.0

### Minimal reproduction case

```js
const { JSDOM } = require( 'jsdom' );

const options = {
    contentType: 'text/xml'
};

const dom = new JSDOM( '<data><_g><b>hey</b></_g></data>', options );

console.log( dom.window.document.querySelector( 'data > _g > b' ).textContent );
// DOMException [SyntaxError]: 'data > _g > b' is not a valid selector
```

It looks like this has to do with the underscore in node `<_g>` but interestingly `querySelector( 'data > _g')` does not fail.

Works without exception, and returns 'hey' using:

```
new DOMParser().parseFromString('<data><_g><b>hey</b></_g></data>', 'text/xml').querySelector('data > _g > b' ).textContent;
```

asamuzaK:
Fixes #3015, #3416, #3427, #3432.
Not all but partially #3362.

Motivation:
I think nwsapi is a very ambitious library, but IMHO there are some concerns for a popular library like jsdom to depend on it.

* Dynamically generate functions from input strings.
  * I don't see input strings being sanitized.
* Relying exclusively on regular expressions to parse selectors.
  CSS Selectors can be nested and repeated, so it is very difficult to deal without creating an AST.
  * There are also expressions that may lead to ReDos.

[selery](https://github.com/danburzo/selery) creates AST and retrieves DOM.
My only concern is that selery hasn't been updated in a long time, but I've heard from the developer that he plan to continue developing it.
See [Question: Still active? · Issue #10 · danburzo/selery](https://github.com/danburzo/selery/issues/10)

About the tests:
I'm new to contributing to jsdom and don't know where or how to write proper tests.
I created a new test file for now, please correct me.


asamuzaK:
Closing for now

PR #3547, opened 2023-05-12
Replace nwsapi

### LINKED ISSUE #2159: querySelector doesn't deal with xml namespaces
### Basic info:

- **Node.js version:** n/a (current Chrome w/browserify)
- **jsdom version:** 11.6.2

### Minimal reproduction case

```js
import {JSDOM} from "jsdom"

let jsdom = new JSDOM()
let parser = new jsdom.jsdom.window.DOMParser()
let string = `<?xml version="1.0"?><cp:coreProperties xmlns:cp="http://schemas.openxmlformats.org/package/2006/metadata/core-properties" xmlns:dc="http://purl.org/dc/elements/1.1/"><dc:title></dc:title></cp:coreProperties>`
let dom = parser.parseFromString(string, 'text/xml')
let el = dom.querySelector('coreProperties')
console.log(el) 
```
The console gives an output of `null`, meaning it cannot find the `cp:coreProperties` element

This works fine in Chrome/Firefox:

```js
let parser = new window.DOMParser()
let string = `<?xml version="1.0"?><cp:coreProperties xmlns:cp="http://schemas.openxmlformats.org/package/2006/metadata/core-properties" xmlns:dc="http://purl.org/dc/elements/1.1/"><dc:title></dc:title></cp:coreProperties>`
let dom = parser.parseFromString(string, 'text/xml')
let el = dom.querySelector('coreProperties')
console.log(el) 
```

The console gives an output of the `cp:coreProperties` element.

asamuzaK:
# Motivation:

I think nwsapi is a very ambitious library and appreciate the developer's hard work, but IMHO there are some concerns for a popular library like jsdom to depend on it.

* It dynamically generates functions from input strings which seems not to be sanitized.
* It relies on regexp to parse CSS selectors, but it's very difficult to cover the spec with regexp.
  CSS selectors can be repeated and nested, and the spec is very complicated.
  For example, `:not()` takes [&lt;complex-real-selector-list&gt;](https://drafts.csswg.org/selectors-4/#typedef-complex-real-selector-list) as an argument while `:is()` takes [ &lt;forgiving-selector-list&gt;](https://drafts.csswg.org/selectors-4/#typedef-forgiving-selector-list ), `:not(:is(foo))` is allowed but `:not(:not(foo))` is not.
  There is also concern that regexp used by nwsapi can lead to ReDos.

I once proposed [selery](https://github.com/danburzo/selery) as an alternative in #3535, but found out that selery is not ready yet.

So, I created a new matcher named [DOM Selector](https://github.com/asamuzaK/domSelector).
It is based on CSS parser [CSSTree](https://github.com/csstree/csstree).
There are still unsupported selectors, but I think it at least covers the basics.

# About tests:

Most of the issues related to CSS selectors in jsdom are tested upstream with a monkey patched jsdom.
See [upstream tests](https://github.com/asamuzaK/domSelector/blob/main/test/index.test.js).


domenic:
> About tests:

To accept this we'll need regression tests which fail beforehand and succeed afterward, in the jsdom repository. See https://github.com/jsdom/jsdom/commit/7512ce988ee1f2d84d567e3c053def63656f77e0 for an example.

yarn.lock
```diff
@@ -864,6 +885,11 @@ loupe@^2.3.1:
   dependencies:
     get-func-name "^2.0.0"
 
+mdn-data@2.0.30:
```
domenic, line 888: Why?

yarn.lock
```diff
@@ -1167,6 +1188,11 @@ shebang-regex@^3.0.0:
   resolved "https://registry.yarnpkg.com/shebang-regex/-/shebang-regex-3.0.0.tgz#ae16f1644d873ecad843b0307b143362d4c42172"
   integrity sha512-7++dFhtcx3353uBaq8DDR4NuxBetBzC7ZQOhmTQInHEd6bSrXdiEyzCvG07Z44UYdLShWUyXt5M/yhz8ekcb1A==
 
+source-map-js@^1.0.1:
```
domenic, line 1191: Why?

yarn.lock
```diff
@@ -84,6 +92,11 @@
   resolved "https://registry.yarnpkg.com/@tootallnate/once/-/once-2.0.0.tgz#f544a148d3ab35801c1f633a7441fd87c2e484bf"
   integrity sha512-XCuKFP5PS55gnMVu3dty8KPatLqUoy/ZYzDzAGCQ8JNFCkLXzmI7vNHCR+XpbZaMWQK/vQubr7PkYq8g470J/A==
 
+"@types/css-tree@^2.3.1":
```
domenic, line 95: Why?

lib/jsdom/living/helpers/selectors.js
```diff
@@ -1,47 +1,13 @@
...
-    });
-  }
-
-  return document._nwsapiDontThrow.match(selector, idlUtils.wrapperForImpl(elImpl));
-};
-
-// nwsapi gets `document.documentElement` at creation-time, so we have to initialize lazily, since in the initial
-// stages of Document initialization, there is no documentElement present yet.
-exports.addNwsapi = parentNode => {
+exports.addDomSelector = parentNode => {
```
domenic, line 5: It looks like this is not needed at all anymore since it doesn't perform any initialization. It's just a global singleton library. In which case this file can be removed and `require("@asamuzakjp/dom-selector")` can be used directly.

domenic (commented):
Another thing we'll need is some benchmarks. See https://github.com/jsdom/jsdom/blob/master/Contributing.md#benchmarks and in particular maybe try to get something like https://github.com/jquery/sizzle/tree/main/speed into the benchmark suite.

yarn.lock
asamuzaK: Same above, css-tree depends on it.

yarn.lock
asamuzaK: Removed.

lib/jsdom/living/helpers/selectors.js
asamuzaK: Removed.

asamuzaK:
> > About tests:
> 
> To accept this we'll need regression tests which fail beforehand and succeed afterward, in the jsdom repository. See [7512ce9](https://github.com/jsdom/jsdom/commit/7512ce988ee1f2d84d567e3c053def63656f77e0) for an example.

I found [jsdom/to-run.yaml](https://github.com/jsdom/jsdom/blob/master/test/web-platform-tests/to-run.yaml#L261)

Should I remove this line?

I'll add `test/web-platform-tests/to-upstream/` tests later (after the benchmark issue is resolved).

Thanks for the quick feedbacks.


> > About tests:
> 
> To accept this we'll need regression tests which fail beforehand and succeed afterward, in the jsdom repository. See [7512ce9](https://github.com/jsdom/jsdom/commit/7512ce988ee1f2d84d567e3c053def63656f77e0) for an example.

I found [jsdom/to-run.yaml](https://github.com/jsdom/jsdom/blob/master/test/web-platform-tests/to-run.yaml#L261)
```
Element-closest.html: [fail, :has is not supported (by all major browsers as well)]
```

Should I remove this line?

I'll add tests in `test/web-platform-tests/to-upstream/` later (after the benchmark issue is resolved).

Thanks for the quick feedbacks.

domenic:
> Should I remove this line?

If you can get the test to pass, then yes, that'd be great!

> stucks on # dom/inner-html/tables #

Hmm, I cannot reproduce this on my machine without your changes. It does appear to be one of the slowest tests though, and the test uses `querySelector()`, so if for you it's only stuck after the change, then that's probably a sign that `querySelector()` is too slow for the scenario tested in https://github.com/jsdom/jsdom/blob/master/benchmark/dom/inner-html.js .

asamuzaK:
I have some questions about testing in wpt.
* There are cases where a SyntaxError exception is thrown as expected but the test fails.
Many of these fail due to mismatched error codes.
But error codes are already obsolete.
Since they are not related to this PR itself, is it okay to ignore them?
* About test/web-platform-tests/tests/dom/nodes/selectors.js.
Some of the cases illustrated as invalidSelectors are parsed by CSSTree without raising a SyntaxError.
These are ignored by Matcher as invalid values ​​such as "Unsupported foo" or "Unknown foo".
How can I write a test for this case where an exception is expected but not raised?

domenic:
> * Many of these fail due to mismatched error codes.
>       But error codes are already obsolete.
>       Since they are not related to this PR itself, is it okay to ignore them?

I don't understand what you mean by "error codes are already obsolete". But no, it is not OK to ignore regressions.

> * Some of the cases illustrated as invalidSelectors are parsed by CSSTree without raising a SyntaxError.
>       These are ignored by Matcher as invalid values ​​such as "Unsupported foo" or "Unknown foo".
>       How can I write a test for this case where an exception is expected but not raised?

I don't understand. Aren't there tests for this already in  test/web-platform-tests/tests/dom/nodes/selectors.js? Why would you need to write new ones?

asamuzaK:
> > * Many of these fail due to mismatched error codes.
> >   But error codes are already obsolete.
> >   Since they are not related to this PR itself, is it okay to ignore them?
> 
> I don't understand what you mean by "error codes are already obsolete". But no, it is not OK to ignore regressions.

See [DOMException: code property - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/DOMException/code)

domenic:
MDN is not the spec; I don't know why they have that notice. Regardless, if you properly throw jsdom DOMExceptions (like nwsapi does due to https://github.com/jsdom/jsdom/blob/e5b8e8afdebee816862584cb7d45bdc6f86d54af/lib/jsdom/living/helpers/selectors.js#L12 ) then this shouldn't be relevant.

asamuzaK:
> MDN is not the spec; I don't know why they have that notice. Regardless, if you properly throw jsdom DOMExceptions (like nwsapi does due to
> 
> https://github.com/jsdom/jsdom/blob/e5b8e8afdebee816862584cb7d45bdc6f86d54af/lib/jsdom/living/helpers/selectors.js#L12
> ) then this shouldn't be relevant.

Got it.
However, it need to be fixed on CssTree side.
[csstree/create-custom-error.js](https://github.com/csstree/csstree/blob/master/lib/utils/create-custom-error.js)

test/web-platform-tests/to-run.yaml
```diff
@@ -258,7 +258,7 @@ Document-characterSet-normalization-2.html: [timeout, Some encodings are not sup
 Document-createEvent-touchevent.window.html: [fail, Pointer events not implemented]
 Document-createEvent.https.html: [fail, We don't support every event interface yet]
 Document-getElementById.html: [fail, We cache IDs in insertion order]
-Element-closest.html: [fail, :has is not supported (by all major browsers as well)]
+Element-closest.html: [fail, test with :invalid pseudo-class is failing]
```
asamuzaK, line 261: 1/29 errors in test.
Only failure is that `fieldset.checkValidity()` returns `true` even if the input element within the fieldset is invalid.
Seem to me a bug of jsdom?

Other pseudo-classes e.g. `:not()`, `:has()` are passing as expected.

test/web-platform-tests/to-run.yaml
```diff
@@ -1190,8 +1189,8 @@ template-element/template-content-hierarcy.html: [fail, "We do not implement htt
 DIR: html/semantics/selectors
 
 pseudo-classes/checked-type-change.html: [fail, CSS computed style computation]
-pseudo-classes/dir-html-input-dynamic-text.html: [fail, need nwsapi fix]
-pseudo-classes/dir.html: [fail, https://github.com/dperini/nwsapi/issues/45]
+pseudo-classes/dir-html-input-dynamic-text.html: [fail, Unknown]
```
asamuzaK, line 1192: I'm not sure what this test is trying to achieve.

test/web-platform-tests/to-run.yaml
```diff
@@ -1190,8 +1189,8 @@ template-element/template-content-hierarcy.html: [fail, "We do not implement htt
...
+pseudo-classes/dir.html: [fail, Unknown]
```
asamuzaK, line 1193: Tests passes in internal tests of dom-selector, but fails when running WPT.
I don't know why...

test/web-platform-tests/to-run.yaml
```diff
@@ -1201,6 +1200,7 @@ pseudo-classes/placeholder-shown-type-change.html: [fail, CSS computed style com
 pseudo-classes/readwrite-readonly-type-change.html: [fail, CSS computed style computation]
 pseudo-classes/readwrite-readonly.html: [fail, Unknown]
 pseudo-classes/required-optional-hidden.html: [fail, CSS computed style computation]
+pseudo-classes/valid-invalid-fieldset-disconnected.html: [fail, Unknown]
```
asamuzaK, line 1203: Same above.
`fieldset.checkValidity()` always returns `true`.

asamuzaK:
Finally PR is ready.
I apologize for spamming pushes as I was unable to get WPT running locally.

About the test:
I can't afford to write tests for WPT.
Please someone else do that.

About the benchmark:
On my PC, running the benchmark on the master branch also got stuck.
So I could not take a benchmark.

domenic:
Ok. I guess we will see if anyone else is willing to finish this PR (benchmarks, web platform tests). Otherwise, we will close it. Let's say in 1 month.

(I don't think it's a good use of my time to review it when it is unfinished.)

asamuzaK:
Benchmark results: (Above: master, Below: PR)

* dom/compare-document-position/compare ancestor
  * jsdom  x 13,564 ops/sec ±4.95% (10 runs sampled)
  * jsdom  x 13,980 ops/sec ±3.48% (9 runs sampled)
* dom/compare-document-position/compare descendant
  * jsdom  x 11,793 ops/sec ±29.39% (8 runs sampled)
  * jsdom  x 13,656 ops/sec ±5.99% (9 runs sampled)
* dom/compare-document-position/compare siblings
  * jsdom  x 731,081 ops/sec ±13.22% (11 runs sampled)
  * jsdom  x 707,733 ops/sec ±23.31% (11 runs sampled)
* dom/construction/createComment
  * jsdom  x 900,515 ops/sec ±8.89% (89 runs sampled)
  * jsdom  x 821,119 ops/sec ±9.85% (72 runs sampled)
* dom/construction/createDocumentFragment
  * jsdom  x 1,015,862 ops/sec ±1.76% (91 runs sampled)
  * jsdom  x 898,812 ops/sec ±6.81% (84 runs sampled)
* dom/construction/createElement
  * jsdom  x 251,444 ops/sec ±0.83% (93 runs sampled)
  * jsdom  x 217,520 ops/sec ±3.46% (79 runs sampled)
* dom/construction/createEvent
  * jsdom  x 244,959 ops/sec ±1.96% (92 runs sampled)
  * jsdom  x 226,752 ops/sec ±5.05% (86 runs sampled)
* dom/construction/createNodeIterator
  * jsdom  x 278,149 ops/sec ±11.86% (70 runs sampled)
  * jsdom  x 231,382 ops/sec ±18.16% (63 runs sampled)
* dom/construction/createProcessingInstruction
  * jsdom  x 740,131 ops/sec ±3.52% (94 runs sampled)
  * jsdom  x 666,575 ops/sec ±11.63% (81 runs sampled)
* dom/construction/createTextNode
  * jsdom  x 621,004 ops/sec ±32.63% (74 runs sampled)
  * jsdom  x 647,597 ops/sec ±37.13% (77 runs sampled)
* dom/inner-html/tables
  * jsdom  x 0.24 ops/sec ±6.41% (5 runs sampled)
  * jsdom  x 0.21 ops/sec ±16.86% (5 runs sampled)
* dom/named-properties/setAttribute(): Remove a named property from window
  * jsdom  x 1,828 ops/sec ±2.30% (49 runs sampled)
  * jsdom  x 1,805 ops/sec ±2.94% (50 runs sampled)
* dom/tree-modification/appendChild: many parents
  * jsdom  x 14.92 ops/sec ±2.00% (38 runs sampled)
  * jsdom  x 12.14 ops/sec ±21.70% (33 runs sampled)
* dom/tree-modification/appendChild: many siblings
  * jsdom  x 217,002 ops/sec ±11.63% (14 runs sampled)
  * jsdom  x 208,877 ops/sec ±17.17% (14 runs sampled)
* dom/tree-modification/appendChild: no siblings
  * jsdom  x 148,337 ops/sec ±18.93% (22 runs sampled)
  * jsdom  x 147,875 ops/sec ±17.98% (24 runs sampled)
* dom/tree-modification/insertBefore: many siblings
  * jsdom  x 140,942 ops/sec ±37.40% (11 runs sampled)
  * jsdom  x 184,490 ops/sec ±16.86% (14 runs sampled)
* dom/tree-modification/removeChild: many parents
  * jsdom  x 179 ops/sec ±11.03% (43 runs sampled)
  * jsdom  x 201 ops/sec ±3.46% (47 runs sampled)
* dom/tree-modification/removeChild: many siblings
  * jsdom  x 190,656 ops/sec ±19.71% (23 runs sampled)
  * jsdom  x 147,534 ops/sec ±36.88% (19 runs sampled)
* dom/tree-modification/removeChild: no siblings
  * jsdom  x 148,285 ops/sec ±27.13% (16 runs sampled)
  * jsdom  x 134,978 ops/sec ±39.02% (14 runs sampled)
* html/parsing/text
  * jsdom  x 13.28 ops/sec ±10.74% (38 runs sampled)
  * jsdom  x 13.42 ops/sec ±8.76% (36 runs sampled)
* jsdom/api/new JSDOM() defaults
  * &lt;Test10&gt; x 152 ops/sec ±5.25% (60 runs sampled)
  * &lt;Test10&gt; x 180 ops/sec ±1.79% (78 runs sampled)
* jsdom/api/new JSDOM() with many elements
  * &lt;Test11&gt; x 13.32 ops/sec ±28.54% (38 runs sampled)
  * &lt;Test11&gt; x 17.48 ops/sec ±12.25% (34 runs sampled)

domenic:
I am glad you were able to get them running!! It looks like none of the currently-benchmarked things regress significantly. But remember we need to add some selectors benchmarks to the suite to really be sure on the difference, per the suggestion in https://github.com/jsdom/jsdom/pull/3547#pullrequestreview-1423923629 .

asamuzaK:
Here are the benchmark results from `jsdom/benchmark/dom/css-selector.js`.

nwsapi do things mainly with regexp, so no wonder it's faster.
But the regexp for catching selectors are very rough, and it seems that it can lead to bugs and possibly vulnerabilities.
However, dom-selector needs more performance tuning.

Above: master, below: PR

* dom/css-selector/closest
  * jsdom  x 8,818 ops/sec ±2.10% (83 runs sampled)
  * jsdom  x 2,704 ops/sec ±9.32% (74 runs sampled)
* dom/css-selector/matches
  * jsdom  x 1,228,765 ops/sec ±1.89% (86 runs sampled)
  * jsdom  x 35.79 ops/sec ±3.21% (57 runs sampled)
* dom/css-selector/querySelector
  * jsdom  x 312 ops/sec ±1.96% (81 runs sampled)
  * jsdom  x 38.67 ops/sec ±1.50% (49 runs sampled)
* dom/css-selector/querySelectorAll
  * jsdom  x 1,110 ops/sec ±1.84% (82 runs sampled)
  * jsdom  x 38.02 ops/sec ±2.99% (49 runs sampled)

test/web-platform-tests/to-run.yaml
```diff
@@ -1191,14 +1190,14 @@ template-element/template-content-hierarcy.html: [fail, "We do not implement htt
...
-pseudo-classes/dir.html: [fail, https://github.com/dperini/nwsapi/issues/45]
+pseudo-classes/dir-html-input-dynamic-text.html: [fail, Unknown]
+pseudo-classes/dir.html: [fail, Unknown]
 pseudo-classes/focus-autofocus.html: [timeout, Unknown]
 pseudo-classes/focus.html: [timeout, Unknown]
-pseudo-classes/indeterminate-type-change.html: [fail, Unknown]
 pseudo-classes/readwrite-readonly-type-change.html: [fail, Unknown]
 pseudo-classes/readwrite-readonly.html: [fail, Unknown]
-pseudo-classes/required-optional-hidden.html: [fail, Unknown]
+pseudo-classes/required-optional-hidden.html: [fail, Incorrect test]
```
asamuzaK, line 1199: There seems to be a bug in the testcase.
<input type="hidden"> should not match `:optional` pseudo-class.
2nd testcase should result `rgb(0, 255, 0)`.

asamuzaK:
Back to draft.
Still needs more fixes.

asamuzaK:
Benchmark results
Above: master (reposted), below: PR (dom-selector v0.15.13)

* dom/css-selector/closest
  * jsdom x 8,818 ops/sec ±2.10% (83 runs sampled)
  * jsdom x 5,312 ops/sec ±6.69% (78 runs sampled)
* dom/css-selector/matches
  * jsdom x 1,228,765 ops/sec ±1.89% (86 runs sampled)
  * jsdom x 153 ops/sec ±1.82% (74 runs sampled)
* dom/css-selector/querySelector
  * jsdom x 312 ops/sec ±1.96% (81 runs sampled)
  * jsdom x 160 ops/sec ±1.21% (75 runs sampled)
* dom/css-selector/querySelectorAll
  * jsdom x 1,110 ops/sec ±1.84% (82 runs sampled)
  * jsdom x 154 ops/sec ±2.27% (75 runs sampled)

asamuzaK:
I will withdraw the fixes for the jsdom issues at the moment, and change this PR to only request a library replacement.

About the jsdom issues, I'm planning to create test cases in WPT format as a follow-up PRs later.
How does that sound?

domenic:
No, sorry. We can't accept a library replacement with that kind of performance degredation, and I'm also not interested in accepting PRs without tests (in particular, tests that follow the guidelines in our CONTRIBUTING.md).

PR #3549, opened 2023-05-15
Beign able to check location changes

cbesa-belike:
Unit tests check inputs and outputs. It is not necessary to have navigation implemented, but you should be able to check the changes in the url, since it is an output.
With this change, it is now allowed to set the url to be able to check the changes in location.
Merge it if you think it's convenient.
Thank you.

### DIFF
--- a/lib/jsdom/living/window/Location-impl.js
+++ b/lib/jsdom/living/window/Location-impl.js
@@ -32,9 +32,9 @@ exports.implementation = class LocationImpl {
   _locationObjectNavigate(url, { replacement = false } = {}) {
     // Not implemented: the setup for calling navigate, which doesn't apply to our stub navigate anyway.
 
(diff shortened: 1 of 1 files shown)


domenic:
We are not willing to violate the spec in this way.

domenic:
We are not willing to violate the spec in this way.

cbesa-belike:
Good morning. Where can I consult the spec? Thanks.

domenic:
https://html.spec.whatwg.org/

cbesa-belike:
ok. I think this commit doesn't violate the specification, because it specifies that location.href **can be set**. So it implements the specification.

`location.[href](https://html.spec.whatwg.org/#dom-location-href)

    Returns the [Location](https://html.spec.whatwg.org/#location) object's URL.

    Can be set, to navigate to the given URL.`
    
But, in the case of not implementing this. In what way can we test functions that change the url?

Thanks again! 😉

PR #3551, opened 2023-05-23
Implement crypto.randomUUID()

jamesbvaughan:
This implements `crypto.randomUUID()` so that it's available in jsdom environments.

This is my first contribution to jsdom, so let me know if I should be doing anything differently here. I mostly followed #3352 as an example.


domenic (approved):
Almost perfect! Just missing `[SecureContext]` in the IDL. I will add that and merge.

Issue #3444, opened 2022-10-11
Support `document.adoptedStyleSheets`


nnmax:
### Basic info:

- **Node.js version: 16.17.0**
- **jsdom version: 20.0.0**

### Minimal reproduction case

```js
const { JSDOM } = require('jsdom')

const dom = new JSDOM(`
  <div>foo</div>
`)

console.log(dom.window.document.adoptedStyleSheets) // `undefined`
```
[adoptedStyleSheets API](https://developer.mozilla.org/en-US/docs/Web/API/Document/adoptedStyleSheets)

### How does similar code behave in browsers?

(Link to a jsbin or similar strongly suggested.)

nnmax:
Do you have any ideas for supporting this API? I'd be happy to submit a PR for support if I could.

domenic:
I'd suggest implementing it, according to its specification.

nnmax:
I need to write tests for my custom components. My custom components use both APIs and reports errors when I run the tests.

**Custom components:**

```TypeScript
export default class Button extends HTMLElement {
  static formAssociated = true

  #internals: ElementInternals

  constructor() {
    super()
    this.attachShadow({ mode: 'open' })
    this.shadowRoot.adoptedStyleSheets.push(styles)
    this.#internals = this.attachInternals()
  }

  // ...some code...

  #render() {
    this.shadowRoot.innerHTML = `
      <button tabindex="-1">
        <slot></slot>
        <max-ripple></max-ripple>
      </button>
    `
  }
}
```

**Test file:**

```TypeScript
describe('Testing <max-button />', () => {
  it('custom elements should be defined', () => {
    const button = window.customElements.get('max-button')
    expect(button).toBeDefined()
  })

  // TODO: await for JSDOM support `HTMLElement.attachInternals()` API
  // https://github.com/jsdom/jsdom/issues/3444
  it.skip.each([
    { attribute: 'variant', value: 'contained' },
    { attribute: 'variant', value: 'outlined' },
    { attribute: 'variant', value: 'text' },
    { attribute: 'color', value: 'primary' },
    { attribute: 'color', value: 'secondary' },
    { attribute: 'type', value: 'button' },
    { attribute: 'type', value: 'submit' },
    { attribute: 'type', value: 'reset' },
    { attribute: 'disabled', value: '' },
    { attribute: 'role', value: 'menuitem' },
    { attribute: 'tabindex', value: '-1' },
[...]

PR #3561, opened 2023-06-02
Implement `HTMLElement.attachInternals()`

zjffun:
- Issue: #3444
- [`HTMLElement.attachInternals()` spec](https://html.spec.whatwg.org/multipage/custom-elements.html#attached-internals)
- [Chromium implement](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/html_element.cc;drc=b44e2b7a8a0df47f5a2d114d82ff8d5375e13bd7;l=2004)


lib/jsdom/living/nodes/HTMLElement-impl.js
```diff
@@ -35,6 +42,54 @@ class HTMLElementImpl extends ElementImpl {
...
+      ]);
+    }
+
+    this._attachedInternals = ElementInternals.new(this._globalObject);
+    const attachedInternalsImpl = wrapperForImpl(this._attachedInternals);
+
+    // https://html.spec.whatwg.org/multipage/custom-elements.html#internals-target
+    attachedInternalsImpl._targetElement = this;
+
+    return attachedInternalsImpl;
```
zjffun, line 90: I'm not sure whether there should use  `ElementInternals.new` and `wrapperForImpl`, or use other method.

zmullett:
+1 Can confirm anecdotally that this PR permits unit-testing [Lit](https://lit.dev/) elements with jsdom 👍

domenic:
I am incorporating a smaller version of this in https://github.com/jsdom/jsdom/pull/3654, with only basic ElementInternals. Then I will try to build a version with ARIAMixin on top, that includes both the changes here and in #3586.

PR #3571, opened 2023-07-03
fix: Handle NaN correctly in Input.valueAsNumber setter

alexandertrefz:
[Relevant part of the Spec](https://html.spec.whatwg.org/multipage/input.html#dom-input-valueasnumber)

NaN is neither finite nor infinite, and requires special handling per the spec, rather than being thrown out with the infinite values at the beginning of the setter.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLInputElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLInputElement-impl.js
@@ -483,7 +483,7 @@ class HTMLInputElementImpl extends HTMLElementImpl {
   }
 
   set valueAsNumber(v) {
-    if (!isFinite(v)) {
+    if (!isFinite(v) && !isNaN(v)) {
       throw new TypeError("Failed to set infinite value as Number");
     }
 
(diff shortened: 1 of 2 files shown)


domenic:
Thanks. Can you be sure to add a test in the to-upstream directory, per CONTRIBUTING.md?

alexandertrefz:
@domenic Like this?

domenic:
No. From [Contributing.md](https://github.com/jsdom/jsdom/blob/master/Contributing.md):

> All tests for web platform features (as opposed to features of jsdom itself, such as the JSDOM() constructor) should be in [web-platform-tests](https://github.com/web-platform-tests/wpt) format.

> If you need to write your own test for a web platform feature, place it in our [to-upstream](https://github.com/jsdom/jsdom/tree/master/test/web-platform-tests/to-upstream) directory.

> Although ideally you should not need to worry about this, there are some tests that are for legacy reasons not in the right format; they use Mocha, but really should be web platform tests. We're keeping them around for coverage until we can convert them.

You added new legacy tests, not the ones in the directory I indicated above and contributing.md mentions.

alexandertrefz:
Yes I did: https://github.com/jsdom/jsdom/pull/3571/files#diff-cb4bf269785adf0cb5642caf787bfef7699bbb3d36e046096e29f1e89beb21fc is in that folder that the contributing.md is pointing at?

I guess I also added the legacy tests, but I can remove those if you like.

alexandertrefz:
@domenic I have removed the legacy tests. Can you check the other ones, as I am unsure that I have the right syntax for the tests there, but I do believe them to be in the right directory.

Issue #3572, opened 2023-07-04
Update tough-cookie package to version 4.1.3 - critical vulnerability


sveem:
Update **tough-cookie** dependency to the latest version 4.1.3 due to a critical vulnerability.

**Reason:** Versions of the package tough-cookie before 4.1.3 are vulnerable to Prototype Pollution due to improper handling of Cookies when using CookieJar in rejectPublicSuffixes=false mode. This issue arises from the manner in which the objects are initialized.

https://github.com/salesforce/tough-cookie/releases/tag/v4.1.3

amonnphillip:
When do you think the PR for this will be merged?

sl-arnaudmathieu:
Hi, a new release will be released to include this fix in the last version ? Cheers

domenic:
jsdom does not set the `rejectPublicSuffixes` flag, so this issue does not affect us.

carledren:
My opinion on the matter. If a vulnerability doesn't pose a direct threat in a specific context, it can still be good practice to update dependencies to their latest secure versions where feasible. This proactive approach can help prevent future vulnerabilities that might be introduced if the usage of the dependency changes, and can also alleviate concerns from users or contributors who might be alarmed to see a known vulnerability in the dependency tree.

As for people looking for a fix to this vulnerability, one can add the latest tough-cookie in an:
`override: {}` field in your package.json. See [overrides field](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides)

domenic:
jsom already uses the `^4.1.2` semver specifier so anyone who installs it will get the latest version of tough-cookie in the 4.x range.

bovas85:
4.1.2 is vulnerable, the semver won't fix that

PR #3574, opened 2023-07-06
Critical fix: Updated tough-cookie package to 4.1.3

sveem:
Critical update for tough-cookie dependency to the latest version 4.1.3 due to a security [vulnerability](https://github.com/salesforce/tough-cookie/releases/tag/v4.1.3)

**Issue**: Update tough-cookie package to version 4.1.3 - critical vulnerability #[3572](https://github.com/jsdom/jsdom/issues/3572)

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "rrweb-cssom": "^0.6.0",
     "saxes": "^6.0.0",
     "symbol-tree": "^3.2.4",
-    "tough-cookie": "^4.1.2",
+    "tough-cookie": "^4.1.3",
     "w3c-xmlserializer": "^4.0.0",
     "webidl-conversions": "^7.0.0",
     "whatwg-encoding": "^2.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1219,10 +1219,10 @@ to-regex-range@^5.0.1:
   dependencies:
     is-number "^7.0.0"
 
-tough-cookie@^4.1.2:
-  version "4.1.2"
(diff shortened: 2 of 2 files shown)


sveem:
Critical update for tough-cookie dependency to the latest version 4.1.3 due to a security [vulnerability](https://github.com/salesforce/tough-cookie/releases/tag/v4.1.3)

**Issue**: Update tough-cookie package to version 4.1.3 - critical vulnerability #[3572](https://github.com/jsdom/jsdom/issues/3572)

yasharzolmajdi (approved):

epreston (approved):

mmncit (approved):

sveem:
@epreston @mmncit @yasharzolmajdi - When do you think this PR will be merged?

epreston:
I am not a maintainer.  I've just pulled the changes, tested them, and linked issues to help out.

domenic:
jsdom does not set the `rejectPublicSuffixes` flag, so this is change is not necessary.

BreakBB:
Does it hurt though to update to a version with one CVE less to worry about @domenic ? 🤔

domenic:
jsom already uses the ^4.1.2 semver specifier so anyone who installs it will get the latest version of tough-cookie in the 4.x range.

> 1. What about future changes, which might not think about this potential security issue?

They will also be unaffected, per the above.

> 2\. Not setting it might result in a `falsy` value and therefore open the breach?

From my testing that is not correct.

> 2\. Maybe the default will change towards `false` as well in the future?

That would be a breaking change so it will not happen in the 4.x line.

BreakBB:
Thanks for the detailed response on this :) That reassures me, this is actually not required 👍🏻

PR #3583, opened 2023-07-27
Add option to display invalid constraints on form validation in console

SamuelWei:
Since #3249 a form must be valid to be submitted.

This PR intends to add a new config option `reportInvalidConstraintsToConsole` to help developers find error in their tests.
In a regular web browser the browser informs the user on the failed form fields.
This option should bring this feature to jsdom to help developers indentify which fields are causing their form to not be submitted during testing.

---

This PR stills needs some work by someone who is familar with the dev. of jsdom. The config option is not correctly passed down to the HTMLFormElement-impl.js file yet. Softwaretests are still missing.

Feel free to edit and modify everything!

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -214,6 +214,8 @@ function transformOptions(options, encoding, mimeType) {
       pretendToBeVisual: false,
       storageQuota: 5000000,
 
+      _reportInvalidConstraintsToConsole: false,
+
       // Defaults filled in later
       resourceLoader: undefined,
       virtualConsole: undefined,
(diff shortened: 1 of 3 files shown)


domenic:
I don't think we want to add options to jsdom to make it behave less like a standards-compliant browser.

I suggest you make your tests pass the same way you would make them pass in a real browser. For example, by setting the novalidate attribute on the form during the test.

domenic:
I'm very sorry. I did not properly read the PR. Reporting to the console is a good idea. In fact I think we should do it by default.

SamuelWei:
@domenic Will you finish implementing the new function?

I don't know how to pass the config variable correctly to the HTMLFormElement-impl.js - otherwise you can also help me how to do that.

domenic:
No, sorry, I don't have time to work on this myself.

Issue #3323, opened 2022-01-21
Add Aria IDL Interface support


ashleyryan:
### Basic info:

jsdom does not support the Aria IDL interface for settings aria attributes and roles via properties instead of `setAttribute`. 

https://www.w3.org/TR/wai-aria-1.2/#idl-interface

- **Node.js version:** 14.6.0
- **jsdom version:** 19.0.0

### Minimal reproduction case

https://stackblitz.com/edit/node-c6cnee?file=index.js

```js
const { window } = new JSDOM(
  `
<!DOCTYPE html>
<html>
<head></head>
<body id="body">
  <span id="test">hello test</span>
  <script>
    document.getElementById('test').setAttribute('aria-label', 'My Aria Label');
    document.getElementById('test').ariaLabel='My Aria Label using a property';


    document.getElementById('test').setAttribute('role', 'button');
    document.getElementById('test').role='presentation';
  </script>
</body>
</html>
`,
  { runScripts: 'dangerously' }
);

console.log(window.document.getElementById('test').outerHTML);
```

### How does similar code behave in browsers?

In the latest Chrome and Safari, the ariaLabel and role are updated using the property.

https://stackblitz.com/edit/web-platform-bube3q

domenic:
Status update: string reflection is implemented, but element reflection is not yet.

PR #3586, opened 2023-08-11
Implement ARIAMixin string reflection

nolanlawson:
Partially addresses #3323. Implements string-based ARIA reflection (e.g. `ariaLabel`, `role`), but not element-based reflection (e.g. `ariaLabelledByElements`, `ariaActiveDescendantElement`).

This has some overlap with #3561, but looks like it should be largely complementary.

With this PR, the following works:

```js
element.setAttribute('aria-label', 'foo')
element.setAttribute('role', 'button')
console.log(element.ariaLabel) // 'foo'
console.log(element.role) // 'button'
```

I confirmed that the [`aria-element-reflection.html` WPT test](https://github.com/web-platform-tests/wpt/blob/master/html/dom/aria-attribute-reflection.html) is now passing.

### DIFF
--- a/lib/jsdom/living/nodes/ARIAMixin-impl.js
+++ b/lib/jsdom/living/nodes/ARIAMixin-impl.js
@@ -1,0 +1,75 @@
+"use strict";
+
+// ARIA Element properties that reflect to strings (not Element or FrozenArray<Element>)
+const ARIA_STRING_PROPS = [
+  "role",
+  "ariaAtomic",
+  "ariaAutoComplete",
+  "ariaBusy",
+  "ariaChecked",
+  "ariaColCount",
+  "ariaColIndex",
+  "ariaColIndexText",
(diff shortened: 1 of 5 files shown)


lib/jsdom/living/nodes/ARIAMixin-impl.js
```diff
@@ -0,0 +1,77 @@
...
+      // differ from the spec and allow undefined as well. Here, we follow the standard.
+      // See: https://github.com/w3c/aria/issues/1858
+      if (value === null) {
+        this.removeAttributeNS(null, attribute);
+      } else {
+        this.setAttributeNS(null, attribute, value);
+      }
+    },
+    configurable: true,
+    enumerable: true
```
nolanlawson, line 71: I couldn't figure out if this is officially specc'ed, but both WebKit and Chromium have these as configurable and enumerable.

PR #3593, opened 2023-09-07
fix CVE-2023-26136 Security fix for Prototype Pollution

micro-chipset:
https://github.com/salesforce/tough-cookie/releases/tag/v4.1.3 
Security fix for Prototype Pollution discovery in https://github.com/salesforce/tough-cookie/issues/282. This is a minor release, although output from the inspect utility is affected by this change, we felt this change was important enough to be pushed into the next patch.


domenic:
Dupe of closed issues and PRs.

PR #3601, opened 2023-09-24
Simplified Array.from in convertToSequenceDOMString

Connormiha:
`Array.from` has optional second argument. Calling `map` is not required for this case.


AlexErrant:
Interesting! j/w, do you know if this has perf benefits? I _imagine_ it might allocate less memory...

Connormiha:
Performance +- the same

AlexErrant:
Hah yeah I guess V8 is smarter than me.

domenic (approved):

Issue #3604, opened 2023-10-06
new URL(…).searchParams.size is not implemented


Ayc0:
### Basic info:

- **Node.js version:** v18.15.0
- **jsdom version:** v22.1.0

### Minimal reproduction case

```js
const { JSDOM } = require('jsdom');
const jsdom = new JSDOM();
console.log(new jsdom.window.URL('https://github.com/jsdom/jsdom').searchParams.size);
// undefined & should be 0

console.log(new jsdom.window.URL('https://github.com/jsdom/jsdom?hello').searchParams.size);
// undefined & should be 1
```

### How does similar code behave in browsers?

In the chrome devtools:

```js
new URL('https://github.com/jsdom/jsdom').searchParams.size
new URL('https://github.com/jsdom/jsdom?hello').searchParams.size
```

<img width="493" alt="image" src="https://github.com/jsdom/jsdom/assets/22725671/fb1a2292-7681-475f-b0fb-95b020681cf9">

Ayc0:
To fix it locally in my codebase, I added this in the `setupFilesAfterEnv` file for my Jest conf:

```js
Object.defineProperty(URLSearchParams.prototype, 'size', {
    get() {
        let size = 0;
        for (const entry of this) {
            size++;
        }
        return size;
    },
});
```

Ayc0:
I can see that `whatwg-url` added this in their last version, so we could just upgrade `whatwg-url` to the v13:

https://github.com/jsdom/whatwg-url/releases/tag/v13.0.0

> Added `URLSearchParams`'s `size` getter, per https://github.com/whatwg/url/commit/12b6f0c456c6df049e2704c92bb3a6d4d1364ec8.

Ayc0:
I opened https://github.com/jsdom/jsdom/pull/3606 to fix this issue

jimmywarting:
Erm... if we would use global built in URL / SearchParams this wouldn't be an issue. 
Maybe we could use built in instead?

PR #3606, opened 2023-10-09
Update `whatwg-url` to the v23

Ayc0:
Updated `whatwg-url` to the v23 for https://github.com/jsdom/jsdom/issues/3604

As per https://github.com/jsdom/whatwg-url/releases/tag/v13.0.0:
 
> Breaking change: removed Node.js v14 support.
> 
> Added URL.canParse(), per https://github.com/whatwg/url/commit/ae3c28b84e3e7122c2807401c26b8a63cb2ab445.
> 
> Added URLSearchParams's size getter, per https://github.com/whatwg/url/commit/12b6f0c456c6df049e2704c92bb3a6d4d1364ec8.
> 
> Added optional second value argument to URLSearchParams's has() and delete() methods, per https://github.com/whatwg/url/commit/bfb9157186c237078cb1ac4998607d88242abe35.
>
> Changed the serialization of the origin of blob: URLs whose inner URLs were not http: or https: to be "null", per https://github.com/whatwg/url/commit/eee49fdf4f99d59f717cbeb0bce29fda930196d4.

JSDOM already only offers support for Node v16+

I also used yarn resolutions to force `whatwg-url` to only be present in the v23 because `data-urls` uses the v22. This can be removed if `data-urls` was also update to be able to use the v23

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,9 +40,12 @@
     "webidl-conversions": "^7.0.0",
     "whatwg-encoding": "^2.0.0",
     "whatwg-mimetype": "^3.0.0",
-    "whatwg-url": "^12.0.1",
+    "whatwg-url": "^13.0.0",
     "ws": "^8.13.0",
     "xml-name-validator": "^4.0.0"
+  },
+  "resolutions": {
+    "whatwg-url": "^13.0.0"
   },
   "peerDependencies": {
     "canvas": "^2.5.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -40,10 +40,13 @@
...
     "whatwg-encoding": "^2.0.0",
     "whatwg-mimetype": "^3.0.0",
-    "whatwg-url": "^12.0.1",
+    "whatwg-url": "^13.0.0",
     "ws": "^8.13.0",
     "xml-name-validator": "^4.0.0"
   },
+  "resolutions": {
+    "whatwg-url": "^13.0.0"
+  },
```
Ayc0, line 49: This resolutions is to say to yarn that even if `data-urls` asks the vesion `^12.0.0`, we should use the version `^13.0.0`

I can remove it but this would create a duplication of the library `whatwg-url` and inconstancies with URL manipulations in different places for JSDOM

Issue #3614, opened 2023-10-27
`getComputedStyle()` doesn't inherit `font-family`


danon:
### Basic info:

- **Node.js version:** 18.13.0
- **jsdom version:** 22.1.0

### Minimal reproduction case

```js
const jsdom = require("jsdom");
const {JSDOM} = jsdom;

const {window} = new JSDOM(`<!DOCTYPE html>
<html>
    <body>
        <div id="parent" style="font-family: Arial;">
            <div id="child">Hello, Dave!</div>
        </div>
    </body>
</html>`);

const parent = window.document.querySelector("#parent");
const child = window.document.querySelector("#child");

console.log(window.getComputedStyle(parent).fontFamily);
console.log(window.getComputedStyle(child).fontFamily);
```

### How does similar code behave in browsers?

In browser the font family is inherited.

PR #3615, opened 2023-10-27
Implement getComputedStyle() for font-family (#3614)

danon:
This is my first to this repository.

I tried to follow example in a8b03afb. @domenic let me know if I did it correctly?

### DIFF
--- a/lib/jsdom/living/helpers/style-rules.js
+++ b/lib/jsdom/living/helpers/style-rules.js
@@ -92,6 +92,12 @@ exports.propertiesWithResolvedValueImplemented = {
     inherited: false,
     initial: "invert",
     computedValue: "computed-color"
+  },
+  // https://drafts.csswg.org/css-fonts-4/#font-family-prop
+  "font-family": {
+    inherited: true,
+    initial: "",
+    computedValue: "as-specified"
   }
 };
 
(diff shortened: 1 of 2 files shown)


lib/jsdom/living/helpers/style-rules.js
```diff
@@ -92,6 +92,12 @@ exports.propertiesWithResolvedValueImplemented = {
     inherited: false,
     initial: "invert",
     computedValue: "computed-color"
+  },
+  // https://drafts.csswg.org/css-fonts-4/#font-family-prop
+  "font-family": {
+    inherited: true,
+    initial: "",
```
domenic, line 99: It looks like this needs to be set to something like `"serif"`. Can you do that, and then update the test? https://jsbin.com/tulemoreka/edit?html,output

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -92,6 +92,12 @@ exports.propertiesWithResolvedValueImplemented = {
...
+    computedValue: "as-specified"
```
domenic, line 100: This doesn't appear correct. Per spec it's

> list, each item a string and/or [<generic-family>](https://drafts.csswg.org/css-fonts-4/#generic-family-value) keywords

The appropriate test to add is something like

```html
<div style="font-family: serif,sans-serif,     'Times New Roman',Arial"></div>
```

and then testing that the computed style is `serif, sans-serif, "Times New Roman", Arial` (with the right quotes and spacing).

Making that work correctly would involve introducing some new infrastructure, probably around here? https://github.com/jsdom/jsdom/blob/907daecb80b33069354d8ce7ee9381c39ffa8dd2/lib/jsdom/living/helpers/style-rules.js#L214-L237

domenic (changes requested):

Jaskaran-Techno (changes requested):
reviewed by me.

PR #3625, opened 2023-11-26
Remove abab dependency

domenic:


lib/jsdom/browser/Window.js
```diff
@@ -757,25 +756,27 @@ function Window(options) {
   this.postMessage = postMessage(window);
 
   this.atob = function (str) {
-    const result = atob(str);
-    if (result === null) {
+    try {
+      return atob(str);
```
DoodleBobBuffPants, line 760: I'm having an issue upgrading to `jsdom` v23 resulting in an `InvalidCharacterError`. It feels more time consuming to go from my toolchain (React, MUI, Vitest) to a minimal reproduction than to point out the source of the issue. If I throw `e` on line 761 I see that it's a `RangeError: Maximum call stack size exceeded`. The call to `atob` on line 760 is recursive, so this always fails?

PR #3629, opened 2023-11-30
Changed examples within the "Executing scripts" section

DanKaplanSES:
Pros:
- For each example, the HTML is exactly the same, allowing the reader to focus on what matters: the effect of different `runScripts` values.
- When a `<script>` is a noop, `0` is logged instead of `1`. IMO, this is more intuitive and reduces cognitive load.
- The examples can be copied and pasted elsewhere and, without modification, output useful information. Before, the last line of each example was something like `x === y`, which won't always show output (e.g., running in a test framework). 

Cons:
- The examples are longer
- I added `<div id="content"></div>` into the body to separate script elements from script outcomes. This might make the HTML harder to read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,22 +65,24 @@ However, this is also highly dangerous when dealing with untrusted content. The 
 
 ```js
 const dom = new JSDOM(`<body>
-  <script>document.body.appendChild(document.createElement("hr"));</script>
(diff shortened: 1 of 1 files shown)


domenic (approved):
Thanks for this; I think it's overall an improvement!

PR #3633, opened 2023-12-04
Update wpt

asamuzaK:
Updated wpt tests and related files.

Some notes:
* `package.json`
  * Updated `update-wpt` script.
    * Added `--init` cmd arg.
    * Changed `python3` to `python`.
* `wpt-config.json` and `tuwpt-config.json`
  * Changed `"log_level": "info",` to
    ```
    "logging": {
      "level": "info"
    },
    ```
* `to-upstream/html/webappapis/timers/settimeout-setinterval-handles.html`
  * Fixed test failure.
    It seems that setTimeout / setInterval is called somewhere outside the test, maybe in `testharness.js`.
* `wpt-server.js`
  * Added new function to avoid infinite server startup attempts.


asamuzaK:
PR ready.
For logs in case of server errors, see [Update wpt · jsdom/jsdom@6b171c5](https://github.com/jsdom/jsdom/actions/runs/7121294425).

test/web-platform-tests/wpt-server.js
```diff
@@ -33,13 +35,14 @@ exports.start = async ({ toUpstream = false } = {}) => {
...
     cwd: wptDir,
     stdio: ["inherit", "pipe", "pipe"]
   });
 
   subprocess.stdout.filter(nonSpammyWPTLog).pipe(process.stdout);
   subprocess.stderr.filter(nonSpammyWPTLog).pipe(process.stderr);
+  subprocess.stderr.on("data", terminateWptOnKeyError);
 
   return new Promise((resolve, reject) => {
     subprocess.on("error", e => {
```
asamuzaK, line 48: In what cases is this error expected to be triggered?
It seems that the format of `new Error()` is wrong in the first place.

asamuzaK: I think we should add an error listener outside the function.
I'll push 1 more commit to fix this.

domenic (approved):
Truly amazing work; thank you so much! I pushed an extra commit with some more runner improvements; after CI completes on that I'll merge.

Issue #3635, opened 2023-12-09
SyntaxError: Unexpected `||=` Operator on Web Service Hosting


JAYAKANTHARUN:
### Basic info:

- **Node.js version:** v18.16.1
- **jsdom version:** ^23.0.1

### Minimal reproduction case

```js
const express = require('express')
const app = express()
const { JSDOM } = require("jsdom");

app.post('/submit2',(req,res)=>{
    let currentTextarea2 = req.body.textarea3 || '';
    currentTextarea2 = currentTextarea2.replace(/<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/gi, '');
    const dom = new JSDOM(currentTextarea2);
    const liElements = dom.window.document.querySelectorAll('li');
    let liArray = [];
    liElements.forEach(li => {
        liArray.push(li.outerHTML);
    });
    console.log(liArray)           //The array of `<li>` elements will be displayed in the terminal 
});

```

### How does similar code behave in browsers?

When running the code on localhost, no errors occur. However, when attempting to host it on render.com, an error arises.

```js
/opt/render/project/src/node_modules/jsdom/lib/jsdom/living/nodes/HTMLInputElement-impl.js:583
      this[filesSymbol] ||= FileList.createImpl(this._globalObject);
                        ^^^
SyntaxError: Unexpected token '||='
    at wrapSafe (internal/modules/cjs/loader.js:984:16)
    at Module._compile (internal/modules/cjs/loader.js:1032:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1097:10)
    at Module.load (internal/modules/cjs/loader.js:933:32)
    at Function.Module._load (internal/modules/cjs/loader.js:774:14)
    at Module.require (internal/modules/cjs/loader.js:957:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/opt/render/project/src/node_modules/jsdom/lib/jsdom/living/generated/HTMLInputElement.js:1875:14)
    at Module._compile (internal/modules/cjs/loader.js:1068:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1097:10)
[nodemon] app crashed - waiting for file changes before starting...
```

asamuzaK:
Set nodejs version and try again.
See https://render.com/docs/node-version

I don't think it's an issue of jsdom.

JAYAKANTHARUN:
[...]

PR #3636, opened 2023-12-09
Issue #3635 : Fix Operator Syntax Error

JAYAKANTHARUN:
#3635 
The code provided includes the usage of the ||= operator, which was introduced in ECMAScript 2021. The error encountered on render.com indicates that the platform might be running an environment that does not support this particular JavaScript feature.

When running similar code in browsers, the behavior depends on the browser's JavaScript engine and its support for ECMAScript features. Most modern browsers support ECMAScript 2021 features, including the ||= operator. However, some older browsers or specific environments may not fully support the latest ECMAScript specifications.

If you encounter a similar issue in a browser, it could be due to using an outdated or less feature-rich browser that does not support ECMAScript 2021. In such cases, you might need to consider alternative approaches or transpile your code to an earlier ECMAScript version to ensure compatibility with a broader range of browsers.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLInputElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLInputElement-impl.js
@@ -580,7 +580,7 @@ class HTMLInputElementImpl extends HTMLElementImpl {
 
   get files() {
     if (this.type === "file") {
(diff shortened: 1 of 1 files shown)


domenic:
https://github.com/jsdom/jsdom/issues/3635#issuecomment-1850253190

PR #3639, opened 2023-12-11
Mark `Element.scrollIntoView` as not implemented

### LINKED ISSUE #1695: scrollIntoView is not a function

Cipscis:
Resolves #1695 

Because JSDom can't handle layout, `Element.scrollIntoView` can't really be implemented. However, it hasn't been marked as not implemented like several other scrolling methods have, for example `window.scrollTo`.

This PR marks `Element.scrollIntoView` as not implemented, with the purpose of improved error reporting when users write tests than invoke `Element.scrollIntoView`.

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -26,6 +26,7 @@ const ShadowRoot = require("../generated/ShadowRoot");
 const Text = require("../generated/Text");
 const { isValidHostElementName } = require("../helpers/shadow-dom");
 const { isValidCustomElementName, lookupCEDefinition } = require("../helpers/custom-elements");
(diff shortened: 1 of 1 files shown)


Benjin:
@domenic could you or another maintainer handle this PR, please?

agrberg (approved):
This would really help. Documented my discovery in https://github.com/jsdom/jsdom/issues/1695#issuecomment-4471849782

PR #3646, opened 2023-12-26
feat: HTML Search Element

danilofuchs:
Recently introduced in Chrome 118, a new semantic element containing search bar and filters. It has no styles or behaviour.

https://html.spec.whatwg.org/multipage/grouping-content.html#the-search-element

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/search


domenic:
This is not enough for an implementation. In particular, you need to pass the tests in https://github.com/web-platform-tests/wpt/pull/39158 and https://github.com/web-platform-tests/wpt/pull/39163, and not modify legacy tests, per CONTRIBUTING.md.

PR #3644, opened 2023-12-23
Proposal: New CSS selector engine for jsdom

asamuzaK:
`nwsapi` has been working great so far, and I appreciate the developer's hard work, but `nwsapi` is unable to support Selectors Level 4 properly at the moment.
You can find many selector-related bug reports in the `jsdom` issues.

Therefore, I recommend changing to a different CSS selector engine [dom-selector](https://github.com/asamuzaK/domSelector).
It supports many features of the Selectors Level 4 and also the Shadow Hosts.
Ref [Supported CSS selectors](https://github.com/asamuzaK/domSelector#supported-css-selectors)

# Benchmark results

```
npm run benchmark
```

|API|nwsapi|dom-selector|
|:--------|:--------|:--------|
|closest()|jsdom x 126,133 ops/sec ±0.75% (73 runs sampled)|jsdom x 7,449 ops/sec ±13.28% (64 runs sampled)|
|matches()|jsdom x 120,369 ops/sec ±8.31% (65 runs sampled)|jsdom x 8,507 ops/sec ±7.84% (60 runs sampled)|
|querySelector()|jsdom x 3.12 ops/sec ±2.60% (12 runs sampled)|jsdom x 14.72 ops/sec ±1.94% (35 runs sampled)|
|querySelectorAll()|jsdom x 2.93 ops/sec ±1.52% (11 runs sampled)|jsdom x 12.41 ops/sec ±2.47% (31 runs sampled)|

For `closest()` and `matches()`, `nwsapi` is about 15 times faster, but in reality, `dom-selector` processed in 0.11 to 0.13 msec, so I don't think it's extremely slow.
Rather, it is `querySelector()` and `querySelectorAll()` that often have speed performance issues, and in this benchmark, `dom-selector` is about 4 times faster than `nwsapi`.

PR #3647, opened 2023-12-27
Add new benchmark for css selectors

asamuzaK:
Split from #3644

The reasons for using `try-catch` are:
* The selector list contains invalid selectors.
* Even if the selector is valid, an exception will be thrown if nwsapi fails to parse.

Benchmark result on my local PC is:
```
# selectors/selector/querySelectorAll #
18/273 fails.
jsdom  x 1.14 ops/sec ±5.54% (7 runs sampled)
```


asamuzaK:
The more errors there are, the *shorter* the processing time is, so I think it to be **slower** if it is tested on dom-selector.

.gitmodules
```diff
@@ -1,3 +1,7 @@
 [submodule "test/web-platform-tests/tests"]
 	path = test/web-platform-tests/tests
 	url = https://github.com/web-platform-tests/wpt.git
+[submodule "benchmark/selectors/sizzle"]
```
domenic, line 4: Since submodules are already a pain, and I don't think we'll need to update this one very often, I think it'd be best to just directly check in the two relevant files. We can add a or modify a README.md in an appropriate directory to explain where they came from.

benchmark/selectors/selector.js
```diff
@@ -0,0 +1,49 @@
...
+
+exports.querySelectorAll = function () {
+  let document, selectors, total;
+  let count = 0;
+  let cycle = 0;
+
+  return suite({
+    setup() {
+      const dir = "./benchmark/selectors/sizzle/speed/";
+      const htmlFile = path.resolve(process.cwd(), dir, "./data/selector.html");
```
domenic, line 17: Resolving relative to `process.cwd()` is a bad idea, since it prevents people from running the test in different places. (E.g., I think this will fail if one does `cd benchmark; node ./runner`.) Instead, use `__dirname`, since that never changes.

.gitmodules
asamuzaK: Done.
I'll leave the README up to you.

benchmark/selectors/selector.js
asamuzaK: Done.

asamuzaK:
Log:
```
# selectors/selector/querySelectorAll #
18/273 fails.
jsdom  x 1.15 ops/sec ±6.53% (7 runs sampled)

# selectors/selector/querySelectorAll only nwsapi supported #
jsdom  x 1.18 ops/sec ±11.85% (7 runs sampled)
```

domenic:
> Log:

Great! And what are the results for dom-selector?

asamuzaK:
I haven't tested it yet.
Once this PR is merged, I plan to rebase #3644 and test it there.

benchmark/selectors/sizzle-speed/images/favicon.ico
domenic, line 1: No need for this favicon.ico file

benchmark/selectors/sizzle-speed/data/selector.html
domenic, line 1: We don't need the data/ subdirectory, IMO; we can just keep this as a sibling of selectors.large.css

benchmark/selectors/selector.js
```diff
@@ -0,0 +1,148 @@
...
+    },
+    onComplete() {
+      // eslint-disable-next-line no-console
+      console.log(`${(count / cycle).toFixed(0)}/${total} fails.`);
+    }
+  });
+};
+
+// This is only for comparison. Can remove later.
+/* eslint-disable quotes, max-len */
```
domenic, line 50: Maybe don't disable `quotes`, and just use `npm run lint -- --fix` to fixup the single quotes?

domenic (approved):
Looks good with nits; please fix those and then we'll merge. (I apologize, I might merge tomorrow, as I need to head to bed soon...)

benchmark/selectors/sizzle-speed/images/favicon.ico
asamuzaK: Done.

benchmark/selectors/sizzle-speed/data/selector.html
asamuzaK: Done.

benchmark/selectors/selector.js
asamuzaK: Done.

asamuzaK:
Latest log:
```
# selectors/selector/querySelectorAll #
18/273 fails.
jsdom  x 1.20 ops/sec ±10.21% (7 runs sampled)

# selectors/selector/querySelectorAll only nwsapi supports #
jsdom  x 1.24 ops/sec ±8.02% (7 runs sampled)
```

PR #3653, opened 2024-01-02
Documentation: Add additional information about the canvas package

DanKaplanSES:
- I removed the information about the canvas package being a "peer dependency" because, as far as I could tell, this was an implementation detail. If I'm right about that assumption, it could confuse users who are unfamiliar with the concept.
- I reiterated information about the relationship between the canvas package, the `` element, and the `ResourceLoader`. I believe this redundancy is warranted because users could spend a lot of time troubleshooting if they miss this detail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -440,7 +440,7 @@ console.log(frag.firstChild.outerHTML); // logs "<p>Hello</p>"
 jsdom includes support for using the [`canvas`](https://www.npmjs.com/package/canvas) package to extend any `<canvas>` elements with the canvas API. To make this work, you need to `npm install canvas@2` as a separate dependency (or devDependency) in your project. If jsdom can find the `canvas` package, it will use it, but if it's not present, then: 
 
 * `<canvas>` elements will behave like `<div>`s.
-* `<img>` elements' `src` attributes will not be loaded by the [`ResourceLoader`](#loading-subresources).
+* The `src` attribute of `<img>` elements will not be loaded by the [`ResourceLoader`](#loading-subresources).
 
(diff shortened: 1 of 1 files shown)


DanKaplanSES:
Not sure how to rerun this but the failure doesn't seem related to my change? LMK if I'm wrong about that.

domenic:
> * I removed the information about the canvas package being a "peer dependency" because, as far as I could tell, this was an implementation detail

This is not correct; a peer dependency is a well-documented concept in the npm ecosystem.

DanKaplanSES:
> > * I removed the information about the canvas package being a "peer dependency" because, as far as I could tell, this was an implementation detail
> 
> This is not correct; a peer dependency is a well-documented concept in the npm ecosystem.

Hi @domenic Apologies for how I worded that; I didn't mean to imply peer dependencies are an obscure topic. Though, maybe I don't understand them as well as I thought I did. To be honest, I just learned about them today by [reading this arti...](https://nodejs.org/en/blog/npm/peer-dependencies) Annnnd I just noticed you're the author. This is surreal. (It totally reminds me of a Futurama reference.)

Anyway, before I go, may I ask a clarifying question about the jsdom readme? [It says](https://github.com/jsdom/jsdom/pull/3653/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L440):

> To make this work, you need to include `canvas` as a dependency in your project, **as a peer of `jsdom`**.

But how does that bold part effect my behavior differently than if it was written like this:

> To make this work, you need to include `canvas` as a dependency in your project.

Thanks for your time!

PR #3668, opened 2024-01-21
Rewrite ID caching for correctness and maybe speed

domenic:
document.getElementById() now properly returns the first element with the given ID in tree order.


domenic:
I believe this is a performance regression at the moment because of the additional root node lookup that becomes necessary. I'm investigating caching root nodes, although maybe there's a way to avoid that sort of thing entirely...

PR #3674, opened 2024-01-27
Use css-color module

### LINKED ISSUE #3657: Improve support for CSS colors
### Basic info:

- **Node.js version:** v20.10.0
- **jsdom version:** 23.1.0

### Minimal reproduction case

```js

const domstr = `<!DOCTYPE html>
<html>
  <head>
    <style>
      #target1 {
        color: green;
      }
      #target2 {
        color: hsl(120 100% 25%);
      }
    </style>
  </head>
  <body>
    <div id="target1"></div>
    <div id="target2"></div>
  </body>
</html>
`;

const options = {
  runScripts: 'dangerously',
  url: 'http://localhost/'
};

const dom = new JSDOM(domstr, options);
const window = dom.window;

const target1 = document.getElementById('target1');
console.log(window.getComputedStyle(target1).color; // 'rgb(0, 128, 0)'

const target2 = document.getElementById('target2');
console.log(window.getComputedStyle(target2).color; // fails, empty string  returned
```

'rgb(0, 128, 0)' is returned as expected.

### Proposal

I took a look at [jsdom/lib/jsdom/living/helpers/colors.js](https://github.com/jsdom/jsdom/blob/main/lib/jsdom/living/helpers/colors.js) and found that CSS color support of jsdom is quite limited.
In another project of mine, I am using an internal module that can parse / convert colors defined in CSS Color Module Level 4.
[sidebarTabs/src/mjs/color.js](https://github.com/asamuzaK/sidebarTabs/blob/master/src/mjs/color.js)
If you are interested in using it with jsdom, I will publish it as a public module on npm.
How about that?

asamuzaK:
Fix #3657 

# Benchmark result

|Benchmark|main|patch|
|:----|:----|:----|
|# dom/get-computed-style/getComputedStyle() on document.body #|jsdom  x 24,373 ops/sec ±2.94% (95 runs sampled)|jsdom  x 29,307 ops/sec ±0.72% (90 runs sampled)|
|# dom/get-computed-style/getComputedStyle() on element with rgb color #|jsdom  x 13,206 ops/sec ±4.41% (86 runs sampled)|jsdom  x 13,911 ops/sec ±1.81% (87 runs sampled)|
|# dom/get-computed-style/getComputedStyle() on element with non-rgb color #|jsdom  x 24,526 ops/sec ±1.32% (93 runs sampled)|jsdom  x 29,089 ops/sec ±0.97% (90 runs sampled)|

The last `getComputedStyle() on element with non-rgb color` benchmark is simply failing, so ignore it.
It looks like jsdom does not pass non-rgb colors e.g. `color-mix()` correctly to `getSpecifiedColor()`.
We need to fix https://github.com/jsdom/jsdom/blob/main/lib/jsdom/living/helpers/style-rules.js#L175 to support them.


asamuzaK:
The commit log is a mess, so I'll close it and submit a new PR.

PR #3675, opened 2024-01-27
Use css-color module

### LINKED ISSUE #3657: Improve support for CSS colors

asamuzaK:
Fix #3657 

# Benchmark result

|Benchmark|main|patch|
|:--------|:--------|:--------|
|# dom/get-computed-style/getComputedStyle() on document.body #|jsdom  x 24,411 ops/sec ±2.90% (88 runs sampled)|jsdom  x 28,467 ops/sec ±1.05% (89 runs sampled)|
|# dom/get-computed-style/getComputedStyle() on element with non-rgb color #|jsdom  x 24,564 ops/sec ±1.24% (89 runs sampled)|jsdom  x 29,178 ops/sec ±0.91% (91 runs sampled)|
|# dom/get-computed-style/getComputedStyle() on element with rgb color #|jsdom  x 14,097 ops/sec ±4.15% (85 runs sampled)|jsdom  x 13,822 ops/sec ±2.08% (84 runs sampled)|

Ignore the 2nd `getComputedStyle() on element with non-rgb color` as it simply failed.
It looks like jsdom does not pass unsupported colors, e.g. `color-mix()`, correctly to `getSpecifiedColor()`.
We need to fix https://github.com/jsdom/jsdom/blob/main/lib/jsdom/living/helpers/style-rules.js#L175 to support them?

### DIFF
--- a/benchmark/dom/get-computed-style.js
+++ b/benchmark/dom/get-computed-style.js
@@ -1,0 +1,48 @@
+"use strict";
+const suite = require("../document-suite");
+
+exports["getComputedStyle() on document.body"] = () => {
+  let window, document;
+
+  return suite({
+    setup(doc) {
+      window = doc.defaultView;
+      document = doc;
+    },
+    fn() {
+      window.getComputedStyle(document.body);
+    }
+  });
+};
+
+exports["getComputedStyle() on element with rgb color"] = () => {
+  let window, node;
+
+  return suite({
+    setup(doc) {
+      window = doc.defaultView;
+      node = doc.createElement("div");
+      node.style.color = "rgb(128, 0, 128)";
+      doc.body.appendChild(node);
+    },
+    fn() {
+      window.getComputedStyle(node);
+    }
+  });
+};
+
+exports["getComputedStyle() on element with non-rgb color"] = () => {
+  let window, node;
+
+  return suite({
+    setup(doc) {
+      window = doc.defaultView;
+      node = doc.createElement("div");
+      node.style.color = "color-mix(in srgb, red, blue)";
+      doc.body.appendChild(node);
+    },
+    fn() {
+      window.getComputedStyle(node);
+    }
+  });
+};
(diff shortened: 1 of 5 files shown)


asamuzaK:
> We need to fix https://github.com/jsdom/jsdom/blob/main/lib/jsdom/living/helpers/style-rules.js#L175 to support them?

I found out that `cssstyle` only supports hex, rgb and hsl.
https://github.com/jsdom/cssstyle/blob/main/lib/parsers.js#L289
I think that `css-color` may also be useful for `cssstyle`.

asamuzaK:
> > We need to fix https://github.com/jsdom/jsdom/blob/main/lib/jsdom/living/helpers/style-rules.js#L175 to support them?
> 
> I found out that `cssstyle` only supports hex, rgb and hsl. https://github.com/jsdom/cssstyle/blob/main/lib/parsers.js#L289 I think that `css-color` may also be useful for `cssstyle`.

Filed a PR on `cssstyle` [Extend CSS color support with css-color module](https://github.com/jsdom/cssstyle/pull/171/files)

PR #3676, opened 2024-01-29
Trigger selectionchange event on document when selection changes (small change - 2 files)

piotr-oles:
I want to test my contenteditable component that relies on `selectionchange` events. When debugging why my tests doesn't reflect browser behaviour, I found this TODO:
https://github.com/jsdom/jsdom/blob/main/lib/jsdom/living/selection/Selection-impl.js#L335

This PR addresses that TODO by triggering `selectionchange` event on `document` as stated in the spec. I added missing tests as well.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLOrSVGElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLOrSVGElement-impl.js
@@ -67,6 +67,7 @@ class HTMLOrSVGElementImpl {
     ownerDocument._lastFocusedElement = this;
     focusing.fireFocusEventWithTargetAdjustment("focus", this, previous);
     focusing.fireFocusEventWithTargetAdjustment("focusin", this, previous, { bubbles: true });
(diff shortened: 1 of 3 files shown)


piotr-oles:
There is a timeout on node-canvas - I'm not sure if it's just flaky test

piotr-oles:
@domenic could you take a look? :)

domenic:
Based on https://jsbin.com/qiyexuxoro/edit?html,output, the test does not seem to be valid.

domenic:
One guess is that the spec says to schedule the event by queuing a task, whereas this fires it immediately. In jsdom we often emulate this sort of task queuing by doing `window.setTimeout(,0)`.

piotr-oles:
Ah, right! I added wait and it passes: https://jsbin.com/kogutovadu/1/edit?html,output
I will update the PR :)

piotr-oles:
@domenic I updated the PR :)

domenic (approved):

Issue #3680, opened 2024-02-09
Implement RegExp v patterns for input validation


stnguyen90:
### Basic info:

- **Node.js version:** v20.8.0
- **jsdom version:** 24.0.0

### Minimal reproduction case

```js
const jsdom = require('jsdom');
const { JSDOM } = jsdom;

const pattern = String.raw`^(?!-)[[A-Za-z0-9\-]]+([[\-\.]]{1}[[a-z0-9]]+)*\.[[A-Za-z]]{2,18}$`;
const value = 'not-a-domain';
const dom = new JSDOM(`<input id="domain" pattern="${pattern}" required value="${value}">`);

const input = dom.window.document.querySelector("#domain");

console.log('pattern:', input.pattern);
console.log('value:', input.value);
input.checkValidity();
console.log('validity.valid:', input.validity.valid); // returns true
console.log('validity.patternMismatch:', input.validity.patternMismatch); // returns false
```

### How does similar code behave in browsers?

In the browser, `input.validity.valid` is `false` and  `input.validity.patternMismatch` is `true`. You can test this by pasting this code:

```
<input id="domain" pattern="^(?!-)[[A-Za-z0-9\-]]+([[\-\.]]{1}[[a-z0-9]]+)*\.[[A-Za-z]]{2,18}$" required value="not-a-domain">

<input id="domain" pattern="^(?!-)[[A-Za-z0-9\-]]+([[\-\.]]{1}[[a-z0-9]]+)*\.[[A-Za-z]]{2,18}$" required value="example.com">
```

into [MDN's pattern docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern):

![image](https://github.com/jsdom/jsdom/assets/1477010/86154922-4231-4c5c-930e-80afcd1d3458)

In jsdom's source code, I noticed the `u` flag being used for the RegExp:

https://github.com/kontomondo/jsdom/blob/c8f6f2504a71aa904f76b5bd4359505fc4820860/lib/jsdom/living/nodes/HTMLInputElement-impl.js#L806

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern#overview):

![image](https://github.com/jsdom/jsdom/assets/1477010/3f984001-1039-4f38-8ba1-6426e736a35a)

Would it be okay to update the source accordingly?

PR #3681, opened 2024-02-09
Fix the RegExp flag used for input pattern validation

stnguyen90:
According to MDN (https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern#overview), the RegExp flag was changed from 'u' to 'v' around mid-2023, so jsdom should update to match that.

Fixes: https://github.com/jsdom/jsdom/issues/3680

### DIFF
--- a/test/web-platform-tests/to-run.yaml
+++ b/test/web-platform-tests/to-run.yaml
@@ -1107,7 +1107,6 @@ formaction.html: [fail, Unknown]
 
 DIR: html/semantics/forms/constraints
 
-form-validation-validity-patternMismatch.html: [fail, Unknown]
 form-validation-validity-valueMissing.html: [fail, 'Spec unclear (see https://github.com/whatwg/html/issues/5202)']
 form-validation-willValidate.html:
   "[INPUT in COLOR status] Must be barred from the constraint validation if it is readonly": [fail, Not implemented]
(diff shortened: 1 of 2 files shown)


stnguyen90:
I think it'd be good to add a test-case too, but I'm not sure what the best approach is. Should I create a new test file or copy https://github.com/web-platform-tests/wpt/blob/master/html/semantics/forms/constraints/form-validation-validity-patternMismatch.html into `test/web-platform-tests/to-upstream/html/semantics/forms/constraints`?

stnguyen90:
> I think it'd be good to add a test-case too, but I'm not sure what the best approach is. Should I create a new test file or copy https://github.com/web-platform-tests/wpt/blob/master/html/semantics/forms/constraints/form-validation-validity-patternMismatch.html into `test/web-platform-tests/to-upstream/html/semantics/forms/constraints`?

Never mind, I guess the test was skipped in jsdom, so I updated `test/web-platform-tests/to-run.yaml` to not expect failure for some of the related tests.

stnguyen90:
Looks like node 18 fails because it doesn't support the v flag:

image

test/web-platform-tests/to-run.yaml
```diff
@@ -1107,7 +1107,6 @@ formaction.html: [fail, Unknown]
 
 DIR: html/semantics/forms/constraints
 
-form-validation-validity-patternMismatch.html: [fail, Unknown]
```
domenic, line 1110: Use `fail-node18` here (and below) to signal the right support matrix, and the tests should pass, I think!

domenic (commented):
Nice work!!

test/web-platform-tests/to-run.yaml
stnguyen90: Thanks for the tip! I added it, but there were more tests that fail. Essentially, any test where the input has a pattern will fail. Should I add more `fail-node18` to those tests?

domenic: Oh, I see. Well, we don't want to make jsdom (or the pattern parts of it) completely unusable on Node 18, at least until it goes out of LTS support. So I guess we need a fallback, that try/catches with "v" and then falls back to "u".

PR #3694, opened 2024-03-19
Implement the getSetCookie method of Headers

ushiboy:
I've implemented the `getSetCookie` method of `Headers`. The `HeaderList` data structure has been changed accordingly.

See here for the specification of the `getSetCookie` method.
https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie

With this fix, the `header-setcookie.any.html` test is resolved except for a failure due to `Response` not being implemented.

before
```
  1) web-platform-tests
       fetch/api/headers
         header-setcookie.any.html:
     18/24 errors in test:

Failed in "Headers iterator does not combine set-cookie headers":
assert_equals: Array length is not equal expected 2 but got 1
...
```

after
```
  1) web-platform-tests
       fetch/api/headers
         header-setcookie.any.html:
     1/24 errors in test:

Failed in "Set-Cookie is a forbidden response header"
Response is not defined
```

### DIFF
--- a/lib/jsdom/living/fetch/Headers-impl.js
+++ b/lib/jsdom/living/fetch/Headers-impl.js
@@ -56,9 +56,17 @@ class HeadersImpl {
     return this.headersList.contains(name);
   }
 
+  getSetCookie() {
+    return this.headersList.get("Set-Cookie") || [];
+  }
+
   get(name) {
     assertName(name);
-    return this.headersList.get(name);
+    const r = this.headersList.get(name);
+    if (!r) {
+      return null;
+    }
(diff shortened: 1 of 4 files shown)


ushiboy:
Why are CI tests for latest failing?

The `overwritten-customElements-global.html` is set up to expect the test to fail, but since node.js v21.7, the test seems to succeed.

This can be resolved by setting `to-run.yaml` to `overwritten-customElements-global.html: [flaky, Unknown]`, but I cannot determine if that should be included in this PR.

Stadly:
Would be great to have this merged!

PR #3701, opened 2024-04-09
downgrade nwsapi to 2.2.4 to solve open issues

brantphoto:
currently having to pin nwsapi to 2.2.4 due to open issues

https://github.com/dperini/nwsapi/issues/95

getting these same symtoms `SyntaxError: missing ) after argument list` when using https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/Popover.tsx

Seems stalled on nwsapi side... consider downgrading?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "http-proxy-agent": "^7.0.0",
         "https-proxy-agent": "^7.0.2",
         "is-potential-custom-element-name": "^1.0.1",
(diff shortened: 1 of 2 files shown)


domenic:
Looking at web platform test rates, nwsapi 2.2.10 seems better than our previous 2.2.7, which was better than 2.2.4.

If you could produce a specific selector that showed a problem, we could reconsider, but for now I will keep nwsapi upgraded.

PR #3705, opened 2024-04-20
Add: Blob.text() method

ArpanSolanki29:
## Description:
Blob: text() implementation

## Changes:

Blob.webidl: interface definition for text
Blob-Impl.js: Text() implementation that returns a Promise that resolves with a string containing the content of the blob in UTF-8 format.


ArpanSolanki29:
Unit Test added for text() method. Please have a look.

domenic:
That is not an appropriate test. See CONTRIBUTING.md. Also, CI is not passing.

PR #3719, opened 2024-05-20
Default type to XPathResult.ANY_TYPE in XPathExpression.evaluate

### LINKED ISSUE #3422: XPathExpression's evaluate method fails if (optional?) type parameter is omitted
### Basic info:

In browsers, XPathExpression's `evaluate` method takes an optional `type` parameter: https://developer.mozilla.org/en-US/docs/Web/API/XPathExpression/evaluate

However, JSDOM's XPathExpression's `evaluate` method fails if the type parameter is ommited, with a "You must provide an XPath result type (0=any)." error: https://github.com/jsdom/jsdom/blob/b1c0072e306e4e9ad83a5e1ddf0e356ba044bd25/lib/jsdom/level3/xpath.js#L1737

Note that the `type` parameter is marked as optional in @types/jsdom, though.

JSDOM's XPathExpression's `evaluate` method use a default type value of "0" if no value provided?

- **Node.js version:** v16.16.0
- **jsdom version:** 20.0.0

### Minimal reproduction case

```js

const window = new JSDOM().window;

const parsedXPath = window.document.createExpression('//*');

parsedXPath.evaluate(window.document);
```

https://jsbin.com/qobeyewoku/edit?js,output

(Link to a jsbin or similar strongly suggested.)

deej-io:
I got a DOMException when using the HTMX library inside JSDOM, as it does not provide an argument for the `type` parameter in XPathExpression.evaluate. This PR allows the `type` parameter to be optional and defaults the value to XPathResult.ANY_TYPE (0).

The whatwg living DOM spec states that the type argument should default to 0 (any): https://dom.spec.whatwg.org/#interface-xpathexpression.

Whilst the IDL in the level 3 XPath spec doesn't state explicitly that the type parameter is optional, it does use the phrase "If a specific type is specified" in the parameter description, which implies optionality:
(https://www.w3.org/TR/DOM-Level-3-XPath/xpath.html#XPathExpression). Therefore, I believe this patch is spec-compliant on both counts.

Fixes #3422

### DIFF
--- a/lib/jsdom/level3/xpath.js
+++ b/lib/jsdom/level3/xpath.js
@@ -1718,6 +1718,9 @@ module.exports = core => {
             core.DOMException.WRONG_DOCUMENT_ERR,
             'The document must be the same as the context node\'s document.');
       }
+      if (type === undefined) {
+        type = XPathResult.ANY_TYPE;
+      }
       var evaluator = new Evaluator(doc);
       var value = evaluator.val(this._ast, {nodes: [contextNode]});
       if (XPathResult.NUMBER_TYPE === type)
@@ -1735,7 +1738,7 @@ module.exports = core => {
                XPathResult.FIRST_ORDERED_NODE_TYPE !== type)
         throw new core.DOMException(
             core.DOMException.NOT_SUPPORTED_ERR,
-            'You must provide an XPath result type (0=any).');
+            'You must provide a valid XPath result type (0=any).');
       else if (XPathResult.ANY_TYPE !== type &&
                'object' !== typeof value)
         throw new XPathException(


domenic:
Can you add a test for this, or find an existing one, per the instructions in CONTRIBUTING.md?

stekyne:
@deej-io @domenic any chance this can be merged? Facing this issue. Is there a workaround?

deej-io:
Sorry, I've not had a chance to take another look at this yet. Will see if I can find some time today to get a test case sorted.

stroiman:
Just ran into this issue, using JSDOM to test an HTMX application. 

Until merged, I have the following workaround:

```javascript
const fixXPathEvaluator = (browser) => {
  const { XPathExpression, XPathResult } = browser.window;
  const evaluate = XPathExpression.prototype.evaluate;
  XPathExpression.prototype.evaluate = function (contextNode, type, result) {
    if (type === undefined) {
      type = XPathResult.ANY_TYPE;
    }
    return evaluate.call(this, contextNode, type, result);
  };
  // YOU PROBABLY NEED TO ADD THIS TOO
  // browser.window.Document.prototype.evaluate = XPathExpression.prototype.evaluate
};
```

From browsing the jsdom code, it appears that each jsdom instance has it's own `XPathEvaluator` class, so you can't fix it in a global setup, but need to call this for every instance of JSDOM you create. In my case (where I use it as a headless browser for a running web server)

```javascript
    const browser = await JSDOM.fromURL(url, { runScripts: "dangerously", resources: "usable" });
    fixXPathEvaluator(browser);
```

I didn't replace `Document.prototype.evaluate` as I don't have any issues with that, but I guess it could be necessary.

PR #3734, opened 2024-06-18
Update ws to 8.17.1 to mitigate CVE-2024-37890

Plsr:
We were notified by dependabot this morning about [CVE-2024-37890](https://github.com/advisories/GHSA-3h5v-q93c-6h6q) that was introduced to our project by a peer-dependency on `ws` by `jsdom`.

This issue was fixed in version `8.17.1` of `ws`, so this PR introduces this update.

This release of `ws` addresses only this fix (see [release notes](https://github.com/websockets/ws/releases/tag/8.17.1))

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2278,9 +2278,9 @@
       "dev": true
     },
     "node_modules/ws": {
-      "version": "8.17.0",
-      "resolved": "https://registry.npmjs.org/ws/-/ws-8.17.0.tgz",
-      "integrity": "sha512-uJq6108EgZMAl20KagGkzCKfMEjxmKvZHG7Tlq0Z6nOky7YF7aq4mOx6xK8TJ/i1LeK4Qus7INktacctDgY8Ow==",
+      "version": "8.17.1",
+      "resolved": "https://registry.npmjs.org/ws/-/ws-8.17.1.tgz",
+      "integrity": "sha512-6XQFvXTkbfUOZOKKILFG1PDK2NDQs4azKQl26T0YS5CxqWLgXajbPZ+h4gZekJyRqFU8pvnbAbbs/3TgRPy+GQ==",
       "engines": {
         "node": ">=10.0.0"
       },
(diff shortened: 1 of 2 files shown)


kevinschaul:
You'll want to update the ws version in package.json too, right? https://github.com/jsdom/jsdom/blob/main/package.json#L42

ibakirov:
@kevinschaul @Plsr already did in new pull request: #3735

Plsr:
@kevinschaul good point, updated `package.json` as well

@ibakirov not sure if it's helpful to have two PRs open for this now

ibakirov:
@Plsr ok, i canceled my request

domenic:
Since jsdom uses `^8.17.0`, 8.17.1 is automatically installed by all compliant package managers, so this is not necessary.

Plsr:
@domenic True, I missed that tbh. Though it would be nice to exclude the version with a known vulnerability explicitly anyways.

Thanks for taking the time to look at this and your work on jsdom, have a great day!

ibakirov:
@domenic I would argue and propose `^8.17.1` to be clearly defined.

am:
> Since jsdom uses `^8.17.0`, 8.17.1 is automatically installed by all compliant package managers, so this is not necessary.

Isn't [package-lock.json definition](https://github.com/jsdom/jsdom/blob/main/package-lock.json#L2281-L2283) preventing that from happening?

jhoermann:
With [npm update](https://docs.npmjs.com/cli/v10/commands/npm-update) it'll automatically update `ws` in your package lock file as mentioned in https://github.com/jsdom/jsdom/pull/3734#issuecomment-2179651959.

bendeutz:
I am a bit confused why now both PRs are closed (and not merged)? Sorry if i missed something.

MarkBerube:
@bendeutz other PR was closed due to dup. PR. This one was closed because the dep definition in `package.json` includes `8.17.1` in the `package.json` already technically: 
```
"ws": "^8.17.0",
``` 

The problem I still face though is that the lack of package lock updates makes many dependency/security managers like dependabot scream because `jsdom`'s package lock is still set at `8.17.0`. `npm update` will not fix that, only modding the lock will.

could this please be reopened @domenic ?

domenic:
Nope, if there are tools that don't understand the `^` syntax, we're not going to work around them in jsdom.

PR #3738, opened 2024-06-26
Add AbortSignal.any() static method

ezzatron:
This PR adds the [`AbortSignal.any()` static method](https://developer.mozilla.org/docs/Web/API/AbortSignal/any_static).

All but one WPT test is passing:

```
  web-platform-tests
    dom/abort
      ✔ AbortSignal.any.html
      ✔ [expected 1 failure] abort-signal-any.any.html
      ✔ abort-signal-timeout.html
      ✔ event.any.html
      ✔ reason-constructor.html
```

The one test that I couldn't pass has to do with event firing order with "dependant" abort signals. See steps 5 and 6 in ["signal abort"](https://dom.spec.whatwg.org/#abortsignal-signal-abort).

As far as I understand, passing this final test would require you to somehow be able to:

- Ensure that the "source" signal's `abort` event handlers have finished firing before dispatching the `abort` event of the "dependent" abort signal, AND
- Do so without resorting to dispatching the dependent signal's `abort` event asynchronously.

I don't know if that's even possible without a larger rewrite of abort signals.

### DIFF
--- a/lib/jsdom/living/aborting/AbortSignal-impl.js
+++ b/lib/jsdom/living/aborting/AbortSignal-impl.js
@@ -37,6 +37,35 @@ class AbortSignalImpl extends EventTargetImpl {
     return abortSignal;
   }
 
+  static any(globalObject, signals) {
+    for (const signal of signals) {
+      if (signal.aborted) {
+        return AbortSignalImpl.abort(globalObject, signal.reason);
+      }
+    }
+
+    const anySignal = AbortSignal.createImpl(globalObject, []);
+    const handlers = new Map();
+
+    function stopListening() {
+      for (const signal of signals) {
+        signal.removeEventListener("abort", handlers.get(signal));
+      }
+    }
+
+    for (const signal of signals) {
+      function handler() {
+        stopListening();
+        anySignal._signalAbort(signal.reason);
+      }
+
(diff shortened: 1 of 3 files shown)


domenic (changes requested):
Thanks. We need to get the last test passing though.

The problem is you're not fully following the specification. Try to do a line-by-line translation of the spec at https://dom.spec.whatwg.org/#create-a-dependent-abort-signal . E.g., you should be directly setting the reason in step 2. You need to add the `dependent` concept, as a boolean, so you can correctly implement step 3. You need to add source signals and dependent signals. And you need to use the dependent signals concept to modify the "signal abort" algorithm (`_signalAbort` in our implementation).

Another thing to note is that no part of the spec talks about adding or removing event listeners, whereas you use `removeEventListener` and `addEventListener`. That's not correct.

What you have here is more like a polyfill. In jsdom, we implement the specs directly instead.

ezzatron:
Thanks for the feedback. Unfortunately, I don't think I can take that on right now. I'll close this PR to make way for someone else to add the feature.

jdbevan:
Cross-posting a 2nd attempt here for visibility https://github.com/jsdom/jsdom/pull/3805

PR #3757, opened 2024-08-03
Update wpt

asamuzaK:
Updated WPT.

Changes have been made to testharness.js since the last WPT update of JSDOM.
testharness.js now throws exceptions in some tests.

I have reported the issue to WPT and have also submitted a PR.
* [`testharness.js` throws in some cases · Issue #46940 · web-platform-tests/wpt](https://github.com/web-platform-tests/wpt/issues/46940)
* [Add a check if `substitutions` exists by asamuzaK · Pull Request #46945 · web-platform-tests/wpt](https://github.com/web-platform-tests/wpt/pull/46945)

It has already been over a month since I submitted the PR but it still hasn't been merged.
So, as a temporary workaround, patched testharness.js is used.
I will revert this change once [#46945](https://github.com/web-platform-tests/wpt/pull/46945) is merged.


asamuzaK:
I briefly tested the following:

* Comment out patched testharness.js and use native testharnes.js
  -> fails
* Use previous testharness.js
  -> fails

From the above, it looks like there is an issue with the Nodejs v22?

asamuzaK:
Testing with Nodejs v22.4.1 passed.

domenic:
I found the Node.js regression and filed https://github.com/nodejs/node/issues/54436 with a relatively-minimal example.

We can add a "fail-node22" reason to the to-run.yml file in the meantime. (See some code for fail-node18 in `run-wpts.js`.)

Since this keeps happening, I'd kind of like it if we had some generic way of marking failing Node versions that didn't need to get updated. E.g. `fail-node:>22.4.1 <23` using the `semver` package to parse the range. Maybe too much trouble though, I dunno.

asamuzaK:
I tried that "fail-node22" before, but it threw an exception when used in subtests.

``` console
Exception during run: Error: Reason "fail-node22" is only supported for files, not subtests (expectation "window-properties.https.html", subtest "EventTarget method: addEventListener")
```

`window-properties.https.html:` originally has 22 errors.
How about:
* marking the entire file as `[fail, Incorrectly implemented or not implemented]`, and for a meantime comment-out the 22 subtests (see below)
* after the issue is closed and a new Node.js is released
* when we update wpt next time
* then remove comment-out and checking it again?

I will watch [nodejs/node#54436](https://github.com/nodejs/node/issues/54436) as well.

```
window-properties.https.html: [fail, Incorrectly implemented or not implemented]
  #"Window attribute: name": [fail, Incorrectly implemented as a data property]
  #"Window attribute: onmousewheel": [fail, Not implemented]
  #"Window attribute: onmove": [fail, Unknown]
  #"Window attribute: opener": [fail, Not implemented]
  #"Window attribute: status": [fail, Incorrectly implemented as a data property]
  #"Window method: createImageBitmap": [fail, Not implemented]
  #"Window method: matchMedia": [fail, Not implemented]
  #"Window readonly attribute: applicationCache": [fail, Not implemented]
  #"Window replaceable attribute: devicePixelRatio": [fail, Incorrectly implemented as a data property]
  #"Window replaceable attribute: innerHeight": [fail, Incorrectly implemented as a data property]
  #"Window replaceable attribute: innerWidth": [fail, Incorrectly implemented as a data property]
  #"Window replaceable attribute: outerHeight": [fail, Incorrectly implemented as a data property]
  #"Window replaceable attribute: outerWidth": [fail, Incorrectly implemented as a data property]
  #"Window replaceable attribute: pageXOffset": [fail, Incorrectly implemented as a data property]
  #"Window replaceable attribute: pageYOffset": [fail, Incorrectly implemented as a data property]
  #"Window replaceable attribute: screenLeft": [fail, Incorrectly implemented as a data property]
  #"Window replaceable attribute: screenTop": [fail, Incorrectly implemented as a data property]
  #"Window replaceable attribute: screenX": [fail, Incorrectly implemented as a data property]
  #"Window replaceable attribute: screenY": [fail, Incorrectly implemented as a data property]
  #"Window replaceable attribute: scrollX": [fail, Incorrectly implemented as a data property]
  #"Window replaceable attribute: scrollY": [fail, Incorrectly implemented as a data property]
  #"constructor": [fail, Unknown]
window-prototype-chain.html:
  #"Window.prototype": [fail, Unknown]
  #"Global scope polluter": [fail, Depends on the named properties object]
  #"EventTarget.prototype": [fail, Depends on the named properties object]
  #"Object.prototype": [fail, Depends on the named properties object]
window-reuse-in-nested-browsing-contexts.tentative.html:
  #"synchronously navigate iframe with no initial src.": [fail, Unknown]
  #"after the first iframe load event, navigate iframe with no initial src.": [fail, Unknown]
  #'synchronously navigate iframe with initial src == "".': [fail, Unknown]
  #'synchronously navigate iframe with initial src == "about:blank".': [fail, Unknown]
```

domenic:
Right now `fail-node22` doesn't exist at all. So we need to implement it. And maybe we can implement it so that it works for subtests as well.

If that's too much work, then we should implement it just for top-level tests. Then your plan is reasonable but we should use `fail-node22` instead of just `fail`.

asamuzaK:
> Right now `fail-node22` doesn't exist at all. So we need to implement it. And maybe we can implement it so that it works for subtests as well.

"constructor" subtest fails on Node 18-20, but incorrectly(?) passes on Node 22.
Then maybe we also need a reason like "pass-node22" or "fail-node18_20"?

domenic:
Oh, yeah, that is annoying. I see what you mean now. Your original solution of using `fail` plus comments seems reasonable to me. It would be best as a separate PR against main so that it is a single commit we can more easily remember and revert in the future.

asamuzaK:
I'll file another PR for that.

asamuzaK:
Filed #3761

domenic (commented):
Can you use wildcards for entire directories that fail? E.g. moveBefore, pagereveal, pageswap, styleable-select, shadow-dom/tentative/, ...

asamuzaK:
Regarding the timeout error on GitHub Actions Attempt #1.

"Element-matches.html" takes time, and I sometimes experience timeouts even in my local testing.

Is it okay to extend the timeout value to 90_000 of [run-single-wpt.js](https://github.com/jsdom/jsdom/blob/main/test/web-platform-tests/run-single-wpt.js#L24)?
I would like to take this opportunity to check other tests marked as `Flaky`.
Some tests may no longer timeout.
On the other hand, I think there are some that are marked as `Flaky` but actually always failing.

domenic:
It is a bit suspicious that it started failing twice in a row on Node latest but not on other versions. I wonder if that vm change also caused a general slowdown for jsdom, or there was some other perf regression in Node. (Or maybe it's just a coincidence.)

It is also suspicious that Element-matches.html is the one that times out and not any of the other tests that also test a ton of different selectors.

---

You can try extending the timeout in run-single-wpt.js. However I suspect the WPT timeout will kick in anyway.

There are several possible fixes we could make to the upstream file to improve the timeout situation:

- We could add [`<meta name=timeout content=long>`](https://web-platform-tests.org/writing-tests/testharness-api.html#harness-timeout) which I believe will extend from 10 seconds to 60 seconds.
- We could break up the file into multiple tests using [variants](https://web-platform-tests.org/writing-tests/testharness.html#specifying-test-variants) and in particular [/common/subset-tests.js](https://github.com/web-platform-tests/wpt/blob/master/common/subset-tests.js); see [this example](https://github.com/web-platform-tests/wpt/blob/master/encoding/legacy-mb-japanese/euc-jp/eucjp-decode.html).

It feels a bit lame to make such changes when only jsdom is slow enough to need them, but probably they would be accepted anyway. I would be happy to approve and merge adding the `<meta>` by myself without any more reviewers.

---

I am happy for you to audit flaky tests, and we could use flaky to get around this in order to merge faster.

asamuzaK:
@domenic 
PR ready and all green.
But maybe we should wait for #3763 to land?

domenic (approved):
Great improvements, thanks. I will work on doing a bunch of merges and maybe releases this weekend.

Issue #3688, opened 2024-03-01
Browsers return DOM property names in `Object.getOwnPropertyNames(window)`. JSDOM does not. 


DanKaplanSES:
### Basic info:

- **Node.js version:** 20.10.0
- **jsdom version:** 24.0.0 (I saw no difference in v22)

### Minimal reproduction case

```js
import { JSDOM } from 'jsdom';

const { window } = new JSDOM('', {
    url: 'http://localhost',
    runScripts: 'dangerously',
    resources: 'usable',
    pretendToBeVisual: true,
})

console.log(`Object.getOwnPropertyNames(window).filter((k) => k === "HTMLElement").length`, 
    Object.getOwnPropertyNames(window).filter((k) => k === "HTMLElement").length); // 0
```

### How does similar code behave in browsers?
```js
console.log(`Object.getOwnPropertyNames(window).filter((k) => k === "HTMLElement").length`, 
    Object.getOwnPropertyNames(window).filter((k) => k === "HTMLElement").length); // 1
```

### Additional information

[I created this StackOverflow question](https://stackoverflow.com/questions/78085257/why-doesnt-jsdoms-object-getownpropertynameswindow-include-dom-objects) before creating this issue. It might provide some helpful context, and it includes additional log statements showing similarities between the two environments. For example, `window.HTMLElement` have equivalent property descriptors.

I don't fully understand the ramifications of these differences, but it made it challenging to troubleshoot the JSDOM error I was experiencing at the bottom of that SO example.

DanKaplanSES:
I've yet to understand why, but the `runScripts: 'dangerously',` is causing this problem. When I comment that out, the first `console.log` prints `1`, just like a browser. Is this expected behavior?

marrus-sh:
We encountered this issue as well, also impacting `Reflect.ownKeys` and `Object.getOwnPropertyDescriptors`. It seems that whatever is proxying `window` is returning correct values for `getOwnPropertyDescriptor`, but not actually listing all of the properties in `ownKeys`.

legendecas:
This should have been fixed by https://github.com/nodejs/node/pull/54522.

PR #3763, opened 2024-08-20
Setup Window prototype in vm context

legendecas:
Refs: https://github.com/nodejs/node/issues/54436#issuecomment-2296539259

At the moment, the global object prototype in the vm context is as follows:

```
Outer Context    |    Vm Context
Object           |    Object
  ^              |      ^
  | prototype    |      | prototype
EventTarget      |    Global
  ^              |      ^
  | prototype    |      | prototype
Window           |      |
  ^              |      |
  | prototype    |      |
window (sandbox) <--- globalThis
```

This PR corrects the prototype in respect to the WebIDL definition as:

```
Outer Context    |    Vm Context
Object           |    Object
  ^              |   
  | prototype    |
EventTarget      |
  ^              |
  | prototype    |
Window (and WindowProperties)
  ^              |      ^
  | prototype    |      | prototype
window (sandbox) <--- globalThis
```


domenic:
Well, this does look promising! We can try doing a release and seeing if anyone's code breaks, but our test suite is pretty exhaustive, so I'm hopeful.

lib/jsdom/browser/Window.js
```diff
@@ -148,6 +156,10 @@ function setupWindow(windowInstance, { runScripts }) {
 
   windowConstructor.prototype = windowPrototype;
   Object.setPrototypeOf(windowInstance, windowPrototype);
+  if (runScripts === "outside-only" || runScripts === "dangerously") {
+    const global = vm.runInContext("this", windowInstance);
+    Object.setPrototypeOf(global, windowPrototype);
```
domenic, line 161: Note to self: we could try also setting `EventTarget.prototype.__proto__` to the `Object.prototype` from the vm context. Might be best to do with a major version bump...

legendecas:
The test fails because html/browsers/the-window-object.constructor expect `window.constructor` is not an own property. But in Node.js versions before v22.5.0, all prototype properties are intercepted as own properties of the global object. This is not fixable in userland unless with some strange polyfills.

Is there a way to specify a test only passes on certion versions?

lib/jsdom/browser/Window.js
legendecas: Yes, I assume this could be a big change so I'd leave it from this PR...

asamuzaK:
> Is there a way to specify a test only passes on certion versions?

Maybe some kind of a reason like [this](https://github.com/jsdom/jsdom/pull/3757#issuecomment-2295545330) in [run-wpts.js](https://github.com/jsdom/jsdom/blob/main/test/web-platform-tests/run-wpts.js#L25)?

legendecas:
> > Is there a way to specify a test only passes on certion versions?
> 
> Maybe some kind of a reason like [this](https://github.com/jsdom/jsdom/pull/3757#issuecomment-2295545330) in [run-wpts.js](https://github.com/jsdom/jsdom/blob/main/test/web-platform-tests/run-wpts.js#L25) and [utils.js](https://github.com/jsdom/jsdom/blob/main/test/web-platform-tests/utils.js#L12)?

Yes, in v22.5.0 the `html/browsers/the-window-object.constructor` passes because `Object.hasOwn` behaves correctly on the global object.

asamuzaK:
Regarding https://github.com/jsdom/jsdom/issues/3688

The test case below fails `test2` and `test3`:

``` html
<!DOCTYPE html>
<meta charset="utf-8">
<title>Access window properties</title>
<link rel="help" href="https://github.com/jsdom/jsdom/issues/3688">

<script>
"use strict";

test(() => {
  assert_true("HTMLElement" in window, "for-in");
}, "test1: window object should have HTMLElement property in for-in loop");

test(() => {
  assert_true(Object.keys(window).includes("HTMLElement"), "Object.keys()");
}, "test2: window object should have HTMLElement property in Object.keys()");

test(() => {
  assert_equals(Object.getOwnPropertyNames(window).includes("HTMLElement"), "Object.getOwnPropertyNames()");
}, "test3: window object should have HTMLElement property in Object.getOwnPropertyNames()");
</script>
```

legendecas:
`HTMLElement` should be an unenumerable property on `window`, so test2 should expect false?

For test3, it fails on either Node.js versions (v20.17.0, v22.4.1, v22.5.0). So I think it is not relevant to this patch?

asamuzaK:
I see.
Sorry for the noise.

legendecas:
No worries, I will try to address it in a follow-up and see if it will break more tests...

domenic (approved):

PR #3765, opened 2024-08-24
Add a proper WindowProperties object

domenic:
The original "named properties tracker" was created before Proxy was available. Now we can implement this in a way much closer to the spec.

As part of this, restructure Window.js to be less convoluted, with a single createWindow function and no fake Window constructor that operates on a this object but then manually sets is prototype chain anyway.


domenic:
Oh, it seems a lot of tests now fail on Node <22 because I guess putting a Proxy in the prototype chain doesn't work as well there... that's discouraging. Hmm.

ExE-Boss:
> Oh, it seems a lot of tests now fail on Node <22 because I guess putting a Proxy in the prototype chain doesn't work as well there... that's discouraging. Hmm.

Yeah, that’s <https://github.com/nodejs/node/issues/30985>, which blocked my last attempts in <https://github.com/jsdom/webidl2js/pull/167> and <https://github.com/jsdom/jsdom/pull/2835>.

domenic:
Pursuing https://github.com/jsdom/jsdom/pull/3779 instead.

PR #3770, opened 2024-08-25
Change various Window data properties to accessors

domenic:
All of them have setters (most via [Replaceable] behavior), so this should not cause any compat issues.


domenic:
https://github.com/nodejs/node/issues/31808 is causing the XHR test to fail, since it defines a function named `status`.

We should probably not actually merge this until that is fixed, and propagated to all the versions we support. Otherwise it's unclear how many peoples' code we would be breaking.

domenic:
Pursuing https://github.com/jsdom/jsdom/pull/3779 instead.

PR #3773, opened 2024-08-25
Update heap size

asamuzaK:


asamuzaK:
I'll close this for now.
If heap error continues to occur, I think you should reconsider.

PR #3779, opened 2024-09-22
Use DONT_CONTEXTIFY to make all Window properties problems go away

domenic:
This new option avoids several of the Node.js bugs that have bitten us so far in other attempts.

Since this is Node.js v22+ only, we can't really use this until ~April 2026 when Node.js v20 falls out of LTS. So I'm mostly throwing up this PR to remind us of the possibility.

This includes #3765 and #3770.


domenic:
The remaining failures seem to largely come from not implementing WindowProxy. We could probably actually implement a WindowProxy, since for same-origin cases it's basically just about the indexed property handling. Currently that's done in a hacky way via HTMLFrameElement-impl.js.

asamuzaK:
FYI: `vm.constants.DONT_CONTEXTIFY` has been backported to Node v20.
https://nodejs.org/docs/latest-v20.x/api/vm.html#vmconstantsdont_contextify

We can move this PR forward?

domenic:
Yeah, although I'd need to actually implement a WindowProxy I think.

domenic:
Actually I'm not sure this approach is compatible with implementing a WindowProxy.

We could create a WindowProxy wrapping the `window` object createed in `Window.js`. But I don't know how we would install it as the global `this` value in the vm context.

Issue #3785, opened 2024-10-22
rrweb-cssom 0.8.0 now supports `@layers`


Exac:
Hi, I noticed that https://github.com/rrweb-io/CSSOM/pull/5 was just merged, so we could upgrade to `rrweb-cssom@0.8.0`.

Examples of issues that this could help resolve:
- https://github.com/jsdom/jsdom/issues/2177#issuecomment-1724971596
- https://github.com/primefaces/primeng/issues/14085
- https://github.com/thymikee/jest-preset-angular/issues/2194

### Basic info:

- **Node.js version:** v20.17.0
- **jsdom version:** v22.0.0

### Minimal reproduction case

https://stackblitz.com/edit/jsdom-rrweb-cssom-0-7-1-errors?file=index.js

```js
const jsdom = require('jsdom');
const { JSDOM } = jsdom;

const dom = new JSDOM(`
<!DOCTYPE html>
<html>
<head>
  <style>
    @layer foo, bar;

    @layer foo {
      p {
        color: yellow;
      }
    }
    
    @layer bar {
      p {
        background-color: black;
      }
    }

    p {
      font-size: 4rem;
    }
  </style>
</head>
<body>
  <p>hello, jsdom</p>
</body>
</html>
`);

const pEl = dom.window.document.querySelector('p');
const { color, backgroundColor, fontSize } = dom.window.getComputedStyle(pEl);

console.log({ color, backgroundColor, fontSize });
```
```text
Error: Could not parse CSS stylesheet
    at exports.createStylesheet (/home/projects/jsdom-rrweb-cssom-0-7-1-errors/node_modules/jsdom/lib/jsdom/living/helpers/stylesheets.js:37:21)
    at HTMLStyleElementImpl._updateAStyleBlock (/home/projects/jsdom-rrweb-cssom-0-7-1-errors/node_modules/jsdom/lib/jsdom/living/nodes/HTMLStyleElement-impl.js:68:5)
    at HTMLStyleElementImpl._poppedOffStackOfOpenElements (/home/projects/jsdom-rrweb-cssom-0-7-1-errors/node_modules/jsdom/lib/jsdom/living/nodes/HTMLStyleElement-impl.js:42:10)
    at JSDOMParse5Adapter.onItemPop (/home/projects/jsdom-rrweb-cssom-0-7-1-errors/node_modules/jsdom/lib/jsdom/browser/parser/html.js:175:43)
[...]

PR #3786, opened 2024-10-22
feat: add support for @layers via cssom

Exac:
- cssom now supports @layers
- this change is particularly useful to users of primeng and jest, who are getting errors in their unit tests

Closes https://github.com/jsdom/jsdom/issues/3785

Edit: In order to run the tests it seems I have to edit my hosts file, so I'll let them run via github actions instead.

```console
$ ./test/web-platform-tests/tests/wpt make-hosts-file | sudo tee -a /etc/hosts
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "is-potential-custom-element-name": "^1.0.1",
         "nwsapi": "^2.2.12",
         "parse5": "^7.1.2",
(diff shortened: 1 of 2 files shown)


Exac:
This seems to have passed the `nesting-layer.html` test 🔥

Exac:
The tests pass, but I'm not sure that the `nesting-layer.html` ran.

PR #3790, opened 2024-11-06
fix(form-controls): faces are labelable

hesxenon:
Form associated custom elements are classified as ["labelable elements"](https://html.spec.whatwg.org/#category-label) - jsdom currently does not consider them as such.


hesxenon:
@domenic I'm trying to understand the tests a bit but having problems. As far as I can tell:
1. there are no custom elements tests in jsdom at all (???)
2. the wpt tests already include a test for this that should have failed

How do wpt tests and jsdom relate?

domenic:
https://github.com/jsdom/jsdom/blob/main/Contributing.md should answer your questions.

hesxenon:
@domenic it did, thank you. However, I found the relevant part to be quite confusing, so in my last commit I changed the wording. _IF_ that wording holds true I suggest also changing `to-run.yaml` to `to-skip.yaml`, since it appears that that is how it acts.

I also added a [todo comment](https://github.com/jsdom/jsdom/pull/3790/files#diff-ad84fb016f66a73bc3cee0dda820c88681c0fe01bdad52036d6fb5c30f9a452bR3) because I don't know what's appropriate to use here. I think I am beginning to understand the difference between those in `generated/` and the actual impls but I don't get when to use what. Would appreciate some answers :)

domenic:
Thanks for your patience. I don't have very much jsdom time these days, and this weekend is no exception, but I suspect with the holidays coming up there will be more time soon. Rest assured that this is in the queue.

lib/jsdom/living/custom-elements/ElementInternals-impl.js
```diff
@@ -1,5 +1,9 @@
 "use strict";
 
+// TODO: when to use generated and when to use impl directly?
```
domenic, line 3: Generally, `require()` the generated files, as they expose the full utility APIs for each Web IDL interface.

lib/jsdom/living/custom-elements/ElementInternals-impl.js
```diff
@@ -15,6 +19,30 @@ class ElementInternalsImpl {
     return shadow;
   }
 
+  get labels() {
+    if (!this._targetElement._ceDefinition.formAssociated) {
+      throw DOMException.create(this._targetElement._globalObject, [
+        "ElementInternals.labels is only supported for form-associated custom elements",
```
domenic, line 25: Nit: don't use `InterfaceName.property` for non-static properties. Prefer something like `ElementInternal's labels property`.

lib/jsdom/living/custom-elements/ElementInternals-impl.js
```diff
@@ -15,6 +19,30 @@ class ElementInternalsImpl {
...
+        "NotSupportedError"
+      ]);
+    }
+    return NodeList.create(this._targetElement._globalObject, [], {
```
domenic, line 29: We should share the implementation with `HTMLInputElement-impl.js`, by using `getLabelsForLabelable`.

lib/jsdom/living/custom-elements/ElementInternals-impl.js
```diff
@@ -15,6 +19,30 @@ class ElementInternalsImpl {
...
+      element: this._targetElement,
+      query: () => {
+        // TODO: why doesn't .id work? What about attrs that don't reflect idl -> content?
```
domenic, line 32: FYI, it doesn't work because the impl class doesn't have an `id` property, and `_targetElement` is an impl instance. The `id` getter is only part of the generated class, since it's auto-generated to reflect the attribute.

lib/jsdom/living/helpers/form-controls.js
```diff
@@ -132,6 +132,9 @@ exports.isLabelable = node => {
     case "input":
       return node.type !== "hidden";
   }
+  if (node._ceDefinition?.formAssociated) {
+    return true; // as per https://html.spec.whatwg.org/#category-label
```
domenic, line 136: Nit: this comment is redundant with the one before the function declaration, so we can omit it.

test/web-platform-tests/to-run.yaml
```diff
@@ -392,7 +392,23 @@ custom-element-reaction-queue.html:
   "Mutating a undefined custom element while upgrading a custom element must not enqueue or invoke reactions on the mutated element": [fail, document.write() implementation is not spec compliant]
 element-internals-aria-element-reflection.html:
   "Getting previously-unset ARIA element reflection properties on ElementInternals should return null.": [fail, Unknown]
-form-associated/**: [fail-slow, Not implemented]
+form-associated/ElementInternals-NotSupportedError.html: [fail-slow, Not implemented]
+form-associated/ElementInternals-form.html: [fail-slow, Not implemented]
+# form-associated/ElementInternals-labels.html: [fail-slow, Not implemented]
+# form-associated/ElementInternals-reportValidity-bubble-notref.html: [fail-slow, Not implemented] # cannot be found, even though html file exists?
```
domenic, line 398: These are ref tests, which means they try to render the result in a web browser and then compare iwth the match/mismatch file. We don't support these in the jsdom testing harness because we don't render anything.

You can just omit them from this file.

lib/jsdom/living/custom-elements/CustomElementRegistry-impl.js
```diff
@@ -185,6 +185,7 @@ class CustomElementRegistryImpl {
       localName,
       constructor,
       objectReference: ctor,
+      formAssociated: ctor.formAssociated,
```
domenic, line 188: Implement this more closely following the spec in https://html.spec.whatwg.org/#element-definition . Steps 10, and 14.11-14.12. It doesn't make a practical difference right now because formAssociated is the last property gotten, but the order would be observable if the spec added another property getting step in the future.

test/web-platform-tests/to-run.yaml
```diff
@@ -354,13 +354,13 @@ DIR: custom-elements
...
-  'customElements.define must get "observedAttributes" property on the constructor prototype when "attributeChangedCallback" is present': [fail, formAssociated not implemented]
-  'customElements.define must rethrow an exception thrown while getting formAssociated on the constructor prototype': [fail, formAssociated not implemented]
-  'customElements.define must get four additional callbacks on the prototype if formAssociated is converted to true': [fail, formAssociated not implemented]
-  'customElements.define must rethrow an exception thrown while getting additional formAssociated callbacks on the constructor prototype': [fail, formAssociated not implemented]
+#  'customElements.define must get "prototype", "disabledFeatures", and "formAssociated" property of the constructor': [fail, formAssociated not implemented]
+#  'customElements.define must get "observedAttributes" property on the constructor prototype when "attributeChangedCallback" is present': [fail, formAssociated not implemented]
+#  'customElements.define must rethrow an exception thrown while getting formAssociated on the constructor prototype': [fail, formAssociated not fully implemented]
```
domenic, line 359: Delete these now that they're passing :)

domenic (commented):
Nice work; this looks pretty good overall. Thank you for your patience.

lib/jsdom/living/custom-elements/ElementInternals-impl.js
hesxenon: interesting, I remember some similar rationale behind some doc tools distinction between `#member` and `.member`... will do

lib/jsdom/living/custom-elements/ElementInternals-impl.js
hesxenon: I think I understand the difference between impl and generated - but why is `_targetElement` not a generated instance when we should generally only require the generated modules? And how could I have known this?

hesxenon:
@domenic I think I've implemented your feedback, hopefully to your satisfaction. I am however unsure on how I could have known these things without your help... is this a repo that one just has to "grow into"?

test/web-platform-tests/to-run.yaml
```diff
@@ -392,7 +388,19 @@ custom-element-reaction-queue.html:
   "Mutating a undefined custom element while upgrading a custom element must not enqueue or invoke reactions on the mutated element": [fail, document.write() implementation is not spec compliant]
 element-internals-aria-element-reflection.html:
   "Getting previously-unset ARIA element reflection properties on ElementInternals should return null.": [fail, Unknown]
-form-associated/**: [fail-slow, Not implemented]
+form-associated/ElementInternals-NotSupportedError.html: [fail-slow, Not implemented]
```
domenic, line 391: It looks like these now fail fast, instead of slow, so I'll switch these to `fail`. (Which causes them to run, and be expected to fail, instead of be skipped. So that in the future if someone makes them pass, we notice.)

Contributing.md
```diff
@@ -44,7 +44,7 @@ In the following sections, we'll give commands for running a subset of the tests
...
+All tests for web platform features (as opposed to features of jsdom itself, such as the `JSDOM()` constructor) should be in [web-platform-tests](https://github.com/web-platform-tests/wpt) format. We have some infrastructure for running these directly against jsdom documents. So ideally, when contributing a bugfix or new feature, you can browse the web-platform-tests repository and find the test covering your work, and then just exclude it from being skipped by making sure it's not matched in [the `to-run.yaml` file](https://github.com/jsdom/jsdom/blob/main/test/web-platform-tests/to-run.yaml). These tests are HTML files which use a special library called [testharness.js](https://web-platform-tests.org/writing-tests/testharness-api.html) to report their results.
```
domenic, line 47: I think this is not a correct change because the to-run.yaml file is actually about which tests to run, and what results to expect for them. I can see why you might be confused though, since most of the "results to expect" are failures, and the "tests to run" are only a very small portion of the file. (E.g. `DIR: custom-elements`.)

I'll try to reword it a bit to make it clearer.

domenic (approved):
Great work, and again, sorry for the delay! I got sick over the holidays :(. I'll push a few small changes and then merge this and release it.

> I am however unsure on how I could have known these things without your help... is this a repo that one just has to "grow into"?

Yeah, I think that's generally common with all projects over a certain size. Ideas welcome...

lib/jsdom/living/custom-elements/ElementInternals-impl.js
domenic: For me it's just that `ElementInternals.labels === undefined`, because there is no static member named `labels` on the `ElementInternals` class.

lib/jsdom/living/custom-elements/ElementInternals-impl.js
domenic: The bindings layer (in the webidl2js project) tries to ensure that the implementation code only has to deal with impls, so it tries to unwrap everything it gives to you.

However, the abstraction is leaky. The most notable leak is the one you've noted, that when you create new objects, sometimes we create generated versions.

I think we can fix this to create impl versions more often, since the generated bindings code will convert from impl -> wrapper for you: see, e.g., `generated/ElementInternals.js` line 117.

However, some leaks will still remain:

- Exceptions need to be created as the generated version, because at least for now ewe don't have a `try`/`catch` block around the call to the impl getter/setter/method that catches any exceptions to convert them from impl -> generated
- We don't do "deep" conversion: see discussion in https://github.com/jsdom/webidl2js?tab=readme-ov-file#methods-implementing-idl-operations .

domenic: > And how could I have known this?

The documentation I linked above might help a bit. But honestly, this impl vs. wrapper (a.k.a. generated) distinction is a pretty strong reason for adopting TypeScript. That'd be a big project, though :)

Contributing.md
hesxenon: ah, now I get your other wording on this topic. So the default (without this file) is to "run all wpt"; this file specifies "run certain wpt tests in a specific way (mostly skip them)", right? Hope you got better btw, missed the holidays for the same reason mostly.

PR #3795, opened 2024-11-25
Allow canvas 3.0.0-rc2

birkskyum:
I'm having install issues using the 3.0.0 rc release with JSDOM because of the strict peer dependencies. As can be seen it's the only release that's been cut in more than a year.

[Canvas Releases](https://github.com/Automattic/node-canvas/releases)

The breaking change in v3.x is the sunset of Node 16, rather than something that makes jsdom incompatible.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,7 +50,7 @@
         "node": ">=18"
       },
       "peerDependencies": {
-        "canvas": "^2.11.2 || 3.0.0-rc2"
+        "canvas": "^2.11.2 || 3.0.0-rc2 || ^3.0.0"
       },
       "peerDependenciesMeta": {
         "canvas": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,7 +50,7 @@
         "node": ">=18"
       },
       "peerDependencies": {
-        "canvas": "^2.11.2"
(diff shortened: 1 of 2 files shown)


domenic:
Thanks, but I'm not willing to officially support multiple versions of the canvas dependency, and especially not one which does not declare itself stable. I look forward to working on 3.0.0 integration whenever they work on releasing it, but as you've noted, they don't seem to be making a lot of progress.

birkskyum:
I fully understand. My frustration came from canvas 2 not being compatible with Node 22. It left me with 

- stay blocked on Node 20 forever
- have canvas 2 be compatible 
- have Node 22 be compatible
- have canvas 3 released and Jsdom updated
- remove all use of canvas
- relax Jsdom peer deps, potentially in a pre release, or patched locally

If it's this annoying it seems like it might be best to test in actual browsers and avoid it altogether.

domenic:
Oh, hmm. Lack of compatibility with Node v22 is pretty serious.

What if we only supported the v3 line? It's not great that it's pre-release, but it seems like the least-bad option.

birkskyum:
The semver check that npm use actually doesn't recognize the RC release, so i believe it's actually a no-op for now

birkskyum:
Canvas 3 is out 🎉

PR #3797, opened 2024-12-02
querySelectorAll Fix `TypeError: list.map is not a function`

sanjarcode:
I am using Vitest, and this error shows up, even when no React/UI code is being rendered.
`TypeError: list.map is not a function`


domenic:
Please create a minimal reproduction following the issue template so we can understand if this is an issue with jsdom vs. Vitest.

PR #3798, opened 2024-12-03
add node 22 to test ci matrix

birkskyum:
Increase test ci matrix to include node 22

Before:
Node 18, 20, 20 ([latest](https://www.npmjs.com/package/node?activeTab=versions))

After:
Node 18, 20, 22

### DIFF
--- a/.github/workflows/jsdom-ci.yml
+++ b/.github/workflows/jsdom-ci.yml
@@ -53,7 +53,7 @@ jobs:
         node-version:
           - 18
           - 20
-          - latest
+          - 22
     steps:
     - uses: actions/checkout@v4
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/jsdom-ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         node-version:
           - 18
           - 20
-          - latest
+          - 22
```
asamuzaK, line 56: It should be
- 18
- 20
- 22
- latest

birkskyum: Added that now

domenic (approved):
Thank you!

PR #3806, opened 2024-12-24
Update optional peer dependency canvas to v3

sebastianwachter:
Canvas v3.0.0 is now out of Release Candidate.

~Should canvas v2.x.x compatibility be removed now that a stable version is available for newer node versions?~ (removed)

### DIFF
--- a/.github/workflows/jsdom-ci.yml
+++ b/.github/workflows/jsdom-ci.yml
@@ -40,10 +40,10 @@ jobs:
         run: sudo apt-get install build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
       - name: Setup HOSTS file for Web Platform Test server
         run: ./test/web-platform-tests/tests/wpt make-hosts-file | sudo tee -a /etc/hosts
-      - name: Install dependencies
-        run: npm ci
+      - name: Install dependencies and canvas
+        run: npm ci && npm i canvas@3
       - name: Run tests
-        run: npm i canvas && npm test -- --retries 1
+        run: npm test -- --retries 1
   build:
     name: Tests
     runs-on: ubuntu-latest
--- a/.github/workflows/jsdom-ci.yml
+++ b/.github/workflows/jsdom-ci.yml
@@ -36,7 +36,7 @@ jobs:
           - node-version: '20'
             canvas-version: '^2.11.2'
           - node-version: '22'
-            canvas-version: '^3.0.0-rc3'
+            canvas-version: '^3.0.0'
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/jsdom-ci.yml
+++ b/.github/workflows/jsdom-ci.yml
@@ -29,29 +29,21 @@ jobs:
     env:
       TEST_SUITE: 'node-canvas'
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - node-version: '20'
-            canvas-version: '^2.11.2'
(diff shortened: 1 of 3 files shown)


birkskyum:
Not release candidate anymore - canvas 3 is out

domenic:
Yes, please remove canvas v2 from the peer dependencies and testing matrix so that we don't have the extra support burden!

sebastianwachter:
I removed canvas v2 from the test matrix (I removed the test matrix) and from the peer dependencies. I set node version for that pipeline step to v22.

.github/workflows/jsdom-ci.yml
```diff
@@ -29,29 +29,21 @@ jobs:
...
       - name: Install required image manipulation packages with APT
         run: sudo apt-get install build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
       - name: Setup HOSTS file for Web Platform Test server
         run: ./test/web-platform-tests/tests/wpt make-hosts-file | sudo tee -a /etc/hosts
       - name: Install dependencies
-        run: npm ci && npm install canvas@${{ matrix.canvas-version }}
+        run: npm ci
       - name: Run tests
-        run: npm test -- --retries 1
+        run: npm i canvas && npm test -- --retries 1
```
alumni, line 46: I know that the test setup is not that great (probably there should be a test project with both canvas and jsdom) but:
1. Installing dependencies should not be done in 2 steps. `canvas` should be installed together with the other dependencies. This will also keep the test output clean.
2. Not specifying a version when installing a package will always install the latest. This will install `canvas@4.0.0` when it will be released.

sebastianwachter: I'm currently on my way to a ski trip and will implement the changes this evening.

I have a question regarding your second point: will `npm i canvas` always install the latest version even when the field `"peerDependencies": { "canvas": "^3.0.0" }` is set? I thought (or hoped) it would install the latest minor or patch version of the v3 major version.

alumni: `npm i canvas` will modify `package.json` and the lockfile and will add the latest version of canvas to the `dependencies` group. I'm not sure if npm will also remove it from `peerDependencies` (might also depend on the npm version), some package managers do this.

_Viel Spaß! :)_

sebastianwachter: Moved the `npm i canvas@3` to the same line as `npm ci` and specified canvas in its installtion with `@3`

birkskyum:
Can we land this, so that jsdom will support node 22?

alumni:
> Can we land this, so that jsdom will support node 22?

There is no change in API between canvas@2.11.2 and canvas@3.0.1, so you can use them interchangeably.

So unless you enable [npm/strict-peer-deps](https://docs.npmjs.com/cli/v11/using-npm/config#strict-peer-deps) or [pnpm/strict-peer-dependencies](https://pnpm.io/npmrc#strict-peer-dependencies), you should be able to provide canvas@3 and jsdom would use it without issues.

sebastianwachter:
The thing causing issues is that canvas v3 is not listed in jsdom's peer dependencies which can cause issues like this one (dependabot wants to update to canvas v3 but can't): https://github.com/sebastianwachter/wc-scratch/actions/runs/12568528588/job/35036163215?pr=212

marklundin:
This would be super helpful. We're currently waiting on an 3.0.1 `canvas` which fixes an issue with Apple Silicon, and currently we can't avoid the strict-peer-deps

domenic (approved):
Thanks!

PR #3809, opened 2025-01-09
Document WPTs and to-run.yaml better

domenic:
@hesxenon, what do you think of this?

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -44,9 +44,15 @@ In the following sections, we'll give commands for running a subset of the tests
 
 ### Web platform feature tests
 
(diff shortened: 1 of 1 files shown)


hesxenon:
I think it's a lot better, however there's one line that still confuses me in combination with the file name:

> To control what tests we run against jsdom, we use [the `to-run.yaml` file](https://github.com/jsdom/jsdom/blob/main/test/web-platform-tests/to-run.yaml)

This isn't correct though afaik? Removing lines altogether _also_ enables the tests? So it seems to me there's some kind of blacklisting mechanism where all tests are run by default _except_ for those that are matched in this file in which case it's not the default but the specified configuration being executed. 

Kinda like a gitignore where you do
```gitignore
*
!lib
```
?

asamuzaK:
> Or, if there's no section in `to-run.yaml` for the directory you're working on, then add that section.

Wouldn't it be better to add `npm run update-wpt` instruction here?
I think the reason why the tests are not included in `to-run.yaml` are:
* It has already been downloaded from WPT, but just not applied.
* Not downloaded yet for some reason, e.g. recently added to WPT.

If it's beyond the scope of this PR, feel free to ignore.

domenic:
> This isn't correct though afaik?

I think it's correct. The file controls what we run.

> Removing lines altogether _also_ enables the tests?

If all lines are removed, no tests will be run. Only the directories with `DIR` lines will be run.

> Wouldn't it be better to add `npm run update-wpt` instruction here?

Good idea, although that requires a lot of work from contributors to add all the new expectations...

asamuzaK:
> Good idea, although that requires a lot of work from contributors to add all the new expectations...

Yeah, it was really a pain...

domenic:
I'll merge this for now, although it sounds like there are indeed some possible improvements.

PR #3641, opened 2023-12-12
Add `customElements.getName()`

mash-graz:
Fixes: #3640

PR #3815, opened 2025-01-11
Update WPT

asamuzaK:
Updated WPT to the latest version.
* Manifest version has been increased to 9
* Added new CSSOM and XPATH directories
* The timeout problem still occurs, so extended timeout value from `90_000` to `120_000`.
  Ref https://github.com/jsdom/jsdom/pull/3757#issuecomment-2298234008

It may include tests related to some PRs, e.g. #3641, #3705, #3710, #3719, #3730


domenic (commented):
Thanks as always for taking on this hard task.

Some initial notes:

- Remember to use blobs to blocklist directories or large sets of tests (e.g. `derive_bits_keys/**`).
- I think we should exclude all shadowrealms tests in the runner code, like we already do worker tests.

asamuzaK:
> Thanks as always for taking on this hard task.

It's tough, but someone has to do it.
And I'm working on it positively, thinking it's kinda penance of Shugendo(修験道).

> Remember to use blobs to blocklist directories or large sets of tests (e.g. derive_bits_keys/**).

Done for `BlobURL/**`.
But for `derive_bits_keys/`, one test passing for some reason (perhaps falsely), so couldn't be gathered.
https://github.com/jsdom/jsdom/actions/runs/12722890009/job/35467444431#step:7:2978

> I think we should exclude all shadowrealms tests in the runner code, like we already do worker tests.

Done.

By the way, is there a reason we put `fail-fast: false`?
https://github.com/jsdom/jsdom/blob/main/.github/workflows/jsdom-ci.yml#L51

domenic:
> But for `derive_bits_keys/`, one test passing for some reason (perhaps falsely), so couldn't be gathered.

It may be possible to use more advanced glob syntax to include only that file? E.g. `derive_bits_keys/!(derive_key_and_encrypt.https.any.html)` or `derive_bits_keys/^derive_key_and_encrypt.https.any.html`.

Or, we could try to fix the test in WPT. It definitely looks like it should not be passing. Maybe change https://github.com/web-platform-tests/wpt/blob/de8a05c11520d1df06f57d2c87e940e5b499fdf9/WebCryptoAPI/derive_bits_keys/derive_key_and_encrypt.js#L6-L26 from `importKeys().then(, uselessFailureHandler); return Promise.resolve("define_tests")` to `return importKeys().then(() => "define_tests")` would suffice.

> Done.

I don't see this in the PR. I meant, adding some line like https://github.com/jsdom/jsdom/blob/bab2213d6f44cb05ff910dcda3afacf0a67c53d3/test/web-platform-tests/wpt-manifest-utils.js#L19-L22 so that we don't have to list them in to-run at all.

> By the way, is there a reason we put `fail-fast: false`?
> https://github.com/jsdom/jsdom/blob/main/.github/workflows/jsdom-ci.yml#L51

So that if it fails only on Node 18, for example, we can still see the results on Node 20.

asamuzaK:
> derive_bits_keys/!(derive_key_and_encrypt.https.any.html) or derive_bits_keys/^derive_key_and_encrypt.https.any.html.

Oh, I didn't know we can use globs. I'll try.

>  I meant, adding some line like 

Got it, will do.

> So that if it fails only on Node 18, for example, we can still see the results on Node 20.

I see

domenic (approved):

Issue #3320, opened 2022-01-15
uninstall form-data dep


jimmywarting:
Seeing as if you have FormData-impl.js shipped already... how about dropping the dependency on the form-data package and using your own now instead?

https://github.com/jsdom/jsdom/blob/9dae17bf0ad09042cfccd82e6a9d06d3a615d9f4/lib/jsdom/living/xhr/xhr-utils.js#L294-L295

domenic:
Pull requests that pass all the tests are always welcome.

jimmywarting:
fyi, nodeJS have spec compliant formdata globally avalible now. doe it still lacks iterating over form elements cuz it works similar to a web worker.
extending it doe could be easy with a own constructor.

domenic:
This has become more urgent because the form-data project has been taken over by the notorious maintainer who bloats packages with a bunch of their own dependencies.

PR #3841, opened 2025-03-15
Update dependencies and dev dependencies

domenic:
Regrettably, the form-data update bloats the package tree due to maintainer takeover; see #3320. However, our users were already getting the updated version anyway due to semver matching, so we should just reflect that in the lockfile as well.


domenic:
More nwsapi regressions... Also I think we need to roll the WPTs to get recent URL test updates.

domenic:
Abandoning for a separate from-scratch PR.

PR #3843, opened 2025-03-16
SVG improvements (rebased)

domenic:
This is a rebase and partial split-up of https://github.com/jsdom/jsdom/pull/2926, with the goal of landing it.


domenic:
Implementation notes for part 1 (AI allowed):

AI was only slightly helpful here. I tried pointing Cursor (default agent mode) at the original PR and asking it to port over changes, but it pretended to do so while instead fabricating new implementations from scratch. (It admitted to doing so after I called it out.) I had more success letting Cursor do some more minor refactoring, like the first commit that moves reflection into its own file.

I briefly used Claude Sonnet 3.7 (non-thinking) to remind myself of how to port changes from one Git branch to another.

domenic:
Implementation notes part 2 (AI not allowed):

AI would definitely have been helpful here, but this ended up being less volume of work than I thought, so the loss wasn't so bad. (Honestly, I was kind of dreading this work for the last week because I thought I'd have to grind out tons of repetitive test code without AI.) After realizing that `<use>` support was not a good idea, that left only two test files to write (for `<metadata>` and `<switch>`), and doing them by hand wasn't that bad.

Issue #3729, opened 2024-06-05
getClientRect is not a function


danieldongit:
### Basic info:

- **Node.js version:** 
- **jsdom version:** 

### Minimal reproduction case

When using Codemirror 6 I get the following error when mocking any user event:
`TypeError: textRange(...).getClientRects is not a function`

I managed to fix it by adding the following code to my test file:

```js
document.createRange = () => {
    const range = new Range();
    range.getClientRects = () => ({
        item: () => null,
        length: 0,
        [Symbol.iterator]: function *() {
            yield* [];
        },
    });
    return range;
};
```

### How does similar code behave in browsers?
This doesnt cause issues in the browser as this is JS native code. 

Would be great if this implmenetation (or something simiar) can be added so that there will be no need to add hacky solutions directly to test files.

Thanks!

aspiers:
I am also experiencing this issue.  Thanks a lot @danieldongit for sharing your solution!

Issue #1322, opened 2015-12-16
Implement a layout engine


yaycmyk:
I was taking a look at https://github.com/facebook/css-layout and it seems fairly complete.

A layout engine is I think the last major piece of the puzzle that JSDOM needs. I know it's a bit of an undertaking...

domenic:
It only does flexbox though, not e.g. relative or absolute or static positioning.

Sebmaster:
css-layout only implements flexbox. It'd not even be possible to parse only newly developed sites people are trying to parse with jsdom. We talked about implementing a layouting engine at jsconf.eu - the will is there, but priority definitely isn't.

The problem is, when you start going down that rabbit hole there's a lot to consider - we'd have to include font rendering (or at least font width measurement) to get even the simplest websites to work. All those things could be a bit reduced (for example only emulating a monospace font), but they're nontheless there and there won't be any correct results without taking all that into consideration.

inikulin:
Can't it be accomplished with native deps like Skia? It's not the best option, but it can be made as an optional extension (like node-canvas at the moment).

yaycmyk:
@domenic It says on the README they support relative and static. The absence of the other `display` types is annoying though

yaycmyk:
There's also this one: https://github.com/tategakibunko/nehan.js, though the docs are a little cryptic.

Sebmaster:
nehan.js:
> Available on IE8+, Firefox3.5+, Google Chrome4+, Safari3+, Opera10+ etc.

Makes me think they're using dom methods to render that stuff, which makes it useless to implement rendering.

>  absolute, relative and static

That's the simple ones, but doesn't take into account overflows, etc. It basically only supports boxes and doesn't take into account their contents (which are a big deal on the web though).

yaycmyk:
@Sebmaster I haven't looked too deep into it, but IE8+ could be due to using certain ES3-5+ things that the older browsers didn't support.

inikulin:
Well, implementing layout engine seems like an interesting hobby task for years =) So, I decided to give it a try when things will settle down a bit with my primary projects. Meanwhile, I've decided to improve my knowledge on internals by reading CSS spec (interesting reading BTW) and made some basic research. Regarding font measurement: seems like we can do it already without native code.
[...]

PR #3844, opened 2025-03-19
add support for getClientRects in Range

iammatthew2:
# Summary

Add support for `range.getClientRects` and reduce cases of "target.getClientRects is not a function" 

# Details

JSDOM does not currently handle getClientRects() for Ranges and this affects users trying to run tests in a variety of situations, including [Tiptap](https://tiptap.dev/), other extensions of ProseMirror and more

Fixes https://github.com/jsdom/jsdom/issues/3729

Conversations related to this issue:
- [Stackoverflow issue](https://stackoverflow.com/questions/68023284/react-testing-library-user-event-throws-error-typeerror-root-elementfrompoint/77219899#77219899)
- [Tiptap discussion](https://github.com/ueberdosis/tiptap/discussions/4008)

### DIFF
--- a/lib/jsdom/living/range/Range-impl.js
+++ b/lib/jsdom/living/range/Range-impl.js
@@ -421,12 +421,12 @@ class RangeImpl extends AbstractRangeImpl {
 
     class FakeDOMRectList {
       constructor(rects) {
-        this._rects = rects;
+        rects.forEach((rect, i) => (this[i] = rect));
         Object.defineProperty(this, "length", { value: rects.length, writable: false });
       }
 
       item(index) {
-        return this._rects[index] || null;
+        return this[index] || null;
       }
 
       get [Symbol.toStringTag]() {
@@ -435,7 +435,6 @@ class RangeImpl extends AbstractRangeImpl {
     }
 
     const rect = new FakeDOMRect(0, 0, 10, 10);
-
     return new FakeDOMRectList([rect]);
   }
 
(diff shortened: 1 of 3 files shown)


domenic:
Sorry, but implementing fakes in this way is not how we work in jsdom. If you want to add mocks in your own tests, you can, but we work by implementing specifications.

iammatthew2:
> Sorry, but implementing fakes in this way is not how we work in jsdom. If you want to add mocks in your own tests, you can, but we work by implementing specifications.

@domenic , the main reason I added more of a detailed mock is so we could successfully run this against the existing web platform test. I'm happy to refactor.

Would you recommend a simpler solution like what already exists in `lib/jsdom/living/nodes/Element-impl.js`? 

It looks like this:
```
  getClientRects() {
    return [];
  }
  ```
  
  And then I could add some sort of simple test against it

domenic:
I don't think we should add any of these methods until we have a full layout engine, really. I think the existing support for `getClientRects()` was a mistake.

iammatthew2:
Is a layout engine on the roadmap? I see this old but active issue: https://github.com/jsdom/jsdom/issues/1322 but not much else. I'm willing to help dev it if it's moving forward

domenic:
A layout engine is extremely hard, but certainly if someone dropped a standards-compliant one in a pull request we'd work on reviewing it.

Probably a better step is to join the discussion in https://github.com/jsdom/jsdom/issues/2751.

PR #3845, opened 2025-03-23
Run WPT lint on to-upstream WPTs

### LINKED ISSUE #2264: Run WPT lint on to-upstream WPTs
https://github.com/web-platform-tests/wpt/issues/9879 is now fixed, so we should be able to lint our to-upstream tests too. Filing this bug to track that work.

domenic:
Closes #2264.

### DIFF
--- a/test/web-platform-tests/run-tuwpts.js
+++ b/test/web-platform-tests/run-tuwpts.js
@@ -1,20 +1,38 @@
 "use strict";
+/* eslint-disable no-console */
 const path = require("path");
 const { describe, before, after } = require("mocha-sugar-free");
 const { spawnSync } = require("child_process");
 const { readManifest, getPossibleTestFilePaths } = require("./wpt-manifest-utils.js");
 const wptServer = require("./wpt-server.js");
-const { killSubprocess } = require("./utils.js");
+const { killSubprocess, spawnSyncFiltered } = require("./utils.js");
 
 const wptPath = path.resolve(__dirname, "tests");
 const testsPath = path.resolve(__dirname, "to-upstream");
-const manifestFilename = path.resolve(__dirname, "tuwpt-manifest.json");
+const relativeTestsPath = path.relative(wptPath, testsPath);
 
 // We can afford to re-generate the manifest each time; we have few enough files that it's cheap.
-const testsRootArg = path.relative(wptPath, testsPath);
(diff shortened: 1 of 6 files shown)


domenic:
Implementation notes:

AI significantly sped up this task in several ways. I used Cursor, for a couple rounds set to auto but I switched to sonnet-3.5 quickly.

- Cursor was helpful in tab-completing some of the simple code in `run-tuwpt.js`, where running the linter is an analogous process to running the manifest generator and so it was able to create the parallel code pretty easily. (Refactoring out the common variables and such nicely.)
- When I needed to fix the lint errors in `document-selectionchange.html`, I didn't dare ask Cursor to just fix them immediately, because I suspected the context about `setTimeout()` vs. `t.step_timeout()` would be too much for it to know about. But, I edited about four key lines, and then told Cursor to do all the mechanical work of threading the `t` variable through to all call sites, and it did so quickly and flawlessly. That would have been quite tedious without AI.
- When I got confusing error messages from the lint script due to some complexities around my directory setup, I switched Cursor to "ask" mode and asked it questions about the WPT project's `lint.py` file. It found the source of the error messages and had good suggestions on how to fix it. I had to explicitly tell it how to navigate to an imported Python module which seemed a bit lame, but whatever.
  - A particularly notable case here was where there was an error message being output that I couldn't find with Ctrl+F. I asked Cursor, which correctly realized that the format of the error message was coming from a git subprocess that the script was spawning, and it found the part of the code that spawned that subprocess.
- When I decided eventually that the useless error message (reported at https://github.com/web-platform-tests/wpt/issues/51534) should just be filtered from the output, I used Cursor agent mode to write the filtering function (`spawnSyncFiltered`). I gave it detailed instructions which it executed on very well, but its initial code was duplicative, with identical code branches for stdout and stderr. I prompted it to deduplicate, which it did with a helper function, but I told it that was just overkill and suggested looping over `["stdout", "stderr"]`. It happily made that change, leading to the code we see here. (And it was amusingly pleased with itself, telling me how clean the new code was.)

Note that several times throughout this I got confused as to why Cursor couldn't "see" my current file, and it needed to be explicitly tagged as context. Toward the end I realized I'm probably supposed to start a new chat every time I want to do a new agent task, and indeed when I did that for the final task it properly auto-populated the context file.

PR #3846, opened 2025-03-23
Add a test expectations file for to-upstream tests

domenic:


domenic:
Implementation report:

This task consisted of a lot of factoring out existing code to make it more general, so it could be reused from two places. I suspect this task could have benefited from AI a decent amount, although it's also possible I would have pushed it beyond its limits and gotten frustrated.

There was a lot of refactoring that was on the border between mechanical and interesting, and I made some nontrivial decisions on how to simplify and layer the components after the refactoring. Just pointing Cursor at it and saying "please factor this out so that it can be called from two places" might have worked, but might have produced something not-quite-right.

I definitely noticed a drag from the lack of autocomplete and Cursor-tab, as even if I do the major refactorings myself, the micro-refactorings of updating moved code for its new environment would have benefited a lot of from those AI tools.

Issue #3750, opened 2024-07-23
Selecting HTML elements became case sensitive


ge-ku:
### Basic info:

- **Node.js version:** 20.13.1
- **jsdom version:** 24.1.1

### Minimal reproduction case

```js
const jsdom = require("jsdom");
const { JSDOM } = jsdom;

const dom = new JSDOM(`<!DOCTYPE html><myElement>Hello world</myElement>`);
console.log(dom.window.document.querySelector("myelement").textContent); // works in 24.1.1, 23.2.0 and 23.1.0
console.log(dom.window.document.querySelector("myElement").textContent); // error in 24.1.1, works in 23.2.0 and 23.1.0
```

### How does similar code behave in browsers?

In browser case doesn't matter, ```document.querySelector('div')``` or ```document.querySelector('dIV')``` leads to the same result.

bakkot:
This [happened previously](https://github.com/jsdom/jsdom/issues/3392), incidentally.

Issue #3544, opened 2023-05-05
Nested selector should not be case-sensitive 


ocavue:
### Basic info:

- **Node.js version:** 18
- **jsdom version:** 21.1.2

After updating JSDOM from 21.1.1 to 21.1.2 (and nwsapi from 2.2.2 to 2.2.4 https://github.com/jsdom/jsdom/commit/f75a9d1f7c27d653800b128ac75e2d9f0e93cb1f), the nested `UL > LI` doesn't work anymore because it becomes case-sensitive .

### Minimal reproduction case

Given the following code, it should print `1 1 1 1`. However it actually prints `1 1 1 0` 

```ts
import { JSDOM } from 'jsdom';

const dom = new JSDOM(`<!DOCTYPE html><body></body>`);
const document = dom.window.document;
const body = document.body;

const div = document.createElement('div');
const ul = document.createElement('ul');
const li = document.createElement('li');

ul.appendChild(li);
div.appendChild(ul);
body.appendChild(div);

console.log(div.querySelectorAll('UL').length);
console.log(div.querySelectorAll('ul').length);
console.log(div.querySelectorAll('ul > li').length);
console.log(div.querySelectorAll('UL > LI').length);
```

Online playground with JSDOM 21.1.1 and nwsapi 2.2.2: https://stackblitz.com/edit/node-sahv91?file=package.json&file=index.mjs

<img width="1087" alt="image" src="https://user-images.githubusercontent.com/24715727/236422828-c5c36309-d025-4c85-b20d-659a41c222b0.png">

Online playground with JSDOM 21.1.2 and nwsapi 2.2.4: https://stackblitz.com/edit/node-rqxlwe?file=package.json&file=index.mjs

<img width="1087" alt="image" src="https://user-images.githubusercontent.com/24715727/236422978-789b3aff-cddd-4214-be52-fd1dd90c937e.png">

 

### How does similar code behave in browsers?

Run the code above in browsers should print `1 1 1 1`

dormesica:
Hey
I see this was tagged as `maybe fixed but I'm getting a similar issue on version `24.0.0`.

Is there an expectation for resolution?

Manvel:
> Hey I see this was tagged as `maybe fixed` but I'm getting a similar issue on version `24.0.0`.
> 
> Is there an expectation for resolution?

Yeah, created a ticket in nwsapi repo: https://github.com/dperini/nwsapi/issues/100
[...]

PR #3848, opened 2025-03-30
Add various selectors tests

domenic:
These are all drawn from the issue tracker.

### DIFF
--- a/test/web-platform-tests/to-upstream-expectations.yaml
+++ b/test/web-platform-tests/to-upstream-expectations.yaml
@@ -1,3 +1,4 @@
+dom/nodes/ParentNode-querySelector-case-insensitive.html: [fail, https://github.com/jsdom/jsdom/issues/3612]
 dom/nodes/ParentNode-querySelector-host-pseudo-class.html: [fail, https://github.com/jsdom/jsdom/issues/3618]
 dom/nodes/ParentNode-querySelector-hyphen-prefix.html: [fail, https://github.com/jsdom/jsdom/issues/3620]
 jsdom-only/failure-example-1.html: [fail, https://github.com/jsdom/jsdom/issues/3835]
--- a/test/web-platform-tests/to-upstream-expectations.yaml
+++ b/test/web-platform-tests/to-upstream-expectations.yaml
@@ -1,4 +1,7 @@
 dom/nodes/Element-matches-active-pseudo-class.html: [fail, https://github.com/jsdom/jsdom/issues/3610]
+dom/nodes/Element-matches-disabled-pseudo-class.html:
+  "querySelectorAll(':enabled') should not select elements in disabled fieldset": [fail, https://github.com/jsdom/jsdom/issues/3603]
+  "querySelector('input:enabled') should return the first enabled input": [fail, https://github.com/jsdom/jsdom/issues/3603]
 dom/nodes/ParentNode-querySelector-case-insensitive.html: [fail, https://github.com/jsdom/jsdom/issues/3612]
 dom/nodes/ParentNode-querySelector-host-pseudo-class.html: [fail, https://github.com/jsdom/jsdom/issues/3618]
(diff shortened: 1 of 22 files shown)


domenic:
Implementation report:

I was excited and thankful that this issue was randomized to the AI-allowed bucket. It was a lot of grungy work, turning user bug reports into standardized-format test cases, and seemed perfect for AI to work on. I estimated 180 minutes without AI, and 60 minutes with AI.

However, in the end I'm not sure AI actually saved much time here. I ended up spending 240 minutes with AI. I think I might have been able to produce the same volume of tests if I were just coding with no AI, as long as I was fully locked in and coding at high speed.

I think I underestimated:

- The volume of issues I was churning through: at ~25 issues, 60 minutes is not realistic unless the AI is flawlessly doing an issue in 2 minutes without human intervention. Which might be possible if I was less picky and the AIs were slightly better at following instructions, but that was not my experience.
- The amount of iteration we'd need to get on the same page about issues like coding style, instruction following, etc.
- Just the time it takes to write the code and run it.

I was right that this task was pretty well-suited to AI, though. Unlike most other tasks so far in the study, this time I wrote ~0 lines of code. I spent the entire time just guiding the AIs through it, giving them reviews and guidance.

- I tried Claude Code first, excited because I felt this was a use case where a fully agentic setup would shine. This was going reasonably, with it autonomously browsing the GitHub issues list and writing the test, until I ran into issues where it wasn't able to run the test server from inside the Windows Subsystem for Linux environment. I wasted about 20 minutes purely trying to troubleshoot my network configuration before giving up and saying that Claude Code is not a usable product at this stage due to its lack of first-class Windows support.

- I then switched to Cursor. I went for Claude 3.7 thinking mode (not max), because I thought this might benefit from some extra iteration. This went well for like 4 or 5 tests.
  - It would do the whole flow: write the test, run it, update the expectations file if the test failed.
  - I tried at first to get it to make the commits, but it was having trouble creating multiline commit messages. (Maybe another Windows problem?) So I took over that part of the work.
  - It was sometimes quite eager, generating a lot of test cases related to the problem instead of just the simple reproduction case the user encountered. Sometimes this was nice, although I scolded it once because it failed to test the user's actual code.
  - Eventually though, it went crazy. It started hallucinating tests that didn't exist, and repeatedly trying to insert them into the expectations file, then debug the problems this caused.
- I decided to switch to trying Cursor's gemini-2.5-pro-exp-03-25 mode, the new kid on the block. This served as my workhorse for the remaining 15+ tests, and I was reasonably happy with it, but it definitely had some dumb moments:
  - I tried at the beginning to tell it not to insert redundant comments or assertion messages that duplicate the content being tested. It kept doing this and I kept having to remind it to remove them.
  - It had real troubles with lexicographically sorting the test filenames within the expectations file, which was surprisingly dumb. Like, it kept moving the line one line backward, running the tests and getting the sorting error, and repeating, instead of just inserting it into the right place.
  - I tried hard to establish a workflow of "write the test, let me suggest improvements, then run it". But it was over-eager and kept trying to run it before I had finished my checks. A bit scary as a failure of instruction-following.
  - It also sometimes added extra complications to the test instead of just testing what the user reported, but not as much as Claude.
  - Toward the end, it totally lost its marbles when I asked it to add both #3750 and #3544 tests to an existing test file. It hallucinated that these issues had something to do with attributes instead of tag names, and the case-insensitivity flag on attributes. I tried to correct it two or three times but it kept just iterating on this false premise.
    - I got it back on track by rejecting everything, and saying "stop. Let's start over on this problem." and then doing just one issue at a time.
    - This was strange because previously it had done two issues at a time without a problem.

I continue to be somewhat plagued by Cursor UI confusion. The two biggest issues are that committing files doesn't seem to "accept" them, and I have to do that separately or else the AI gets confused; and, typing `@` then pasting links into Cursor will often have it mention the current file instead of the link text, for some reason.

PR #3849, opened 2025-03-30
Support indexed properties on HTMLFormElement

### LINKED ISSUE #3616: Implement `HTMLFormElement` indexed getter
### Basic info:

- **Node.js version:** 18 
- **jsdom version:** 22 

### Minimal reproduction case

```js
import jsdom from "jsdom";
const { JSDOM } = jsdom;

const dom = new JSDOM(`
  <!doctype html>
  <html>
    <head></head>
    <body>
      <form>
        <input name="one" value="" />
        <input name="two" value="" />
      </form>
    </body>
  </html>
`);

console.log(dom.window.document.forms.length); // 1
console.log(dom.window.document.forms[0].length); // 2

// this should report `HTMLInputElement {}` but doesn't
console.log(dom.window.document.forms[0][0]); // undefined <--- !!!
```

Index access on a form object to access its form elements in browsers returns the form's element node found at that index. Attempting the same thing in JSDOM returns `undefined`.

### Relevant portion from the spec

https://html.spec.whatwg.org/multipage/forms.html#htmlformelement

formelement

domenic:
Closes #3616.


domenic:
Implementation report: this was relatively fine without AI, taking 25 minutes. Most of the time was spent re-familiarizing myself with the webidl2js infrastructure and the relevant parts of the spec, which AI wouldn't have necessarily helped much with.

I spent 10 of those minutes trying to implement the named getter too, but got confused when browsers don't seem to match the spec for `Object.getOwnPropertyNames()` with named properties. So I decided to just tie it off and implement only the original feature request.

Issue #2149, opened 2018-02-10
Single Quotes in URL  = Bad Request - Invalid URL


khadlock:
We're attempting to load working site URLs and are receiving a "Bad Request - Invalid URL" response from jsdom. Here's an example: http://www.dominionpost.com/‘Rise-up’-Sen--Capito-vis--its-M

You can see that the site address loads in a browser, but if you attempt to load via jsdom you'll get this error.

Is there a URL option that would resolve this for us?

Thank you and we appreciate your assistance!

domenic:
Unfortunately this issue did not follow the issue template, so we cannot help with it. Please update to follow the template, then drop another message in this thread, and we can reopen it.

khadlock:
We're attempting to load working site URLs and are receiving a "Bad Request - Invalid URL" response from jsdom. Here's an example: http://www.dominionpost.com/‘Rise-up’-Sen--Capito-vis--its-M

You can see that the site address loads in a browser, but if you attempt to load via jsdom you'll get this error.

Is there a URL option that would resolve this for us?

Please see the sample code below.

- **Node.js version:** 9.4.0
- **jsdom version:** 9.12.0

```js
const { JSDOM } = require("jsdom");

jsdom.env({
                    url: "http://www.dominionpost.com/‘Rise-up’-Sen--Capito-vis--its-M",
                    contentType: "text/html",
                    scripts: ["http://code.jquery.com/jquery.js"],
                    done: function (err, window) {});

```

Zirro:
That's an outdated version of jsdom. Can you reproduce it with the latest release, 11.6.2?

khadlock:
The new version seems to be working, thank you. However, I have a new question; how can we inject scripts using the fromURL? We were using the "scripts" option in the "env" function in the previous versions of JSDOM.

Issue #3659, opened 2024-01-08
Performance impact of jsdom `23.2.0`


DoodleBobBuffPants:
### Basic info:

- **Node.js version:** `21.4.0`
- **jsdom version:** `23.2.0`

### Minimal reproduction case

- Vitest tests with jsdom `23.1.0` (mm:ss):
  - Without coverage: 3:30
  - With coverage: 5:30
- Vitest tests with jsdom `23.2.0` (mm:ss):
  - Without coverage: 7:30
  - With coverage: 15:30

The hit to performance in `23.2.0` is quite significant, to the point that a handful of our unit tests doing no more than clicking and selecting using [RTL](https://testing-library.com/docs/react-testing-library/intro/) are consistently timing out too now

I wanted to raise this as an issue as requested in the release notes for `23.2.0`

### How does similar code behave in browsers?
N/A

kpala:
Seeing even more dramatic impact on one of our test suites using RTL and mocha:

jsdom `23.1.0`:

```
436 passing (1m)
```

jsdom `23.2.0`:

```
433 passing (12m)
3 failing                    <-- timed out
```

That's 12x time spent (and the timeouts).

domenic:
Thanks both for your reports. It looks likely that we'll have to revert. /cc @asamuzaK

If either of you are able to produce public repro cases, as standalone repositories that we can clone and run, I imagine that will make our work on reintroducing the selection engine in the future easier.

asamuzaK:
Sorry for the inconvenience.

DoodleBobBuffPants:
I started looking at a repro but in my case I don't believe it will add anything new - in `23.1.0` our tests which timeout were under the default timeout limit of 5s, and in `23.2.0` our tests exceed that timeout. Increasing the timeout resolves this, so it comes down to performance again

SamuelWei:
My tests are also failing due to TypeError since `23.2.0`
```
TypeError: Cannot read properties of null (reading 'createTreeWalker')
    at q._setup (/var/www/html/node_modules/@asamuzakjp/dom-selector/src/js/matcher.js:173:29)
    at new selector (/var/www/html/node_modules/@asamuzakjp/dom-selector/src/js/matcher.js:95:67)
    at Object.y (/var/www/html/node_modules/@asamuzakjp/dom-selector/src/index.js:41:31)
    at exports.closest (/var/www/html/node_modules/jsdom/lib/jsdom/living/helpers/selectors.js:22:22)
[...]

Issue #3154, opened 2021-03-23
Web platform tests that are worryingly slow


domenic:
We've been getting several CI failures recently from `Node-cloneNode.html` taking longer than 70 seconds to pass. (Even on my local machine which is pretty beefy, it takes around 45 seconds.) I will disable it for now.

Similarly, many tests involving ranges (e.g. under `dom/ranges` or `selection`) take a very long time to pass. I will not be disabling those yet, although we might have to at some point.

It'd be lovely if we could dig in to what is slow about these tests. I suspect that jsdom could be faster; maybe we have some O(n^2) algorithm or something. This seems especially likely for ranges.

If we think a given case is just a very slow test, then we could try to split it up by modifying the upstream WPT repo, e.g. like https://github.com/web-platform-tests/wpt/pull/21881 did.

Narretz:
One issue I've found is that selections that use setEndBefore (and probably also setStartBefore) iterate over the whole DOM in some cases.

Consider this example, where the range ends before the second paragraph.

The end node of this range is the body element, which kind of makes sense because the end is between two nodes.

However, the iteration code https://github.com/jsdom/jsdom/blob/04f6c13f4a4d387c7fc979b8f62c6f68d8a0c639/lib/jsdom/living/range/Range-impl.js#L394 makes it so that the endNode is `null` which means the loop can never terminate and iterates over all nodes in the DOM. This isn't an issue with small documents, but big documents become very slow.

To me the obvious solution would be to introduce an explicit boundary node for selections that end between nodes.

```
import { JSDOM } from 'jsdom';

const html = `<!DOCTYPE HTML>
<html>

<head>
    <meta charset="utf-8">
    <title>AHTML 2.1 Document</title>
    <style type="text/css"></style>
</head>

<body id="0">
    <p id="1"><span id="start" class="cs1" lang="und" dir="ltr">text</span>
      <span id="2601" class="cs2" lang="und" dir="ltr">inside</span>
      <span id="2602" class="cs1" lang="und" dir="ltr">selection</span>
    </p>
    <p id="end">
      <span id="2612" class="cs1" lang="und" dir="ltr">text</span>
      <span id="2613" class="cs2" lang="und" dir="ltr">outside</span>
      <span id="2614" class="cs1" lang="und" dir="ltr">selection</span></p>
</body>
`;

const dom = new JSDOM(html);

const bodyEl = dom.window.document.body;

const domRange = bodyEl.ownerDocument.createRange();

[...]

PR #3854, opened 2025-04-03
Add dom-selector CSS engine once again

### LINKED ISSUE #2159: querySelector doesn't deal with xml namespaces

### LINKED ISSUE #3321: :scope selector fails when element has id beginning with number
### Basic info:

- **Node.js version:** 16.13.1
- **jsdom version:** 19.0.0

### Minimal reproduction case

```js

const options = {
  contentType: "application/xml"
};
const dom = new JSDOM('<a id="9a"><b/></a>', options);

dom.window.document.documentElement.querySelector(":scope>b"); // DOMException
```

```
new DOMParser().parseFromString('<a id="9a"><b/></a>', "application/xml")
  .documentElement.querySelector(":scope>b");
```

No DOMException, returns `<b/>` as expected.

### LINKED ISSUE #3370: querySelectorAll() fails on case-insensitive attribute flag `I`
### Basic info:

- **Node.js version**: 16.14.0
- **jest version:** 27.4.7

### Minimal reproduction case

```js
/**
 * @jest-environment jsdom
 */
describe('case-insensitive attributes ', () => {
    it('1: lower case flag', () => {
        const selector = 'div[class=CasE i]';
        expect(document.querySelectorAll(selector).length).toBe(0);
    });

    // fails with error:
    // SyntaxError: 'div[class=CasE I]' is not a valid selector
    it('2: upper case flag', () => {
        const selector = 'div[class=CasE I]';
        expect(document.querySelectorAll(selector).length).toBe(0);
    });
});
```

standard document.querySelectorAll() method does not fail on upper case flag `I`
https://jsbin.com/jukodeluso/edit?html,js,console

it is allowed and described in docs
https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors#syntax

### LINKED ISSUE #3432: Nested `:is` selectors are treated as invalid

### LINKED ISSUE #3466: `.closest('#null')` or `.closest('.null')` returns self
### Basic info:

- **Node.js version:** 14.19.3
- **jsdom version:** 20.0.3

### Minimal reproduction case

"null" selectors like `#null` and `.null` that shouldn't match anything are matching the original node when queried with `.closest()`.

```js

const { window } = new JSDOM(`<p>foo</p>`);
const p = window.document.querySelector('p');

// These return null, as expected
console.log(p.closest('#whatever'));
console.log(p.closest('.whatever'));

// These return the HTMLParagraphElement
console.log(p.closest('#null'));
console.log(p.closest('.null'));
```

Returns `null` when there is no match.

### LINKED ISSUE #3469: svg, querySelector(All) and :scope
### Basic info:

- **Node.js version:** 16.14
- **jsdom version:** 20.0.3

### Minimal reproduction case

```js

const svgEl = JSDOM.fragment('<?xml version="1.0" encoding="UTF-8"?><svg><defs></defs><g></g></svg>').firstElementChild;

svgEl.tagName;  // "svg" as expected
svgEl.childNodes.length;  // 2 as expected
svgEl.childNodes[1].tagName;  // "g" as expected
svgEl.querySelector('g');  // "SVGElement" as expected
// BUT
svgEl.querySelector(':scope g');  // null - not as expected !!!
```

In Chrome/Firefox the query ":scope g" returns the "SVGElement" just as "g" does.

Cheers

### LINKED ISSUE #3544: Nested selector should not be case-sensitive

### LINKED ISSUE #3603: :disabled pseudo-class does not select elements in <fieldset disabled>
### Basic info:

- **jsdom version:** 22.1.0

### Minimal reproduction case

```js

import { JSDOM } from 'jsdom';

const dom = new JSDOM(`
<body>
  <fieldset disabled>
    <input id="not-focusable-in-fieldset">
  </fieldset>
  <input id="not-focusable-outside" disabled>
  <input id="focusable">
</body>
`, {});

const disabledInput = dom.window.document.body.querySelector('input:disabled');
console.log(disabledInput.id);
if (disabledInput.id === 'not-focusable-in-fieldset') {
  console.log(':disabled in fieldset: correct');
} else {
  console.log(':disabled in fieldset: incorrect');
}

const notDisabledInput = dom.window.document.body.querySelector('input:not(:disabled)');
console.log(notDisabledInput.id);
if (notDisabledInput.id === 'focusable') {
  console.log(':disabled in fieldset: correct');
} else {
  console.log(':disabled in fieldset: incorrect');
}
```

Inputs inside disabled fieldsets can be selected using `querySelector(':disabled')`:
https://jsbin.com/vaxozexedu/edit?html,console,output

The `:disabled` pseudo-class (https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled) represents any disabled element, which includes all inputs inside a `<fieldset>` element with the `disabled` attribute.

My tests only consider `querySelector`, I do think JSDOM handles these inputs as disabled correctly w.r.t. other properties. For example, I think JSDOM does not allow elements inside a disabled fieldset to be focused.

### LINKED ISSUE #3610: :active always matching
### Basic info:

- **Node.js version:** 16.20.0
- **jsdom version:** 22.1.0

### Minimal reproduction case

```js

const dom = new JSDOM('<div>hello</div>');

const element = dom.window.document.querySelector('div');

console.log('active', element.matches(':active'));
```

It logs `active true`, but should log `active false`. This behaves correctly in the browser.

**edit:** I first thought it was due to the update from 21 to 22, but it might not be related. Investigating further
**edit2:** After some digging, I found nwsapi 2.2.7 to be the culprit. The fix (a revert) seems to be already in the master branch, but it has not been released yet. I can locally override to 2.2.6 to make this work again. You might want to as well until the next version is released.

https://jsbin.com/monetalayo/edit?html,js,console

### LINKED ISSUE #3612: selector not work when use uppercase node name
### Basic info:

- **Node.js version: v20.6.1** 
- **jsdom version: 22.1.0** 

### Minimal reproduction case

```js
test("check jsdom uppercase nodeName", () => {
  const mockDocument = new JSDOM(`
  <!DOCTYPE html>
  <html lang="en">
  <body>
    <h1>test mask solution</h1>
    <div class='box'>
      <div>
        <p id="p1"></p>
        <p id="p2"></p>
      </div>
    </div>
  </body>
  </html> `);
  const { document } = mockDocument.window;

  const result = document.querySelector("div.box");
  expect(result).not.null;
  const result1 = document.querySelector("DIV.box");
  expect(result1).to.null;
});
```
if I use `div.box` then the result was expected, but `DIV.box` yield a null 

works in both case

(Link to a jsbin or similar strongly suggested.)

### LINKED ISSUE #3618: Unknown pseudo-class selector ':host>*'
### Basic info:

It is not possible to use the [:host pseudo class] inside the selector parameter of `querySelector` or `querySelectorAll` in `jsdom`.

- **Node.js version:** 18.17.1
- **jsdom version:** 22.1.0

### Minimal reproduction case

```js
const dom = new jsdom.JSDOM(`
<!DOCTYPE html>
<html lang="en">
    <body></body>
</html>
`);

const section = document.createElement('section');
const article = document.createElement('article');
const shadowSection = section.attachShadow({ mode: 'open' });

shadowSection.appendChild(article);
document.body.appendChild(section);

const OK = document.querySelector('section').shadowRoot.querySelector('article');
const KO = document.querySelector('section').shadowRoot.querySelector(':host > article');
```

In browsers, both queries return the same element, the first one just queries for an `article` in any place inside the shadow DOM subtree, the second one specificaly queries for an article that is an inmmediate child of the `shadowRoot` using the [:host pseudo class].

https://jsbin.com/nifuso/edit?html,js,console

[:host pseudo class]: https://developer.mozilla.org/en-US/docs/Web/CSS/:host

### LINKED ISSUE #3620: Class and ID selector's prefixed with hyphen returning null with document.querySelector
### Basic info:

- **Node.js version:** 18.18.2
- **jsdom version:** ^22.1.0

### Minimal reproduction case

```js
const { document } = (new JSDOM('')).window;

try {
     console.log(document.querySelector('#-123')); // Returns null - incorrect
     console.log(document.querySelector('.-123')); // Returns null - incorrect
} catch(err) {
     console.error(err); // Should throw error
}

```

https://jsbin.com/coromew/edit?js,console,output

Verified Chrome / Firefox both throw errors

### LINKED ISSUE #3686: `querySelector(":is([a],b):not(.c)")` throws a "not a valid selector" error
### Basic info:

- **Node.js version:** v20.11.1
- **jsdom version:** 24.0.0

jsdom v23.2.0 didn't show this issue.

### Minimal reproduction case

```js
const dom = new JSDOM('');

dom.window.document.querySelector(`:is([a],b):not(.c)`)
// Uncaught DOMException [SyntaxError]: '[a],b):not(.c' is not a valid selector

dom.window.document.querySelector(`:is(a,[b]):not(.c)`);
// Uncaught DOMException [SyntaxError]: 'a,[b]):not(.c' is not a valid selector

// the followings work
dom.window.document.querySelector(`:is(a,b):not(.c)`);
dom.window.document.querySelector(`:is(a,b):not([c])`);
dom.window.document.querySelector(`:is(a):not([b],.c)`);
dom.window.document.querySelector(`:is(a,[b])`);
dom.window.document.querySelector(`:is(a,[b]):first-child`);
dom.window.document.querySelector(`:is([b]):not(.c)`);
```

They don't throw the error.

https://codepen.io/ginpei/pen/BabXQzv

### LINKED ISSUE #3692: Colon-Prefixed Classes Break :scope Selectors in querySelector/querySelectorAll
### Basic info:

- **Node.js version:** `18.15.0`
- **jsdom version:** `24.0.0`

### Minimal reproduction case

```js
const { JSDOM } = jsdom;
const string =  `<div class="sm:block"><span></span></div>`;

const html = new JSDOM(string);
const htmldoc = html.window.document;

const div = htmldoc.querySelector('div');
const span = div.querySelectorAll(':scope > span'); // Also breaks if using querySelector
```

This code works fine in web browsers.

### Other Details

`jsdom` appears to substitute the `:scope` selector with whatever part of the `X:`-prefixed class name comes after `X`.

If I change the `class="sm:block"` to e.g. `class="anything-not-colon-prefixed sm:block"`, it works. It seems the problem only occurs when the only classes defined on the element are those with a `:` prefix.

### Stack Trace

```
~/jsdom-issue/node_modules/nwsapi/src/nwsapi.js:559
        throw err;
        ^
DOMException [SyntaxError]: unknown pseudo-class selector ':block>*'
    at emit (~/jsdom-issue/node_modules/nwsapi/src/nwsapi.js:557:17)
    at compileSelector (~/jsdom-issue/node_modules/nwsapi/src/nwsapi.js:1273:17)
    at compile (~/jsdom-issue/node_modules/nwsapi/src/nwsapi.js:745:16)
    at collect (~/jsdom-issue/node_modules/nwsapi/src/nwsapi.js:1539:22)
    at Object._querySelectorAll [as select] (~/jsdom-issue/node_modules/nwsapi/src/nwsapi.js:1499:36)
    at HTMLDivElementImpl.querySelectorAll (~/jsdom-issue/node_modules/jsdom/lib/jsdom/living/nodes/ParentNode-impl.js:78:26)
    at HTMLDivElement.querySelectorAll (~/jsdom-issue/node_modules/jsdom/lib/jsdom/living/generated/Element.js:1119:58)
    at Object.<anonymous> (~/jsdom-issue/jsdom.js:9:18)
    at Module._compile (node:internal/modules/cjs/loader:1254:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
```

### LINKED ISSUE #3750: Selecting HTML elements became case sensitive

### LINKED ISSUE #3780: Implement the `:defined` selector
### Basic info:

- **Node.js version:** 22.8.0
- **jsdom version:** 25.0.1

Jsdom uses `nwsapi.js` for their query selectors, which does not support the `:defined` selector. This issue is mainly created to notify that it would be very useful to have this functionality available.

### Minimal reproduction case

```js

const dom = new JSDOM(`
<!DOCTYPE html>
<html lang="en">
  <body>
    <p>Hello World</p>
  </body>
</html>
`, { contentType: "text/html" });

dom.window.document.querySelectorAll(":defined"); // DOMException [SyntaxError]: unknown pseudo-class selector ':defined'
```

```js
// [object NodeList] (16)
[<html/>,<head/>,<meta/>,<title/>,<meta/>,<link/>,<link/>,<link/>,<style/>,<script/>,<script/>,<script/>,<script/>,<body/>,<p/>,<script/>]
```

### LINKED ISSUE #3792: Syntax error thrown from getComputedStyle()
### Basic info:

- v20.18.0
- 24.0.0

### Minimal reproduction case

```js

    const html =
      '<html><head><style type="text/css">  tr.svelte-1ob8zmv:has(input:where(.svelte-1ob8zmv):checked) td:where(.svelte-1ob8zmv) {\n    \n    background-color: rgb(226 232 240);\n  }\n\n  </style></head><body><div><div class="w-full h-full relative svelte-1ob8zmv"><div class="sm:flex svelte-1ob8zmv"> <div class="w-full svelte-1ob8zmv"><div class="flex justify-end h-full items-center mt-2 bg-white svelte-1ob8zmv"> </div></div></div>   <table tabindex="-1" class="w-full grid border-collapse items-stretch overflow-auto mt-2 outline-none bg-white svelte-1ob8zmv" style="max-height: -72px; grid-template-columns: minmax(auto, 1fr) minmax(auto, 1fr) ; grid-template-rows: repeat(1, 48px);"><thead class="svelte-1ob8zmv"><tr id="table-row-header" class="svelte-1ob8zmv"><th class="group table-header-name flex items-center px-2 sm:px-6 h-12 font-semibold text-gray-900 sticky top-0 text-left\n            overflow-hidden z-10 border-b border-gray-300 bg-white backdrop-blur backdrop-filter gap-x-2\n            ease-in-out duration-200 svelte-1ob8zmv" id="header-1" title="Name" tabindex="0">  </th><th class="group table-header-name flex items-center px-2 sm:px-6 h-12 font-semibold text-gray-900 sticky top-0 text-left\n            overflow-hidden z-10 border-b border-gray-300 bg-white backdrop-blur backdrop-filter gap-x-2\n            ease-in-out duration-200 svelte-1ob8zmv" id="header-2" title="Total Number" tabindex="-1">  </th></tr></thead> <tbody class="svelte-1ob8zmv"><tr class="svelte-1ob8zmv"><td class="col-span-full svelte-1ob8zmv"></td></tr></tbody></table> </div> </div></body></html>'

    const dom = new JSDOM(html)
    const th = dom.window.document.querySelector('th')
    dom.window.getComputedStyle(th) // this line throws the error "SyntaxError: 'input:where(.svelte-1ob8zmv):checked) td:where(.svelte-1ob8zmv' is not a valid selector"
```

No error: https://jsbin.com/lopukojanu/edit?html,js,console,output

### LINKED ISSUE #3802: Inconsitent appendChild behavior from Web implementation
### Basic info:

- **Node.js version:** 22.12.0
- **jsdom version:** 25.0.1

### Minimal reproduction case

Given a document with the following XML:

```xml
<bar>
  <bar id="theBar"><child-bars/></bar>
  <child-bars/>
</bar>
```

using `appendChild` to move `#theBar` into the parent `child-bars` element will generate the error `[DOMException [HierarchyRequestError]: The operation would yield an incorrect node tree.]`.

```js
const xml = `
  <bar>
    <bar id="theBar"><child-bars/></bar>
    <child-bars/>
  </bar>
`.trim();
const { document } = new JSDOM(xml, { contentType: 'text/xml' }).window;
for (const bar of [...document.querySelectorAll('bar')]) {
  if (!bar.parentElement) continue;
  bar.parentElement.querySelector(':scope > child-bars').appendChild(bar);
}
```

Comparatively running the same in FireFox works as expected.

```js
const xml = `
  <bar>
    <bar id="theBar"><child-bars/></bar>
    <child-bars/>
  </bar>
`.trim();
const doc = new DOMParser().parseFromString(xml, 'text/xml');
for (const bar of [...doc.querySelectorAll('bar')]) {
  if (!bar.parentElement) continue;
  bar.parentElement.querySelector(':scope > child-bars').appendChild(bar);
}
```

[JSFiddle Link](https://jsfiddle.net/te4jLf8r/)

### LINKED ISSUE #3818: query selector fails with 26.0.0 to find for `button#react-aria-:r22:.css-ofddhz svg`
Since version 26.0.0 I get `SyntaxError: 'button#react-aria-:r22:.css-ofddhz svg' is not a valid selector`.

I do:

`await userEvent.click(screen.getByTestId('delete-button-0'));`

The generated HTML looks like that:

```
<button
        aria-label="Delete"
        class="css-18cu5b5"
        data-rac=""
        data-testid="delete-button-0"
        id="react-aria-:r6:"
        type="button"
      >
        <svg
          fill="none"
          height="24"
          viewBox="0 0 24 24"
          width="24"
          xmlns="http://www.w3.org/2000/svg"
        >
         ...
        </svg>
      </button>
```

The id seems to be valid and is generated by react-aria.

I also opened an issue in nwsapi but it seems it is not related?

https://github.com/dperini/nwsapi/issues/140#issuecomment-2596018169

asamuzaK:
Ref #3667, #3644
This is another attempt to replace the CSS selector engine.
I believe there will be few performance issues this time.
However, do not merge until we can confirm there is no issue in a real project.
Until then, I would like you to publish a development release.

Issues fixed by this PR:
fix #2159, fix #3321, fix #3370, fix #3432, fix #3466,
fix #3469, fix #3544, fix #3603, fix #3610, fix #3612,
fix #3618, fix #3620, fix #3686, fix #3692, fix #3750,
fix #3780, fix #3792, fix #3802, fix #3818


asamuzaK:
Benchmark result:

|Selector|A: main (nwsapi)|B: PR (dom-selector)|A/B|
|:----|:----|:----|:----|
|closest|19,367 ops/sec ±9.02% (66 runs sampled)|18,060 ops/sec ±4.97% (67 runs sampled)|1.07|
|matches|14,894 ops/sec ±5.70% (62 runs sampled)|13,673 ops/sec ±5.66% (63 runs sampled)|1.09|
|querySelector|1.51 ops/sec ±1.76% (8 runs sampled)|907 ops/sec ±6.19% (54 runs sampled)|0.002 (dom-selector is faster)|
|querySelectorAll|1.46 ops/sec ±1.69% (8 runs sampled)|1.22 ops/sec ±1.00% (7 runs sampled)|1.20|
|sizzle querySelector|0.46 ops/sec ±5.96% (6 runs sampled)|0.37 ops/sec ±4.76% (5 runs sampled)|1.35|
|sizzle querySelectorAll|0.49 ops/sec ±0.91% (6 runs sampled)|0.32 ops/sec ±4.96% (5 runs sampled)|1.53|

asamuzaK:
The test failures appears to be related to #3154, not this PR.

asamuzaK:
Ref #3659
Tested locally with [ariakit](https://github.com/ariakit/ariakit)

jsdom + nwsapi
```
 Test Files  132 passed (132)
      Tests  665 passed (665)
   Start at  22:54:55
   Duration  545.00s (transform 8.73s, setup 124.53s, collect 2.44s, tests 1112.49s, environment 277.38s, prepare 39.44s)
```

jsdom + dom-selector
```
 Test Files  132 passed (132)
      Tests  665 passed (665)
   Start at  23:11:20
   Duration  595.00s (transform 8.69s, setup 128.84s, collect 2.57s, tests 1209.64s, environment 321.15s, prepare 39.84s)
```

asamuzaK:
Re: https://github.com/jsdom/jsdom/actions/runs/14283858797/job/40036577613#step:4:139

```
Err:58 http://security.ubuntu.com/ubuntu noble-updates/main amd64 liblzma-dev amd64 5.6.1+really5.4.5-1build0.1
  404  Not Found [IP: 52.252.75.106 80]
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/x/xz-utils/liblzma-dev_5.6.1%2breally5.4.5-1build0.1_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
Fetched 13.3 MB in 3s (4013 kB/s)
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Error: Process completed with exit code 100.
```

It appears to have been changed to [liblzma5_5.6.1+really5.4.5-1ubuntu0.2_amd64.deb](https://security.ubuntu.com/ubuntu/pool/main/x/xz-utils/liblzma5_5.6.1+really5.4.5-1ubuntu0.2_amd64.deb)

Not sure what to do here.

asamuzaK:
Issues fixed by this PR:
fix #2149, fix #3321, fix #3370, fix #3432, fix #3466,
fix #3469, fix #3544, fix #3603, fix #3610, fix #3618,
fix #3620, fix #3686, fix #3692, fix #3750, fix #3780,
fix #3792, fix #3802, fix #3818,

asamuzaK:
@domenic
I think I've done everything I can for now.
If there are no problems with the implementation, I'd like you to publish a pre-release package and have users to test with it.
What do you think?

domenic:
I will try to work on this on my next jsdom weekend, although I have a pretty packed schedule for the next couple of weeks so it might be some time.

I will not be able to merge until we get CI passing, one way or another. I guess the problem is that `Range-mutations-dataChange.html` is failing, which is disappointing because I thought I made it significantly faster in https://github.com/jsdom/jsdom/compare/8da5ea8e3cb4bd3e586056dc306616d7a783f3ed...f54e8e719a79d6a8bf93d0e1441ef2d106dd6a69 . I will want to do some local profiling to see if there is any major change between dom-selector and nwsapi for that test, I guess.

benchmark/selectors/complex-selectors.js
```diff
@@ -43,8 +43,12 @@ exports.matches = () => {
       node = document.getElementById(`p${x - 1}-${y - 1}-${z - 1}`);
     },
     fn() {
+      const div = node.ownerDocument.createElement("div");
+      node.ownerDocument.body.appendChild(div);
```
domenic, line 47: Why make these changes to the benchmarks? How much do they effect the results, and in what direction?

test/web-platform-tests/to-upstream/css/selectors/dir-selector-auto-dont-upstream.html
```diff
@@ -0,0 +1,95 @@
+<!doctype html>
+<html>
+<head>
+  <link rel="help" href="http://www.w3.org/TR/selectors4/#dir-pseudo">
+  <script src="/resources/testharness.js"></script>
+  <script src="/resources/testharnessreport.js"></script>
+  <!-- jsdom only test without innerText -->
```
domenic, line 7: For this test you have a helpful comment explaining why we have a dont-upstream version.

Can you do the same for others?

Also maybe include the comment somewhere very high up like after `<!doctype html>`. That makes it easier to spot.

test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-All-content.html
```diff
@@ -0,0 +1,377 @@
+<!DOCTYPE html>
```
domenic, line 1: Are these all from-scratch tests that we should work to upstream?? Or are derived from existing tests?

asamuzaK: They are copied from /dom/nodes/
Should we renamethem too?

benchmark/selectors/complex-selectors.js
domenic: Can you tell me how much these changes affect the results, and in what direction?

test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-All-content.html
domenic: Why can't we just use the original dom/nodes ones?

benchmark/selectors/complex-selectors.js
asamuzaK: Does this answer your question?

nwsapi:
|Selector|A: main|B: PR|A/B|
|:----|:----|:----|:----|
|closest|106,163 ops/sec ±2.01% (72 runs sampled)|19,367 ops/sec ±9.02% (66 runs sampled)|5.48|
|matches|103,763 ops/sec ±5.65% (65 runs sampled)|14,894 ops/sec ±5.70% (62 runs sampled)|6.97|
|querySelector|2.85 ops/sec ±3.32% (11 runs sampled)|1.51 ops/sec ±1.76% (8 runs sampled)|1.89|
|querySelectorAll|2.70 ops/sec ±4.10% (11 runs sampled)|1.46 ops/sec ±1.69% (8 runs sampled)|1.85|
|sizzle querySelector|18/273 fails. 1.16 ops/sec ±8.44% (7 runs sampled)|18/273 fails. 0.46 ops/sec ±5.96% (6 runs sampled)|2.52|
|sizzle querySelectorAll|18/273 fails. 1.08 ops/sec ±35.05% (7 runs sampled)|18/273 fails. 0.49 ops/sec ±0.91% (6 runs sampled)|2.20|

dom-selector
|Selector|A: main|B: PR|A/B|
|:----|:----|:----|:----|
|closest|93,173 ops/sec ±2.83% (73 runs sampled)|18,060 ops/sec ±4.97% (67 runs sampled)|5.16|
|matches|90,166 ops/sec ±8.41% (64 runs sampled)|13,673 ops/sec ±5.66% (63 runs sampled)|6.59|
|querySelector|1,871 ops/sec ±5.19% (59 runs sampled)|907 ops/sec ±6.19% (54 runs sampled)|2.06|
|querySelectorAll|2.27 ops/sec ±1.80% (10 runs sampled)|1.22 ops/sec ±1.00% (7 runs sampled)|1.86|
|sizzle querySelector|7/273 fails. 0.49 ops/sec ±5.81% (6 runs sampled)|7/273 fails. 0.37 ops/sec ±4.76% (5 runs sampled)|1.32|
|sizzle querySelectorAll|7/273 fails. 0.50 ops/sec ±1.56% (6 runs sampled)|7/273 fails. 0.32 ops/sec ±4.96% (5 runs sampled)|1.56|

test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-All-content.html
asamuzaK: Both files are called within `to-upstream/dom/nodes/ParentNode-querySelector-All-dont-upstream.html`, `to-upstream/dom/nodes/ParentNode-querySelector-All-xht-dont-upstream.xht`.

```
<script src="selectors.js"></script>
<script src="ParentNode-querySelector-All.js"></script>
```

Is the following what you are suggesting?:

* Change those two paths to absolute paths like
  ```
  <script src="/dom/nodes/selectors.js"></script>
  <script src="/dom/nodes/ParentNode-querySelector-All.js"></script>
  ```
* In `run-single-wpt.js`, make modifications in the same way as `/resources/testharness.js` so that they are loaded from the `./tests` folder

domenic: The file I am commenting on here is `ParentNode-querySelector-All-content.html`, not `selectors.js` or `ParentNode-querySelector-All.js`. I am asking whether `ParentNode-querySelector-All-content.html` (and `ParentNode-querySelector-All-content.xht`) are completely new tests, or are derived from existing tests.

---

Separately, if we can reduce duplication for `selectors.js` and `ParentNode-querySelector-All.js` using a technique like what you describe, I think that would be good.

benchmark/selectors/complex-selectors.js
domenic: Thank you, that answers the question! So the changes to the benchmark in this PR favor dom-selector over nwsapi: they show a lower performance gap compared to the one we would see on `main`.

I guess you are explaining that this is because nwsapi has better caching than dom-selector, and you don't want to test the caching.

I think keeping at least a small benchmark which shows the differences in caching would be helpful? It could be used as a target to optimize dom-selector's caching strategy.

I don't know how important caching is for our users. So probably it is OK to not make it part of the main benchmark. But I am still a bit worried...

test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-All-content.html
asamuzaK: Ah, they are loaded in iframe.

ParentNode-querySelector-All-dont-upstream.html
L19:
```
frame.src = "ParentNode-querySelector-All-content.html#target";
```

ParentNode-querySelector-All-xht-dont-upstream.xht
L21:
```
frame.src = "ParentNode-querySelector-All-content.xht#target";
```

I think we can remove them and load from `./tests` too.

benchmark/selectors/complex-selectors.js
asamuzaK: I'm a little confused by what you're saying.

First, I didn't make the changes because it would be advantageous to dom-selector.
This is a benchmark, so if each operation was affected by the cache, it couldn't be measured accurately and that's not good.
That's why I made the change to the benchmark.

Secondly, if there is a big difference between having reflow and not having it (nwsapi: 2.2 - 2.5, dom-selector: 1.3 - 1.5), it probably means that the cache isn't being cleared correctly when it should be.
That is, dom-selector is handling the cache better.
This is true even when pseudo-classes like `input:checked` are still cached.
In fact, in `to-run.yaml`, tests like `pseudo-classes/indeterminate-type-change.html` and `pseudo-classes/readwrite-readonly-type-change.html` that failed in nwsapi are now passing.

test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-All-content.html
asamuzaK: Loading xht (xhtml) files from `./tests` didn't work well.
It seemed to be a content-type issue, but maybe I'm just doing things wrong.

asamuzaK:
> I will not be able to merge until we get CI passing, one way or another. I guess the problem is that `Range-mutations-dataChange.html` is failing, which is disappointing because I thought I made it significantly faster in [8da5ea8...f54e8e7](https://github.com/jsdom/jsdom/compare/8da5ea8e3cb4bd3e586056dc306616d7a783f3ed...f54e8e719a79d6a8bf93d0e1441ef2d106dd6a69) . I will want to do some local profiling to see if there is any major change between dom-selector and nwsapi for that test, I guess.

All green now.

The timeout is likely due to memory swap.

Ref: https://github.com/web-platform-tests/wpt/blob/9574a65b8b265c971aa15fa54500eb8f4a763119/dom/common.js#L18

This function creates and adds a lot of element nodes, many times.

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -168,8 +167,23 @@ exports.getDeclarationForElement = elementImpl => {
   return declaration;
 };
 
-function matches(rule, element) {
-  return matchesDontThrow(element, rule.selectorText);
+function matches(rule, elementImpl) {
+  try {
+    if (rule.selectorText) {
+      const domSelector = elementImpl._ownerDocument._domSelector;
+      const selector = rule.selectorText.split(";").pop().trim();
```
domenic, line 174: Can you explain and/or add a comment for what this is doing? It was not necessary with the old code so I wonder why it is necessary here.

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -196,6 +197,12 @@ class DocumentImpl extends NodeImpl {
     // https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#throw-on-dynamic-markup-insertion-counter
     this._throwOnDynamicMarkupInsertionCounter = 0;
 
+    // Create DOMSelector instance
+    this._domSelector = new DOMSelector(this._globalObject, this._ownerDocument, {
+      domSymbolTree,
+      idlUtils
+    });
```
domenic, line 204: This moves us from lazily initializing nwsapi/domSelector to always initializing them.

Is that OK? The `DOMSelector` constructor looks cheap but I wonder if the nwsapi initialization is expensive.

It's certainly much simpler to just always initialize, so I like your new code for simplicity. But it might be worth checking the impact on very simple benchmarks like `benchmark/dom/construction.js`.

test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-xml-namespaces.html
domenic, line 1: This test does not pass in browsers, I believe: https://jsbin.com/bibutenoko/1/edit?html,output

test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-All-content-dont-upstream.xht
```diff
@@ -0,0 +1,372 @@
+<!DOCTYPE html>
```
domenic, line 1: Can you add a comment explaining why we have a dont-upstream copy of this?

domenic (commented):
Overall this looks very nice. I really appreciate the attention to test coverage.

I left a few minor comments, and we should separate out the test-splitting work into https://github.com/jsdom/jsdom/pull/3860 which I would appreciate your review on. But I think we should be able to merge this soon.

I guess you are suggesting we do a prerelease test first? I am a bit worried we won't get enough testing coverage because a lot of people use jsdom as a transitive dependency (via Jest or Vitest or something, from what I understand). But maybe we can make it work if we:

- Make a list of all issues where people are discussing either selector functionality or selector performance. Probably, that list = your list of fixed issues in the OP, plus https://github.com/jsdom/jsdom/pull/3667, https://github.com/jsdom/jsdom/pull/3644, and https://github.com/jsdom/jsdom/issues/3659
- For each issue we add a small message asking people to test and giving them instructions on how to use [package.json overrides](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides) to upgrade to a prerelease version.

lib/jsdom/living/helpers/style-rules.js
asamuzaK: There is a bug in cases like:
``` css
@namaspace(https://example.com);

body {
  color: red;
}
```
We get `@namaspace(https://example.com);\n\nbody` as selectorText.

There are also some bugs in getComputedStyle(), and I'm planning to fix them altogether later.

asamuzaK:
What I have in mind is:

1. Leave this PR open
2. Prerelease jsdom@27.0.0-b.1
3. Contact jest-dom (@testing-library) and vitest developers to test with them
4. Add fixes on both jsdom and dom-selector if needed
5. Release dom-selector@6.5.0
6. Close this PR
7. Release jsdom@27.0.0

I think I can take care of contacting jest-dom, vitest.
Thoughts?

test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-xml-namespaces.html
asamuzaK: Thanks.
I'll take a deeper look.

lib/jsdom/living/helpers/style-rules.js
domenic: I guess this is a bug in the cssstyle library? A comment pointing to a bug report on that library would be helpful.

test/web-platform-tests/run-single-wpt.js
```diff
@@ -46,8 +46,20 @@ class CustomResourceLoader extends ResourceLoader {
...
+    } else if (url.pathname.startsWith("/dom/nodes/")) {
+      // Some tests require extra resources.
+      // Add them from the one in ./tests.
+      const extraResources = [
+        "/dom/nodes/ParentNode-querySelector-All-content.html",
+        "/dom/nodes/ParentNode-querySelector-All.js",
+        "/dom/nodes/selectors.js"
+      ];
+      if (extraResources.includes(url.pathname)) {
+        const filePath = path.resolve(__dirname, "tests" + url.pathname + (url.hash ? url.hash : ""));
```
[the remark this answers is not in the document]
asamuzaK: I'll fix url.hash part.
hash is needed in, for example, test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-All-dont-upstream.html L19
It loads iframe content with hash.

domenic:
I find leaving PRs open, so that releases on npm correspond to nothing clear on GitHub, to be a bit confusing. So I would prefer a slightly modified flow for that part. Let's merge to main and do a release from main, just with a prerelease version number.

I am unsure whether it should be a new major version or a new minor version. It is theoretically not a breaking change so a new minor version is more appropriate. But maybe it would be more cautious to do a major version?

I like the idea of contacting jest-dom and vitest to have them test. I think asking for feedback from people in https://github.com/jsdom/jsdom/issues/3659 by commenting on that thread would also be good.

test/web-platform-tests/run-single-wpt.js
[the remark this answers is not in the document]
domenic: It's surprising that this makes a difference to the `super.fetch()` call. The file on disk does not have a hash in its name...

lib/jsdom/living/nodes/Document-impl.js
asamuzaK: Should be OK.

* I modified @asamuzakjp/nwsapi not to assume document.documentElement at install time.
  Instead look for document.documentElement at runtime when needed.
  https://www.npmjs.com/package/@asamuzakjp/nwsapi?activeTab=code L7
* On dom-selector side, it does not pass selector to @asamuzakjp/nwsapi if document.documentElement is not present.
  https://github.com/asamuzaK/domSelector/blob/main/src/index.js#L73

domenic: Right, I guess the lazy initialization was not for performance, but due to the nwsapi architecture. Thanks!

test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-All-content-dont-upstream.xht
asamuzaK: Done

lib/jsdom/living/helpers/style-rules.js
asamuzaK: Added a comment.
I'll file issue on cssstyle later.

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -168,8 +167,23 @@ exports.getDeclarationForElement = elementImpl => {
...
+      const { match, pseudoElement } = domSelector.check(selector, elementImpl);
+      // `pseudoElement` is a pseudo-element selector (e.g. `::before`).
+      // However, we do not support getComputedStyle(element, pseudoElement), so return `false`.
+      if (pseudoElement) {
+        return false;
+      }
+      return match;
+    }
+  } catch {
+    // fall through;
```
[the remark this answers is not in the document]
asamuzaK: Done

asamuzaK:
TODO:
* rebase after ##3860 merge and fix conflicts
* release dom-selector@6.5.0

test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-xml-namespaces.html
domenic: Can we instead add a version of the test case that matches browsers? If dom-selector is doing the incorrect thing here, it's good to keep it tracked.

asamuzaK: Note to myself:
Ref https://dom.spec.whatwg.org/#scope-match-a-selectors-string
> Support for namespaces within selectors is not planned and will not be added.

lib/jsdom/living/helpers/style-rules.js
asamuzaK: Filed regression test

test/web-platform-tests/to-upstream/dom/nodes/ParentNode-querySelector-xml-namespaces.html
asamuzaK: Fixed issue upstream.
Added passing test too.

domenic (approved):

PR #3860, opened 2025-04-13
Split "npm test" into more subcommands

domenic:
Helps with #3154 by creating a new Node process for WPTs, separate from all the other tests.

Get rid of the index.js file that was necessary to list all the tests.

Move test/jquery-fixtures into test/to-port-to-wpts/ since those are the only tests that use it.

Update npm scripts to follow the convention of prefix:subcommand.

This is a version of what @asamuzaK did in https://github.com/jsdom/jsdom/pull/3854 but I think it's a bit cleaner as it doesn't introduce new test-listing files. @asamuzaK, let me know what you think.


asamuzaK:
Basically, I have no objections.

The only thing I want to make clear is that this approach makes tests fail-fast.

When I asked you before why you don't use github actions to make it fail-fast, you replied that, for example, even if Node 18 failed, we want to get the test results for other Node(s).

This new approach still doesn't make the tests fail-fast for Node versions, but it makes the individual tests fail-fast.

The reason I combined test:tuwpt and test:wpt in the PoC was because that new tests would often be written in test:tuwpt, so even if they failed, we could continue running the test.

I personally think it's better to keep them separate and fail-fast, though.

asamuzaK:
Also, some of the tests that are `timeout` or `fail-slow` may pass after this fix.
However, we can check them the next time we update wpt.

domenic:
Thank you very much for the careful review and drawing my attention to that. I decided to use the `npm-run-all` package to get continue-on-error behavior while also splitting the suites.

Basically, I feel that if people want to do a smaller set of tests in isolation, like during development, `npm run test:tuwpt` or similar is best. Whereas if they are doing `npm test`, getting all results is best.

PR #3865, opened 2025-04-20
Fix ElementInternals accessibility getters/setters

### LINKED ISSUE #3732: Attempting to set ARIA properties on ElementInternals results in an error.
### Basic info:

Attempting to assign ARIA properties on ElementInternals for custom elements results in an error:

```txt
Error: Uncaught [TypeError: Cannot read properties of undefined (reading 'set')]
```

- **Node.js version:** 21.6.2
- **jsdom version:** 24.1.0

### Minimal reproduction case

```js

new JSDOM(
  `
<!DOCTYPE html>
<title></title>
<script>
class ExampleElement extends HTMLElement {
  constructor() {
    super();

    const internals = this.attachInternals();
    internals.role = 'button';
  }
}
customElements.define('example-element', ExampleElement);
</script>
<example-element></example-element>
`,
  {
    runScripts: "dangerously",
  }
);
```

[CodePen example][1]

[1]: https://codepen.io/zzzzBov/pen/jOoaKOm

domenic:
The web platform test coverage for these was woefully incomplete, so when we added them in d6c0ab258a84a0ce01ee1d0ae78195ddeadaa3aa they did not actually work at all.

Closes #3732.

### DIFF
--- a/lib/jsdom/living/custom-elements/ElementInternals-impl.js
+++ b/lib/jsdom/living/custom-elements/ElementInternals-impl.js
@@ -35,15 +35,19 @@ class ElementInternalsImpl {
   }
 
   _reflectGetTheContentAttribute(reflectedContentAttributeName) {
-    return this._targetElement._internalContentAttributeMap.get(reflectedContentAttributeName) ?? null;
+    return this._targetElement._internalContentAttributeMap?.get(reflectedContentAttributeName) ?? null;
   }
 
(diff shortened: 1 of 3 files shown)


domenic:
Implementation report: this was relatively easy without AI. Sure, autocomplete could have helped sometimes, and I might have been tempted to get it to implement some of the setters tests. (I'm not sure that would have saved any time though.) But this ended up being a pretty simple issue that I was able to bang out unassisted without problems.

PR #3866, opened 2025-04-20
Re-do jsdom errors in the virtual console

### LINKED ISSUE #3596: Enumerate the possible `jsdomError.type` strings
### Basic info:

- **Node.js version:** N/A
- **jsdom version:** latest

### Minimal reproduction case

```js
const { VirtualConsole } = require("jsdom");

virtualConsole.on("jsdomError", error => {
  if (error.type === "this error type doesn't exist") {
    console.log("Typescript should warn me here!");
  }
});
```

I filed a PR with `@types/jsdom` at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66630. I would like the jsdom team to take a look and discuss how we can better enumerate the jsdomError types. In my PR I note "There is no single enumeration of error type strings. It is not clear if jsdom considers this part of a stable API surface." Please give commentary on this. Are the strings going to be "stable" within a non-major version bump? It's not part of the API surface and not defined in DefinitelyTyped, so it is not guaranteed.

domenic:
* Rename sendTo() to forwardTo(). This helps people realize the method now has new semantics.

* Changed from { omitJSDOMErrors: true } to { jsdomErrors } which can be undefined, "none", or a list of jsdom error types to forward.

* Changed and consolidated error types and properties, from their previous undocumented form (usually a type and a detail) into a well-documented form with more structured properties.

* Stopped emitting jsdom errors for failed XHRs, similar to how browsers no longer display those in the console.

* Generally expanded test coverage for the jsdom errors.

Closes #3596.


domenic:
Implementation report:

It was definitely nice to have AI for this task. I was using Cursor, but mostly in tab-complete mode. For this sort of situation where I was doing a lot of updating from the old style to the new, or renaming things, tab-complete mode was great. In particular it was good (but not perfect) at anticipating what the next change I wanted to make was.

I briefly used the agent mode, with gemini-2.5-pro-exp-03-25, to adapt some old tests to the new API.

PR #3867, opened 2025-04-20
Use the HTML Standard UA stylesheet

### LINKED ISSUE #2642: LGPL protected file.
https://github.com/jsdom/jsdom/blob/master/lib/jsdom/browser/default-stylesheet.js#L10-L31
contains a GPL license which makes it really difficult to include this package (legally) into any web application.

### LINKED ISSUE #3836: Use HTML Standard UA stylesheet instead of Chrome's
It might need some adaptation to deal with selectors we don't support, but it seems like a good idea.

This consists of collecting all the CSS blocks from https://html.spec.whatwg.org/multipage/rendering.html#rendering and replacing the contents of https://github.com/jsdom/jsdom/blob/main/lib/jsdom/browser/default-stylesheet.js with it.

Then, we run jsdom and see if we get any CSS parsing errors. If so, we comment out the responsible parts of the stylesheet.

Finally, we need to compare the diff to see if there are things in Chrome's UA stylesheet that are not expressed in the spec, but are important somehow.

domenic:
This is a bit messy since we don't have any specificity support, or various other features. But it's probably about as good as our existing stylesheet, which was pulled from a very old version of Chromium.

Closes #3836. Closes #2642.


domenic:
Implementation report:

This task was pretty annoying to do without AI. Not for the coding itself, but for the repetitive task of scraping https://html.spec.whatwg.org/multipage/rendering.html into a CSS file, commenting about where each bit came from, commenting about the omitted bits, etc. I'm honestly not sure I got it 100% right, and might go back and have AI check it after the experiment.

Similarly, the original issue (#3836) envisioned

> Finally, we need to compare the diff to see if there are things in Chrome's UA stylesheet that are not expressed in the spec, but are important somehow.

But doing this by hand is kind of a fool's errand. If I was allowed to use AI for this task, I would love to throw this comparison problem at a couple of models and see what we got out. But without AI, I kind of just didn't do this. To be fair, what I realized while working on this issue in general is that both the old and new UA stylesheets were not working that well, due to limitations elsewhere in jsdom's CSS support. So I suspect that isn't wasn't too much that people might miss in the new UA stylesheet, because so much of the old one wasn't working as intended anyway. But it still would have been at least interesting, even if less actionable than I originally thought.

I similarly am interested in trying this comparison after the experiment.

asamuzaK:
Can we add a default font size something like this?
```
:root {
  font-size: 16px;
}
```
I'm planning to tackle `getComputedStyle()` issues, and we may need the font size in pixels for that.

domenic:
> Can we add a default font size something like this?

I think that should be done separately from the default UA stylesheet. Maybe `getComputedValue`?

In the specs they [say](https://drafts.csswg.org/css-fonts/#font-size-prop) the initial value is `medium`, but I can't find where in the specs that resolves to `16px`.

asamuzaK:
As far as I know, it's not explicitly stated in the spec.
However, all browsers seem to use 16px internally as the default value, although they allow the user to change it.

domenic:
Yeah, it's just a question of where we implement it. I don't think it's a UA stylesheet rule, but instead a computed-value time thing.

Issue #3565, opened 2023-06-07
HTMLSelectElement won't work with Object.defineProperty: TypeError: 'set' on proxy: trap returned falsish for property <name>


patrickpissurno:
### Basic info:

- **Node.js version:**  18.12.0
- **jsdom version:**  latest (22.1.0)

### Minimal reproduction case

```js
'use strict';

const { JSDOM } = require('jsdom');

const dom = new JSDOM('<select />', {});

const select = dom.window.document.querySelectorAll('select')[0];

const prop = 'test123';

const setter = (v) => {
    console.log('HERE:', v);
    return true;
};

Object.defineProperty(select, prop, {
    set: setter,
});

select.test123 = 123;

// test.js:20
// select.test123 = 123;
//                ^

// TypeError: 'set' on proxy: trap returned falsish for property 'test123'
//     at Object.<anonymous> (/Users/patrickpissurno/Desktop/bug/test.js:20:16)
//     at Module._compile (node:internal/modules/cjs/loader:1159:14)
//     at Module._extensions..js (node:internal/modules/cjs/loader:1213:10)
//     at Module.load (node:internal/modules/cjs/loader:1037:32)
//     at Module._load (node:internal/modules/cjs/loader:878:12)
//     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
//     at node:internal/main/run_main_module:23:47
```

### How does similar code behave in browsers?

The setter is called and no errors are returned. See [here](https://jsbin.com/cirazofaxi/edit?html,console,output):
```html
<!DOCTYPE html>
<html>
<body>
  <select/>

  <script>
    'use strict';
    
    const select = window.document.querySelectorAll('select')[0];

    const prop = 'test123';

    const setter = (v) => {
        console.log('HERE:', v);
        return true;
    };

    Object.defineProperty(select, prop, {
        set: setter,
    });

    select.test123 = 123;

    // "HERE:"
    // 123
  </script>
</body>
</html>
```

I've traced the issue to be in the `node_modules/jsdom/lib/jsdom/living/generated/HTMLSelectElement.js` file, inside of its `ProxyHandler` class.

There's a piece of code in the `ProxyHandler.set(target, P, V, receiver)` implementation that does this:

```js
if (!ownDesc.writable) {
  return false;
}
```

If I change that to:
```js
if (!ownDesc.writable) {
  return Reflect.set(target, P, V, receiver);
}
```

Everything works as expected.

[...]

PR #3868, opened 2025-04-20
Roll webidl2js and add a test for the defineProperty bug

domenic:
See #3565.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,7 @@
         "npm-run-all2": "^7.0.2",
         "pngjs": "^7.0.0",
         "server-destroy": "^1.0.1",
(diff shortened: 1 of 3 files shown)


domenic:
Implementation report:

(Note that the more important work happened in https://github.com/jsdom/webidl2js/pull/272.)

This was slightly annoying to do without AI, but was not that hard of a task regardless. Some of the refactorings and cleanups could have benefited from autocomplete. And I might have been tempted to use AI to generate the test case, but I think that would have taken more time than just typing it out manually. Similarly I might have been tempted to tell AI "look at the spec and fix my implementation", but it was pretty quick to just put the spec on the other monitor and fix the implementation manually.

PR #3873, opened 2025-05-06
Remove `fail-node18` reason and add `fail-lt-node24`

asamuzaK:


domenic:
Nice update, thanks. Yeah, feel free to make that change in this PR as well.

asamuzaK:
2**25 = 32 * 1024 * 1024 = 33554432 > 10000000, but it passes.

PR #3876, opened 2025-05-09
Update `tough-cookie@6.x`

colincasey:
The upcoming release of `tough-cookie` (currently available as a [pre-release](https://www.npmjs.com/package/tough-cookie/v/6.0.0-rc.0)) contains a [breaking change](https://github.com/salesforce/tough-cookie/pull/498) for how `Secure` cookies are treated in secure contexts such as `localhost`. This change causes several test failures in `jsdom` when used without modifying the default configuration for `CookieJar`.

This PR updates the cookie processing tests with:
- A `CookieJar` configured to simulate an insecure context to match the expected behavior of the `http`-based server tests
- A correction for the `https`-based server test which should allow `Secure` cookies

> [!Note]
> I'm using a draft PR here to give a heads up and gather potential feedback around this breaking change before publishing an official `v6.0.0` release.


domenic:
Can you explain more why this level of change is necessary? Why would localhost now accepting Secure cookies, cause all our non-Secure cookies that are set in tests to stop working as they have previously?

colincasey:
@domenic Using the first test as an example:

* describe `Cookie processing`
  * describe `document.cookie`
    * it `reflects back what is set to it`

This test starts a server running on a random port at `http://127.0.0.1:${port}`. The test then proceeds to set the following cookies for the url `/TestPath/test-page` using the `window.document.cookie=${cookieStr}` API for JSDOM:

```
"Test1=Basic; expires=Wed, 13-Jan-2051 22:23:01 GMT",
"Test2=PathMatch; expires=Wed, 13-Jan-2051 22:23:01 GMT; path=/TestPath",
"Test3=PathNotMatch; expires=Wed, 13-Jan-2051 22:23:01 GMT; path=/SomePath/",
"Test4=DomainMatchButPubSuffixRejected; expires=Wed, 13-Jan-2051 22:23:01 GMT; domain=.com",
"Test5=DomainNotMatch; expires=Wed, 13-Jan-2051 22:23:01 GMT; domain=.example.com",
"Test6=HttpOnly; expires=Wed, 13-Jan-2051 22:23:01 GMT; path=/; HttpOnly",
"Test7=Secure; expires=Wed, 13-Jan-2051 22:23:01 GMT; path=/; Secure",
"Test8=Expired; expires=Wed, 13-Jan-1977 22:23:01 GMT; path=/",
"Test9=Duplicate; One=More; expires=Wed, 13-Jan-2051 22:23:01 GMT; path=/",
"Test10={\"prop1\":5,\"prop2\":\"value\"}; expires=Wed, 13-Jan-2051 22:23:01 GMT; path=/",
"Malformed; expires=Wed, 13-Jan-2051 22:23:01 GMT; path=/"
``` 

This test expects the following cookies to have been set:
```
"Test1=Basic",
"Test2=PathMatch",
"Test9=Duplicate",
"Test10={\"prop1\":5,\"prop2\":\"value\"}",
"Malformed"
```

With `tough-cookie@5.x`, this test passes but, with `tough-cookie@6.x`, it will fail because the default configuration of the `CookieJar` used now allows the cookie with `"Test7=Secure; expires=Wed, 13-Jan-2051 22:23:01 GMT; path=/; Secure"` to be set as it treats the `localhost` domain under `http` as potentially trustworthy. All the tests using the `http`-based local server suffer from this same failure.

This new behavior in `tough-cookie@6.x` matches how Chrome, Firefox, and Edge treat Secure cookies on `localhost`. This behavior can vary by user-agent though, as [RFC6265bis (draft 19)](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19#section-5.8.3-2.1.2.3.2) states:

> NOTE: The notion of a "secure" connection is not defined by this document. Typically, user agents consider a connection secure if the connection makes use of transport-layer security, such as SSL or TLS, or if the host is trusted. For example, most user agents consider "https" to be a scheme that denotes a secure protocol and "localhost" to be trusted host.

So this PR is reconfiguring the `CookieJar` used to match the existing behavior of an `http` server with an untrustworthy host to align with how these tests were originally written. An alternative to this approach would be to update the test assertions to include the Secure cookie. I wasn't sure which was preferable.

domenic:
Thanks! I am embarrassed to admit that I tried to review this just by reading the diff, and forgot we had a big block of test cookies (including one secure one) at the top of the file.

My preference would be that we switch most tests to expecting the secure cookie, and then have a single test with the fake insecure server. Or maybe we could even use `http://web-platform.test`, which people need to have aliased to jsdom in the hosts file to run our test suite anyway.

colincasey:
> My preference would be that we switch most tests to expecting the secure cookie

@domenic the tests have been updated to expect the secure cookie. The Web Platform Tests suite is also passing with the latest `tough-cookie` release candidate (`6.0.0-rc.1`).

One other major change we added in our v6 release candidates is migrating to ESM. I noticed that `jsdom` supports Node.js [>=20](https://github.com/jsdom/jsdom/blob/main/package.json#L91). Would it be possible to update that to `>=20.19`? That way `jsdom` could take advantage of [backported ESM support](https://github.com/nodejs/node/releases/tag/v20.19.0) which will help with adoption of ESM packages.

domenic:
I'm not sure. jsdom's intention is to support all LTS-supported Node.js versions. If you can find docs showing that versions <= 20.19 are not LTS-supported, then we could consider that.

colincasey:
Understandable @domenic, was just curious since I'm starting to see some libraries (e.g.; [yargs](https://github.com/yargs/yargs/pull/2451)) adopt that tighter engine restriction to promote the move to ESM-only. For now, I've added dual-mode ESM / CJS publishing for `tough-cookie` so that broader LTS support is retained. 

I should have an official release published sometime within the next week or two and will update this PR once that's available.

domenic:
I rebased this on main to fix the dependency update conflicts, and was just about to merge this, but the part of the PR that tests the non-secure behavior (discussed in https://github.com/jsdom/jsdom/pull/3876#issuecomment-2875036595) seems to have disappeared.

Was that intentional, or did it get lost accidentally?

colincasey:
@domenic that might be an oversight on my part. Let me double-check what I left out after I reverted most of the original changes in this PR.

PR #3880, opened 2025-05-19
Add regression test for #3552

### LINKED ISSUE #3552: CSS: background-image/style.backgroundImage has no effect
### Basic info:

- **Node.js version v18.16.0:** 
- **jsdom version: 22.0.0** 

### Minimal reproduction case
```js
// setup
const {document} = new JSDOM().window;

// can be executed in browser console as-well
document.body.insertAdjacentHTML("afterbegin", `<div id="a" style="background-image: linear-gradient(90deg, red 0%, green 100%); color: blue;">`);
const aStyle = document.getElementById("a").style;
console.assert(aStyle.color === "blue", "color should be blue"); // OK
console.assert(aStyle.backgroundImage === "linear-gradient(90deg, red 0%, green 100%)", "backgroundImage should be 'linear-gradient(90deg, red 0%, green 100%)'"); // FAILS

aStyle.backgroundImage = "linear-gradient(270deg, blue 0%, red 100%)"
console.assert(aStyle.backgroundImage === "linear-gradient(270deg, blue 0%, red 100%)", "backgroundImage should be 'linear-gradient(270deg, blue 0%, red 100%)'"); // STILL FAILS
```

```js
// just copy this to the browsers console (same code as above)
document.body.insertAdjacentHTML("afterbegin", `<div id="a" style="background-image: linear-gradient(90deg, red 0%, green 100%); color: blue;">`);
const aStyle = document.getElementById("a").style;
console.assert(aStyle.color === "blue", "color should be blue"); // OK
console.assert(aStyle.backgroundImage === "linear-gradient(90deg, red 0%, green 100%)", "backgroundImage should be 'linear-gradient(90deg, red 0%, green 100%)'"); // FAILS (not in browser)

aStyle.backgroundImage = "linear-gradient(270deg, blue 0%, red 100%)"
console.assert(aStyle.backgroundImage === "linear-gradient(270deg, blue 0%, red 100%)", "backgroundImage should be 'linear-gradient(270deg, blue 0%, red 100%)'"); // STILL FAILS (not in browser)
```

asamuzaK:
Fixes #3552

### DIFF
--- a/test/web-platform-tests/to-upstream/css/cssom/style-backgroundimage-lineargradient.html
+++ b/test/web-platform-tests/to-upstream/css/cssom/style-backgroundimage-lineargradient.html
@@ -1,0 +1,22 @@
+<!DOCTYPE html>
+<title>CSSOM: Setting style.backgroundColor to oklch()</title>
+<meta charset="utf-8">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<link rel="help" href="https://drafts.csswg.org/cssom/#the-elementcssinlinestyle-mixin">
+<!-- Regression test for https://github.com/jsdom/jsdom/issues/3552 -->
+
+<div id="test"></div>
+
+<script>
+"use strict";
+
+test(() => {
+  const div = document.getElementById("test");
+
+  div.style.backgroundImage = "linear-gradient(green, blue)";
+  assert_equals(div.style.backgroundImage, "linear-gradient(green, blue)");
+
+}, "Setting style.backgroundImage using linear-gradient()");
+
+</script>


domenic (commented):
Can we use the exact code that is claimed to fail in that issue? I think it is `linear-gradient(270deg, blue 0%, red 100%)`.

domenic (approved):

test/web-platform-tests/to-upstream/css/cssom/style-backgroundimage-lineargradient.html
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
+<title>CSSOM: Setting style.backgroundColor to oklch()</title>
```
domenic, line 2: Ah, the `<title>` is off. Put the correct title here and remove it from the `test()` function. (If it's a single test in a file, you don't need the label.)

domenic (changes requested):

Issue #3887, opened 2025-06-08
Several to-upstream CSS related tests do not match browsers


domenic:
See, e.g. https://jsbin.com/riqexagori/1/edit?html,output . Currently jsdom returns `""` for some colors. After upgrading cssstyle to 4.4.0, it returns `"CanvasText"`. But it should apparently return `"rgb(0, 0, 0)"`.

There are quite a lot of these problems in `css/cssom` and in `html/semantics/document-metadata/{the-link-element,the-style-element}`.

I will fix a few while upgrading to v4.4.0 but there might be more.

@asamuzaK

asamuzaK:
I haven't looked into the details of this issue, and I'm not sure why `CanvasText` is being returned, but I imagine it has something to do with recent changes to the default CSS?

Apart from that, there is a lot of problems with jsdom's implementation of getComputedStyle().
* For dimensions, if the value is relative, it will not be returned as an absolute value.
* URLs will remain relative URLs.
* Colors are not fully resolved ​​either.
* Same with calc() and var().

To resolve them, I think we need to add significant modifications to jsdom, cssstyle, and cssom.

domenic:
> I'm not sure why `CanvasText` is being returned, but I imagine it has something to do with recent changes to the default CSS?

Yes, probably. But it's surprising that CanvasText only shows up after upgrading from cssstyle 4.3.1 to 4.4.0. I am not sure what would cause that.

asamuzaK:
Found it.
We are setting `CanvasText` as initial value here.
https://github.com/jsdom/jsdom/blob/main/lib/jsdom/living/helpers/style-rules.js#L92

And a spec.
https://drafts.csswg.org/css-color-4/#the-color-property

So it seems we are getting the expected result now?

domenic:
I think the issue is around style resolution. getComputedStyle() is supposed to [return the resolved value](https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle). The [resolved value of a color is its used value](https://drafts.csswg.org/css-color-4/#resolving-color-values). The [used value is basically the computed value](https://drafts.csswg.org/css-cascade-5/#used-value). The [computed value](https://drafts.csswg.org/css-cascade-5/#computed-value) is computed from the "specified value" of `CanvasText` via https://drafts.csswg.org/css-color-4/#resolving-color-values , which states

[...]

PR #3892, opened 2025-06-21
Add color map for system colors

asamuzaK:
Related to #3887

There is no clear specification for how system color values ​​should be set.
In this implementation, the system colors are set to ensure sufficient contrast between foreground and background colors.

The failure on `css/cssom/style-color-values.html` is fixed on cssstyle side and should be okay once we update cssstyle to next version.

### DIFF
--- a/lib/jsdom/living/helpers/style-rules.js
+++ b/lib/jsdom/living/helpers/style-rules.js
@@ -248,12 +248,12 @@ function getComputedValue(element, property) {
     }
     let value = "";
     if (specifiedValue === "initial") {
-      value = initial;
+      value = asciiLowercase(initial);
     } else if (specifiedValue === "inherit") {
       if (element.parentElement) {
         return getComputedValue(element.parentElement, property);
       }
-      value = initial;
+      value = asciiLowercase(initial);
     } else {
       value = asciiLowercase(specifiedValue);
     }


lib/jsdom/living/helpers/style-rules.js
```diff
@@ -245,7 +267,15 @@ function getComputedValue(element, property) {
...
     }
-
+    if (specifiedValue === "inherit") {
+      if (element.parentElement !== null) {
+        return getComputedValue(element.parentElement, "color");
+      }
+      return initial;
+    }
+    if (systemColors.has(specifiedValue.toLowerCase())) {
+      return systemColors.get(specifiedValue.toLowerCase());
```
domenic: Here is a fun fact: `"mar\u212A".toLowerCase() === "mark"`.

CSS (and other web specs) are ASCII case-insensitive, so this technically is not correct behavior.

Please update to use `asciiLowercase` from the `strings.js` helper file, and add a test for this case that fails before and passes after :). Maybe in its own file.

asamuzaK:
> Can we try passing the half of https://wpt.fyi/results/css/css-color/system-color-compute.html?label=experimental&label=master&aligned that does not use `getComputedStyleMap`?

I'm not sure what this https://wpt.fyi/results/css/css-color/system-color-compute.html is testing against.
`Menu` is a system-font keyword, not a system-color?

lib/jsdom/living/helpers/style-rules.js
asamuzaK: We need to fix cssstyle first.
Sent a PR for that.

domenic:
Ah, I think you forgot to implement https://drafts.csswg.org/css-color-4/#deprecated-system-colors :)

asamuzaK:
Oh my...

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -100,6 +102,264 @@ exports.propertiesWithResolvedValueImplemented = {
...
+    }
+  ],
+  [
+    "visitedtext", {
+      light: "rgb(128, 0, 128)",
+      dark: "rgb(255, 51, 255)"
+    }
+  ],
+  // Deprecated system colors
+  // https://drafts.csswg.org/css-color-4/#deprecated-system-colors
```
domenic: Can we instead programmatically set these to aliases, following the spec? Something like

```js
const deprecatedAliases = new Map([["activeborder", "buttonborder"], ...);

for (const [deprecated, systemColor] of deprecatedAliases) {
  systemColors.set(deprecated, systemColors.get(systemColor));
}
```

This will add a small amount of startup time, but better guarantees that we're matching the spec, and is more maintainable. (And saves a bit of memory, I guess.)

test/web-platform-tests/to-run.yaml
```diff
@@ -72,6 +72,56 @@ simple-requests.htm: [timeout, Maybe https://github.com/jsdom/jsdom/issues/1833
...
+parsing/color-valid-contrast-color-function.html: [fail, Unknown]
+parsing/color-valid-hsl.html: [fail, Unknown]
+parsing/color-valid-hwb.html: [fail, Unknown]
+parsing/color-valid-lab.html: [fail, Unknown]
+parsing/color-valid-relative-color.html: [fail, Unknown]
+parsing/color-valid-rgb.html: [fail, Unknown]
+parsing/color-valid-system-color.html: [fail, Unknown]
+parsing/opacity-computed.html: [fail, Unknown]
+parsing/opacity-valid.html: [fail, Unknown]
+parsing/relative-color-out-of-gamut.html: [fail, Unknown]
```
domenic, line 102: It's a bit surprising that we fail these given all the work you put into the color parser!

domenic:
Let me know if a new cssstyle release would be helpful to make this PR work better, and I will create one.

test/web-platform-tests/to-run.yaml
asamuzaK: Not sure, but probably we need to map `/css/support/computed-testcommon.js`

asamuzaK:
> Let me know if a new cssstyle release would be helpful to make this PR work better, and I will create one.

That would be a great help.
Thanks!

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -100,6 +102,152 @@ exports.propertiesWithResolvedValueImplemented = {
   }
 };
 
+// https://drafts.csswg.org/css-color-4/#css-system-colors
```
domenic: Can you put these two variables into their own module, to make it easier to read?

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -235,6 +383,29 @@ function getComputedValue(element, property) {
     const specifiedValue = getSpecifiedValue(element, property);
 
     // https://drafts.csswg.org/css-color-4/#resolving-other-colors
+    let value = "";
+    if (specifiedValue === "initial") {
+      value = initial;
+    } else if (specifiedValue === "inherit") {
+      if (element?.parentElement?.nodeType === NODE_TYPE.ELEMENT_NODE) {
```
domenic: I think the node type is always is guaranteed to be element, because you are using `parentElement` instead of `parentNode`.

Also, I don't think `element` should ever be null.

So the following seems more correct:

```suggestion
      if (element.parentElement) {
```

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -235,6 +383,29 @@ function getComputedValue(element, property) {
...
+        return getComputedValue(element.parentElement, property);
+      }
+      value = initial;
+    } else if (/^[a-z]+/i.test(specifiedValue)) {
+      value = asciiLowercase(specifiedValue);
+    }
+    if (value && (systemColors.has(value) || deprecatedAliases.has(value))) {
```
domenic: This is redundant.

```suggestion
    if (systemColors.has(value) || deprecatedAliases.has(value)) {
```

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -235,6 +383,29 @@ function getComputedValue(element, property) {
...
     // https://drafts.csswg.org/css-color-4/#resolving-other-colors
+    let value = "";
+    if (specifiedValue === "initial") {
+      value = initial;
+    } else if (specifiedValue === "inherit") {
+      if (element?.parentElement?.nodeType === NODE_TYPE.ELEMENT_NODE) {
+        return getComputedValue(element.parentElement, property);
+      }
+      value = initial;
+    } else if (/^[a-z]+/i.test(specifiedValue)) {
```
domenic: I don't think this guard is necessary. Since `value` is only used for lookup in the `systemColors` and `deprecatedAliases` maps, we can just always lowercase it.

test/web-platform-tests/to-run.yaml
domenic: (Not blocking this PR.)

Mapping should not really be necessary for most support files. But looking at https://github.com/web-platform-tests/wpt/blob/master/css/support/computed-testcommon.js my guess is that this using `CSS.supports()` and we don't implement that.

test/web-platform-tests/to-upstream-expectations.yaml
```diff
@@ -2,14 +2,15 @@ css/css-cascade/layers-basic.html: [fail, https://github.com/jsdom/jsdom/issues/
...
   "System color keyword": [fail, https://github.com/jsdom/jsdom/issues/3887 also needs cssstyle fix]
-  "Global keyword": [fail, https://github.com/jsdom/jsdom/issues/3887]
 css/cssom/style-font-shorthand.html: [fail, https://github.com/jsdom/jsdom/issues/3021]
 css/cssom/style-set-property.html:
   "set null for cssRule.style.setProperty": [fail, Need cssstyle fix; https://github.com/jsdom/cssstyle/issues/196]
   "set undefined for cssRule.style.setProperty": [fail, Need cssstyle fix; https://github.com/jsdom/cssstyle/issues/196]
   "set toString for div.style.setProperty custom property": [fail, Need cssstyle fix; https://github.com/jsdom/cssstyle/issues/129]
+css/cssom/system-color-lowercase.html:
+  "Color with upper cased letter": [fail, Need cssstyle fix]
+  "Color with upper cased kelvin sign": [fail, Need cssstyle fix]
```
domenic: I published a cssstyle update. Feel free to update cssstyle in this PR as well, so we can remove these.

test/web-platform-tests/to-run.yaml
asamuzaK: I'll take a look at them after all regressions caused by cssstyle 4.4.0 and/or 4.5.0 are addressed.

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -235,6 +237,29 @@ function getComputedValue(element, property) {
...
+    if (specifiedValue === "initial") {
+      value = initial;
+    } else if (specifiedValue === "inherit") {
+      if (element.parentElement) {
+        return getComputedValue(element.parentElement, property);
+      }
+      value = initial;
+    } else {
+      value = asciiLowercase(specifiedValue);
+    }
```
domenic, line 259: Shouldn't we call `asciiLowercase()` even if `specifiedValue === "initial"` or `"inherit"`? What if we inherit a system color?

If I am right about this being a bug, please add a test that fails before and passes after.

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -235,6 +237,29 @@ function getComputedValue(element, property) {
...
+        key = deprecatedAliases.get(value);
+      }
+      const { light, dark } = systemColors.get(key);
+      const colorScheme = getCascadedPropertyValue(element, "color-scheme");
+      if (colorScheme === "dark") {
+        return dark;
+      }
+      return light;
+    }
     if (specifiedValue === "currentcolor") {
```
domenic: As a style issue, can we move this to be right after the other `specifiedValue` tests? It should not make any functional difference, but I like the flow of handle special `specifiedValue` => handle system color specified values => return computed value.

lib/jsdom/living/helpers/style-rules.js
asamuzaK: All the `initial` values ​​in `propertiesWithResolvedValueImplemented` are in lowercase, so there shouldn't be any problems without `asciiLowercase()`.
See test passing in https://github.com/jsdom/jsdom/pull/3892/commits/968d0f94c0eba6d1e490eaeec49d920795c96db8

However, it would be safer to apply `asciiLowercase()` for future considerations.
For example, `propertiesWithResolvedValueImplemented` is extremely limited, so when we decide to use third party libraries instead.

Should we apply `asciiLowercase()` from now?

asamuzaK: Applied asciiLowercase()

asamuzaK:
I believe we've addressed all of the regressions pointed out in #3887. 
Please let me know if we've missed anything.

domenic:
Thank you! I will try to review soon. I apologize the last couple of weekends I have not had much time, because I was preparing for 日本語能力試験.

asamuzaK:
試験頑張ってください！

domenic (approved):
Thanks for your patience!

PR #3893, opened 2025-06-24
Add regression test for negative margins

asamuzaK:
Regression test for https://github.com/jsdom/cssstyle/issues/222

### DIFF
--- a/test/web-platform-tests/to-upstream/css/css-box/negative-margin.html
+++ b/test/web-platform-tests/to-upstream/css/css-box/negative-margin.html
@@ -1,0 +1,74 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Negative margins</title>
+<link rel="help" href="https://drafts.csswg.org/css-box/#margin">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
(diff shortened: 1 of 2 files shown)


test/web-platform-tests/to-upstream/css/css-box/negative-margin.html
```diff
@@ -0,0 +1,74 @@
...
+<div id="div"></div>
+
+<script>
+"use strict";
+const div = document.getElementById("div");
+
+test(() => {
+  div.style.marginTop = "-10px";
+  assert_equals(div.style.getPropertyValue("margin-top"), "-10px");
+  div.style.marginTop = null;
```
domenic:
```suggestion
test(t => {
  t.add_cleanup(() => { div.style.marginTop = null; });
  div.style.marginTop = "-10px";
  assert_equals(div.style.getPropertyValue("margin-top"), "-10px");
```

here and below. This is best practice to ensure that one test failing does not cause all other tests to fail.

domenic (changes requested):

test/web-platform-tests/to-upstream/css/css-box/negative-margin.html
asamuzaK: Done

domenic (approved):

PR #3896, opened 2025-06-28
Avoid a ReDos in string.js

bastien-roucaries:
[ ab]+$ is a ReDoS and crash a regression test on debian

Could you please apply

### DIFF
--- a/lib/jsdom/living/helpers/strings.js
+++ b/lib/jsdom/living/helpers/strings.js
@@ -35,12 +35,42 @@ exports.stripNewlines = s => {
 
 // https://infra.spec.whatwg.org/#strip-leading-and-trailing-ascii-whitespace
 exports.stripLeadingAndTrailingASCIIWhitespace = s => {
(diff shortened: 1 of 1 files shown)


domenic:
See https://github.com/jsdom/.github/blob/master/SECURITY.md#redos

bastien-roucaries:
@domenic Could you reconsider, it block some regression test on debian (we see it is not security supported on your side).

The whole story is we use jsdom to close and test security bug on debian side (ReDoS on angular) and we use jsdom for unit testing, but we hit this redos before the angular redos...

bastien-roucaries:
Test case here  https://salsa.debian.org/js-team/angular.js/-/blob/master/debian/tests/CVE-2024-21490-PoC.js?ref_type=heads

domenic:
Can you create a benchmark that does not use external libraries, as a pull request to https://github.com/jsdom/jsdom/tree/main/benchmark ?

PR #3900, opened 2025-07-08
3899: Fixes a bug where faketimers did not work for expiring cookies

DamianPereira:
I reported the original issue in jest, but I think the actual issue is this line in jsdom.

Original issue: https://github.com/jestjs/jest/issues/15727

This should make fake timers expire cookies correctly. Code comes from the original issue in jest repo: https://github.com/jestjs/jest/issues/11585, by @g2guy-aefxx

### DIFF
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -925,7 +925,7 @@ class DocumentImpl extends NodeImpl {
   }
 
   get cookie() {
-    return this._cookieJar.getCookieStringSync(this.URL, { http: false });
+    return this._cookieJar.getCookieStringSync(this.URL, { http: false, now: new Date() });
   }
   set cookie(cookieStr) {
     cookieStr = String(cookieStr);


domenic:
This needs tests if we are to accept it. See CONTRIBUTING.md.

DamianPereira:
Sorry, just realized by adding specs that the problem is not actually in jsdom itself (it was working as expected since current date is actually the default for cookieJar). The problem is that jest-environment-jsdom package creates the new JSDOM before actually creating the fake timers. I'll try to arrive at a solution there.

For reference for future googlers, when using jest, adding this to the setup makes jsdom use the fake timers clock, but you also have to use jest-environment-jsdom-global as environment in jest.config.js:

```js
Object.defineProperty(document, 'cookie', {
  get () {
    return jsdom.cookieJar.getCookieStringSync(this.URL, { http: false, now: new Date() })
  },

  set (cookieStr) {
    jsdom.cookieJar.setCookieSync(String(cookieStr), this.URL, { http: false, ignoreError: true })
  },

  configurable: true
})
```

PR #3902, opened 2025-07-12
Split `to-run.yaml` into directory based files

asamuzaK:
**Motivation**
`to-run.yaml` is quite huge, so even if you want to test only a part of it, it takes a long time to complete the test.
If we split `to-run.yaml` into several files, it can potentially make our development more efficient.


domenic:
I'm not sure on the motivation here.

Is the motivation to reduce the time it takes to run tests? To do that, you can filter which tests you run, by using `npm run test:wpt -- --fgrep <...test pattern..>`.

Is the motivation to make it easier to edit `to-run.yaml`?

I also think this particular approach has a lot more code duplication than I would prefer.

asamuzaK:
Oh, I didn't know about `--fgrep`.
Thanks, I'll use it.

PR #3904, opened 2025-07-13
Regression tests for cssstyle

asamuzaK:
Regression tests for https://github.com/jsdom/cssstyle/issues/209

Re: `style-background-shorthand.html`, there are some bugs in browsers?
* When background shorthand is set to `none`, i.e. `div.style.background = "none";`, Firefox resolves default `transparent` keyword of the backgroundColor to `rgba(0, 0, 0, 0)`.
* Chrome resolves default values to `initial` keyword.

--
Deleted previous PR by mistake.
Sorry for a noise.


domenic (approved):
The specs are a bit hard to read, but I think your expectation of `transparent` is correct. Thanks!

PR #3907, opened 2025-07-15
Regression test for cssstyle

asamuzaK:
https://github.com/jsdom/cssstyle/pull/232


test/web-platform-tests/to-upstream/css/css-images/gradient-specified-values.html
```diff
@@ -0,0 +1,80 @@
...
+
+<script>
+"use strict";
+
+const div = document.getElementById("div");
+
+test(t => {
+  t.add_cleanup(() => {
+    div.style.backgroundImage = null;
+  });
```
domenic, line 20: I wonder if it might be cleaner to have each test replace these three lines with `const div = document.createElement("div");`. Either way works though.

PR #3912, opened 2025-07-22
Port part of style.js to wpt

asamuzaK:
Original test in style.js seems to be incorrect.


domenic (approved):
I love moving things out of to-port-to-wpts!! And making them more correct while we do is even better!!

Issue #3906, opened 2025-07-15
:scope selector don't work after used once on a query anymore


mmis1000:
### Node.js version

v20.12.0

### jsdom version

27.0.0-beta.2

### Minimal reproduction case

```js
import { JSDOM } from "jsdom";
import assert from "assert";

const data = `<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>PayloadType</key>
    <string>Configuration</string>
    <key>PayloadContent</key>
    <array>
        <dict>
            <key>FullScreen</key>
            <true/>
        </dict>
    </array>
</dict>
</plist>
`;

const dom = new JSDOM(data, {
  contentType: "text/xml",
});

const xmlDocument = dom.window.document;

console.log(xmlDocument.querySelectorAll(":root "));

const rootCollection = xmlDocument.querySelector(":root > dict");

console.log(rootCollection);

const keysInRootCollection = rootCollection.querySelectorAll(":scope > key");

console.log(keysInRootCollection.length);

const subCollection = rootCollection.children[3].children[0];
const keysInSubCollection = subCollection.querySelectorAll(":scope > key");

console.log(keysInSubCollection.length);
const clonedSubCollection = subCollection.cloneNode(true);
const keysInClonedSubCollection =
  clonedSubCollection.querySelectorAll(":scope > key");

console.log(keysInClonedSubCollection.length);

assert(subCollection.outerHTML === clonedSubCollection.outerHTML);
assert(keysInSubCollection.length === keysInClonedSubCollection.length);
```

### How does similar code behave in browsers?

The exact same code will pass the assert in browser environments

change the

```js
const dom = new JSDOM(data, {
  contentType: "text/xml",
});

const xmlDocument = dom.window.document;
```

to

```js
const assert = (v, m = "assert failure") => {
  if (!v) {
    throw new Error(m);
  }
};
const parser = new DOMParser();
const xmlDocument = parser.parseFromString(data, "application/xml");
```

to see the result in browser terminal

[code sandbox playground link](https://codesandbox.io/p/devbox/r5kc53)

## Expected

The `:scope > key` in child collection selects the child `<key>` element and hehaves exactly like the cloned node

## Actual

The `:scope > key` in child collection selects nothing

## Note

[...]

Issue #3895, opened 2025-06-26
27.0.0-beta.2: window.getComputedStyle(q).getPropertyValue("display") of <q> is an empty string


jdufresne:
### Node.js version

v22.16.0

### jsdom version

27.0.0-beta.2

### Minimal reproduction case

```js
import { JSDOM } from "jsdom";
const { window } = new JSDOM();
const { document } = window;
const q = document.createElement("q");
document.body.appendChild(q);
console.log(window.getComputedStyle(q).getPropertyValue("display"));
```

### How does similar code behave in browsers?

Browser: outputs "inline"
JSDom v26.1.0: outputs "inline"
JSDom 27.0.0-beta.2: outputs empty string

### What is the problem?

Like JSDom v26.1.0 and the browser, I expect this code to output "inline".

The 3rd party module dom-accessibility-api uses the computed style to calculate the element's accessible name. See:
https://github.com/eps1lon/dom-accessibility-api/blob/27845ffc0513a9c5e20236d3c3e6aa7ffdf04674/sources/accessible-name-and-description.ts#L413

The change in JSDom behavior when combined with this third party module resulted in test suite failures in my project.

asamuzaK:
We had q specified
https://github.com/jsdom/jsdom/blob/26.1.0/lib/jsdom/browser/default-stylesheet.js#L123

Not  now, including other inline elements
https://github.com/jsdom/jsdom/blob/main/lib/jsdom/browser/default-stylesheet.css

```
* { display: inline; }
```
before 
https://github.com/jsdom/jsdom/blob/main/lib/jsdom/browser/default-stylesheet.css#L9 should fix this?

domenic:
Hmm. I suspect the correct fix is instead to add some sort of default value for `display` as part of style computation. I don't think it should ever return the empty string.

domenic:
I forgot to say, thank you so much for testing with the beta! We will fix this before doing the final release.

asamuzaK:
> Hmm. I suspect the correct fix is instead to add some sort of default value for `display` as part of style computation. I don't think it should ever return the empty string.

I don't disagree with your point, but looking at the spec of `display`, it seems like it will require a somewhat complicated implementation.

I think it can be done with cssstyle, but it will take time.

Until then, how about applying
```
* { display: inline; }
```
as a temporary workaround?

domenic:
[...]

PR #3914, opened 2025-07-28
Address regressions in v27 beta

asamuzaK:
* Confirmed fix https://github.com/jsdom/cssstyle/issues/209
* Confirmed fix https://github.com/jsdom/cssstyle/pull/232
* Fixes https://github.com/jsdom/jsdom/issues/3895
* Fixes https://github.com/jsdom/jsdom/issues/3906
* Correctly support CSS-wide `initial`, `inherit`, `unset` keywords.
  NOTE: `revert` and `revert-layer` need extra work to getComputedStyle()


asamuzaK:
@domenic 

It appears that cssstyle v5.2.0 is published with file paths using Windows delimiters in `lib/generated/properties.js`

cssstyle v5.2.0:
```js
var external_dependency_strings_1 = require("../utils\\strings.js");
```

cssstyle v5.1.0:
```js
var external_dependency_strings_1 = require("../utils/strings.js");
```

Can you fix it?

asamuzaK:
Fixed generator script.
See https://github.com/jsdom/cssstyle/pull/238

domenic:
Published; sorry about that!

Will #3895 be part of this PR?

asamuzaK:
> Published; sorry about that!

Thanks!

> Will #3895 be part of this PR?

Yes, and maybe some others  too.

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -225,37 +232,56 @@ function getSpecifiedValue(element, property) {
...
+    case "unset": {
+      if (inherited && element.parentElement !== null) {
+        specifiedValue = getComputedValue(element.parentElement, property);
+      } else {
+        specifiedValue = initialValue;
+      }
+      break;
+    }
+    case "revert":
+    case "revert-layer":
```
domenic: Can we add clarifying comments for these three cases? My best guess is that revert/revert-layer are supposed to be future TODOs, and the default case is that the specified value is suitable to return as-is.

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -225,37 +232,56 @@ function getSpecifiedValue(element, property) {
...
 }
 
 // https://drafts.csswg.org/css-cascade-4/#computed-value
 function getComputedValue(element, property) {
-  const { computedValue, initial } = exports.propertiesWithResolvedValueImplemented[property];
+  const { computedValue, inherited, initial } = exports.propertiesWithResolvedValueImplemented[property];
+  const initialValue = asciiLowercase(initial);
+  let specifiedValue = getSpecifiedValue(element, property);
+  // https://drafts.csswg.org/css-cascade/#defaulting-keywords
+  switch (specifiedValue) {
```
domenic, line 243: Is it possible for getSpecifiedValue to return `"INITIAL"`?

Overall it feels like we are ASCII lowercasing many times in this file and I wonder if we could do it earlier.

test/web-platform-tests/to-run.yaml
```diff
@@ -215,17 +218,14 @@ animations/display-interpolation.tentative.html: [fail, Depends on CSS]
...
 display-math-on-non-mathml-elements.html: [fail, Unknown]
 display-math-on-pseudo-elements-001.html: [fail, Unknown]
 display-with-float-dynamic.html: [fail, Unknown]
-inheritance.html: [fail, Unknown]
 parsing/display-computed.html: [fail, Depends on CSS]
-parsing/display-invalid.html: [fail, Unknown]
-parsing/display-valid.html: [fail, Unknown]
-parsing/tentative/display-computed.html: [fail, Unknown]
+parsing/tentative/display-computed.html: [fail, Unsupported value]
+parsing/tentative/display-valid.html: [fail, Unsupported value]
```
domenic: Can you explain this a bit more? "Unsupported value" isn't that much more helpful than "Unknown"; which value is unsupported, by what?

test/web-platform-tests/to-upstream/css/css-display/display-initial.html
```diff
@@ -0,0 +1,136 @@
...
+
+test(() => {
+  assert_equals(div.style.getPropertyValue("display"), "", "Specified value of div");
+  assert_equals(window.getComputedStyle(div).getPropertyValue("display"), "block", "Computed value of div");
+
+  assert_equals(q.style.getPropertyValue("display"), "", "Specified value of q");
+  assert_equals(window.getComputedStyle(q).getPropertyValue("display"), "inline", "Computed value of q");
+
+  assert_equals(q2.style.getPropertyValue("display"), "", "Specified value of q2");
+  assert_equals(window.getComputedStyle(q2).getPropertyValue("display"), "inline-block", "Computed value of q2");
```
domenic, line 34: Wow, this is surprising to me, but I guess it is correct. https://drafts.csswg.org/css-display/#the-display-properties says

> a pair of keywords...

and https://drafts.csswg.org/css-display/#legacy-display says inline-block "Computes to `inline flow-root`."

but then I guess https://drafts.csswg.org/cssom/#serializing-css-values has some super-vague language saying

> If component values can be omitted or replaced with a shorter representation without changing the meaning of the value, omit/replace them.
>
> If either of the above syntactic translations would be less backwards-compatible, do not perform them.

CSS specs are so terrible...

lib/jsdom/living/helpers/style-rules.js
asamuzaK: Done

test/web-platform-tests/to-run.yaml
asamuzaK: How about this?

lib/jsdom/living/helpers/style-rules.js
asamuzaK: After receiving your comment, I realized that uppercase and lowercase letters are already handled by cssstyle, there is no need to convert to lowercase on jsdom side.

asamuzaK: The only thing I had to do was change `CanvasText` to lowercase, but since it's an internal object, this change shouldn't cause any problems.
By the way, in the future, I plan to import supported property information from cssstyle instead of `propertiesWithResolvedValueImplemented()`.

asamuzaK: > After receiving your comment, I realized that uppercase and lowercase letters are already handled by cssstyle, there is no need to convert to lowercase on jsdom side.

Found that there is still one place where we need to change the value to lowercase.

asamuzaK:
https://github.com/jsdom/jsdom/actions/runs/17174516163/job/48728600145?pr=3914#step:6:9000
This failure appears to be unrelated to this PR.

domenic (approved):

PR #3910, opened 2025-07-19
swap rweb-cssom to @acemir/cssom for better css nesting support

acemir:
The updated dependency package enhances the parser's ability to correctly process complex CSS structures, including nested selectors, nested declarations, layer statements and at-rules validations.

### Introducing CSS Nesting Module support
- Added support for [CSSNestedDeclarations](https://drafts.csswg.org/css-nesting-1/#cssnesteddeclarations)
- [CSSStyleRule](https://drafts.csswg.org/cssom/#the-cssstylerule-interface) now inherits properties from [CSSGroupingRule](https://drafts.csswg.org/cssom/#the-cssgroupingrule-interface)

### Improved At-Rule Validation
- Refactored at-rule validation logic ensuring that invalid or malformed at-rules are ignored more reliably.
- Improved handling of edge cases, such as unbalanced braces and complex at-rule conditions.
- Enhanced handling of at-rules that cannot be nested.

### Improvements in CSS Cascading
- Added support for [CSSLayerStatement](https://drafts.csswg.org/css-cascade-5/#csslayerstatementrule)
- Improves [CSSLayerBlockRule ](https://drafts.csswg.org/css-cascade-5/#csslayerblockrule)interface and layer name validation
- Added support for layerName and supportsText to [CSSImportRule](https://drafts.csswg.org/cssom/#the-cssimportrule-interface)

PR #3916, opened 2025-08-15
Add some tests

asamuzaK:
* Add regression tests for https://github.com/jsdom/cssstyle/issues/228
* Rewrite part of to-port-to-wpts test in wpt format
* Updated to-run.yaml to check errors in detail for https://github.com/jsdom/cssstyle/issues/228
* Enabled `css/css-cascading` and `css/css-nesting` directories in to-run.yaml for https://github.com/jsdom/jsdom/pull/3910


domenic:
I'm back from vacation! Can you explain why `border-style-var` would be tentative?

asamuzaK:
Because https://bugzilla.mozilla.org/show_bug.cgi?id=1982973 is still open.
Should I remove it from PR?

domenic:
Well, a test should only be marked tentative if it

> makes assertions not yet required by any specification, or in contradiction to some specification. 

[source](https://web-platform-tests.org/writing-tests/file-names.html)

They should not be marked tentative just because a browser has a bug.

So, do you know if the test matches the specification?

asamuzaK:
I did not use `tentative` with that meaning.
Removed it from the PR for now.

domenic:
Oh, but it's sad to lose coverage. Are you deleting it because you're unsure what the spec says? If you think it matches the spec, then I think it would be good to include...

asamuzaK:
I'm not sure about this part.

``` js
test(() => {
  div.style.borderTop = "none";
  div.style.borderStyle = "var(--foo)";
  assert_equals(
    div.style.cssText,
    "border-top-width: medium; border-style: var(--foo); border-top-color: currentcolor;",
    "cssText"
  );
}, "Set border-top as none and change border-style to var()");
```

Anyway, Mozilla will probably upload their test for https://bugzilla.mozilla.org/show_bug.cgi?id=1982973, and if I think the test is insufficient, then I'll write a follow-up test.

domenic (approved):
Safari fails some of border-style-none, but in different ways than Chrome. Very interesting that there are still browser interop bugs in this area. I look forward to upstreaming these tests and filing the bugs on browsers!

PR #3920, opened 2025-08-27
Fix #3882 add search tag and ensure all non-inherited tags are tested

### LINKED ISSUE #3882: Search tag is not recognised
### Node.js version

v20.10.0

### jsdom version

jsdom@26.1.0

### Minimal reproduction case

```js

const options = {
};
const dom = new JSDOM(`
  <search>whatever</search>
`, options);

console.log(Object.prototype.toString.call(dom.window.document.querySelector('search'))); // [object HTMLUnknownElement]
```

[object HTMLElement]

### What is the problem?

jsdom doesn't recognise the `search` tag, creating an element based on `HTMLUnknownElement` instead of `HTMLElement`.

See #3881 for fix

ashsearle:
* Add support for the `<search>` tag (fixes #3882)
* Align tests with `create-element.js` to ensure all non-inherited tags are tested (this required adding a `<body>` wrapper to stop the test failing for the `noscript` tag.)


domenic:
See CONTRIBUTING.md; we do not add new tests to to-port-to-wpts.

Issue #3931, opened 2025-09-15
Possible regressions regarding adding/removing nodes and user events


asamuzaK:
I'm not sure yet, but have v27.0.0 made any changes related to adding and/or removing DOM nodes, or user-oriented events like mouse clicks?

After upgrading jsdom to 27.0.0 and testing domSelector, tests that previously passed now hang.
https://github.com/asamuzaK/domSelector/actions/runs/17736554937/job/50406715638#step:6:2309
`css/selectors/invalidation/is-where-pseudo-containing-hard-pseudo.html` is a modified wpt test and it adds/removes nodes many times.
In jsdom, original wpt test of this is skipped as fail-slow.

https://github.com/jsdom/jsdom/issues/3930
In this issue too, when I look at the test source, it seems that the change event and click event are triggered before the error occurs.
https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/6e28800e512e088cd7120c5284e97a351d868e36/app/components/formElements/autoSuggestInput/__test__/AutoSuggestInput.test.tsx#L172

I will write up an issue once I've narrowed down the cause a bit more, but for now I'll leave this here as a note.

asamuzaK:
Note to myself:
Add --parallel to mocha fixed domSelector tests.

cooperwalbrun:
I also suspect a change in DOM manipulation behavior that seems to impact [@testing-library/svelte](https://www.npmjs.com/package/@testing-library/svelte)'s [rerender()](https://testing-library.com/docs/svelte-testing-library/api#rerender) functionality. In one of my tests I render the DOM via [render()](https://testing-library.com/docs/svelte-testing-library/api#render) and then perform assertions/rerenders on it via code that looks something like this:
```typescript
const dom = render(Component, ...);
expect(dom.container.querySelector(...)).toBeInTheDocument();
await dom.rerender(...);
expect(dom.container.querySelector(...)).not.toBeInTheDocument();
```
The `rerender()` call replaces the node being checked in the first `expect()` with a node that has different classes, so the node is expected to be absent after the rerender. However, the second `expect()` assertion fails when I upgrade `jsdom` to version `27.0.0` with no other code/dependency changes, so it seems that the node manipulation/replacement functionality of `jsdom` may not be working as it did previously.

domenic:
> made any changes related to adding and/or removing DOM nodes, or user-oriented events like mouse clicks?

[...]

PR #3932, opened 2025-09-16
Use new DOMSelector instance for internal querySelector()

### LINKED ISSUE #3928: 27.0.0 broke multi-level query selecting in vitest/testing-library
### Node.js version

22.15.1

### jsdom version

27.0.0

### What is the problem?

As per the issue template, I expect this will be closed, and I wouldn't want to waste your time, but I thought you might want to at least know that upgrading from 26.1.0 to 27.0.0 introduced a major undocumented change/regression that broke tests that rely on nested query selection. It seems that `document.querySelector` now always operates from root now, so nested calls ignore all previous ones.

### Minimal reproduction case

```html
<div>
  <span>Actions</span>
  <ul>
    <li>
      <a href="/">
        <span>Link</span>
      </a>
    </li>
  </ul>
</div>
```

```js
const anchor = document.querySelector('a')
const span = anchor.querySelector('span')
expect(span.innerHTML).toBe('Link')
```

Fails, `span.innerHTML` == 'Actions'

It scopes queries based on previous selectors

asamuzaK:
Fix #3928
See also https://github.com/jsdom/jsdom/issues/3931#issuecomment-3299518202

### DIFF
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -226,7 +226,10 @@ class DocumentImpl extends NodeImpl {
       return this._baseURLCache;
     }
 
-    const firstBase = this.querySelector("base[href]");
+    const firstBase = new DOMSelector(this._globalObject, this._ownerDocument, {
+      domSymbolTree,
+      idlUtils
+    }).querySelector("base[href]", this);
 
     this._baseURLCache = firstBase === null ?
       this._fallbackBaseURL() :


domenic:
This does not seem like an acceptable fix. It needs to be acceptable to use querySelector() internally without causing bugs. Why is querySelector() stateful in this way? That isn't how the spec says to behave.

asamuzaK:
Why?

The results differ depending on whether you use console.log() or not is probably because internal and external processing is not properly separated.

Also, while the spec says "in tree order", I don't think this means you should use querySelector().

domenic:
If querySelector() is fundamentally unstable for usage in this way, it's too buggy to be used in jsdom. Users could run into the same bug without jsdom using it. As far as we can tell, the bug is that if you use querySelector() twice, it returns different results than using it once.

It should be resilient to any usage, just like the web platform's version of it is.

querySelector() is a simple and easy way to implement "in tree order", so if our querySelector() is unable to provide that functionality, then it's not a very good querySelector().

asamuzaK:
To be honest, I'm not convinced, but I'll look into it further.
I'll close this PR for now.

PR #3933, opened 2025-09-17
Add cssstyle regression tests for CSS flex shorthand

matmannion:
As discussed in https://github.com/jsdom/cssstyle/pull/245, adds a regression test for CSS `flex` shorthand in cssstyle


domenic (approved):
Great stuff, thank you!

PR #3934, opened 2025-09-17
Update @asamuzakjp/dom-selector

### LINKED ISSUE #3928: 27.0.0 broke multi-level query selecting in vitest/testing-library

asamuzaK:
Fix #3928 
See also #3931

It's not testable in wpt format; please run:
``` console
>node test/api/fixtures/console.js
```

Results:
jsdom + domSelector@6.5.4:
``` console
>node test/api/fixtures/console.js
Action
```

jsdom + domSelector@6.5.5:
``` console
>node test/api/fixtures/console.js
Link
```


domenic:
Nice, so glad you found the cause. Is this testable by adding a test to https://github.com/jsdom/jsdom/blob/main/test/api/from-outside.js ?

asamuzaK:
Fail log with jsdom  + domSelector@6.5.4:
``` console
1) Test cases only possible to test from the outside
       should switch node context on querySelector() (GH-388):

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

'Actions' !== 'Link'

      + expected - actual

      -Actions
      +Link
```

conbrad (approved):

domenic (approved):

PR #3938, opened 2025-09-30
Update @asamuzakjp/dom-selector

asamuzaK:
See #3931

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "27.0.0",
       "license": "MIT",
       "dependencies": {
-        "@asamuzakjp/dom-selector": "^6.5.5",
+        "@asamuzakjp/dom-selector": "^6.5.7",
         "cssstyle": "^5.3.0",
         "data-urls": "^6.0.0",
         "decimal.js": "^10.6.0",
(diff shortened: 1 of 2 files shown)


birkskyum:
Your test PR passes for both jsdom 26+27 because it's checking `container.classList.contains('is-red') === false/true`, which I see locally work as expected as well.

It should instead be checking on `document.querySelectorAll('.is-red').length === 0/1`, which I anticipate will make it break for jsdom 27.

test/web-platform-tests/to-upstream/dom/events/element-click.html
```diff
@@ -1,65 +1,51 @@
...
-  });
-
-  const target = document.querySelector("#target");
-  assert_equals(target.classList.contains("foo"), false, "initial");
+  const targetQuerySelectorAll = document.querySelectorAll(".foo");
+  assert_equals(targetQuerySelectorAll.length, 0, "initial");
 
   button.click();
-  assert_equals(target.classList.contains("foo"), true, "1st clicked");
+  assert_equals(targetQuerySelectorAll.length, 0, "1st click");
```
birkskyum, line 32: Should this have length == 1?

asamuzaK: querySelectorAll returns static nodeList, so it should be 0

domenic:
Did this get closed accidentally?

asamuzaK:
Yes. Reopened

domenic (approved):
Confirmed test fails without the new version and passes with. Thank you!

PR #3954, opened 2025-10-17
Call domSelector.clear() on class change

### LINKED ISSUE #3950: Regression: JSDOM-27.0.0 (and recent dependency update) not selecting by class
### Node.js version

v22.20.0

### jsdom version

27.0.0 (and the recent dependency update)

### Minimal reproduction case

```js
const {JSDOM} = require("jsdom");
const svgns = 'http://www.w3.org/2000/svg';
const dom = new JSDOM(
  '<!doctype html>' +
    '<html>' +
    '<head>' +
    '<title>JSDOM bug test</title>' +
    '</head>' +
    '<body>' +
    '<svg>' +
    '<g class="a" />' +
    '</svg>' +
    '<div class="one">' +
    '</div>' +
    '</body' +
    '</html>',
);
const window = dom.window;
const svg = document.querySelector('svg');
let g = document.querySelector('g.a');
console.log(`g === null:${g === null}`);
console.log(`g.getAttributeNS(null, 'class'):${g.getAttributeNS(null, 'class')}`);

g.setAttributeNS(null, 'class', 'b');
g = null;
g = document.querySelector('g.a');
console.log(`g === null:${g === null}`);
if (g !== null) {
  console.log(`g.getAttributeNS(null, 'class'):${g.getAttributeNS(null, 'class')}`);
} else {
  console.log("selector 'g.a' did not find 'g.b'");
}

let d = document.querySelector('div.one');
console.log(`d === null:${d === null}`);
console.log(`d.getAttributeNS(null, 'class'):${d.getAttributeNS(null, 'class')}`);
d.setAttributeNS(null, 'class', 'two');
d = null;
d = document.querySelector('div.one');
console.log(`d === null:${d === null}`);
if (d !== null) {
  console.log(`d.getAttributeNS(null, 'class'):${d.getAttributeNS(null, 'class')}`);
} else {
  console.log("selector 'div.one' did not find 'div.two'");
}
```

In browsers, and in JSDOM 26.1.0, the output of the above is:

jsdom-26.1.0
g === null:false
g.getAttributeNS(null, 'class'):a
g === null:true
selector 'g.a' did not find 'g.b'
d === null:false
d.getAttributeNS(null, 'class'):one
d === null:true
selector 'div.one' did not find 'div.two'

jsdom-27.0.0
g === null:false
g.getAttributeNS(null, 'class'):a
g === null:false
g.getAttributeNS(null, 'class'):b
d === null:false
d.getAttributeNS(null, 'class'):one
d === null:false
d.getAttributeNS(null, 'class'):two

jsdom-27.0.0 with dependency update
g === null:false
g.getAttributeNS(null, 'class'):a
g === null:false
g.getAttributeNS(null, 'class'):b
d === null:false
d.getAttributeNS(null, 'class'):one
d === null:false
d.getAttributeNS(null, 'class'):two

### What is the problem?

The regression is a selector like 'g.a', or 'div.one' is selecting items with the same nodeName and a different class.

It should (and previously did) select a g element with class="a" and NOT select a g element with class="b" when given selector 'g.a'.
Similarly it should (and did) select a div element with class="one" and NOT select a div with class="two" when given selector 'div.one'.

### LINKED ISSUE #3953: JSDOM 27 regression - querySelector() returns stale results after classList.remove() with compound selectors
### Node.js version

24.5.0

### jsdom version

27.0.0

### Minimal reproduction case

```js

const dom = new JSDOM(`<!DOCTYPE html><html><body></body></html>`);
const {document} = dom.window;

// Create element with two classes
div.className = 'class-one class-two';

// Verify it can be found
console.log(document.querySelector('.class-one.class-two')); // Element found ✓

// Remove one class
div.classList.remove('class-two');

// Check the class was removed
console.log(div.className);                        // "class-one" ✓
console.log(div.classList.contains('class-two')); // false ✓

// BUG: querySelector still finds it with both classes!
console.log(document.querySelector('.class-one.class-two')); // STILL RETURNS THE ELEMENT ✗

// However, querySelectorAll works correctly
console.log(document.querySelectorAll('.class-one.class-two').length); // 0 ✓
```

**Expected output (Works with JSDOM 26):**
```

HTMLDivElement {}
class-one
false
null
0

```

**Actual output with JSDOM 27**

```sh

HTMLDivElement {
  Symbol(SameObject caches): [Object: null prototype] { classList: DOMTokenList {} }
}
class-one
false
HTMLDivElement {
  Symbol(SameObject caches): [Object: null prototype] { classList: DOMTokenList {} }
}
0
```

### What is the problem?

**Expected Behavior (Works with JSDOM 26):**

  After classList.remove('class-two') is called:
  - querySelector('.class-one.class-two') should return null
  - The element no longer has both classes, so the compound selector shouldn't match

  **Actual Behavior with JSDOM 27**

  After classList.remove('class-two') is called:
  - div.className correctly shows "class-one"
  - div.classList.contains('class-two') correctly returns false
  - BUT querySelector('.class-one.class-two') still returns the element
  - querySelectorAll('.class-one.class-two') correctly returns empty array

  **Why This is a Bug**

  There's an internal inconsistency in JSDOM:
  - The element's classList has been updated correctly
  - Single-class selectors work: .class-one finds it, .class-two returns null
  - But the compound selector .class-one.class-two is cached/stale
  - querySelector and querySelectorAll return different results for the same selector!

asamuzaK:
Fix #3950, fix #3953

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -119,7 +119,10 @@ class ElementImpl extends NodeImpl {
     // update classList
     if (name === "class" && this._classList !== undefined) {
       this._classList.attrModified();
-      // clear domSelector cached results on class change
+    }
+
+    // clear domSelector cached results on class change
+    if (name === "class") {
       this._ownerDocument._domSelector.clear();
     }
 


domenic:
Is it only necessary to do this on class changes? Or would it be good to do this on any modifications?

asamuzaK:
It might be a good idea to clear the cache if there are any attribute changes that affect the layout, but I haven't researched exactly which attributes require this.
Specifically, `hidden` and `style`.
Let's add some tests to verify this.

domenic:
Sounds like a good idea. FYI, jsdom has a number of hooks for its own internal cache clearing. The most general is [`_modified()`](https://github.com/jsdom/jsdom/blob/894151baf0603c5d2203261ce9f4748295e6681c/lib/jsdom/living/nodes/Node-impl.js#L215), which currently does some stuff related to style rule clearing already.

asamuzaK:
Added a test.
It looks like we only need to do this on `class` changes.

By the way, I've checked `invalidateStyleCache()` and don't think we need to call `domSelector.clear()` there.

lib/jsdom/living/nodes/Element-impl.js
```diff
@@ -119,6 +119,8 @@ class ElementImpl extends NodeImpl {
     // update classList
     if (name === "class" && this._classList !== undefined) {
       this._classList.attrModified();
+      // clear domSelector cached results on class change
+      this._ownerDocument._domSelector.clear();
```
domenic, line 126: Let's also do this even if `this._classList === undefined`. I guess right now this is working, because domSelector is accessing classList. But I don't want to depend on that implementation detail of domSelector. (domSelector would might be faster if it used className + regexps in the future).

asamuzaK: Done.

PR #3957, opened 2025-10-19
Prepare CSSStyleDeclaration-impl.js and CSSStyleProperties-impl.js

### LINKED ISSUE #113: Add mechanism to handle [CEReactions] extended IDL attributes
Custom elements lifecycle hooks invocation depend on `[CEReactions]` extended IDL attributes. The [`CSSStyleDeclaration`](https://drafts.csswg.org/cssom/#cssstyledeclaration) interface has multiple occurrences of this extended attribute:
- `cssText`
- `setProperty`
- `removeProperty`
- `cssFloat`
- all the camel case, dashed case and webkit cases properties.

Adding support for such hooks would allow jsdom to invoke the `attributeChangedCallback` whenever the `style` attribute changes.

### LINKED ISSUE #129: Value as is not converted to String
I'm not sure that the following behavior tested in Chrome and Firefox is described in the [specification](https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-setproperty)...

```
class Prop {
  toString() {
    return 'red'
  }
}

target.style.setProperty('background-color', new Prop)
// Or: target.style.backgroundColor = new Prop()

console.log(target.style.backgroundColor) // 'red'
```

... meaning the value should be converted to String, but with `cssstyle` it returns `''`, ie. it results to an invalid value.

### LINKED ISSUE #247: Regression: Unable to change borderWidth when border contains a CSS variable
Reproducible test case:

```js
it("should allow changing a single property on a border, when border contains a css variable", () => {
  const style = new CSSStyleDeclaration();
  style.border = "0.1rem solid var(--my-color-value)";
  style.borderWidth = "0.2rem";
  assert.strictEqual(style.borderWidth, "0.2rem");
  assert.strictEqual(style.border, "0.2rem solid var(--my-color-value)");
});
```

It fails on the `style.borderWidth = "0.2rem";` line with:

```
TypeError [Error]: Cannot create property 'border-width' on string '0.1rem solid var(--my-color-value)'
```

This doesn't crash in cssstyle@v4.6.0 (though technically the individual property didn't change either, but at least it didn't crash).

### LINKED ISSUE #3169: Cannot set style.background with "/" in it
### Basic info:

- **Node.js version:** 15.12.0
- **jsdom version:** 16.5.2

### Minimal reproduction case

```js
const dom = new JSDOM(`<!DOCTYPE html><img>`);
const img = dom.window.document.querySelector("img");

img.style.background = 'url("foo.jpg") center/20px no-repeat';
console.log(img.style.background); // ""

img.style.background = 'url("foo.jpg") center no-repeat';
console.log(img.style.background); // 'url("foo.jpg") center no-repeat'
```
- Here's a RunKit: https://runkit.com/embed/4vwdswg1owxi

- On FF 87, it sets to: `"rgba(0, 0, 0, 0) url(\"foo.jpg\") no-repeat scroll center center / 20px"`
- On Chrome, it sets to `"url(\"foo.jpg\") center center / 20px no-repeat"`

Here's a jsbin: https://jsbin.com/dokuhogabi/edit?html,js,console

### LINKED ISSUE #3339: getComputedStyle doesn't return a default opacity value "1", but empty string
### Basic info:

- **Node.js version: v16.13.2
- **jsdom version: 19.0.0

### Minimal reproduction case

```js
const options = {
  contentType: 'text/html'
};
const dom = new JSDOM('<html><head></head><body></body></html>', options);
const input = dom.window.document.createElement('input');
input.type = 'text';
dom.window.document.body.append(input);
const styles = dom.window.getComputedStyle(input);
console.log(`opacity=${styles.getPropertyValue('opacity')}
display=${styles.getPropertyValue('display')}`);
```

actual output
```
opacity=
display=inline-block
```

expected output
```
opacity=1
display=inline-block
```

in browser (latest Chrome and Firefox), if an element doesn't specify the opacity, `getPropertyValue('opacity')` would return `'1'`
While in current jsdom, it return empty string

### LINKED ISSUE #3663: Inline background with border-box value will be accidentally stripped
### Basic info:

- **Node.js version:** v18.17.1
- **jsdom version:** v23.2.0

### Minimal reproduction case

```js
const { JSDOM } = jsdom;

const dom = new JSDOM(`<div id="test" style="background: border-box;"></div>`)
const { document } = dom.window
const value = document.querySelector('#test').style.background
console.log(value); // value is '', which is expected to be 'border-box'
```

### LINKED ISSUE #3939: Regression: for `el.style.transform = "translate((10%), 10%)"` whitespace is removed (worked in v26, fails in v27)
### Node.js version

v22

### jsdom version

v27

### Minimal reproduction case

```js

const dom = new JSDOM(``, {});

const p = document.createElement("p");

const transform = "translate(calc(10%), 10%)";
p.style.transform = transform;
console.assert(p.style.transform === transform, "p.style.transform should be\n", transform, "\nbut is\n", p.style.transform);
```

In Chrome this assertion just passes.

### What is the problem?

The assertion fails with:

```
Assertion failed: p.style.transform should be
 translate(calc(10%), 10%)
but is
 translate(calc(10%),10%)
```
In v26 there were no semantic checks, but at least my tests did not fail.
If I update my tests, I need to update them again as soon as jsdom aligns with the browser again.

### LINKED ISSUE #3940: CSSStyleDeclaration `getPropertyPriority` method return nothing
### Node.js version

22.13.0

### jsdom version

27.0.0

### Minimal reproduction case

```js
import { JSDOM } from 'jsdom' 

const dom = new JSDOM(
  `
  <!DOCTYPE html>
  <html lang="en">
  <head>
  </head>
    <body>
    <p class='element-with-inline-style' style='word-spacing: 1px !important;'>Inline Style Test</p>
    </body>
  </html>
`
);

console.log(dom.window.document.querySelector('.element-with-inline-style').style.getPropertyPriority('word-spacing'));
// empty
```

Property priority is returned based on inline styles

### What is the problem?

It worked as expected on 26.1.0

### LINKED ISSUE #3941: `getComputedStyle` can calculate inherited styles based on document context, but cannot beyond shadow dom boundary
### Node.js version

22.13.0

### jsdom version

27.0.0

### Minimal reproduction case

```js
import { JSDOM } from 'jsdom' 

const dom = new JSDOM(
  `
  <!DOCTYPE html>
  <html lang="en">
  <head>
  </head>
    <body>
        <div id="shadow-root" style="visibility:hidden;"></div>
    </body>
  </html>
`
);

const window = dom.window;

document.addEventListener('load', async () => {
    const shadowRoot = document.querySelector('#shadow-root').attachShadow({mode: 'open'});
    shadowRoot.innerHTML = 
        `<div class="container">
            <span class="shadow-element-with-inline-style" style="visibility:inherit;"></span>
        </div>`

    const innerEl = document.querySelector('#shadow-root').shadowRoot.querySelector('.shadow-element-with-inline-style');

    console.log(window.getComputedStyle(innerEl).visibility); // visible

    shadowRoot.innerHTML = 
        `<div class="container" style="visibility:hidden;">
            <span class="shadow-element-with-inline-style" style="visibility:inherit;"></span>
        </div>`

    const innerEl2 = document.querySelector('#shadow-root').shadowRoot.querySelector('.shadow-element-with-inline-style');
    console.log(window.getComputedStyle(innerEl2).visibility); // hidden
});
```

`getComputedStyle` API returns inherited value from all container ancestor types

### What is the problem?

JSDom 26.1.0 returned directly set neutral style values (unset, inherit) and custom recursive implementation at least could make assumptions on inherited style values. Is it expected limitation that style dynamic retrieval should work correctly only in scope of document?

### LINKED ISSUE #3972: `getComputedStyle` of a `border-color` repeats the color thrice
### Node.js version

22.15.0

### jsdom version

27.0.1

### Minimal reproduction case

```js
const { JSDOM, VirtualConsole } = require("jsdom");

const { window } = new JSDOM(undefined, {
  virtualConsole: new VirtualConsole(),
});
const { document } = window;

div.style.setProperty("border-color", "gray");

console.log(window.getComputedStyle(div).getPropertyValue("border-color"));
```

https://jsfiddle.net/s5dL1n93/

Chrome and Firefox outputs `"rgb(128, 128, 128)"`.

### What is the problem?

jsdom outputs `"rgb(128, 128, 128) rgb(128, 128, 128) rgb(128, 128, 128)"` instead.

asamuzaK:
WIP
Depends on
* https://github.com/jsdom/cssstyle/pull/244
* https://github.com/acemir/CSSOM/pull/58

Fixes
* Fix https://github.com/jsdom/cssstyle/issues/113
* Fix https://github.com/jsdom/cssstyle/issues/129
* Fix https://github.com/jsdom/cssstyle/issues/247
* Fix https://github.com/jsdom/jsdom/issues/3169
* Fix https://github.com/jsdom/jsdom/issues/3339
* Fix https://github.com/jsdom/jsdom/issues/3663
* Fix https://github.com/jsdom/jsdom/issues/3939
* Fix https://github.com/jsdom/jsdom/issues/3940
* Fix https://github.com/jsdom/jsdom/issues/3941
* Fix https://github.com/jsdom/jsdom/issues/3972


test/web-platform-tests/to-run.yaml
```diff
@@ -201,6 +196,13 @@ system-color-compute.html:
   "Inherited system color keyword is observable on caret-color": [fail, computedStyleMap not implemented]
   "Inherited system color keyword is observable on fill": [fail, computedStyleMap not implemented]
   "Inherited system color keyword is observable on stroke": [fail, computedStyleMap not implemented]
+  "System color doesn't compute to itself on box-shadow": [fail, Need cssstyle fix]
```
domenic, line 177: Why do these regress?

test/web-platform-tests/to-upstream-expectations.yaml
```diff
@@ -3,21 +3,14 @@ css/css-conditional/container-queries-basic.html: [fail, https://github.com/jsdo
 css/css-display/display-initial.html:
   "Revert value of display": [fail, Revert keyword is not supported]
   "Revert-layer value of display": [fail, Revert-layer keyword is not supported]
+css/cssom/CSSStyleProperties.html:
+  "CSSStyleRule.style": [fail, Need rrweb-cssom fix]
```
domenic: Why does this regress? Does https://github.com/jsdom/jsdom/pull/3910 help?

test/web-platform-tests/to-upstream/css/cssom/inline-style.html
```diff
@@ -0,0 +1,51 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Inline style attribute</title>
+<link rel="help" href="https://drafts.csswg.org/cssom/#dom-elementcssinlinestyle-style">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<!-- Rewritten from https://github.com/jsdom/jsdom/blob/main/test/to-port-to-wpts/level2/style.js -->
```
domenic: This comment can be removed since it is not relevant for the future or for WPT maintainers. But it is helpful for code review, so thank you for writing it. In the future it can go in the commit message :)

test/web-platform-tests/to-upstream/css/cssom/CSSStyleProperties.html
```diff
@@ -0,0 +1,38 @@
...
+  assert_equals(typeof window.CSSStyleDeclaration, "function", "declaration");
+  assert_equals(typeof window.CSSStyleProperties, "function", "properties");
+}, "Sanity check");
+
+test(() => {
+  const div = document.createElement("div");
+  const style = window.getComputedStyle(div);
+  assert_equals(style instanceof window.CSSStyleDeclaration, true, "declaration");
+  assert_equals(style instanceof window.CSSStyleProperties, true, "properties");
+}, "Window.getComputedStyle");
```
domenic: Nit: please avoid `ClassName.staticProperty` syntax for non-static properties.

lib/jsdom/living/cssom/CSSStyleProperties-impl.js
```diff
@@ -0,0 +1,40 @@
+"use strict";
+
+const cssstyle = require("cssstyle");
+const CSSStyleDeclaration = require("./CSSStyleDeclaration-impl").implementation;
+
+class CSSStyleProperties extends CSSStyleDeclaration {
+  constructor(globalObject, args, privateData = {}) {
+    super(globalObject, args, privateData);
+    const { options = {} } = privateData;
+    this._style = new cssstyle.CSSStyleProperties(globalObject, options);
```
domenic: It feels a bit wasteful to reinitialize `this._style` twice. Maybe we should not call `super()`?

lib/jsdom/living/cssom/CSSStyleDeclaration.webidl
```diff
@@ -0,0 +1,12 @@
+// https://drafts.csswg.org/cssom/#the-cssstyledeclaration-interface
+[Exposed=Window]
+interface CSSStyleDeclaration {
+  [CEReactions] attribute CSSOMString cssText;
+  readonly attribute unsigned long length;
+  getter CSSOMString item(unsigned long index);
```
domenic, line 6: If I am reading the diffs correctly, this means that `el.style[0]` will now work, whereas it never did before. Am I correct? If so, adding a quick test case or two to the WPTs would be nice.

lib/jsdom/living/cssom/CSSStyleDeclaration-impl.js
```diff
@@ -0,0 +1,54 @@
+"use strict";
+
+const cssstyle = require("cssstyle");
+
+class CSSStyleDeclaration {
+  constructor(globalObject, args, privateData = {}) {
+    const { options = {} } = privateData;
```
domenic: Nit: here and elsewhere I think it's nicer to put the options in the constructor declaration, e.g.

```js
constructor(globalObject, args, { options = {} } = {})
```

lib/jsdom/browser/Window.js
```diff
@@ -906,40 +904,20 @@ function installOwnProperties(window, options) {
...
-    });
-
-    // https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle
-    const declarations = Object.keys(propertiesWithResolvedValueImplemented);
-    forEach.call(declarations, property => {
-      declaration.setProperty(property, getResolvedValue(elt, property));
-    });
-
-    return declaration;
+    const elementImpl = Element.convert(window, elt);
```
domenic: Nit: move this to the top of the function, while you're here.

domenic (commented):
Thanks for doing this. The overall approach looks very promising.

test/web-platform-tests/to-upstream/css/cssom/inline-style.html
asamuzaK: Should I remove it from this PR?

test/web-platform-tests/to-run.yaml
asamuzaK: It is because cssstyle does not yet properly implement parsers for `box-shadow` and other color related properties.

test/web-platform-tests/to-upstream-expectations.yaml
asamuzaK: > Does https://github.com/jsdom/jsdom/pull/3910 help?

I'll take another look.

test/web-platform-tests/to-run.yaml
domenic: But why did they pass before, and fail with this PR?

test/web-platform-tests/to-upstream/css/cssom/inline-style.html
domenic: Yes please.

test/web-platform-tests/to-run.yaml
asamuzaK: I haven't looked into the details, but I suspect that this may have coincidentally matched a color conversion table that jsdom had locally before.
My personal opinion is that it's expected to fail at this point.

asamuzaK: Found the reason.
For example, in the case of box-shadow, the propertyList.js contains:
```
[
  "box-shadow",
  {
    "href": "https://drafts.csswg.org/css-borders-4/#propdef-box-shadow",
    "initial": "none",
    "inherited": "no",
    "computedValue": "see individual properties",
    "name": "box-shadow",
    "styleDeclaration": [
      "box-shadow",
      "boxShadow"
    ]
  }
],
```

"computedValue" is set to "see individual properties" and not to "a/the computed color"

See lib/jsdom/living/helpers/style-rules.js#161

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -1,193 +1,40 @@
...
-  });
-
-  styleCache.set(elementImpl, declaration);
-
-  return declaration;
-};
+  return key;
+}).filter(key => key));
+const customProperties = new Map();
+const colorProperties = new Set();
```
domenic: What are these global variables doing? It seems surprising to me that sharing custom properties and color properties across all jsdom instances would be correct. They should probably be document-scoped, if not element-scoped, right?

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -1,193 +1,40 @@
...
+let styleCache = null;
```
domenic: Similarly this should not be a global, shared-across-all-jsdoms variable.

lib/jsdom/living/cssom/CSSStyleProperties-impl.js
asamuzaK: If we remove super(), we get these errors.
```
7:3  error  Expected to call 'super()'              constructor-super
8:5  error  'this' is not allowed before 'super()'  no-this-before-super
```

test/web-platform-tests/to-upstream/css/cssom/CSSStyleProperties.html
asamuzaK: If we remove `window.`, we get
```
20:34  error  'CSSStyleProperties' is not defined  no-undef
29:39  error  'CSSStyleProperties' is not defined  no-undef
35:38  error  'CSSStyleProperties' is not defined  no-undef
```

How should we fix this?

lib/jsdom/living/cssom/CSSStyleDeclaration.webidl
asamuzaK: That's true in theory.
But how do we set a getter for this[i]?

asamuzaK: Got it fixed for Element.style (but not for CSSStyleRule.style)

lib/jsdom/living/helpers/style-rules.js
asamuzaK: Done.

lib/jsdom/living/helpers/style-rules.js
asamuzaK: Moved them into getDeclarationForElement()

test/web-platform-tests/to-upstream/css/cssom/inline-style.html
asamuzaK: Done

lib/jsdom/browser/Window.js
asamuzaK: Done

lib/jsdom/living/cssom/CSSStyleDeclaration-impl.js
asamuzaK: Done

test/web-platform-tests/to-upstream-expectations.yaml
asamuzaK: It is expected to fail, since @acemir/cssom use it's own CSSStyleDeclaration.
Fixed comment.

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -273,47 +150,172 @@ function getComputedValue(element, property) {
...
+  const { parsedDefaultStyleSheet } = opt;
+  handleSheet(parsedDefaultStyleSheet, elementImpl, declaration, opt);
+  for (const sheet of elementImpl._ownerDocument.styleSheets._list) {
+    handleSheet(sheet, elementImpl, declaration, opt);
+  }
+  return declaration;
+}
 
-  throw new TypeError(`Internal error: unrecognized computed value instruction '${computedValue}'`);
+function setDeclarationForElement(elementImpl, opt, initialize = false) {
```
asamuzaK: Per https://github.com/jsdom/cssstyle/pull/244#discussion_r2496928748

<a href="https://github.com/jsdom/cssstyle/pull/244#:~:text=(Note%20that%20in%20general%2C%20I%20find%20setDeclarationForElement%20having%20two%20very-different%20behaviors%20in%20the%20initialize%20vs.%20not-initialize%20case%20confusing.%20It%20would%20be%20nice%20if%20there%20were%20a%20way%20to%20split%20that%20into%20two%20functions%2C%20probably%20by%20pulling%20out%20other%20parts%20of%20it%20into%20smaller%20functions." rel="noopener">(Note that in general, I find setDeclarationForElement having two very-different behaviors in the initialize vs. not-initialize case confusing. It would be nice if there were a way to split that into two functions, probably by pulling out other parts of it into smaller functions.</a>

test/web-platform-tests/to-upstream/css/cssom/CSSStyleProperties.html
domenic: I was just saying `Window.getComputedStyle` should be `window.getComputedStyle` or `getComputedStyle`, since `getComputedStyle` is not a static method of `Window`.

lib/jsdom/living/cssom/CSSStyleProperties-impl.js
domenic: Oh, very interesting. After a bit more research, probably it is a bad idea to use inheritance while not calling `super()`. But perhaps there are still better architectures.

- One option: Can you remind me why cssstyle exports `CSSStyleProperties` at all? What if we just had cssstyle only export `CSSStyleDeclaration`, and jsdom does all of the extra work for `CSSStyleProperties`? That is basically move https://github.com/jsdom/cssstyle/pull/244/files#diff-6d50434d88279a329253c2442b5510b674bfff939d6105ee3b2775307c3419efR21-R34 into jsdom. In my opinion, this works well with the program of "CSS complexity in cssstyle; wrapper object complexity in jsdom".

- Another option: we could have a privateData option in the jsdom `CSSStyleDeclaration` constructor. that lets the subclass pass in the value for `_style`.

test/web-platform-tests/to-upstream/css/cssom/CSSStyleProperties.html
asamuzaK: Oh, got it

lib/jsdom/living/cssom/CSSStyleProperties-impl.js
asamuzaK: > * One option: Can you remind me why cssstyle exports `CSSStyleProperties` at all? What if we just had cssstyle only export `CSSStyleDeclaration`, and jsdom does all of the extra work for `CSSStyleProperties`?

In line with current spec changes, we have separated it into CSSStyleDeclaration and CSSStyleProperties.
If this is not necessary and you want jsdom to comply with the spec, then let's revert to the original single CSSStyleDeclaration class.

domenic: We can still be spec-compliant, we just don't need to implement CSSStyleProperties in cssstyle. Only CSSStyleDeclaration.

In other words, there does not need to be any class in the cssstyle package with properties for `backgroundPosition`, etc.

test/web-platform-tests/to-upstream/css/cssom/CSSStyleProperties.html
asamuzaK: Done

lib/jsdom/living/cssom/CSSStyleProperties-impl.js
asamuzaK: Removed CSSStyleProperties from cssstyle.
And removed `    this._style = new cssstyle.CSSStyleProperties(globalObject, options);` line.

lib/jsdom/living/cssom/CSSStyleDeclaration.webidl
asamuzaK: When https://github.com/acemir/CSSOM/pull/58 is merged, CSSStyleRule.style will also be fixed.

lib/jsdom/living/helpers/style-rules.js
asamuzaK: Updated the function

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -1,331 +1,340 @@
...
-        }
-      } else if (matches(rule, elementImpl)) {
-        handleRule(rule);
-      }
-    });
-  }
-
-  if (!parsedDefaultStyleSheet) {
-    parsedDefaultStyleSheet = cssom.parse(defaultStyleSheet);
+function getDefaultStyleSheet() {
```
domenic: Please let's continue loading the file at startup, not lazily. You can parse lazily if that's helpful, but keep the existing code for loading the file as-is.

lib/jsdom/living/nodes/ElementCSSInlineStyle-impl.js
```diff
@@ -1,19 +1,36 @@
...
+      }
+      this.setAttributeNS(null, "style", newCssText);
+      this._settingCssText = false;
+    }
+  }
   get style() {
-    return this._style;
+    return idlUtils.wrapperForImpl(this._style);
+  }
+  set style(value) {
```
domenic: This should be able to be omitted; it's generated by the `[PutForwards]`.

lib/jsdom/living/nodes/ElementCSSInlineStyle-impl.js
```diff
@@ -1,19 +1,36 @@
...
+      options: {
+        context: idlUtils.wrapperForImpl(this),
+        onChange: this._styleChangeCallback.bind(this)
       }
     });
   }
+  _styleChangeCallback(newCssText) {
+    if (!this._settingCssText) {
+      this._settingCssText = true;
+      this._style._list = [];
```
domenic: Why do we need to reach into the internals of `_style` to duplicate its properties in two places? It seems better if we could design this to have a single source of truth, and to be responsible for updating itself.

lib/jsdom/living/helpers/style-rules.js
asamuzaK: Reverted.

lib/jsdom/living/nodes/ElementCSSInlineStyle-impl.js
asamuzaK: Done.

lib/jsdom/living/nodes/ElementCSSInlineStyle-impl.js
asamuzaK: Sorry, that was a mistake.

lib/jsdom/living/nodes/ElementCSSInlineStyle-impl.js
```diff
@@ -1,19 +1,32 @@
...
 
 class ElementCSSInlineStyle {
   _initElementCSSInlineStyle() {
     this._settingCssText = false;
-    this._style = new cssstyle.CSSStyleDeclaration(newCssText => {
-      if (!this._settingCssText) {
-        this._settingCssText = true;
-        this.setAttributeNS(null, "style", newCssText);
-        this._settingCssText = false;
+    this._list = [];
```
domenic: I still don't understand what `this._list` is and why we need to copy a bunch of stuff from `this._style` into `this._list`.

asamuzaK: Per https://github.com/jsdom/jsdom/pull/3957#discussion_r2463319198
We need a list to save keys.
It's called by https://github.com/jsdom/jsdom/pull/3957/files#diff-8c09836153dabcf0ec22015575059c90e9ff4862bac2e4761527c089546b67abR52

domenic: We only need the keys; we don't need a copy of everything else. So that can be much simpler: just 

```js
get [idlUtils.supportedPropertyIndices]() {
  let indices = new Array(this._style.length);
  for (let i = 0; i < this._style.length; ++i) {
    indices.push(i);
  }
  return indices;
}
```

asamuzaK: Done https://github.com/jsdom/jsdom/pull/3957/commits/c07acad6ecbccdeaaae26b01972383996b57662d

PR #3961, opened 2025-10-20
Revert parse5

### LINKED ISSUE #3959: Error [ERR_REQUIRE_ESM] on require() after updating to jsdom@27.0.1
### Node.js version

v20.18.1

### jsdom version

27.0.1

### Minimal reproduction case

```js

const window = new JSDOM('').window;
```

Doesn't even load running node command in console. Console output below from code above

"/var/www/node/jsdomtest/node_modules/jsdom/lib/jsdom/browser/parser/html.js:3
const parse5 = require("parse5");
               ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /var/www/node/jsdomtest/node_modules/parse5/dist/index.js from /var/www/node/jsdomtest/node_modules/jsdom/lib/jsdom/browser/parser/html.js not supported.
Instead change the require of index.js in /var/www/node/jsdomtest/node_modules/jsdom/lib/jsdom/browser/parser/html.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/var/www/node/jsdomtest/node_modules/jsdom/lib/jsdom/browser/parser/html.js:3:16) {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v20.18.1"

### What is the problem?

Upgrading from 27.0.0 to 27.0.1 causes the module to break on loading

asamuzaK:
Fix #3959

### DIFF
--- a/.github/workflows/jsdom-ci.yml
+++ b/.github/workflows/jsdom-ci.yml
@@ -51,7 +51,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          - 20.0.x
+          - 20
           - 22
           - latest
     steps:
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,7 +44,7 @@
         "pngjs": "^7.0.0",
         "server-destroy": "^1.0.1",
         "webidl2js": "^19.1.0",
-        "yargs": "^17.7.2"
+        "yargs": "^18.0.0"
       },
       "engines": {
         "node": ">=20"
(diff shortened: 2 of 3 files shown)


asamuzaK:
Supporting v20.0.x is not realistic.
We need to revert our devDependencies if we want to do this.
See https://github.com/jsdom/jsdom/actions/runs/18660207203/job/53198654750 https://github.com/jsdom/jsdom/actions/runs/18661148531/job/53201697639

domenic:
Thanks for testing this. It looks like enough of the ecosystem has moved on that jsdom needs to as well. It's unfortunate we didn't do the upgrade in 27.0.0, but I think we just have to accept it as-is for now.

We should update the "engines" field, and the README, and https://github.com/jsdom/jsdom/blob/main/.github/ISSUE_TEMPLATE/bug_report.yml , to have a more precise requirement, I guess.

asamuzaK:
It seems true that there are some third-party libraries that don't support `require(esm)`.
e.g.) https://github.com/vercel/vercel/issues/11102

So how about reverting `parse5` for now, and then explicitly state something like "jsdom is guaranteed to work only with the latest version of each Node.js release line"?

In any case, feel free to close this PR without accepting my suggestion.

domenic:
Yeah, I was thinking about that. But if we cannot maintain a realistic testing environment for these cases, then such best-effort support will just set us up for pain and failure in the future. And in general I prefer to minimize the amount of configurations we support since this project is so bandwidth-constrained already.

Issue #1122, opened 2015-05-06
Documents are not getting GCed, I think


domenic:
Repro: clone https://github.com/domenic/worm-ebook-scraper/tree/127b6612c90bb6227ce97d48f68dba577ee930cf and run `iojs lib/worm-scraper.js`. Remove the `rawChapterDoc.defaultView.close()` to get a faster leak, but it's still there even without it. I've been watching my memory grow slowly toward 1.5 GiB, at which point V8 falls over.

So that sucks. vm, I'd guess...

kylezeeuwen:
If it helps I'm seeing the same gradual increase of memory usage over time. I am use node-jsdom in a loop that is fetching and parsing URLs. After about 200 URLs node stops allocataing memory and crashes.

Of note I am seeing this issue using "node-jsdom": "3.1.5" running on nodejs 0.12.

Not looking for help, just pointing out the memory leak is occuring on a variety of environments.

kylezeeuwen:
Sorry wasn't calling window.close() . I did notice that the memory footprint still gradually increases after adding a window.close() call but the rate is more gradual. Are there any general usage guidelines that will slow this down / make it go away ?

Joris-van-der-Wel:
> Of note I am seeing this issue using "node-jsdom": "3.1.5" running on nodejs 0.12.

That is a pretty old version, so it might be a different issue than noted in the opening post. (That version also does not use `vm` afaik). Have you tried your script with the latest jsdom and io.js?

Also note that memory usage can be difficult to measure with a Garbage Collector, they can be lazy.

kylezeeuwen:
Hi Joris,

When calling node with --expose-gc and explicitly calling `global.gc()` I see a stable memory footprint so I'm happy. I did not get a chance to run latest with io.js.

I have some unrelated questions about using jsdom concurrently - specifically making multiple calls to `jsdom.env` without waiting for the `done` callback between subsequent calls. What is the best forum to pose questions like this to the maintainers ?

Thanks for your help !

Joris-van-der-Wel:
There's an irc channel we are all on, #jsdom on freenode. Stackoverflow is also sometimes used: http://stackoverflow.com/questions/tagged/jsdom, although I am not sure any of us actively follow that.

daemon1981:
Too bad this memory leak :( 

Simple use of jsdom results in a memory leak :
```
    jsdom.env('<html><head></head><body></body></html>', {
        scripts: [],
[...]

Issue #1089, opened 2015-04-20
memory leaks on large URLs


rickdog:
jsdom.env({
  url: "http://www.nytimes.com/2015/04/19/business/energy-environment/solar-power-battle-puts-hawaii-at-forefront-of-worldwide-changes.html",

C:\Users\rickdog>npm -v jsdom
2.1.18

please advise

```
(node) warning: possible EventEmitter memory leak detected. 11 pipe listeners added. Use emitter.setMaxListeners() to increase limit.
Trace
    at Request.addListener (events.js:161:15)
    at Request.init (C:\Users\rickdog\node_modules\jsdom\node_modules\request\request.js:667:8)
    at Redirect.onResponse (C:\Users\rickdog\node_modules\jsdom\node_modules\request\lib\redirect.js:149:11)
    at Request.onRequestResponse (C:\Users\rickdog\node_modules\jsdom\node_modules\request\request.js:1108:22)
    at ClientRequest.emit (events.js:95:17)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:399:21)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:89:23)
    at Socket.socketOnData (_http_client.js:290:20)
    at Socket.emit (events.js:95:17)
    at readableAddChunk (_stream_readable.js:143:16)
(node) warning: possible EventEmitter memory leak detected. 11 end listeners added. Use emitter.setMaxListeners() to increase limit.
Trace
    at Request.addListener (events.js:161:15)
    at Request.start (C:\Users\rickdog\node_modules\jsdom\node_modules\request\request.js:1001:8)
    at Request.end (C:\Users\rickdog\node_modules\jsdom\node_modules\request\request.js:1531:10)
    at end (C:\Users\rickdog\node_modules\jsdom\node_modules\request\request.js:734:14)
    at Immediate._onImmediate (C:\Users\rickdog\node_modules\jsdom\node_modules\request\request.js:748:7)
    at processImmediate [as _immediateCallback] (timers.js:321:17)
```

inikulin:
Seems like it's not caused by the URL length, but by the multiple nested redirects. According to the fact that we have more than 11 redirects, seems like that page or some resource goes into infinite redirect loop. 
@rickdog Do you have [loading of the external resources](https://github.com/tmpvar/jsdom#external-resources) enabled?

rickdog:
should I enable or disable external resources, and if to enable, which resources?
[...]

Issue #855, opened 2014-08-13
Event listeners in the capture phase should also fire on the event target


Joris-van-der-Wel:
I noticed that if I add an event listener in the capture phase on a certain element, and I subsequently dispatch that event on that same element, the listener is never called. Any listeners on parents are called as expected.

As far as I know, the capture & bubble phase should behave as in this example:

``` html
<div>
     <p>
            <span>abc</span>
     </p>
</div>
```

``` javascript
spanElement.dispatchEvent(...);
```
1. Listeners in the Capture phase on the &lt;div&gt; are called
2. Capture phase &lt;p&gt;
3. Capture phase &lt;span&gt;
4. Bubble phase &lt;span&gt;
5. Bubble phase &lt;p&gt;
6. Bubble phase &lt;div&gt;

If I test this in firefox the results are like in this example. jsdom skips step 3.

I wanted to update the test cases in jsdom and submit a fix but I noticed the following test case:

``` javascript
// An event is dispatched to the document with a capture listener attached.
  // A capturing EventListener will not be triggered by events dispatched directly to the EventTarget upon which it is registered.
  'EventListener with capture true': function (test) {
    var monitor = new EventMonitor();
    this.doc.addEventListener("foo", monitor.handleEvent, true);
    var event = this.doc.createEvent("Events");
    event.initEvent("foo",true,false);
    this.doc.dispatchEvent(event);
    test.expect(3);
    test.equal(monitor.atEvents.length, 0, 'should not receive atEvent');
    test.equal(monitor.bubbledEvents.length, 0, 'should not receive at bubble phase');
    test.equal(monitor.capturedEvents.length, 0, 'should not receive at capture phase');
    test.done();
  },
```

( https://github.com/tmpvar/jsdom/blame/638882c203a7e212cfd1082d8625dff41b558033/test/level2/events.js#L141 )

It seems that this behavior of jsdom was put in intentionally. Was there a reason for this? Or is it a misinterpretation of the specifications? (or maybe I am misinterpreting them)

PR #3963, opened 2025-10-21
Note more strict minimum version requirement

### LINKED ISSUE #3959: Error [ERR_REQUIRE_ESM] on require() after updating to jsdom@27.0.1

domenic:
As discussed in the changelog and in #3961, 894151baf0603c5d2203261ce9f4748295e6681c inadvertently increased the minimum Node.js version to v20.19.0.

Set a new minimum Node.js version of v20.19.5 (the current v20.x at this time); document this minimum in various places (including the issue template which was not correctly updated); and test this minimum on CI.

Closes #3959. Closes #3961.


asamuzaK:
See https://github.com/jsdom/jsdom/issues/3959#issuecomment-3428748801
https://nodejs.org/en/blog/release/v22.12.0

asamuzaK:
Re: jsdom-ci.yml:
With this fix, we only test the minimum version of v20, and will miss out on testing the latest version of v20?


Node.js is currently in a version transition period, so how about the following approach?

* Merge #3961 and release jsdom v27.0.2.
* Node.js v24 will enter LTS on October 28th, so at that time we will determine the minimum supported version for each Node.js release line and release jsdom v28.0.0.
  jsdom-ci.yml:
  ``` yml
  matrix:
    node-version:
      - 20.19.5
      - 20
      - 22.21.0
      - 22
      - 24.10.0
      - lts/*
      - latest
  ```
  
  package.json:
  ``` json
  "engines": {
    "node": "^20.19.5 || ^22.21.0 || >=24.10.0"
  }
  ```

For a while, 20.19.5 and 20, 22.21.0 and 22,  and 24.10.0 and lts/* are the same, so unnecessary tests will be run, but it will be almost maintenance-free for the next six months.

Refs:
* https://github.com/nodejs/Release
* [Release plan - v25.x Current](https://github.com/nodejs/Release/issues/1122 "Release plan - v25.x Current · Issue #1122 · nodejs/Release")
* [Release plan - v24.x Current](https://github.com/nodejs/Release/issues/1089 "Release plan - v24.x Current · Issue #1089 · nodejs/Release")
* [Release plan - v22.x Maintenance LTS](https://github.com/nodejs/Release/issues/1001 "Release plan - v22.x Maintenance LTS · Issue #1001 · nodejs/Release")
* [Release plan - v20.x Maintenance LTS](https://github.com/nodejs/Release/issues/855 "Release plan - v20.x Maintenance LTS · Issue #855 · nodejs/Release")

MattIPv4:
This is, in itself, a breaking change and should be versioned as such.

domenic:
> This is, in itself, a breaking change and should be versioned as such.

I'm happy to do this, as version numbers are free, but I don't understand why. jsdom 27.0.1 doesn't work in Node.js v20.18.x. What good does releasing 28.0.0, with the exact same Node.js version requirements and no code changes, do for anyone?

domenic:
> * Merge https://github.com/jsdom/jsdom/pull/3961 and release jsdom v27.0.2.

I'd prefer not to do this.

I only have bandwidth to support one release line of jsdom. From now on, that is going to be a version of jsdom which uses parse5 ^8.0.0, and thus relies on `require(esm)`.

We already have one unsupported release line which does not rely on `require(esm)`: v26.x. I don't think making v27.x another such release line is beneficial.

So I think we have the following choices:

* Merge this PR and release v27.0.2 with an `"engines"` field that reflects reality.

* Merge this PR and release v28.0.0 with an `"engines"` field that reflects reality.

Per my response to @MattIPv4 above, I don't really understand the benefit of the second option over the first. In both cases, the breaking change actually happened in v27.0.1, and there's no way to erase that. Releasing a v28 with no breaking changes compared to v27.0.1 seems confusing to me. But maybe it's helpful to someone in a way I don't understand? I'd love to hear the of the specific scenario where it matters.

Another tool in our toolbox is we could use `npm deprecate` to cause warnings when installing jsdom v27.0.1. But again I'm not sure exactly who that helps.

> * Node.js v24 will enter LTS on October 28th, so at that time we will determine the minimum supported version for each Node.js release line and release jsdom v28.0.0.

I've done something similar, although with lower lower-bounds. I'm curious what you think. In particular I didn't quite understand the benefit of waiting for October 28th.

asamuzaK:
> I've done something similar, although with lower lower-bounds. I'm curious what you think. In particular I didn't quite understand the benefit of waiting for October 28th.

It is a very minor thing, so don't worry about it.

Until October 27:
lts/* -> v22
latest -> v25

-> v24 is still a Current release but doesn't fall under either lts/* or latest.

After October 28:
lts/* -> v24
latest -> v25

---
I think it's fine to merge PR as-is.

Feel free to decide on the version.
If v28.0.0 seems a bit odd, how about v27.1.0?

.github/workflows/jsdom-ci.yml
```diff
@@ -51,9 +51,14 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
+          # Explicitly test minimum Node.js versions. Keep in sync with package.json.
+          - 20.19.0
           - 20
+          - 22.13.0
```
asamuzaK: Nit
Probably 22.12.0 ?

package.json
```diff
@@ -88,6 +88,6 @@
   "main": "./lib/api.js",
   "type": "commonjs",
   "engines": {
-    "node": ">=20"
+    "node": "^20.19.0 || ^22.13.0 || >=24.0.0"
```
asamuzaK: Nit
Probably ^22.12.0 ?

Issue #3979, opened 2025-11-04
SyntaxError: Failed to execute 'insertRule' on 'CSSStyleSheet'


martijnpieters:
### Node.js version

24.2.0

### jsdom version

27.1.0

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const dom = new JSDOM(`<!DOCTYPE html><html><head><style></style></head><body><h1>Hello world</h1></body></html>`);
const styleSheet = dom.window.document.styleSheets[0];

try {
[...]

PR #3981, opened 2025-11-06
Update `@acemir/cssom` dependency

acemir:
The updated dependency package with new features and improvements.

## Key Changes

- Added support for [CSSScopeRule](https://drafts.csswg.org/css-cascade-6/#cssscoperule)
- Added support for [CSSRuleList](https://drafts.csswg.org/cssom/#the-cssrulelist-interface)
- Improves validation for nested selectors and at-rules
- Added support for [legacy CSSStyleSheet methods: addRule, removeRule](https://drafts.csswg.org/cssom/#legacy-css-style-sheet-members)


acemir:
Also fixes the `insertRule` issue related to #3979.

package-lock.json
```diff
@@ -180,7 +180,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
```
domenic, line 183: You keep deleting these lines with your dependency updates, probably because you are using an older version of npm? I doubt it matters for anything important, so I will just merge it, but maybe you would find it useful to update your npm version.

domenic (approved):
Amazing work!

You should also feel free to release a v1.x of your cssom package. Then, when you make minor or patch version updates, people will get the benefits automatically on any fresh new jsdom install.

(But, even then, these sorts of PRs are still useful to help us understand the improvements.)

Issue #3350, opened 2022-04-09
CSS @import causes parent stylesheet to be overwritten in document.styleSheets


redneb:
### Basic info:

- **Node.js version:** 16.14.1
- **jsdom version:** 19.0.0

When JSDOM encounters a stylesheet that imports some other stylesheet with `@import`, then the imported stylesheet overwrittes the original one in the `document.styleSheets` array. In fact, the original stylesheet cannot be found at all in `document.styleSheets`.

### Minimal reproduction case

Suppose that you have the following files:

`index.html`:
```html
<!DOCTYPE html>
<html>
	<head>
		<meta charset="utf-8">
		<link rel="stylesheet" href="index.css">
	</head>
	<body></body>
</html>
```

`index.css`:
```css
@import url("aux.css");

.test1 {
	display: none;
}
```

and `aux.css`:
```css
.test2 {
	display: none;
}
```

Then if you run the following code
```js
const {JSDOM} = require("jsdom");
const {once} = require("events");

(async () => {
	const dom = await JSDOM.fromFile("index.html", {resources: "usable"});
	await once(dom.window, "load");
	for (const [idx, styleSheet] of [...dom.window.document.styleSheets].entries()) {
		console.log(`======== ${idx} ========`);
		console.log([...styleSheet.cssRules].map(c => c.cssText + "\n").join(""));
	}
})();
```
it will only find a single element in `document.styleSheets` that corresponds to the `@import`ed stylesheet.

I have put the above files in a [gist](https://gist.github.com/redneb/32e0217588e5c8a1eaddb777b6807912) and the bug can be easily reproduced as follows:
```bash
mkdir /tmp/jsdom-css-import-bug
cd /tmp/jsdom-css-import-bug
wget https://gist.github.com/redneb/32e0217588e5c8a1eaddb777b6807912/archive/1b876d9ed1506c151037a733e1831b233ae88fd4.zip
unzip 1b876d9ed1506c151037a733e1831b233ae88fd4.zip
cd */
npm install
node run.js
```

### How does similar code behave in browsers?

With the above example example code, when opened in a browser, you will find that `document.styleSheets` contains a single stylesheet that corresponds to `index.css` above and not `aux.css`. You can see that in action [here](https://codesandbox.io/s/relaxed-voice-5pd5qo) (look at the console tab).

Issue #2124, opened 2018-01-24
Issues with <style>/<link> processsing


domenic:
From @TimothyGu's research in https://github.com/tmpvar/jsdom/issues/2122#issuecomment-359992111, tracked here separately:

- [x] Empty `<style>` has a null `sheet` property; regressed in 8c84ebf3b7f3420b6e6cc68612ed3e0b9c1f4455. (#2122)
- [x] `<style>` with spaces will cause parse errors; probably regressed in 8c84ebf3b7f3420b6e6cc68612ed3e0b9c1f4455. (#2123)
- [x] `<style>` elements that are not connected (or got disconnected) but with inner text still have a non-null `sheet` (this sheet is also in `document.styleSheets` when it shouldn't be)
- [ ] `@import`s in style don't work (this is an issue in new API, as the default FetchExternalResources does not have "style", only "link")
- [ ] `link` with `@import` has `sheet` pointed to the imported sheet
- [ ] `link` can get into wrong order in `document.styleSheets`

I'm working on an incoming PR to fix some, but not all of these.

lovetingyuan:
how to disable fetch `@import` style? 
the document said jsdom won't try to fetch sub resource by default

Splaktar:
Any plans to fix the `@import` issue?

PR #3982, opened 2025-11-08
Fix css `@import` overwriting the sheet on elementImpl and cssom improvements

acemir:
## Key Changes
- Exports `addStylesheet` from stylesheets helpers
- Changes `createStylesheet` to return the parsed sheet
- Passes the `CSSImportRule` as argument to `fetchStylesheet`
- Populates `CSSImportRule.styleSheet.cssRules` with the parsed sheet cssRules
- Handles `CSSImportRule` on style-rules helpers
- Updates `@acemir/cssom` dependency and exposes CSSOM Rules on `addToCore`

Related #2124
Related #3350

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "27.1.0",
       "license": "MIT",
       "dependencies": {
-        "@acemir/cssom": "^0.9.22",
+        "@acemir/cssom": "^0.9.23",
         "@asamuzakjp/dom-selector": "^6.7.3",
         "cssstyle": "^5.3.2",
         "data-urls": "^6.0.0",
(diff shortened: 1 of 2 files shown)


domenic:
This work overall looks great, thank you! I'm especially happy to see that this fixes #3350, which has been around for a very long time.

I just want to double-check your impression on the test coverage. Do you think the WPTs that you got to start passing cover all of the improvements you made here? From skimming the contents of the tests you got passing, I don't think any of them test that the imported style rules actually apply to the page. (Like, if you import a stylesheet with `p { color: green; }`, and then test that the computed style of `document.querySelector('p')` is green.) Or maybe we already have coverage for that somewhere?

acemir:
@domenic There is a test in to-port-to-wpts that I believe should already cover this behavior: https://github.com/jsdom/jsdom/blob/70741dab9a9733c8c359ae82745269c5de607e44/test/to-port-to-wpts/level2/style.js#L179-L205
While debugging this test, I verified that the test was already passing precisely because the of the issue.

The root `CSSStyleSheet`, instead of containing the `CSSImportRule` item in its `cssRules,` actually contained the `cssRules` that should have been inside this `CSSImportRule.styleSheet`, meaning that the `CSSImportRule `ended up being completely removed along with any other rules declared in the root `CSSStyleSheet`.

Now with the fixed behavior, in this test case, the root `CSSStyleSheet` really have only one `CSSImportRule` with it's rules correctly populated inside  `CSSImportRule.styleSheet.cssRules`, and `getComputedStyle` also takes into account the content of this `CSSImportRule.styleSheet.cssRules`.

So it turns out that the test is now testing exactly this fixed behavior, checking weather a `<link>` element pointing to .css file with a `@import` pointing to another .css file with `body { color: red; }` applies correctly.

PR #3990, opened 2025-12-05
CSSOM Improvements

acemir:
The latest version of `@acemir/CSSOM` the dependency package comes with new features, improvements and bug fixes.

## Impact on jsdom
- `CSSOM.setup` method for globalObject configuration
- `CSSOM.parse` accepts `ownerNode` (for `<style>` and `<link>` stylesheets) and `ownerRule` (for `@import` stylesheets) to support unconstruced `CSSStyleSheet`

# Key changes
- Add `StyleSheet` missing properties according to the interface definition
- Add `CSSStyleSheet` constructor options and missing properties according to the interface definition
- Fix `CSSStylesheet.cssRules` and `CSSGroupingRule.cssRules` implementation according to the interface definition
- Fix `CSSGroupingRule.insertRule` to handle inserting `CSSNestedDeclarations`
- Fix `CSSRule.cssText` interface implementation according to the interface definition
- Fix `MediaList` implementation according to the interface definition
- Improvements in parse validations:
  - Improve validations on pseudo-class with selectors list suport (such as `:not`, `:is`, `:has`, `:where)`
  - Improve auto-close for all unclosed nested structures
  - Handle `CSSLayerStatementRule` inside `CSSLayerBlockRule` and other grouping rules
  - Handle `CSSFontFaceRule`,  `CSSKeyframeRule` and `CSSPageRule` nesting validations

### DIFF
--- a/test/to-port-to-wpts/level2/style.js
+++ b/test/to-port-to-wpts/level2/style.js
@@ -352,45 +352,4 @@ describe("level2/style", () => {
     assert.equal(document.body.style.width, "auto");
     assert.equal(document.body.style.cssText, "width: auto;");
   });
-  specify("ensure cssom uses the correct jsdom instance's globalObject", () => {
-    const dom1 = new JSDOM(`<html><head><style></style></head><body>`);
-    const dom2 = new JSDOM(`<html><head><style></style></head><body>`);
-  
-    const dom1ConstructedStylesheet = new dom1.window.CSSStyleSheet();
-    let dom1ConstructedStylesheetThrownError;
-    try {
-      dom1ConstructedStylesheet.insertRule("@layer;");
-    } catch (err) {
-      dom1ConstructedStylesheetThrownError = err;
-    }
-
-    const dom1UnconstructedStylesheet = dom1.window.document.head.lastChild.sheet;
-    let dom1UnconstructedStylesheetThrownError;
-    try {
-      dom1UnconstructedStylesheet.insertRule("@layer;");
-    } catch (err) {
-      dom1UnconstructedStylesheetThrownError = err;
-    }
-
-    const dom2ConstructedStylesheet = new dom2.window.CSSStyleSheet();
-    let dom2ConstructedStylesheetThrownError;
-    try {
-      dom2ConstructedStylesheet.insertRule("@layer;");
-    } catch (err) {
-      dom2ConstructedStylesheetThrownError = err;
-    }
-
-    const dom2UnconstructedStylesheet = dom2.window.document.head.lastChild.sheet;
-    let dom2UnconstructedStylesheetThrownError;
-    try {
-      dom2UnconstructedStylesheet.insertRule("@layer;");
-    } catch (err) {
-      dom2UnconstructedStylesheetThrownError = err;
-    }
-
-    assert.equal(dom1ConstructedStylesheetThrownError instanceof dom1.window.DOMException, true);
-    assert.equal(dom1UnconstructedStylesheetThrownError instanceof dom1.window.DOMException, true);
-    assert.equal(dom2ConstructedStylesheetThrownError instanceof dom2.window.DOMException, true);
-    assert.equal(dom2UnconstructedStylesheetThrownError instanceof dom2.window.DOMException, true);
-  });
 });
--- a/test/api/from-outside.js
+++ b/test/api/from-outside.js
@@ -113,4 +113,51 @@ describe("Test cases only possible to test from the outside", () => {
     const span = anchor.querySelector("span");
     assert.equal(span.innerHTML, "Link");
   });
+
+  it("ensure cssom uses the correct jsdom instance's globalObject", () => {
+    const dom1 = new JSDOM(`<html><head><style></style></head><body>`);
+    const dom2 = new JSDOM(`<html><head><style></style></head><body>`);
+
+    assert(
+      dom1.window.DOMException !== dom2.window.DOMException,
+      "Precondition: DOMException should differ between jsdom instances"
+    );
+
+    const dom1ConstructedStylesheet = new dom1.window.CSSStyleSheet();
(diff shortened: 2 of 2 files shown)


lib/jsdom/level2/style.js
```diff
@@ -19,7 +19,9 @@ exports.addToCore = core => {
   // - Load events are not in the specs, but apparently some browsers
   //   implement something. Should onload only fire after all @imports have been
   //   loaded, or only the primary sheet?
-
+  cssom.setup({
+    globalObject: core._globalObject
+  });
```
domenic: I don't understand this design. cssom is a single module, shared across the entire Node.js process. Whereas each jsdom instance has a separate `_globalObject`.

From what I can tell, this will cause CSSOM's `globalObject` reference to get overridden if someone creates two jsdom instances, e.g., via

```js
const dom1 = new JSDOM();
const dom2 = new JSDOM();
```

That seems like it'll be problematic.

acemir: You're right! I ran a test case that proves this behavior on this PR.

```js
    const dom1 = new JSDOM(`<html><head><style></style></head><body>`);
    const dom2 = new JSDOM(`<html><head><style></style></head><body>`);
  
    const dom1ConstructedStylesheet = new dom1.window.CSSStyleSheet();
    let dom1ConstructedStylesheetThrownError;
    try {
      dom1ConstructedStylesheet.insertRule("@layer;"); // SyntaxError
    } catch (err) {
      dom1ConstructedStylesheetThrownError = err;
    }

    const dom1UnconstructedStylesheet = dom1.window.document.head.lastChild.sheet;
    let dom1UnconstructedStylesheetThrownError;
    try {
      dom1UnconstructedStylesheet.insertRule("@layer;"); // SyntaxError
    } catch (err) {
      dom1UnconstructedStylesheetThrownError = err;
    }

    const dom2ConstructedStylesheet = new dom2.window.CSSStyleSheet();
    let dom2ConstructedStylesheetThrownError;
    try {
      dom2ConstructedStylesheet.insertRule("@layer;"); // SyntaxError
    } catch (err) {
      dom2ConstructedStylesheetThrownError = err;
    }

    const dom2UnconstructedStylesheet = dom2.window.document.head.lastChild.sheet;
    let dom2UnconstructedStylesheetThrownError;
    try {
      dom2UnconstructedStylesheet.insertRule("@layer;"); // SyntaxError
    } catch (err) {
      dom2UnconstructedStylesheetThrownError = err;
    }

    // FAIL (It's instanceof dom2.window.DOMException instead)
    assert.equal(dom1ConstructedStylesheetThrownError instanceof dom1.window.DOMException, true); 
    // FAIL:  (It's instanceof dom2.window.DOMException instead)
    assert.equal(dom1UnconstructedStylesheetThrownError instanceof dom1.window.DOMException, true);
    // SUCCESS
    assert.equal(dom2ConstructedStylesheetThrownError instanceof dom2.window.DOMException, true);
    // SUCCESS
    assert.equal(dom2UnconstructedStylesheetThrownError instanceof dom2.window.DOMException, true);
```

And if we run the same test on the latest version of jsdom 27.2.0, we get:
```js
    // FAIL (It's instanceof <nodeinternals>DOMException instead)
    assert.equal(dom1ConstructedStylesheetThrownError instanceof dom1.window.DOMException, true); 
    // SUCCESS
    assert.equal(dom1UnconstructedStylesheetThrownError instanceof dom1.window.DOMException, true);
    // FAIL (It's instanceof <nodeinternals>DOMException instead)
    assert.equal(dom2ConstructedStylesheetThrownError instanceof dom2.window.DOMException, true);
    // SUCCESS
    assert.equal(dom2UnconstructedStylesheetThrownError instanceof dom2.window.DOMException, true);
```
The problem I'm trying to solve is that we need to expose jsdom _globalObject also for constructed CSSStylesheet and other CSSOM objects.

Any suggestions on how to do this in a more appropriate way?

domenic: The module needs to be able to create one copy of CSSStyleSheet, etc., per jsdom instance, via a factory function that returns them.

acemir: Thanks for the suggestion, it was very helpful. I made the changes and now it's working correctly.

test/to-port-to-wpts/level2/style.js
```diff
@@ -352,4 +352,45 @@ describe("level2/style", () => {
     assert.equal(document.body.style.width, "auto");
     assert.equal(document.body.style.cssText, "width: auto;");
   });
+  specify("ensure cssom uses the correct jsdom instance's globalObject", () => {
```
domenic: Create new tests that cannot be WPTs in https://github.com/jsdom/jsdom/blob/main/test/api/from-outside.js

test/to-port-to-wpts/level2/style.js
```diff
@@ -352,4 +352,45 @@ describe("level2/style", () => {
...
+
+    const dom2UnconstructedStylesheet = dom2.window.document.head.lastChild.sheet;
+    let dom2UnconstructedStylesheetThrownError;
+    try {
+      dom2UnconstructedStylesheet.insertRule("@layer;");
+    } catch (err) {
+      dom2UnconstructedStylesheetThrownError = err;
+    }
+
+    assert.equal(dom1ConstructedStylesheetThrownError instanceof dom1.window.DOMException, true);
```
domenic: Add a sanity check that `dom1.window.DOMException` is not equal to `dom2.window.DOMException`. I think this sanity check might fail unless you pass `{ runScripts: "outside-only" }` (or `"dangerously"`)?

domenic: Style nit: since instanceof can only ever return true/false, just `assert()` is simpler than `assert.equal(x, true)`.

domenic (commented):
Looks promising, just some test nits!

PR #3994, opened 2025-12-09
Move to caller-provided HTTP(S) agents

domenic:
Previously, jsdom would create new HTTP and HTTPS agents for every fetch that it did (via ResourceLoader).

This was quite wasteful, and led to memory leaks whenever external resources were requested, because HTTP(S) agents require manual teardown with their destroy() method—we cannot simply rely on GC to clean them up.

Stepping back, jsdom's pattern of trying to encapsulate the complexity of HTTP(S) agents, proxies, connections, etc. behind simple { proxy, strictSSL } options was not a good one. In idiomatic Node.js, HTTP(S) agents are created by the application developer, and then fed into the libraries they use (like jsdom). This lets the developer make decisions about, e.g., agent reuse, keepalive, etc.

I am hopeful this is the root of the memory leak that causes WPT timeouts on CI (#3154), since running WPT involves many external resource requests.

### DIFF
--- a/test/web-platform-tests/run-single-wpt.js
+++ b/test/web-platform-tests/run-single-wpt.js
@@ -14,19 +14,6 @@ const reporterPathname = "/resources/testharnessreport.js";
 // that's fine, since the tests are run sequentially.
 const httpsAgent = new https.Agent({ rejectUnauthorized: false, keepAlive: true, timeout: 5_000 });
 
-// Memory leak debugging
-let testCount = 0;
-function logMemory(testPath) {
-  testCount++;
-  if (global.gc) {
-    global.gc();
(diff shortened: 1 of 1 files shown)


domenic:
> I am hopeful this is the root of the memory leak that causes WPT timeouts on CI (#3154), since running WPT involves many external resource requests.

This doesn't seem likely to be the case, unfortunately. Maybe the hypothesis of there being a memory leak is not correct? The pattern in the control run (and the maybe-fixed run) appears to be closer to what we'd expect, with lots of heap usage building up but then the GC always coming in to clean things up eventually.

e20cc353-37ec-499d-9493-d5f76c56d6bc

PR #4003, opened 2025-12-21
improve asciiLowercase / asciiUppercase perf on ascii

ChALkeR:
Coming from ce779560c10abf4bbe851c6520053a2cc02e5068 and https://github.com/jsdom/whatwg-encoding/issues/22#issuecomment-3652806899

If the input is ASCII most of the time, this is ~3x-10x faster

Also, if this is used for tagname, a direct comparison to the two top-frequent-tag names (or a map) could improve perf further


domenic:
It still feels like there should be a single-pass way to do this instead of two, but I can see that the current code is also two-pass and less efficient, so I'll merge this. Maybe there are further improvements possible.

We try not to use this for comparisons, but rather for canonicalization, so a caching system should not be necessary.

PR #4004, opened 2025-12-22
Switch from iconv-lite to exodus/bytes for decoding

ChALkeR:
See https://github.com/jsdom/whatwg-encoding/issues/22 and https://github.com/jsdom/whatwg-encoding/pull/24#issuecomment-3672970117

This should be non-breaking:
1. `legacyHookDecode` now accepts everything that `whatwg-encoding.decode()` accepts (as it supports all labels)
2. Added `labelToName` that returns cased names for the DOM API
3. Engines support remains the same

This fixes handling of all encodings that were previously incorrect, including: `UTF-8`, `UTF-16`, half of single-byte, all of multi-byte, and adds `ISO-8859-8-I`, `x-mac-cyrillic`, `ISO-2022-JP`, `replacement`

Calls to `legacyHookDecode` don't check that the encoding is not `'replacement'`. Should they?

### DIFF
--- a/lib/jsdom/living/file-api/FileReader-impl.js
+++ b/lib/jsdom/living/file-api/FileReader-impl.js
@@ -39,8 +39,8 @@ class FileReaderImpl extends EventTargetImpl {
   readAsDataURL(file) {
     this._readFile(file, "dataURL");
   }
-  readAsText(file, encoding) {
-    this._readFile(file, "text", labelToName(encoding) || "UTF-8");
+  readAsText(file, encodingLabel) {
+    this._readFile(file, "text", labelToName(encodingLabel) || "UTF-8");
   }
 
   abort() {
@@ -63,7 +63,7 @@ class FileReaderImpl extends EventTargetImpl {
     fireAnEvent(name, this, ProgressEvent, props);
   }
 
-  _readFile(file, format, encoding) {
+  _readFile(file, format, encodingLabel) {
     if (this.readyState === READY_STATES.LOADING) {
       throw DOMException.create(this._globalObject, [
         "The object is in an invalid state.",
@@ -109,7 +109,7 @@ class FileReaderImpl extends EventTargetImpl {
             break;
           }
           case "text": {
-            this.result = legacyHookDecode(data, encoding);
+            this.result = legacyHookDecode(data, encodingLabel);
             break;
           }
           case "buffer":
--- a/lib/jsdom/living/helpers/stylesheets.js
+++ b/lib/jsdom/living/helpers/stylesheets.js
@@ -8,11 +8,11 @@ const { invalidateStyleCache } = require("./style-rules");
 // It (and the things it calls) is nowhere close right now.
 exports.fetchStylesheet = (elementImpl, urlString, importRule) => {
   const document = elementImpl._ownerDocument;
-  let defaultEncoding = document._encoding;
+  let defaultEncodingLabel = document._encoding;
   const resourceLoader = document._resourceLoader;
 
   if (elementImpl.localName === "link" && elementImpl.hasAttributeNS(null, "charset")) {
-    defaultEncoding = elementImpl.getAttributeNS(null, "charset");
+    defaultEncodingLabel = elementImpl.getAttributeNS(null, "charset");
   }
 
   function onStylesheetLoad(data) {
@@ -21,7 +21,7 @@ exports.fetchStylesheet = (elementImpl, urlString, importRule) => {
       return;
     }
 
-    const css = legacyHookDecode(data, defaultEncoding);
+    const css = legacyHookDecode(data, defaultEncodingLabel);
 
     // TODO: MIME type checking?
     if (!importRule && elementImpl.sheet) {


domenic:
I did the new v5 html-encoding-sniffer release so we can roll that into this too.

> 1. `labelToName` usage blocks some encodings that are actually supported

Can you expand on what you mean by this? What was the previous behavior, and what is the new behavior?

> 2\. Calls to `legacyHookDecode` don't check that the encoding is not `'replacement'`. Should they?

I suspect this is fine as-is, as there is no such check in the spec. For example, if replacement shows up in the charset= parameter of a Content-Type for an iframe, style, or script, I suspect this is supposed to work.

ChALkeR:
> Can you expand on what you mean by this? What was the previous behavior, and what is the new behavior?

That's fixed already with the unification with `html-encoding-sniffer` v5, I updated the PR description.
Now that just adds support for encoding names that whatwg-encoding did not support but the spec has

domenic:
> 2\. Internal lowercase encodings are transformed to cased with a map to keep API compat, as they are exposed e.g. via `dom.window.document.characterSet`. That could be removed later separately if waned, just no reason to make this PR breaking.

I don't think this can be removed, because the spec requires exposing the encoding name: https://dom.spec.whatwg.org/#dom-document-characterset

This actually points out an unfortunate deficit of `@exodus/bytes`'s decision to use the non-standard lowercased name/"canonical label" concept, instead of using the standard name concept. I wish `@exodus/bytes` had stuck to the standard :(.

If you are set on keeping the nonstandard concept in `@exodus/bytes`'s API, then I guess converting back in jsdom is fine. I'll just add some suggestions on how to do so as code review comments.

lib/jsdom/living/nodes/HTMLFrameElement-impl.js
```diff
@@ -51,10 +52,10 @@ function fetchFrame(serializedURL, frame, document, contentDoc) {
       }
     }
 
-    const encoding = sniffHTMLEncoding(data, sniffOptions);
+    const encoding = toCasedEncodingName(sniffHTMLEncoding(data, sniffOptions));
```
domenic: We should stick to storing canonical encoding labels, not names, throughout jsdom, and only convert to names at the boundary.

To implement this, remove the mapping here. And for clarity, rename `_encoding` to `_canonicalizedEncodingLabel`.

lib/jsdom/living/file-api/FileReader-impl.js
```diff
@@ -40,7 +40,7 @@ class FileReaderImpl extends EventTargetImpl {
     this._readFile(file, "dataURL");
   }
   readAsText(file, encoding) {
```
domenic: Let's rename this parameter encodingLabel

lib/jsdom/living/file-api/FileReader-impl.js
domenic: Let's rename this parameter `canonicalEncodingLabel`

lib/jsdom/living/helpers/stylesheets.js
domenic: Let's rename this parameter `defaultCanonicalEncodingLabel`

lib/jsdom/living/helpers/stylesheets.js
```diff
@@ -12,7 +12,7 @@ exports.fetchStylesheet = (elementImpl, urlString, importRule) => {
   const resourceLoader = document._resourceLoader;
 
   if (elementImpl.localName === "link" && elementImpl.hasAttributeNS(null, "charset")) {
-    defaultEncoding = whatwgEncoding.labelToName(elementImpl.getAttributeNS(null, "charset"));
+    defaultEncoding = elementImpl.getAttributeNS(null, "charset");
```
domenic: This needs normalization from an encoding label to a canonical encoding label, right?

lib/jsdom/living/nodes/HTMLScriptElement-impl.js
```diff
@@ -62,7 +62,7 @@ class HTMLScriptElementImpl extends HTMLElementImpl {
   _fetchExternalScript() {
     const document = this._ownerDocument;
     const resourceLoader = document._resourceLoader;
-    const defaultEncoding = whatwgEncoding.labelToName(this.getAttributeNS(null, "charset")) || document._encoding;
+    const defaultEncoding = normalizeEncoding(this.getAttributeNS(null, "charset")) || document._encoding;
```
domenic: Rename to `defaultCanonicalizedEncodingLabel`

lib/jsdom/living/nodes/HTMLScriptElement-impl.js
```diff
@@ -88,10 +88,10 @@ class HTMLScriptElementImpl extends HTMLElementImpl {
         contentType = MIMEType.parse(response.headers["content-type"]) || new MIMEType("text/plain");
       }
 
-      const encoding = whatwgEncoding.getBOMEncoding(data) ||
-        (contentType && whatwgEncoding.labelToName(contentType.parameters.get("charset"))) ||
+      const encoding = getBOMEncoding(data) ||
```
domenic: Rename to `canonicalizedEncodingLabel`

lib/jsdom/living/xhr/XMLHttpRequest-impl.js
```diff
@@ -248,8 +249,8 @@ class XMLHttpRequestImpl extends XMLHttpRequestEventTargetImpl {
       return "";
     }
 
-    const fallbackEncoding = finalCharset(this) || whatwgEncoding.getBOMEncoding(responseBuffer) || "UTF-8";
-    const res = whatwgEncoding.decode(responseBuffer, fallbackEncoding);
+    const fallbackEncoding = finalCharset(this) || getBOMEncoding(responseBuffer) || "utf-8";
```
domenic: Rename to `fallbackEncodingCanonicalizedLabel`

lib/jsdom/living/xhr/XMLHttpRequest-impl.js
```diff
@@ -289,8 +290,8 @@ class XMLHttpRequestImpl extends XMLHttpRequestEventTargetImpl {
       return null;
     }
 
-    const encoding = finalCharset(this) || whatwgEncoding.getBOMEncoding(responseBuffer) || "UTF-8";
-    const resText = whatwgEncoding.decode(responseBuffer, encoding);
+    const encoding = finalCharset(this) || getBOMEncoding(responseBuffer) || "utf-8";
```
domenic: Rename to `canonicalizedEncodingLabel`

lib/jsdom/living/xhr/XMLHttpRequest-impl.js
```diff
@@ -301,7 +302,7 @@ class XMLHttpRequestImpl extends XMLHttpRequestEventTargetImpl {
         lastModified: new Date(getResponseHeader(this, "last-modified")),
         parsingMode: isHTML ? "html" : "xml",
         cookieJar: { setCookieSync: () => undefined, getCookieStringSync: () => "" },
-        encoding,
+        encoding: toCasedEncodingName(encoding),
```
domenic: Rename the constructor privateData option to `Document` to `canonicalizedEncodingLabel` (which sets `_canonicalizedEncodingLabel`), and then we can just pass it in as-is.

lib/jsdom/encoding-map.js
```diff
@@ -0,0 +1,20 @@
...
+  "IBM866", "KOI8-R", "KOI8-U",
+  "ISO-8859-2", "ISO-8859-3", "ISO-8859-4", "ISO-8859-5", "ISO-8859-6", "ISO-8859-7", "ISO-8859-8",
+  "ISO-8859-8-I", "ISO-8859-10", "ISO-8859-13", "ISO-8859-14", "ISO-8859-15", "ISO-8859-16",
+  "GBK", "Big5", "EUC-JP", "ISO-2022-JP", "Shift_JIS", "EUC-KR"
+].map(name => [name.toLowerCase(), name]));
+
+// Map lowercase @exodus/bytes/encoding@1 encoding names to cased whatwg-encoding@3 names
+// Does not perform any input validation, only output from @exodus/bytes/encoding or html-encoding-sniffer is expected
+// null falls through
+// This logic is only for compat and can be removed by switching exposed encoding properties to lowercase
```
domenic:
```suggestion
// Maps canonicalized encoding labels, used by @exodus/bytes/encoding@1, to encoding names as given in the spec, as in
// https://encoding.spec.whatwg.org/#names-and-labels. jsdom internally uses canonicalized encoding labels, since that's
// what our dependency uses; however, some public APIs, notably `document.characterSet` and its aliases, require
// exposing the actual names. We do just-in-time conversion to names for such endpoints via this function.
//
// This function does not perform any input validation; only output from @exodus/bytes/encoding or html-encoding-sniffer
// is expected.
```

lib/jsdom/encoding-map.js
```diff
@@ -0,0 +1,20 @@
...
+function toCasedEncodingName(name) {
```
domenic: Rename to `canonicalizedEncodingLabelToEncodingName`

lib/api.js
domenic, line 291:
```suggestion
  let canonicalizedEncodingLabel = "utf-8";
```

lib/api.js
```diff
@@ -301,7 +302,8 @@ function normalizeHTML(html, mimeType) {
       defaultEncoding: mimeType.isXML() ? "UTF-8" : "windows-1252",
       transportLayerEncodingLabel: mimeType.parameters.get("charset")
     });
-    html = whatwgEncoding.decode(html, encoding);
+    html = legacyHookDecode(html, encoding);
+    encoding = toCasedEncodingName(encoding);
```
domenic: Remove this since per previous discussion we'll store the canonicalized encoding labels.

ChALkeR:
> I don't think this can be removed, because the spec requires exposing the encoding name:

Huh, I see. I missed that those names are exposed in string form in DOM API, not just the internal structure.
I might rethink that

domenic:
Yeah, I learned today for the first time that the getters on `Document` return different values than the ones on `TextDecoder` and `TextDecoderStream`. The web is so wonderful... 🙃

lib/jsdom/living/helpers/stylesheets.js
ChALkeR: Not anymore, any valid label is supported in `legacyHookDecode` now

ChALkeR:
I added `labelToName` and wrapped most calls to `getBOMEncoding` and `sniffHTMLEncoding` in it
Except one case where label can be used directly, there I renamed the var.

`normalizeEncoding` is not used anymore.

lib/jsdom/living/file-api/FileReader-impl.js
ChALkeR: Done

lib/jsdom/living/file-api/FileReader-impl.js
ChALkeR: Renamed to `encodingLabel`, does not have to be canonical.

lib/jsdom/living/nodes/HTMLFrameElement-impl.js
ChALkeR: After https://github.com/jsdom/jsdom/pull/4004#issuecomment-3686023238, I just added `labelToName` and kept the old approach which stores case-sensitive names everywhere (which seems to be what DOM API wants).

`legacyHookDecode` can accept any valid label now and doesn't need them to be normalized, so passing names is fine.

lib/jsdom/living/file-api/FileReader-impl.js
```diff
@@ -39,8 +39,8 @@ class FileReaderImpl extends EventTargetImpl {
   readAsDataURL(file) {
     this._readFile(file, "dataURL");
   }
-  readAsText(file, encoding) {
-    this._readFile(file, "text", whatwgEncoding.labelToName(encoding) || "UTF-8");
+  readAsText(file, encodingLabel) {
+    this._readFile(file, "text", labelToName(encodingLabel) || "UTF-8");
```
domenic, line 43: We can remove this wrapping now then I guess?

lib/jsdom/living/helpers/stylesheets.js
ChALkeR: `document._encoding` is case-sensitive name and that wasn't changed

lib/jsdom/living/nodes/HTMLScriptElement-impl.js
ChALkeR: It's back to being case-sensitive name

lib/jsdom/living/nodes/HTMLScriptElement-impl.js
```diff
@@ -88,10 +88,10 @@ class HTMLScriptElementImpl extends HTMLElementImpl {
         contentType = MIMEType.parse(response.headers["content-type"]) || new MIMEType("text/plain");
       }
 
-      const encoding = whatwgEncoding.getBOMEncoding(data) ||
-        (contentType && whatwgEncoding.labelToName(contentType.parameters.get("charset"))) ||
+      const encoding = labelToName(getBOMEncoding(data)) ||
+        (contentType && labelToName(contentType.parameters.get("charset"))) ||
         defaultEncoding;
```
domenic, line 93: I guess labelToName is not needed here either? We can call this local variable encodingLabel?

lib/jsdom/living/nodes/HTMLScriptElement-impl.js
ChALkeR: This is also back to being a case-sensitive name

lib/jsdom/living/helpers/stylesheets.js
domenic: Alright, so the `defaultEncoding` local variable should probably become `defaultEncodingLabel`, relying on the fact that `_encoding` is a name and names are also labels.

lib/jsdom/living/xhr/XMLHttpRequest-impl.js
ChALkeR: Renamed to `fallbackEncodingLabel`, doesn't have to be canonical.

lib/jsdom/living/xhr/XMLHttpRequest-impl.js
ChALkeR: This one is back to being a case-sensitive name.

lib/jsdom/living/file-api/FileReader-impl.js
domenic, line 43:
```suggestion
    this._readFile(file, "text", encodingLabel || "utf-8");
```

lib/jsdom/living/xhr/XMLHttpRequest-impl.js
ChALkeR: This was reverted

lib/jsdom/living/nodes/HTMLScriptElement-impl.js
```diff
@@ -62,7 +62,7 @@ class HTMLScriptElementImpl extends HTMLElementImpl {
   _fetchExternalScript() {
     const document = this._ownerDocument;
     const resourceLoader = document._resourceLoader;
-    const defaultEncoding = whatwgEncoding.labelToName(this.getAttributeNS(null, "charset")) || document._encoding;
+    const defaultEncoding = labelToName(this.getAttributeNS(null, "charset")) || document._encoding;
```
domenic, line 65: We can change this to defaultEncodingLabel and remove labelToName, I think?

lib/jsdom/encoding-map.js
ChALkeR: Based on https://github.com/jsdom/jsdom/pull/4004#issuecomment-3686023238, I just added `labelToName` to `@exodus/bytes`, which solves this.

domenic (commented):
Alright, so now `@exodus/bytes`'s `legacyDecodeHook` accepts all labels, and labels are a superset of names. So:

- We can continue storing the names as `_encoding` on `Document` objects.
- We can otherwise just remove almost all canonicalization code, operating on raw labels we get from e.g. `charset=""` attributes or `Content-Type` headers or whatever, since it doesn't really matter what we pass in to the underlying package.

I think my latest round of suggestions/comments gets us to that state. Let me know what you think.

This seems like a pretty nice architecture in the end.

lib/jsdom/living/file-api/FileReader-impl.js
ChALkeR: This is different.


`legacyHookDecode` accepts _any valid label_ and throws otherwise.
`labelToName` maps unrecognized labels to `null`

So, what this code does is maps any unrecognized labels to `UTF-8`.
This matches the old behavior, changing the behavior on unrecognized labels would be out of scope of this PR.

```js
> require('whatwg-encoding').decode(Uint8Array.of(), 'what')
Uncaught RangeError: "what" is not a supported encoding name
> encoding = require('whatwg-encoding').labelToName('what') || 'UTF-8'
> require('whatwg-encoding').decode(Uint8Array.of(), encoding)
''

> require('@exodus/bytes/encoding.js').legacyHookDecode(Uint8Array.of(), 'what')
Uncaught RangeError: Unknown encoding
> encoding = require('@exodus/bytes/encoding.js').labelToName('what') || 'UTF-8'
> require('@exodus/bytes/encoding.js').legacyHookDecode(Uint8Array.of(), encoding)
''

lib/jsdom/living/nodes/HTMLScriptElement-impl.js
ChALkeR: ` || defaultEncoding` changes the behavior on unrecognized labels
I kept that for compat.

lib/jsdom/living/file-api/FileReader-impl.js
ChALkeR: This changes the behavior on unrecognized labels.

domenic (approved):
Thanks for walking me through this!

PR #4005, opened 2025-12-22
Unify API tests testsuite usage, fix async

ChALkeR:
1. Two instances of `assert.rejects` were incorrect, one actually failed due to invalid assertion usage
    Test runner failed to notice this
2. Replace mocha-style `{ async: true }` with returning a promise, which mocha also supports
3. Strip colors from the subprocess stdout before parsing it as a number

With this, you can just use `node:test` as a drop-in replacement for `mocha-sugar-free` inside `test/api` directory
The switch itself is not done by this PR though

### DIFF
--- a/test/api/file-url-functionality.js
+++ b/test/api/file-url-functionality.js
@@ -6,27 +6,40 @@ const toFileUrl = require("../util.js").toFileUrl(__dirname);
 
 const { JSDOM } = require("../..");
 
+function withResolvers() {
+  if (Promise.withResolvers) {
+    return Promise.withResolvers();
+  }
+  const r = {};
+  // eslint-disable-next-line no-promise-executor-return
+  r.promise = new Promise((resolve, reject) => Object.assign(r, { resolve, reject }));
+  return r;
+}
+
 const thisFileURL = toFileUrl(__filename);
 
 describe("Test cases for the interaction with file: URLs", () => {
   describe("XMLHttpRequest", () => {
-    it("should retrieve the same file: URL's contents, as both response and responseText", { async: true }, t => {
+    it("should retrieve the same file: URL's contents, as both response and responseText", () => {
       const { window } = new JSDOM(``, { url: thisFileURL });
+      const { promise, resolve } = withResolvers();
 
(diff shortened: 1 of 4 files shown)


domenic:
> 1. Two instances of `assert.rejects` were incorrect, one actually failed due to invalid assertion usage
>        Test runner failed to notice this

Nice, thank you!

> 2\. Replace mocha-style `{ async: true }` with returning a promise, which mocha also supports

Can you undo this? I find the duplicate `Promise.withResolvers()` shim to be worse than the code it replaces.

> 3\. Strip colors from the subprocess stdout before parsing it as a number

Can you explain why this is necessary? What was failing before?

> With this, you can just use `node:test` as a drop-in replacement for `mocha-sugar-free` inside `test/api` directory
> The switch itself is not done by this PR though

We don't want to move from 2 test frameworks (WPT-based-on-Mocha and Mocha) to 3, so this is not a goal we're working towards. If we could move en-masse to `node:test`, that would be nice; see https://github.com/jsdom/jsdom/pull/3650 for some previous attempts. But that's blocked on github.com/nodejs/node/issues/51384 at the very least.

ChALkeR:
> Can you explain why this is necessary? What was failing before?

Again, `node --test` compat.

Unfortunately, this default behavior of `node --test` affects spawned subprocesses: https://github.com/nodejs/node/blob/55600e6153b9d8a23f6365737123ab26e54b12db/lib/internal/test_runner/runner.js#L411-L413

And the stdout of the script is not `"-0.0009765028989929814"`, but `"\u001b[33m-0.0009765028989929814\u001b[39m\n"`.

But that's likely not needed if `node --test` compat is not a goal

The same behavior can be reproduced on main with `FORCE_COLOR=1 npm t` though

> https://github.com/nodejs/node/issues/51384

👀

domenic:
Closing, as I pulled the `assert.rejects()` fixes out into https://github.com/jsdom/jsdom/pull/4049#issuecomment-3882275631. This will remain a valuable reference if we ever are able to move to the native `node:test` runner.

PR #3791, opened 2024-11-14
Implement window.TextEncoder & window.TextDecoder

Ayc0:
Attempt to fix https://github.com/jsdom/jsdom/issues/2524
This just forwards `TextEncoder` & `TextDecoder` from `node:util` to the top level variables defined on the window

This is my 1st contribution to JSDOM so Idk if I'm missing something.
I'm also not sure if I have to add a test (because it seems that we already have a lot of tests with those in WPT and in `to-run.yaml`) and if so, where

PR #4006, opened 2025-12-24
Attempt to implement TextEncoder and TextDecoder

domenic:
/cc @ChALkeR. This currently fails the web platform tests because of the various `instanceof` checks in `@exodus/bytes`, which don't work when the jsdom vm contexts are involved. That's https://github.com/ExodusOSS/bytes/issues/2.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@acemir/cssom": "^0.9.28",
         "@asamuzakjp/dom-selector": "^6.7.6",
-        "@exodus/bytes": "^1.5.0",
+        "@exodus/bytes": "^1.6.0",
         "cssstyle": "^5.3.4",
         "data-urls": "^6.0.0",
         "decimal.js": "^10.6.0",
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   "dependencies": {
     "@acemir/cssom": "^0.9.28",
     "@asamuzakjp/dom-selector": "^6.7.6",
-    "@exodus/bytes": "^1.5.0",
+    "@exodus/bytes": "^1.6.0",
     "cssstyle": "^5.3.4",
     "data-urls": "^6.0.0",
     "decimal.js": "^10.6.0",


ChALkeR:
Looking at it

ChALkeR:
https://github.com/ExodusOSS/bytes/commit/a5766855b05e49e95c23290252ad324c9990ac73 seems to pass

Without looking at the test results (which are not ready yet), this is likely failing because of ArrayBuffer / SharedArrayBuffer instanceof checks on these lines specifically: https://github.com/ExodusOSS/bytes/blob/a5766855/fallback/encoding.js#L52-L54

Will fix that

`Uint8Array` passes because it passes `ArrayBuffer.isView`

ChALkeR:
@domenic Try 1.4.0 for TextDecoder

domenic:
Seems to have worked, still lots of failing tests for TextEncoder though due to the remaining instanceof checks there.

https://github.com/jsdom/webidl-conversions/blob/7d0cfd34a056a6d8b2ba20ea2c3782dfc3a220a9/lib/index.js#L301-L460 might be of interest (although I think these days we can use the `detached` getter to replace my `isArrayBufferDetached()`).

For jsdom's purposes all that is already handled by the webidl2js generated wrappers, so ideally `@exodus/bytes` wouldn't do those checks at all, but I can understand that's a pretty niche use case.

ChALkeR:
> still lots of failing tests for TextEncoder though

`TextEncoder.encodeInto` still has an instanceof check, will fix that too

ChALkeR:
I published [1.5.0](https://github.com/ExodusOSS/bytes/releases/tag/v1.5.0)
It should cover TextEncoder, though the approach is not resistant against something faking `[Symbol.toStringTag]`
(Not sure if we should care about that at all)

It is checked to be a single-byte wide view anyway

domenic:
Alright, encodeInto tests are now working.

Currently no action items for `@exodus/bytes`. I'll slowly work through the WPTs to see which we can afford to run on CI. Some of the tests depend on whatwg-url supporting more encodings than UTF-8, which it doesn't. Some of them are timing out, likely due to jsdom's iframe-loading stuff not being perfectly spec compliant... good times.

ChALkeR:
`Float16Array` does not exist in Node.js 20/22

I added a hack for that here in `@exodus/bytes` tests: https://github.com/ExodusOSS/bytes/blob/8844004/tests/wpt/fixtures/encoding/encodeInto.any.js#L137

Perhaps make that test conditional on float16arr?

Alternatively upstream that check to WPT? Missing Float16Array is not a TextEncoder issue

pjg:
Is there a reason `TextDecoder`/`TextEncoder` aren't exposed at the browser's `window` object as the other PR tried to do (https://github.com/jsdom/jsdom/pull/3791)? Without that exposure I am having a hard time figuring out how to make jsdom's ones available for my jest tests.

domenic:
They are exposed. If you can produce a minimal example following the issue template showing that they are not, please file it. (Not using any third-party libraries like Jest.)

pjg:
Upon further investigation it seems that my code relies on `jest-environment-jsdom`, which in turn relies on jsdom v26. When I added an override in package.json for `jest-environment-jsdom` to use jsdom v27, TextEncoder/TextDecoder became automatically available. However, my tests started running 10x slower (with some taking >60s to run). This looks like some incompatibility inside `jest-environment-jsdom` itself as I was unable to pinpoint it to a specific area in my app/tests. I don't think there's anything to be done here apart from `jest-environment-jsdom` upgrading/requiring jsdom v27. Sorry for confusion.

PR #4001, opened 2025-12-18
Add style performance benchmark

asamuzaK:
# **Description**

Related \#3985

This PR ports the benchmark test provided in issue \#3985 to the jsdom codebase. The goal is to facilitate testing and measuring the performance regression related to pure DOM style operations.

## **Implementation Details**

* **Ported Logic**: The benchmark logic from the issue has been adapted to run within the jsdom repository context.  
* **Standalone Execution**: Currently, this benchmark setup differs significantly from the existing environment in `benchmark/*`. Therefore, it has not been integrated into the main `benchmark/index.js` runner.  
* **Future availability**: It is currently undecided whether this benchmark should be permanently maintained in the repository after the resolution of \#3985. I am open to suggestions on this matter.

## **How to Run**

Since it is not hooked into the main runner, execute the script directly:  
```
node benchmark/style
```

PR #4008, opened 2025-12-27
Add style performance benchmark

domenic:
This is a port of #4001 to use our existing benchmark framework.

@asamuzaK I asked Claude Code to adapt your PR into one that uses the existing framework. Can you check to see if it matches your expectations?

It can be run with

```
npm run benchmark -- --suites style
```


asamuzaK:
LGTM
Thanks!

PR #4018, opened 2026-01-13
Make History API path handling defensive for future whatwg-url changes

ricardoBentoIQGeo:
# Make History API path handling defensive for future whatwg-url changes

Makes the `canHaveItsURLRewritten()` function handle both array and string path types to future-proof against potential `whatwg-url` implementation changes.

## Motivation

The `whatwg-url` package currently represents `URL.path` as an array, but this is an internal implementation detail. The History API code assumes this will always be an array and directly calls `.join("/")` on it. This could break if `whatwg-url` changes the internal representation in the future.

## Key changes

- Updated `canHaveItsURLRewritten()` to defensively check `Array.isArray(path)` before calling `.join("/")`
- Added test coverage for History API with custom URL schemes (non-http/https)

### DIFF
--- a/lib/jsdom/living/window/History-impl.js
+++ b/lib/jsdom/living/window/History-impl.js
@@ -140,7 +140,12 @@ function canHaveItsURLRewritten(document, targetURL) {
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


domenic:
This was not desired or requested. Please do not send such pull requests.

Issue #1393, opened 2016-02-25
XHR and resourceLoader


gustavnikolaj:
I was trying to use jsdom to facilitate server rendering of an application built with systemjs and react. I ended up doing that, as I needed to hook in with some existing tools for internationalization which were not built for running in the server.

My plan was to use jsdom to generate a window object and load the scripts, and then go load reacts renderToString method from that context. But I hit a problem when I realized that the custom resource loader is not used with XHR.

Is there a reason why the resource loader is not used in the XHR implementation?

domenic:
Nope; sounds like a bug! PR with tests certainly welcome.

kapouer:
Any update on this ? I'm willing to submit a PR, hopefully it's not that hard to setup.

domenic:
No updates; PR welcome!

kapouer:
Actually i only need to be able to hook into the "realClient" - maybe @gustavnikolaj too ? Replacing it asks for the custom resourceLoader to support so many things that it is bound to be useless.
Calling the custom resourceLoader with the realClient from xhr-utils would be just fine, allowing one to modify the client at will.

kapouer:
After a few tries: it's not complicated to add that feature indeed, but it asks for a change regarding what  `defaultFetch` is supposed to return.
The most powerful solution is to return the current request, but that's not what was done in the currently implemented version for `download`.
That allows one to have full control over the properly setup default request - and in the case of xhr it's needed because it's difficult to build a custom xhr request without missing something.
Also if `defaultFetch` must return a request instance, then `responseLoader` should too.
Maybe `responseLoader` could be polymorphic: return request if no callback is specified.

domenic:
I'd like to keep things backward compatible. However, if you have input into the new API resource loader, that would be good to consider.

gagern:
Would it make sense to add a method `fetchXHR` or similar to the `ResourceLoader` class? And document that to be compatible with the single argument version of `request`, in that it must return an `EventEmitter` which uses `response` and the `stream.Readable` events.

Having a new method with a suitable default implementation would ensure perfect backwards compatibility, but at the same time offer implementers an extension point to tweak behavior. 

The code for handling the `file:` and `data:` protocols could be moved to the implementation of that method. For `http:` and `https:` protocols, the new method would merely delegate to `request` with no further processing. It could therefore serve as a drop-in replacement for all the `request` calls in `xhr-utils.js`.

domenic:
[...]

PR #4023, opened 2026-01-18
Overhaul the resource loader API

### LINKED ISSUE #2500: Please support async usage of ResourceLoader.fetch
### Basic info:

- **Node.js version**: 8+
- **jsdom version**: jsdom@13.1.0

### Minimal reproduction case

```js

class ResourceLoaderAdapter extends jsdom.ResourceLoader {
    fetch(url, options) {
        return Promise.resolve(super.fetch(url, options));
    }
    
}

const options = {resources: new ResourceLoaderAdapter()};
const dom = jsdom.JSDOM.fromURL('http://google.com', options);
```

### Description

**Note**: the problem appears to be a show-stopper only within the implementation of `JSDOM.fromURL` since cancellation can still be supported externally.

Because the value returned by `fetch` is a "promise" overloaded with additional features that JSDOM.fromURL and other internals that support cancellation expect to directly consume, it is not possible to perform asynchronous work in fetch.  Wrapping the return value in another promise results in this error:
```
TypeError: Cannot read property 'headers' of undefined
    at req.then.body (*****/node_modules/jsdom/lib/api.js:133:35)
```
The ideal API definition would reduce external dependencies and avoid the complexity of asychronously returning a more-than-promise thenable like so:
```
fetch :: (url, options) → Promise(http.ClientRequest)
```
Users could then assume however much control they desire over the resolution of that request, including taking over the actual execution of the request using a wider range of popular request libraries.  (Though there's still the matter of externally using `_getRequestOptions`.)

A simpler change producing an also viable alternative would be:
```
fetch :: (url, options) → Promise(void → RequestPromise)
```
(i.e. fetch return a promise resolving to a 0-argument function that when executed returns the messy object created by `request-promise-native`).  The internals would still expose a heavier footprint, but at least async operations would be possible.

-----

This lack of async support is a considerable difficulty for me because I am trying to build a separate API to control and directly observe a (potentially mock and/or proxy) pluggable server (real `http.Server`) to which all requests are routed.  Without the ability to synchronize the client request with server request events - ensuring a matching 1-to-1 correspondence between them - I am unable to reliably link client requests to their server counterparts without excessive complexity _and_ hacks.

I consider it problematic in general that the mechanism is forcing me to extend an already complex enough class that hooks around the processing of an undocumented/externally defined object (the extended promise from `request-promise-native`) and doesn't seem to intentionally support overriding the actual resolution of the request (undocumented return value and implied privacy of `_getRequestOptions`).  The more I work with it, the more I wish I could just inject a pure function directly receiving all required data (including internal resolving function and requestOptions or a getter for it), sort of like middleware for Koa or Express.  That would really help with the composibility of independent extensions and save me a lot of grief with maintaining separation of concerns, keeping internals truly private, and preventing external mutation.

### LINKED ISSUE #3960: `JSDOM.fromURL` breaks with custom ResourceLoader.fetch when initial request is not handled by the base fetch method
### Node.js version

24.8.0

### jsdom version

27.0.1

### Minimal reproduction case

```js
let testStr = `<!doctype html><meta charset=utf-8><title>test</title>`;
class CustomResourceLoader extends jsdom.ResourceLoader {
  fetch(url, options) {
    //when initial request is not handled by super.fetch
    if(url=="https://example.com/") {
        return Promise.resolve(Buffer.from(testStr));
    }
    return super.fetch(url, options);
  }
}
const {JSDOM} = jsdom;

JSDOM.fromURL("https://example.com", {
  resources: new CustomResourceLoader()
}).then((dom) => {
    console.log(dom.serialize())
})
```

N/A

### What is the problem?

The `fetch` method of the original `ResourceLoader` attaches a `response` property to the returned `Promise`. `JSDOM.fromURL` relies on the properties' existence to set the `contentType` option, which breaks if the custom `ResourceLoader` has not replicated this behavior.  While the documentation mentions that the resulting jsdom's content type is determined from the response, it might be more robust for fromURL to not depend on this property directly.

Expected result:
```
<!DOCTYPE html><html><head><meta charset="utf-8"><title>test</title></head><body></body></html>
```
Actual result:
```
TypeError: Cannot read properties of undefined (reading 'headers')
```

domenic:
This is not yet ready, but it's time to start doing full CI runs and writing a proper PR description, so here's a draft PR.

----

Change the `ResourceLoader` API to be less fragile and more modern. Notably:

* Instead of returning promises for `Uint8Array`s, which sometimes had extra properties tacked on to the promises like `abort()` or `getHeader()` or `response`, return a promise for a Node.js global `Response` object. Fixes #3960 and fixes #2500.
* Change the arguments to be more like the arguments of the standard `fetch()` API, i.e., a url and some options.
* Change connection-level customization to no longer be done via `strictSSL` and `proxy` options, but instead via custom `undici` `Dispatcher`s, which appears to be the modern way this is done in Node.js. Closes #3994 by superseding it. Notably, previously jsdom would create new HTTP/HTTPS `Agent`s for every request, which is wasteful, especially since agents cannot be cleaned up via garbage collection.

Stemming from this fundamental API change come a few other breaking changes:

* We now set the correct referrer for redirect chains, e.g. when A fetches B which redirects to C, the redirect is left as A, instead of incorrectly being set to B.
* Sync XHR no longer supports connection customization, since we cannot send a `Dispatcher` object across processes.
* `WebSocket`s are no longer properly throttled to a single connection per host, due to a bug in the underlying library which we are moving to (https://github.com/nodejs/undici/issues/4743).

Under the hood, a lot of stuff needs to change, mainly because of the decision to accept undici `Dispatcher`s instead of just `{ proxy, strictSSL }` options, or older-style Node.js `Agent`s. We need to rebase all our fetching on top of undici and/or native Node.js `fetch()`:

* This is relatively straightforward for fetch-the-full-body resources like stylesheets, scripts, and images.
* This required replacing the `ws` package with undici's `WebSocket`.
* This required much more work for `XMLHttpRequest`. In particular, the XHR API unfortunately requires some functionality not provided by `fetch()`, such as setting headers with trailing whitespace. (And we couldn't use undici's `request()` either, since it doesn't expose `statusText`. We had to drop all the way down to `dispatch()`.)

---

Still TODO: try to make XHR go through the resource loader's fetch hook. I'm not sure how possible that will be honestly... see #1393. XHR needs much more lower-level stuff than the rest of jsdom.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -46,6 +46,9 @@ export default [
       "prefer-template": "off",
       "new-cap": ["error", { capIsNewExceptions: ["ByteString", "USVString", "DOMString"] }],
 
+      // probably disable this in @domenic/eslint-config; throwing from async functions is convenient
+      "require-await": "off",
+
       // Custom rules
       "jsdom-internal/hook-super-invocation": [
         "error",
--- a/lib/jsdom/browser/resources/resource-loader.js
+++ b/lib/jsdom/browser/resources/resource-loader.js
@@ -48,15 +48,13 @@ class ResourceLoader {
    * @returns {Promise<Response>} A promise resolving to a `Response` (from the Node.js global)
    */
   // eslint-disable-next-line no-unused-vars -- element is for subclasses
-  fetch(url, { method, headers, body, referrer, signal, cookieJar, element } = {}) {
-    if (signal?.aborted) {
-      return Promise.reject(signal.reason ?? new DOMException(`Fetch for ${url} was aborted`, "AbortError"));
-    }
+  async fetch(url, { method, headers, body, referrer, signal, cookieJar, element } = {}) {
+    signal?.throwIfAborted();
 
     const urlRecord = parseURL(url);
 
     if (!urlRecord) {
-      return Promise.reject(new Error(`Tried to fetch invalid URL ${url}`));
+      throw new Error(`Tried to fetch invalid URL ${url}`);
     }
 
     switch (urlRecord.scheme) {
@@ -74,7 +72,7 @@ class ResourceLoader {
       }
 
       default: {
-        return Promise.reject(new TypeError(`Tried to fetch URL ${url} with invalid scheme ${urlRecord.scheme}`));
+        throw new TypeError(`Tried to fetch URL ${url} with invalid scheme ${urlRecord.scheme}`);
       }
     }
   }
@@ -83,6 +81,7 @@ class ResourceLoader {
     // Add an artificial delay to match the specs, where data: URL processing is done from a full task (not just a
     // microtask).
     await delay(0, { signal });
+    signal?.throwIfAborted();
 
     const dataURL = dataURLFromRecord(urlRecord);
     if (dataURL === null) {
@@ -100,6 +99,7 @@ class ResourceLoader {
   async _readFile(url, signal) {
     const filePath = fileURLToPath(url);
     const body = await fsPromises.readFile(filePath, { signal });
+    signal?.throwIfAborted();
 
     return new Response(body, {
       status: 200,


domenic:
Alright. This code is mostly good now, and all tests pass (with several new progressions).

However, I think we can get more ambitious, by getting rid of ResourceLoader in favor of letting developers use undici interceptors directly. That will also ensure XHRs are interceptable.

PR #4024, opened 2026-01-19
Reimplement XHR from scratch on top of resource loader overhaul

domenic:
This is Codex 5.2's attempt when given the following prompt

> As you can see from previous commits on this resource-loader-overhaul branch, a previous agent has been making a lot of progress on trying to migrate our XMLHttpRequest implementation to a new framework.
> 
> However, it's been getting quite messy. In this worktree, we're going to try a different approach and re-code all of XHR from scratch, with nicer abstraction boundaries.
> 
> In particular, what you should be proudly doing (and I've seen you do well at before) is creating jsdom code which matches the specification steps as exactly as possible. Don't go off of your memory of what the spec says, as the specs are Living Standards and change relatively often. Instead, use the source files I've mentioned below.
> 
> The strategy I suggest is to gut everything and start fresh from a skeleton XMLHttpRequest-impl.js. Then, implement https://xhr.spec.whatwg.org/ (source code https://raw.githubusercontent.com/whatwg/xhr/refs/heads/main/xhr.bs) from the API layer downward. You should almost certainly throw away everything in the current XMLHttpRequest-impl.js, xhr-utils.js, and xhr-sync-worker.js.
> 
> This will eventual include implementing much of https://fetch.spec.whatwg.org/ (source code https://raw.githubusercontent.com/whatwg/fetch/refs/heads/main/fetch.bs). Not the public API, but the underlying primitives.
> 
> One relevant file that is probably usable as-is is lib\\jsdom\\living\\fetch\\header-types.js. You'll likely add some code in there as you need it.
> 
> To be a bit more concrete, I think there will be basically the following main sources of complexity:
> 
> - Incidental API-level stuff in implementing XHR, e.g. all the readyState checking, input normalization, etc. Like in setRequestHeader() or overrideMimeType() or similar.
> 
> - The main send() method and how it interfaces with the Fetch spec. Notice how it uses Fetch's hooks: processBodyChunk, processEndOfBody, processBodyError.
> 
> - Synchronous XHR. The only way to really make this work is to make most of the XHR object's guts serializable, then transfer them to a separate worker process and use childProcess.spawnSync() to get the results back. This is the same strategy our current code uses (see xhr-sync-worker.js), although I suspect the marshalling code could be a bit cleaner.
> 
> - Actually implementing the relevant parts of the Fetch, in a way that provides the appropriate processBodyChunk, processEndOfBody, processBodyError hooks. This will require pretty low-level access through the undici library. https://raw.githubusercontent.com/nodejs/undici/250efc8eab72086384fe91b5ee3adbcd9d45dac0/lib/web/fetch/index.js which implements the `fetch()` API using undici's low-level abilities is probably a good reference. We should feel free to copy a bunch of it. The current working directory's code, from the previous agent, does some of this, but in an ad-hoc fashion with lots of confusing conversion layers.
> 
> A failure mode to avoid here is too much converting between different data formats. E.g. jsdom `Headers`, Node.js global `Headers`, undici `Headers`, jsdom `HeadersList`, or different JSON serializations. Pick a single one and stick with it. (In this particular case, you'll need something flexible that preserves case, does not canonicalize, and can be transferred across subprocess boundaries, so probably something like raw arrays of `[[name1, [value1, value2]], [name2, [value2]], ...]`.) Other cases are Node.js/WHATWG `ReadableStream`, Node.js `Readable`, and anything jsdom-specific; I suspect just sticking to Node.js `Readable` will be the best bet, since it lets us leverage, e.g., the `zlib` module for decompression.
> 
> Running tests is a delicate operation here. Please read CLAUDE.md for some instructions on this, geared at the other agent, but also applicable to you. Our north star is to get all of `npm run test:wpt -- --fgrep xhr` passing, but that takes a long time so during initial iteration that won't be the right move.

The internet rumors are true: Codex is fairly good at putting its head down and just churning out code in beast mode. It got even more tests passing. I just repeatedly kept saying "approve" (because [its permissions model is terrible](https://x.com/domenic/status/2013134645156630968)).

I haven't looked at the code in detail yet, and from what I've skimmed it's not quite mergeable. But the high-level directive of reducing abstraction layers and creating clean hooks seems to have been followed.


domenic:
An interesting experiment, but after iterating on a few rounds of simplification with Claude, including telling it to look at this branch and adopt what it thinks is valuable, I don't think I'll end up using any of this.

PR #4040, opened 2026-02-05
test: add regression test for window instanceof Window (GH-2740)

### LINKED ISSUE #2740: `window instanceof Window` is false
### Basic info:

- **Node.js version:** 10.17.0
- **jsdom version:** 15.2.1

### Minimal reproduction case

```js
const { JSDOM, VirtualConsole } = require('jsdom');

virtualConsole.sendTo(console);
const dom = new JSDOM(`
  <body>
    <script>
      console.log(window instanceof Window);
    </script>
  </body>
`, {
  runScripts: 'dangerously',
  virtualConsole,
});
```

The above script outputs `false` to the console.

`window instanceof Window` outputs `true` in Chrome.

Smerly:
Adds a regression test for [GH-2740](https://github.com/jsdom/jsdom/issues/2740): `window instanceof Window` is false when scripts run with `runScripts: 'dangerously'`.

## Background

In browsers, `window instanceof Window` is `true`. In older jsdom/Node setups, it could be `false` because of how the VM global proxy and the Window prototype were set up. With `vm.createContext(vm.constants.DONT_CONTEXTIFY)` (Node 20.18+), the context global and `globalThis` are the same object, so the existing prototype setup in `lib/jsdom/browser/Window.js` already makes `window instanceof Window` true.

## Changes

- **Test:** In `test/api/options-run-scripts.js`, add a case that runs a `<script>` with `runScripts: 'dangerously'` which sets `window.__windowInstanceofWindow = window instanceof Window`, then assert `dom.window.__windowInstanceofWindow === true`.

This locks in the correct behavior so future changes don’t regress GH-2740.

fixes #2740.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -181,7 +181,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -220,7 +219,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -559,7 +557,6 @@
       "integrity": "sha512-NZyJarBfL7nWwIq+FDL6Zp/yHEhePMNnnJ0y3qfieCrmNvYct8uvtiV41UvlSe6apAfk0fY1FbWx+NwfmpvtTg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "acorn": "bin/acorn"
       },
@@ -1089,7 +1086,6 @@
       "integrity": "sha512-LEyamqS7W5HB3ujJyvi0HQK/dtVINZvd5mAAp9eT5S/ujByGjiZLCzPcHVzuXbpJDJF/cxwHlfceVUDZ2lnSTw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@eslint-community/eslint-utils": "^4.8.0",
         "@eslint-community/regexpp": "^4.12.1",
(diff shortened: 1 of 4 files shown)


scripts/webidl/convert.js
```diff
@@ -73,8 +73,50 @@ const outputDir = path.resolve(__dirname, "../../lib/jsdom/living/generated/");
 fs.rmSync(outputDir, { force: true, recursive: true, maxRetries: 2 });
 fs.mkdirSync(outputDir);
 
+function applyGH2265Patches() {
```
domenic, line 76: What the heck is this code doing. Maybe some kind of supply chain hack attempt?

test-gh-2265-proof.js
```diff
@@ -0,0 +1,52 @@
+/**
```
domenic, line 1: This is not how we write tests in jsdom.

package-lock.json
domenic, line 1: Don't modify this file.

test/api/options-run-scripts.js
```diff
@@ -36,6 +36,25 @@ describe("API: runScripts constructor option", () => {
...
     });
 
+    it("should make window instanceof Window true from inside scripts (GH-2740)", () => {
+      const dom = new JSDOM(`<body>
+        <script>window.__windowInstanceofWindow = window instanceof Window;</script>
+      </body>`, { runScripts: "dangerously" });
+      assert.strictEqual(dom.window.__windowInstanceofWindow, true, "window instanceof Window must be true in browsers");
+    });
+
+    it("should throw when DOM methods are called through a Proxy (GH-2265)", () => {
```
domenic, line 46: Per CONTRIBUTING.md, this should be a to-upstream web platform test.

PR #4046, opened 2026-02-07
Refactor style handling: Improve code organization and cache initialization

asamuzaK:
1. **Window.js improvements:**
   - Updated `vm` import to use Node.js built-in `node:` prefix
   - Renamed `elt` to `elementImpl` for clarity and consistency
   - Restructured pseudo-element validation logic
   - Changed `getComputedStyle` declaration to be read-only
   - Removed shadow DOM pseudo-element handler
2. **Document-impl.js improvements:**
   - Initialize `_styleCache` as a `WeakMap` instead of `null`, improving cache consistency from the start
3. **style-rules.js refactoring:**
   - Updated `fs` and `path` imports to use Node.js built-in `node:` prefix
   - Reordered `require()`s
   - Flattened nested functions
   - Changed module exports to use explicit `module.exports` object
   - Updated `invalidateStyleCache` to properly initialize the cache as a `WeakMap` instead of `null`

### DIFF
--- a/lib/jsdom/living/helpers/style-rules.js
+++ b/lib/jsdom/living/helpers/style-rules.js
@@ -4,6 +4,7 @@ const fs = require("node:fs");
 const path = require("node:path");
 const cssom = require("@acemir/cssom");
 const { CSSStyleDeclaration } = require("cssstyle");
+const { wrapperForImpl } = require("../generated/utils");
 const { getSpecifiedColor, getComputedOrUsedColor } = require("./colors");
 const { asciiLowercase } = require("./strings");
 const { deprecatedAliases, systemColors } = require("./system-colors");
@@ -108,17 +109,16 @@ const propertiesWithResolvedValueImplemented = {
 };
 
 function getDeclarationForElement(elementImpl) {
-  // Prepare options for cssstyle.
+  // TODO: Prepare options for cssstyle.
   const opts = {
-    context: elementImpl,
-    computed: true
+    context: wrapperForImpl(elementImpl)
   };
 
   const styleCache = elementImpl._ownerDocument._styleCache;
   const cachedDeclaration = styleCache.get(elementImpl);
   if (cachedDeclaration) {
     const clonedDeclaration = new CSSStyleDeclaration(null, opts);
-    // TODO: Remove later.
+    // TODO: Remove when options are ready.
     clonedDeclaration._global = elementImpl._globalObject;
     clonedDeclaration._computed = true;
 
@@ -133,7 +133,7 @@ function getDeclarationForElement(elementImpl) {
   }
 
   const declaration = new CSSStyleDeclaration(null, opts);
-  // TODO: Remove later.
+  // TODO: Remove when options are ready.
   declaration._global = elementImpl._globalObject;
   declaration._computed = true;
 


asamuzaK:
Re https://github.com/jsdom/jsdom/pull/2918
``` js
if (SHADOW_DOM_PSEUDO_REGEXP.test(pseudoElt)) {
  throw new TypeError("Tried to get the computed style of a Shadow DOM pseudo-element.");
}
```
This implementation appears to be incorrect.
Neither Firefox nor Chrome throws an exception in ::part() or ::slotted().
Should we remove?

asamuzaK:
<a href="https://drafts.csswg.org/cssom/#:~:text=If%20type%20is%20failure%2C%20or%20is%20a%20%3A%3Aslotted()%20or%20%3A%3Apart()%20pseudo-element%2C%20let%20obj%20be%20null.%20" rel="noopener">If type is failure, or is a ::slotted() or ::part() pseudo-element, let obj be null.</a>

So, it does not say to throw.
Removed code block and test.

lib/jsdom/browser/Window.js
```diff
@@ -908,38 +907,33 @@ function installOwnProperties(window, options) {
...
+    const elementImpl = Element.convert(window, elt);
+
     if (pseudoElt !== undefined && pseudoElt !== null) {
       pseudoElt = webIDLConversions.DOMString(pseudoElt);
-    }
-
-    if (pseudoElt !== undefined && pseudoElt !== null && pseudoElt !== "") {
-      // TODO: Parse pseudoElt
 
-      if (SHADOW_DOM_PSEUDO_REGEXP.test(pseudoElt)) {
```
domenic, line 919: I think this was trying to implement https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle step 3.2.

Maybe in the past the spec required a TypeError.

It would be best if you separated the parts of this PR that changed behavior into a separate PR.

lib/jsdom/browser/Window.js
```diff
@@ -908,38 +907,33 @@ function installOwnProperties(window, options) {
...
 
-    const elementDeclaration = getDeclarationForElement(elt);
-    forEach.call(elementDeclaration, property => {
-      declaration.setProperty(
-        property,
-        elementDeclaration.getPropertyValue(property),
-        elementDeclaration.getPropertyPriority(property)
-      );
-    });
+    // Set up a temporary method to toggle the read-only flag
```
domenic: Why do you have a temporary method instead of just setting `declaration._readonly` directly?

lib/jsdom/browser/Window.js
```diff
@@ -908,38 +907,33 @@ function installOwnProperties(window, options) {
...
+    declaration.readOnly = function (bool = false) {
+      this._readonly = bool;
+    };
+    declaration.readOnly(false);
 
     // https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle
     const declarations = Object.keys(propertiesWithResolvedValueImplemented);
```
domenic: Maybe while here we should hoist out this `Object.keys()` call to the top of the file. (Although I guess you are planning to refactor this soon anyway.)

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -11,15 +12,12 @@ const defaultStyleSheet = fs.readFileSync(
   path.resolve(__dirname, "../../browser/default-stylesheet.css"),
   { encoding: "utf-8" }
 );
-
-const { forEach, indexOf } = Array.prototype;
-
-let parsedDefaultStyleSheet;
+const parsedDefaultStyleSheet = cssom.parse(defaultStyleSheet);
```
domenic: Please keep this as lazily parsed, to reduce jsdom startup time.

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -109,90 +107,91 @@ exports.propertiesWithResolvedValueImplemented = {
...
-    } else {
-      declaration.setProperty(
-        property,
-        value,
-        style.getPropertyPriority(property)
-      );
-    }
-  }
+  const declaration = new CSSStyleDeclaration(null, {
+    context: elementImpl._globalObject
```
domenic: I am surprised we don't set it to the element wrapper, but, I guess that would be a behavior change we are not ready for.

domenic (commented):
Sorry for the delay; I have caught a cold.

asamuzaK:
> Sorry for the delay; I have caught a cold.

I don't mind being late.
More importantly, are you feeling better now?

lib/jsdom/browser/Window.js
asamuzaK: You're right. I guess I was overthinking it.

lib/jsdom/living/helpers/style-rules.js
asamuzaK: I don't quite understand what you're trying to say here.
Are you saying we need something to be changed from the proposal?

lib/jsdom/browser/Window.js
asamuzaK: Reverted.

lib/jsdom/living/helpers/style-rules.js
asamuzaK: Reverted.

lib/jsdom/browser/Window.js
asamuzaK: Done.

lib/jsdom/living/helpers/style-rules.js
domenic: Sorry, let me give more detail. Related to https://github.com/jsdom/cssstyle/blob/93ec4c6ab0bcdbeabc11da2def1a52989e215226/lib/CSSStyleDeclaration.js#L49-L67 .

Before we did `new CSSStyleDeclaration()`. There is no context association.

Now we do `new CSSStyleDeclaration(null, { context: globalObject })`. This sets the `_global`, and sets `_computed` to true.

I thought the better option would be `new CSSStyleDeclaration(null, { context: elementWrapper })`. That would set `_global` and `_ownerNode`. But, I guess it would not set `_computed` to true, which is bad...

The spec https://drafts.csswg.org/cssom/#the-cssstyledeclaration-interface says we should set computed to true and owner node to the element. I guess that is not possible with the current `cssstyle` package.

---

So, your proposal in this PR is an improvement over the current jsdom code. But, maybe we should also work on setting `_ownerNode` correctly. We can do that in a separate PR, since this PR is supposed to be mostly refactoring/style changes.

Does that sound like a good plan? Or am I misunderstanding something?

domenic: Wait. Actually, I think this PR changes something pretty critical. Now you are returning the same `CSSStyleDeclaration` object from `getComputedStyle()` and from `element.style`, because of the cache. That is not correct. `getComputedStyle()` is supposed to return a new object every time. And that object is supposed to be different from the `element.style` object.

- `element.style` object: computed = false, readonly = false, parent CSS rule = null, owner node = element
- `getComputedStyle()` object: computed = true, readonly = true, parent CSS rule = null, owner node = element

domenic:
> I don't mind being late.
> More importantly, are you feeling better now?

Thanks for your understanding. I am slightly better than yesterday :)

lib/jsdom/living/helpers/style-rules.js
asamuzaK: Re: style cache
* The file name style-rules.js can be confusing to begin with, but that aside, this file does specific process to getComputedStyle(). 
  It tries to resolve specified values ​​to computed values ​​(or used values).
  * I'm planning to rename this file to something like computed-style.js later.
* styleCache is only set within getDeclarationForElement(), and this method is only called from getComputedStyle().
  Element.style is not cached in _ownerDocument._styleCache from anywhere else.
  _ownerDocument._styleCache only caches computed style declarations.
* Currently, the value of getComputedStyle() is resolved based on propertiesWithResolvedValueImplemented, which is very limited. 
  Because of this limitation, in many cases the specified value is returned as is, but this will be fixed in the future.
* Computed values should actually be resolved within cssstyle.
  And at some point in future updates, propertiesWithResolvedValueImplemented will become completely unnecessary.

Re: context
* Thanks, now I understand what you're saying.
  We need to change the options on the cssstyle side, such as `{ context: ElementImpl, computed: true }`.
* The final form of CSSStyleDeclaration will be as follows:
  ```
  constructor (globalObject, args = [], privateData = {
    // cssstyle options here
  }) {}
  ```
* As a preliminary step, I plan to figure out and collect the necessary options while maintaining compatibility with the current cssstyle implementation:
  ```
  constructor(onChangeCallback, {
    context = {}, // Element, CSSRule
    computed = false, // True if window.getComputedStyle
    // and more options
  }) {}
  ```

domenic: Thanks for explaining the current system in more detail. I see now that `element.style` is completely separate and doesn't seem to use the style cache at all. So that is good. (I was confused by the name `getStyleDeclarationForElement()`... I thought it would be related to `element.style`.)

But, I think there still might be a regression in this PR. By my reading of the code, previously, `window.getComputedStyle(document.body) !== window.getComputedStyle(document.body)` would hold. That is correct behavior. Because in the previous code, every call to `window.getComputedStyle()` made a new call to `new CSSStyleDeclaration()`. It would use the cache to get some information to copy over, but it would always create a new object.

In the new code after this PR, the style cache can be used directly. So it will not correctly return a new object each time.

Can you add a regression test for this as part of the PR, and make sure it passes?

asamuzaK: > window.getComputedStyle(document.body) !== window.getComputedStyle(document.body) would hold. That is correct behavior.

Ah, this PR does not do so.
I'll fix it and add test.

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -109,90 +107,113 @@ exports.propertiesWithResolvedValueImplemented = {
...
-
-exports.getDeclarationForElement = elementImpl => {
-  let styleCache = elementImpl._ownerDocument._styleCache;
-  if (!styleCache) {
-    styleCache = elementImpl._ownerDocument._styleCache = new WeakMap();
-  }
+function getDeclarationForElement(elementImpl) {
+  // Prepare options for cssstyle.
+  const opts = {
+    context: elementImpl,
```
domenic: It probably will want the wrapper, not the impl?

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -109,90 +107,113 @@ exports.propertiesWithResolvedValueImplemented = {
...
+    computed: true
```
domenic: This doesn't do anything right now, right? It's related to the TODO. Probably best to remove it for now.

PR #4049, opened 2026-02-11
Fix assert.rejects() calls that were not being awaited

domenic:
Extracted from https://github.com/jsdom/jsdom/pull/4005 with a small fix to match the actual error message.


asamuzaK:
We can close #4005 ?

Issue #4047, opened 2026-02-10
Dependency problems with Node and Undici


KarlGe:
Node version: 22.13.1
jsdom version: 28.0.0

We recently updated jsdom to the latest version that is using undici. We're using jsdom in testing with vitest.

Sometimes in one of our tests we get this error thrown from https://github.com/nodejs/undici/blob/0ce57bad2c171a16c043c8234317ef5383f86e0b/lib/handler/unwrap-handler.js#L90

```
InvalidArgumentError: invalid onError method
 ❯ Agent.dispatch node:internal/deps/undici/undici:591:19
 ❯ innerDispatch node_modules/jsdom/lib/jsdom/browser/resources/jsdom-dispatcher.js:573:35
 ❯ node_modules/jsdom/lib/jsdom/browser/resources/decompress-interceptor.js:179:14
 ❯ innerDispatch node_modules/jsdom/lib/jsdom/browser/resources/jsdom-dispatcher.js:580:61
 ❯ node_modules/jsdom/lib/jsdom/browser/resources/jsdom-dispatcher.js:478:7
 ❯ JSDOMDispatcher.#doSingleRequest node_modules/jsdom/lib/jsdom/browser/resources/jsdom-dispatcher.js:471:12
 ❯ JSDOMDispatcher.#dispatchHTTP node_modules/jsdom/lib/jsdom/browser/resources/jsdom-dispatcher.js:265:49
 ❯ JSDOMDispatcher.dispatch node_modules/jsdom/lib/jsdom/browser/resources/jsdom-dispatcher.js:92:23
 ❯ node_modules/jsdom/lib/jsdom/living/xhr/xhr-utils.js:158:16
```

It seems to me the problem is that [getGlobalDispatcher in api.js](https://github.com/jsdom/jsdom/blob/4d9461b42021ca08722eb5fcf802c4a61224b765/lib/api.js#L194) gets the dispatcher defined by the version of undici from Node instead of the version defined in jsdom.

I've created a simple repro with a node project that just installs the latest version on undici, using Node 22.13.1
```
{
  "name": "jsdom-undici",
  "version": "1.0.0",
  "main": "index.js",
  "scripts": {
    "start": "node ./index.js"
  },
  "dependencies": {
    "undici": "^7.21.0"
  },
  "type": "module"
}
```

```
import { DecoratorHandler, getGlobalDispatcher } from "undici";

const handler = new DecoratorHandler({});

const opts = {
  method: "GET",
[...]

PR #4056, opened 2026-02-12
Fix undici v6/v7 cross-version global dispatcher compatibility

domenic:
Fixes https://github.com/jsdom/jsdom/issues/4047. Still discussing at https://github.com/nodejs/undici/discussions/4822 where the best fix should live.


domenic:
The tradeoff here is taking on one more dependency on undocumented Undici private APIs, in favor of a smaller diff. (The alternative was duplicating WrapHandler and UnwrapHandler in the jsdom source tree.)

I think this is acceptable as a temporary workaround, until Undici vNext.

Issue #3318, opened 2022-01-15
`getComputedStyle()` ignores specificity priority


danon:
### Basic info:

- **Node.js version:** 16.13.0
- **jsdom version:** 19.0.0

1. Use any styles, that depend on specificy
   ```css
   <style>
    .parent p { color: black; }  // this should take precedense, since it's more specific, according to CSS
    p { color: red; }
   </style>
   ```
2. Use them in HTML

The CSS specificity rules are described: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

### Minimal reproduction case

```js
var JSDOM = require("jsdom").JSDOM;

const dom = new JSDOM(`<!DOCTYPE html>
<style>
 div.parent p { color: black; }
 p { color: red; }
</style>

<div class="parent">
    <p>Hi</p>
</div>
`);

const style = dom.window.getComputedStyle(dom.window.document.querySelector("p"));

console.log(style);

```

### How does similar code behave in browsers?

In every browser, `getComputedStyle()` returns `color: black`, in JSDOM it's `color: red`.

![image](https://user-images.githubusercontent.com/13367735/149621483-a63f4730-1779-409b-ab4d-48be33aaf4eb.png)
![image](https://user-images.githubusercontent.com/13367735/149621510-628a5909-bc7a-4e3f-a58a-e0f7729f1f30.png)
![image](https://user-images.githubusercontent.com/13367735/149621526-696dea3a-1633-4d87-a035-cfaf6c127f1b.png)

PR #4059, opened 2026-02-14
Apply CSS specificity when computing styles

asamuzaK:
Fix https://github.com/jsdom/jsdom/issues/3318 and some wpt tests regarding specificity.

* Add @bramus/specificity to compute and compare selector specificities so stylesheet rules are applied in specificity order.
* Refactor handleRule/handleProperty to consider inline declarations, property priorities, and selector specificity.
* Simplify matches() selector checking.
* Includes minor formatting and argument-passing cleanup.
* Replaced `test/web-platform-tests/to-upstream/css/cssom/getComputedStyle-specificity.html` with `test/web-platform-tests/to-upstream/css/selectors/selector-specificity.html`.

### DIFF
--- a/lib/jsdom/living/helpers/style-rules.js
+++ b/lib/jsdom/living/helpers/style-rules.js
@@ -148,11 +148,11 @@ function prepareComputedStyleDeclaration(elementImpl, { styleCache }) {
   declaration._global = elementImpl._globalObject;
   declaration._computed = true;
 
-  applyStyleSheetRules(elementImpl, declaration, new Map());
+  applyStyleSheetRules(elementImpl, declaration);
 
   for (let i = 0; i < elementImpl.style.length; i++) {
-    const property = elementImpl.style[i];
-    handleProperty(property, declaration, elementImpl.style);
+    const { style } = elementImpl;
+    handlePropertyForInlineStyle(style.item(i), declaration, style);
   }
 
   styleCache.set(elementImpl, declaration);
@@ -169,11 +169,12 @@ function prepareCssstyleOpts(elementImpl) {
   return opts;
 }
 
-function applyStyleSheetRules(elementImpl, declaration, specificities) {
+function applyStyleSheetRules(elementImpl, declaration) {
   if (!parsedDefaultStyleSheet) {
     parsedDefaultStyleSheet = cssom.parse(defaultStyleSheet);
   }
 
+  const specificities = new Map();
   handleSheet(parsedDefaultStyleSheet, elementImpl, declaration, specificities);
   for (const sheet of elementImpl._ownerDocument.styleSheets._list) {
     handleSheet(sheet, elementImpl, declaration, specificities);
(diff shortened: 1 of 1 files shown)


domenic (commented):
Amazing work!! I'm so excited to get specificity support in jsdom after all these years.

I have one higher-level organization issue we should sort out as part of this. I'd like us to move away from the `opt` argument pattern. The existing code had a simple version of this, with `{ declaration }` threaded through the call chain, but this PR significantly expands it. This makes the data flow hard to follow and fragile:

- `prepareComputedStyleDeclaration` creates `{ declaration, specificities }` and passes it to `applyStyleSheetRules`
- `applyStyleSheetRules` passes the same `opt` through to `handleSheet`
- `handleSheet` mutates `opt.ast = ast` before passing to `handleRule`
- `handleRule` mutates `opt.style = rule.style` before passing to `handleProperty`
- `handleProperty` finally destructures `{ ast, declaration, inline, specificities, style }`: five fields, set at four different layers

I think a cleaner approach would be to use explicit parameters for each function, so each one receives only what it needs. Here's how I'd suggest restructuring the existing functions:

**`prepareComputedStyleDeclaration`**: Pass just `declaration` to `applyStyleSheetRules` (no opts bag at all). Handle inline styles separately afterward. Since inline styles always win over non-`!important` stylesheet declarations, they don't need specificity tracking, so you can just call `declaration.setProperty` directly in the loop (with appropriate `!important` handling) instead of routing through `handleProperty` with an `inline: true` flag.

**`applyStyleSheetRules`**: Accept `(elementImpl, declaration)` as direct parameters. Create `specificities = new Map()` internally. That's an implementation detail, not something the caller should need to know about. Pass `declaration` and `specificities` down to `handleSheet`.

**`handleSheet`**: Accept `(sheet, elementImpl, declaration, specificities)`. When a rule matches, pass the `ast` returned by `matches()` directly to `handleRule` as a parameter, instead of stuffing it into a shared object.

**`handleRule`**: Accept `(rule, ast, declaration, specificities)`. Pass `rule.style` directly to `handleProperty` as a parameter, instead of mutating a shared object.

**`handleProperty`**: Accept `(property, style, ast, declaration, specificities)`. All the data it needs arrives as explicit parameters, with no reliance on a shared mutable bag.

This way each function's signature documents exactly what data it needs, there's no shared mutable object being threaded and mutated across layers, and implementation details like `specificities` are created where they're used rather than by a distant caller.

WDYT?

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -149,14 +148,11 @@ function prepareComputedStyleDeclaration(elementImpl, { styleCache }) {
   declaration._global = elementImpl._globalObject;
   declaration._computed = true;
 
-  applyStyleSheetRules(elementImpl, { declaration });
+  applyStyleSheetRules(elementImpl, declaration, new Map());
```
domenic: The caller should not need to create the map for this function.

lib/jsdom/living/helpers/style-rules.js
```diff
@@ -201,71 +197,81 @@ function handleSheet(sheet, elementImpl, opt) {
...
-    // https://drafts.csswg.org/css-cascade-4/#valdef-all-unset
-    if (value === "unset") {
-      declaration.removeProperty(property);
-    } else {
-      declaration.setProperty(
-        property,
-        value,
-        stylePriority
-      );
+function handleProperty(property, declaration, style, specificities, ast) {
```
domenic, line 233: I think this would be clearer split into two functions, handleProperty() and handlePropertyForInlineStyle().

domenic (commented):
Awesome, only minor suggestions remaining.

lib/jsdom/living/helpers/style-rules.js
asamuzaK: Is this what you mean?
```
function applyStyleSheetRules(elementImpl, declaration) {
  if (!parsedDefaultStyleSheet) {
    parsedDefaultStyleSheet = cssom.parse(defaultStyleSheet);
  }

  const specificities = new Map();
  handleSheet(parsedDefaultStyleSheet, elementImpl, declaration, specificities);
  for (const sheet of elementImpl._ownerDocument.styleSheets._list) {
    handleSheet(sheet, elementImpl, declaration, specificities);
  }
}
```

domenic: Yes, exactly.

domenic (approved):

Issue #3563, opened 2023-06-06
getComputedStyle does not work with css variables on the :root


SubliemeSiem:
### Basic info:

- **Node.js version:** 16.16.0
- **jsdom version:** 22.0.0

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");
const window = jsDom.window;
const style = window.document.createElement('style');
style.innerHTML = ':root { --test-var: red }';
window.document.body.appendChild(style);
const computedStyles = window.getComputedStyle(window.document.body);
console.log('css var: ' + computedStyles.getPropertyValue('--test-var'));
```

Output in console:

```
css var:
```

### How does similar code behave in browsers?

Same code in the browser shows
```
css var: red;
```

code pen: https://codepen.io/S-Peeters/pen/zYMYYaY?editors=1111

SubliemeSiem:
Also, using the variable on an element property and getting the computed style for that element property does not work either.
```
window.document.body.style.setProperty('color', '--test-var');
console.log(getComputedStyle(window.document.body).color);
```
does not log "red" or "rgba(255, 0, 0)", which it does log in the browser

PR #4068, opened 2026-02-15
Gather custom properties from :root rules

asamuzaK:
Fixes https://github.com/jsdom/jsdom/issues/3563

### DIFF
--- a/lib/jsdom/living/helpers/style-rules.js
+++ b/lib/jsdom/living/helpers/style-rules.js
@@ -182,6 +182,7 @@ function applyStyleSheetRules(elementImpl, declaration) {
 }
 
 function handleSheet(sheet, elementImpl, declaration, specificities) {
+  const rootElementImpl = elementImpl._ownerDocument.documentElement;
   for (const rule of sheet.cssRules) {
     if (rule.media) {
       let hasMediaScreen = false;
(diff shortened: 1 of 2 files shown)


domenic:
Does this also work with a slight variant of that issue, where someone does e.g. `html { --test-var: red }` or `body { --test-var: red }`?

The point is you're supposed to gather custom properties from the entire ancestor chain, using the cascade, not hard-code `:root` specifically.

asamuzaK:
Not included in this PR.
Inheritance requires separate, detailed handling.

asamuzaK:
Common properties are already known in propertyDefinitions, but custom properties must be found by scraping stylesheets.

This PR aims to collect custom property names, starting with :root.
Resolving custom property values ​​is not the goal of this PR.

asamuzaK:
Perhaps in the future handleRule() will look something like this:
``` js
const inheritableProperties = new Set([
  // property list
]);

function handleRule(rule, declaration, specificities, ast, inherit) {
  const { style } = rule;
  for (let i = 0; i < style.length; i++) {
    const property = style[i];
    if (inherit) {
      if (inheritableProperties.has(property) || property.startsWith("--")) {
        handleProperty(property, declaration, style, specificities, ast);
      }
    } else {
      handleProperty(property, declaration, style, specificities, ast);
    }
  }
}
```

asamuzaK:
I've come up with another approach so I'll close this.

PR #4069, opened 2026-02-15
Split rule processing into handleRule() and handleStyle()

asamuzaK:
* <del>Fixed potential bug regarding AST.</del>
  <ins>Will add fix on dom-selector side.</ins>
* <del>Passing property, value, and priority to the style handlers.</del>
* <del>Simplify rule traversal by destructuring rule fields so `media`/`@import`/`@media` branches correctly process nested cssRules, and reduce repeated matches()/handleRule() code paths.</del>
* Split rule processing into handleRule() and handleStyle(); matches() now accepts selectorText, and then handleStyle() applies.
* <del>Overall cleanup improves clarity and fixes inconsistencies in how properties and priorities were extracted and applied.</del>

### DIFF
--- a/lib/jsdom/living/helpers/style-rules.js
+++ b/lib/jsdom/living/helpers/style-rules.js
@@ -152,10 +152,7 @@ function prepareComputedStyleDeclaration(elementImpl, { styleCache }) {
   applyStyleSheetRules(elementImpl, declaration);
 
   for (let i = 0; i < style.length; i++) {
-    const property = style.item(i);
-    const value = style.getPropertyValue(property);
-    const priority = style.getPropertyPriority(property);
-    handlePropertyForInlineStyle(property, value, priority, declaration);
+    handlePropertyForInlineStyle(style.item(i), declaration, style);
   }
 
   styleCache.set(elementImpl, declaration);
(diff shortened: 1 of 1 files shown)


asamuzaK:
Re potential bug:
I can't write a reproducible test for this.
I'm getting an error in Github Actions for some unknown reason.
For example, this:
https://github.com/jsdom/jsdom/actions/runs/22032623276/job/63660200768#step:6:10189
> TypeError: Cannot destructure property 'value' of 'Specificity.max(...)' as it is undefined.

One possible reason why `Specificity.max(...Specificity.calculate(ast));` becomes undefined is when ast is null.

Although ast is nullable, in theory ast should never be null if selectorText matches, but it does occur in practice (this seems like a bug in a specific Node.js version).
Specificity.calculate() is designed to accept selectorText as well as the css-tree AST, so as a fallback, I've changed it to apply selectorText in case that ast is null.

asamuzaK:
Other than that, the only other major change this time is applying what eslint calls `prefer object destructuring`.
So I don't think it's `worse`.
However, if you want me to revert it, I'll do it.
Please let me know.

domenic:
Thanks for explaining.

I am not 100% sure, but I think it is better to avoid the defensive code without a test.

> Other than that, the only other major change this time is applying what eslint calls `prefer object destructuring`.
> So I don't think it's `worse`.

There's a difference between "prefer object destructuring" and what is done in this PR.

"Prefer object destructuring" is a minor change within a single function to get rid of redundancy:

```js
const x = something.x; // redundant
const { x } = something; // better
```

This is always good. When you can do a simple substitution of line 2 in place of line 1, line 2 is clearly better.

Most of the changes in this PR are much broader, and are instead about moving the responsibility for knowing the details of an object from the function that operates on that object (e.g. `handlePropertyForInlineStyle` operates on `style`) to its callers (now `prepareComputedStyleDeclaration` needs to know that the important parts of `style` for inline property handling are `style.getPropertyValue(property)` and `style.getPropertyPriority(property)`). I think this is slightly worse, because it makes it so that someone has to read both `prepareComputedStyleDeclaration` and `handlePropertyForInlineStyle` to understand how we handle properties for inline style.

I will repeat that the change for `handleRule` is good, because it means less code for the callers. In other words, the callers need to understand less about how `handleRule` works. So please keep that part :)

domenic (approved):
This is great, a clear improvement. Thanks for the iteration.

PR #4070, opened 2026-02-15
Add custom property handler

### LINKED ISSUE #3563: getComputedStyle does not work with css variables on the :root

asamuzaK:
Built on top of #4069
Fixes #3563

### DIFF
--- a/test/web-platform-tests/to-upstream/css/css-variables/custom-properties-inherit.html
+++ b/test/web-platform-tests/to-upstream/css/css-variables/custom-properties-inherit.html
@@ -23,6 +23,14 @@ div {
 .three {
   --font-size: 30px;
 }
+
+#five {
+  --font-size: 50px;
+}
+
+.six {
+  --font-size: 60px;
+}
 </style>
 
 <div class="one" id="one">
@@ -37,7 +45,12 @@ div {
     </div>
   </div>
 </div>
-
+<div class="five" id="five">
+  <p>Five</p>
+  <div class="six" id="six">
+    <p>Six</p>
+  </div>
+</div>
 <script>
 "use strict";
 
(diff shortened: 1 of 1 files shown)


domenic:
Before I do a full review, can you explain the approach you are trying to take here?

My understanding of how CSS variables are supposed to work is that they are inherited properties. So, they should go down the same code path as other inherited properties like `visibility`, which we already have code for.

Some cases that I think will not work well with your approach:

```html
<style>
#grandparent { --color: blue; }
.parent { --color: red; }
</style>
<div id="grandparent">
  <div class="parent">
    <div class="child">
    </div>
  </div>
</div>
```

The child should inherit `--color: red` from `.parent`. But `handleRuleForCustomProperty` walks all ancestors and applies both rules to the child's declaration via `handleProperty`, which compares their specificities. `#grandparent` has specificity `(1,0,0)` vs `.parent` at `(0,1,0)`, so `--color: blue` wins (which is wrong).

Other issues:

- Hard-coding a `:root` special case is generally a sign we're not doing the right thing, IMO. The CSS variables spec doesn't have anything about `:root` being special in it.
- `cssText.includes("--")` would trigger on stuff like `[data-foo--bar] { color: blue }` or `x { background: url("--"); }`.

domenic:
> My understanding of how CSS variables are supposed to work is that they are inherited properties. So, they should go down the same code path as other inherited properties like `visibility`, which we already have code for.

In particular, it might be best to wait on fixing https://github.com/jsdom/jsdom/issues/3563 until later. If I understand correctly, you are working to centralize property inheritance in cssstyle and removed the `propertiesWithResolvedValueImplemented` from jsdom. Hopefully, eventually, custom properties can go down the same path as all other inherited properties. I don't think there is a special rush to fix #3563; we should take the time to get the architecture right instead of hard-coding something for custom properties and/or hard-coding something for `:root`.

asamuzaK:
> so --color: blue wins (which is wrong).

Your right.
Thanks for pointing that out.

> In particular, it might be best to wait on fixing #3563 until later.

Agreed.
I'll close this for now and come back to it later.

PR #4071, opened 2026-02-16
Avoid redundant cloning in getCascadedPropertyValue

domenic:
## Summary

`getCascadedPropertyValue()` is the internal workhorse called during style resolution: for every property on every element, it retrieves the cascaded value. Previously it called `getComputedStyleDeclaration()`, the public API entry point, which clones the entire cached declaration into a new `CSSStyleDeclaration` on every call. This meant resolving N properties on a single element performed N full-declaration clones, each iterating every property just to read back a single value.

The fix is minimal: `getCascadedPropertyValue()` now checks the style cache directly and reads from it when populated, falling back to `getComputedStyleDeclaration()` when not. The caching strategy is unchanged — no new caches, no restructuring.

## Benchmark results

`npm run benchmark -- -s style/get-computed-style`, median latency:

**Warm path** (DOM stable, cache populated — the common case):

| Benchmark | main (ns) | optimized (ns) | Ratio |
|---|---|---|---|
| flat (20 siblings) | 649,600 | 651,100 | ~1.0x |
| deep (10-level nesting) | 32,400 | 33,700 | ~1.0x |
| deep, all levels | 321,200 | 330,400 | ~1.0x |

**Cold path** (fresh DOM each iteration — first call after DOM mutation):

| Benchmark | main (ns) | optimized (ns) | Speedup |
|---|---|---|---|
| flat (20 siblings) | 15,906,400 | 4,490,350 | 3.5x |
| deep (10-level nesting) | 8,247,500 | 2,494,800 | 3.3x |
| deep, all levels | 8,726,400 | 2,561,350 | 3.4x |

No warm-path regression; 3.3–3.5x cold-path speedup.

### DIFF
--- a/benchmark/style/get-computed-style-cold.js
+++ b/benchmark/style/get-computed-style-cold.js
@@ -1,0 +1,80 @@
+"use strict";
+const documentBench = require("../document-bench");
+
+// Cold-path benchmark: DOM is rebuilt before each iteration via beforeEach,
+// so the style cache is always empty and every iteration measures the full
+// cascade computation and property resolution.
+
+const SIBLINGS = 20;
+const DEPTH = 10;
+
+module.exports = () => {
+  const { document, bench } = documentBench();
+  const window = document.defaultView;
+
+  {
+    let divs = [];
+
+    bench.add("flat (20 siblings)", () => {
+      for (const div of divs) {
+        window.getComputedStyle(div);
+      }
+    }, {
+      beforeEach() {
+        document.body.innerHTML = "";
+        divs = [];
+        for (let i = 0; i < SIBLINGS; i++) {
+          const div = document.createElement("div");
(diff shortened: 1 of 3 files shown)


domenic:
There's probably something worthwhile here but this code seems pretty confused as to how it's using the cache and what entries are going into it. Needs more baking.

The core problem statement is correct: `getCascadedPropertyValue` is part of the inner machinery, but it's taking a dependency on `getComputedStyleDeclaration` whose goal is to prepare a public copy of `CSSStyleDeclaration` for jsdom users, and thus does a bunch of extra work.

domenic:
Much better now, although lint and naming issues remain.

domenic:
Alright, this ended up being much less exciting than the original version. But, still a speedup, and it's nice to get some benchmarks.

Issue #4077, opened 2026-02-20
Color like `rgb(255,0,0)` no longer working for SVG path element in the style


linev:
### Node.js version

24.13.0

### jsdom version

28.1.0

### Minimal reproduction case

```js
import { JSDOM } from 'jsdom';

const document = new JSDOM('<!DOCTYPE html><p>Hello world</p>').window.document,
      xmlns = 'http://www.w3.org/2000/svg';

const svg = document.createElementNS(xmlns, 'svg');
svg.setAttribute('width', 100);
svg.setAttribute('height', 100);

const path1 = document.createElementNS(xmlns, 'path');
path1.setAttribute('d', 'M10,10L90,10');
path1.style.stroke = 'rgb(255,0,0)'; // this is ignored
svg.appendChild(path1);

const path2 = document.createElementNS(xmlns, 'path');
path2.setAttribute('d', 'M90,10L90,90');
path2.style.stroke = 'red';    // this works!
svg.appendChild(path2);

console.log(svg.outerHTML);
```

### How does similar code behave in browsers?

Color like `rgb(255,0,0)` works of course.

### What is the problem?

By migration from jsdom ~~28.0.0~~ 27.4.0 to 28.1.0 `stroke` style for svg path element can not be assign if contains color like `rgb(255,0,0)`. Other variants like `red` or `#ff0000` are working.

asamuzaK:
Upstream issue
See https://github.com/w3c/webref/issues/625

asamuzaK:
But I think we can workaround.
I'll take a look.

PR #4078, opened 2026-02-20
Add failing test for stroke property

asamuzaK:
Regression test for https://github.com/jsdom/jsdom/issues/4077

### DIFF
--- a/test/web-platform-tests/to-upstream/svg/painting/paint-stroke.html
+++ b/test/web-platform-tests/to-upstream/svg/painting/paint-stroke.html
@@ -76,5 +76,5 @@ test(t => {
   });
   target.style.stroke = "linear-gradient(to top, green, blue)";
   assert_equals(target.style.getPropertyValue("stroke"), "");
-}, "Gradient; browsers don't support <gradient> in <paint>");
+}, "Gradient, gradient is not included in SVG paint");
 </script>
--- a/test/web-platform-tests/to-upstream-expectations.yaml
+++ b/test/web-platform-tests/to-upstream-expectations.yaml
@@ -38,4 +38,4 @@ jsdom-only/failure-example-2.html:
   "failing test": [fail, https://github.com/jsdom/jsdom/issues/3835]
 svg/painting/paint-stroke.html:
   "RGB color": [fail, https://github.com/jsdom/jsdom/issues/4077]
-  "Gradient; browsers don't support <gradient> in <paint>": [fail, https://github.com/jsdom/jsdom/issues/4077]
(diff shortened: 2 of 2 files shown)


test/web-platform-tests/to-upstream/svg/painting/paint-stroke.html
```diff
@@ -0,0 +1,80 @@
...
+  assert_equals(target.style.getPropertyValue("stroke"), 'url("foo/bar") green');
+}, "URL with color as fallback");
+
+test(t => {
+  t.add_cleanup(() => {
+    target.style.stroke = null;
+  });
+  target.style.stroke = "linear-gradient(to top, green, blue)";
+  assert_equals(target.style.getPropertyValue("stroke"), "");
+}, "Gradient; browsers don't support <gradient> in <paint>");
```
domenic: The tests we write should match the specs, not browsers.

Our implementation is probably best matching the specs, but if it is easier to match browsers then matching browsers is OK.

asamuzaK: Per https://svgwg.org/svg2-draft/painting.html#SpecifyingPaint:
`<paint> = none | <color> | <url> [none | <color>]? | context-fill | context-stroke`
Note that `<gradient>` is not included, and browsers are following this spec.
And this test too.

Also note that:
<a href="https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/stroke#:~:text=Note%3A%20According%20to%20the%204%20April%202017%20draft%20of%20the%20CSS%20Fill%20and%20Stroke%20Module%20Level%203%20specification%2C%20the%20stroke%20property%20is%20a%20shorthand%20for%20a%20number%20of%20other%20stroke%20properties.%20In%20practice%2C%20as%20of%20August%202024%2C%20browsers%20do%20not%20support%20the%20setting%20of%20other%20stroke-related%20values%20such%20as%20width%20or%20dash%20patterns%20via%20the%20stroke%20property%2C%20treating%20it%20instead%20as%20a%20direct%20analogue%20of%20the%20SVG%20stroke%20attribute." rel="noopener">Note: According to the 4 April 2017 draft of the CSS Fill and Stroke Module Level 3 specification, the stroke property is a shorthand for a number of other stroke properties. In practice, as of August 2024, browsers do not support the setting of other stroke-related values such as width or dash patterns via the stroke property, treating it instead as a direct analogue of the SVG stroke attribute.</a>

domenic: Got it, thank you!

domenic (approved):

PR #4079, opened 2026-02-22
Add inherited properties handler

### LINKED ISSUE #3563: getComputedStyle does not work with css variables on the :root

asamuzaK:
Fixes #3563.
Also fixes some WPTs.

Note: There are new failures that were previously incorrectly passing.

### DIFF
--- a/test/web-platform-tests/to-upstream/css/css-variables/custom-properties-inherit.html
+++ b/test/web-platform-tests/to-upstream/css/css-variables/custom-properties-inherit.html
@@ -1,0 +1,72 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Custom properties inheritance</title>
+<link rel="help" href="https://drafts.csswg.org/css-variables/#defining-variables">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<!-- Regression test for https://github.com/jsdom/jsdom/issues/3563 -->
+
+<!-- Inspired from https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascading_variables/Using_custom_properties -->
+<style>
+:root {
+  --font-size: 10px;
+}
+
+div {
+  font-size: var(--font-size);
+}
+
+.two {
+  --font-size: 20px;
+}
+
+.three {
+  --font-size: 30px;
+}
+
+#five {
+  --font-size: 50px;
+}
+
+.six {
+  --font-size: 60px;
+}
+</style>
+
+<div class="one" id="one">
+  <p>One</p>
+  <div class="two" id="two">
+    <p>Two</p>
+    <div class="three" id="three">
+      <p>Three</p>
+    </div>
+    <div class="four" id="four">
+      <p>Four</p>
+    </div>
+  </div>
+</div>
+<div class="five" id="five">
+  <p>Five</p>
+  <div class="six" id="six">
+    <p>Six</p>
+  </div>
+</div>
+<script>
+"use strict";
+
+const one = document.getElementById("one");
+const two = document.getElementById("two");
+const three = document.getElementById("three");
+const four = document.getElementById("four");
+const five = document.getElementById("five");
+const six = document.getElementById("six");
+
+test(() => {
+  assert_equals(window.getComputedStyle(one).getPropertyValue("--font-size"), "10px", "one");
+  assert_equals(window.getComputedStyle(two).getPropertyValue("--font-size"), "20px", "two");
+  assert_equals(window.getComputedStyle(three).getPropertyValue("--font-size"), "30px", "three");
+  assert_equals(window.getComputedStyle(four).getPropertyValue("--font-size"), "20px", "four");
+  assert_equals(window.getComputedStyle(five).getPropertyValue("--font-size"), "50px", "five");
+  assert_equals(window.getComputedStyle(six).getPropertyValue("--font-size"), "60px", "six");
+}, "Inheritance of custom properties");
+</script>


domenic:
Thanks for working on this! I'm very excited to get property inheritance working.

From what I can tell, the core approach here is to do inheritance inside `prepareComputedStyleDeclaration()` (the cascade step), and then rely on `getResolvedValue()` to overwrite the values for implemented properties. The problem with this is that it creates two independent inheritance mechanisms: `applyInheritedStyles()` for all properties, and the existing `getSpecifiedValue()` chain for the ~15 implemented properties. And the latter silently overwrites the former.

In particular, this means `getCascadedPropertyValue()` no longer returns cascaded values. Because the cache now stores declarations with inherited values baked in, `getSpecifiedValue()`'s own inheritance logic (the "if inherited, get from parent" branch) becomes dead code, because the "cascade" value is never empty for inherited properties. The separation of cascade → specified → computed → resolved is short-circuited, going against the spec and the existing architecture.

The approach I'd expected you to take for implementing inheritance was to generalize `getSpecifiedValue()`/`getComputedValue()` to work with cssstyle's `propertyDefinitions`, rather than adding a separate inheritance step at the cascade level. That way there's one inheritance mechanism (the existing pipeline, but driven by `propertyDefinitions` instead of a hardcoded table), the cascade stays pure, and each property goes through the full cascade → specified → computed → resolved chain.

Also note that the `inheritedProperties` filter `(!/\s/.test(initial))` isn't very good. E.g. it drops longhand properties whose initial values happen to contain spaces (`border-spacing`, `font-family`, `font-synthesis`, etc.), and misses shorthands with single-word initial values (`caret`, `text-align`, etc.). Does webref contain any helpful data on "is a shorthand or not"? The best I could come up with from `propertyDefinitions` is checking `initial` or `computedValue` for `"individual properties"`... but maybe hard-coding a list, or opening an issue on webref, would be better.

Finally, for tests, we should be sure to at least include https://github.com/jsdom/jsdom/pull/4070#issuecomment-3906474186 .

WDYT?

asamuzaK:
> The problem with this is that it creates two independent inheritance mechanisms: applyInheritedStyles() for all properties, and the existing getSpecifiedValue() chain for the ~15 implemented properties. And the latter silently overwrites the former.

getSpecifiedValue() and related functions are remaining for compatibility reasons, but I'm planning to remove them in the near future.
These should all be handled within cssstyle.

> Also note that the inheritedProperties filter (!/\s/.test(initial)) isn't very good.

This was meant to be a stopgap until https://github.com/w3c/webref/pull/1821 get merged.
It appears to have already been merged, so I'll fix it when the new @w3c/webref is released.
See https://github.com/w3c/webref/pull/1842

> Finally, for tests, we should be sure to at least include #4070 (comment) .

Will do.

asamuzaK:
Will #4089 or #4092 be merged soon?
I'm pausing work on this PR until one of them get merged.

domenic:
I'm sorry, I was working very actively on them, but then I got a new puppy. She requires lots of attention so bigger projects are hard. But she is starting to self-regulate so I am hopeful that I will be able to finish and merge those PRs within ... 2 or 3 days?

domenic:
It is now day 3. It is clear my original timeline is unrealistic, at least for https://github.com/jsdom/jsdom/pull/4092, which contains a lot of AI-generated code that still needs review. Although I can find ~0.5 hour blocks, longer concentration periods are hard, and I am not sure when they will become easier.

So, let's discuss what options we have for making incremental progress, and what works best for unblocking you.

- Maybe, we just proceed with those two PRs on hold. E.g. we continue with this PR. I work on merging things like https://github.com/jsdom/jsdom/pull/4097, maybe doing an update-deps PR to get the test passing improvements in cssstyle, etc. It's still possible I could finish https://github.com/jsdom/jsdom/pull/4092 soon, but my estimate has widened to something like 2-21 days at this point.

- Maybe we work on merging in https://github.com/jsdom/jsdom/pull/4089. It is almost merge-ready already. My worry is that the overhead of `lib/jsdom/living/cssom/upgrade-stylesheet.js` will be a performance regression that is bad for users. (I haven't benchmarked yet.)

- Maybe we work on a version of https://github.com/jsdom/jsdom/pull/4089 that vendors `@acemir/cssom` into the source tree, with minimal changes (e.g. https://github.com/acemir/CSSOM/pull/97 or something simpler) to avoid the wrapping. We add a README clearly explaining that this is temporary, and the code in that directory is not jsdom quality and will be superseded soon.

Do you have any thoughts on the best path? Other options are also welcome.

----

Regarding the contents of this PR: I'm not sure my biggest concern from https://github.com/jsdom/jsdom/pull/4079#issuecomment-3956211012 is addressed. To restate it, the spec has a separation of cascade → specified → computed → resolved. The existing code mostly matches that. https://github.com/jsdom/jsdom/pull/4085 is an extension of that existing code path.

The approach in this PR has two problems:

1. It introduces a second path. Having two paths is not great. It would be best if we completely replaced the old path with the new path, in the same PR. This is easy enough to fix.

2. **The second path does not match the spec.** There is no cascade → specified → computed → resolved chain. This is harder to fix.

Maybe (2) is not a real problem? Maybe we don't need separate cascaded, specified, computed, and resolve styles, since all we care about is `getComputedStyle()` being correct? But I am unsure whether that will work well in the longer term.

I am open to being convinced that this PR's approach is fine, and (2) is not a problem. But I would appreciate a longer explanation of what your strategy is.

asamuzaK:
>     * Maybe we work on merging in [Integrate cssstyle into jsdom CSSOM #4089](https://github.com/jsdom/jsdom/pull/4089). It is almost merge-ready already. My worry is that the overhead of `lib/jsdom/living/cssom/upgrade-stylesheet.js` will be a performance regression that is bad for users. (I haven't benchmarked yet.)

Since this isn't a project with a deadline, I'm generally open to any approach. 
If I had to choose, the second approach seems the most realistic.

<del>
> Regarding the contents of this PR: I'm not sure my biggest concern from [#4079 (comment)](https://github.com/jsdom/jsdom/pull/4079#issuecomment-3956211012) is addressed. To restate it, the spec has a separation of cascade → specified → computed → resolved. The existing code mostly matches that. #4085 is an extension of that existing code path.

The spec does indeed require processing in the order cascade → specified.
However, in practice, the inline value of Element.style takes precedence.
This means that the cascade only needs to be considered when Element.style is not specified.

If you take a look at the PR, you'll find that:
1. If a style value already exists:
1.1. If the value is `inherit`, get the value from the parent element.
1.2. Otherwise, do nothing.
2. If there is no style value:
2.1. If the property is inheritable, get the value from the parent element.
2.2. Otherwise, do nothing.

This makes it possible to complete the cascade processing in one go.

I understand your point that having two paths is bad, but trying to solve this all at once would require a pretty big PR.
My guess is that we can get rid of the old path in the next 4 or maybe 5 PRs.

1. Step 4: Apply font size in pixel. This resolves `em` in pixels.
2. Replace CSS keyword to computed value within CSSStyleDeclarations.
3. Replace colors to computed value within CSSStyleDeclarations.
4. Get rid of old path from style-rule.js
</del>

asamuzaK:
My previous comment was confusing inheritance and cascade. Sorry for that.

1. We do cascade in Step 1 and Step 2
2. In Step 3
  2.1. Replace specified values ​​with the `inherit` keyword with the value of the parent element
  2.2. Replace inheritable properties with no value set with the value of the parent element

My plan after this PR:
1. Add Step 4: Apply font size in pixel. This resolves em in pixels.
2. Update CSSStyleDeclaration to remove old path.
  2.1. Replace CSS keyword to computed value within CSSStyleDeclaration.
  2.2. Replace colors to computed value within CSSStyleDeclaration.
  2.3. Get rid of old path from style-rule.js
3. Update CSSStyleDeclaration to resolve specified value to computed value.

domenic:
Thanks for being patient while I worked on 4aaacc67ccba17093578b3538402ec43dc0f94aa.

I still am not sure I understand your approach. You want to put all the logic into CSSStyleDeclaration, I guess? Why is that a better place to put it than inside `computed-style.js`? (Formerly `style-rules.js`.) Maybe because it will be more efficient than the current architecture of cloning declarations, setting properties, etc.?

Now that the repos are merged, https://github.com/jsdom/jsdom/pull/4085 shows that an alternate approach can fix many tests at once, with no regressions. Your PR here fixes a few tests, but has a few regressions too. It also prolongs the state where the codebase has two paths.

I am willing to trust your vision here, but I want to make sure you have a path toward doing better than https://github.com/jsdom/jsdom/pull/4085. Or maybe we should merge https://github.com/jsdom/jsdom/pull/4085 first, and then try to make sure all future work has no regressions from that baseline?

asamuzaK:
Since cssstyle is now included in jsdom, I think this PR can be closed.

Apart from that, the current post-processing part in computed-style.js only handles property values that are either as-specified or computed-color.
And only a very limited number of properties.
This should eventually be fixed, and I believe it can be fixed within CSSStyleDeclaration.

What I plan to implement in CSSStyleDeclaration is something like this:

Getters in `properties/*` and `propertyDescriptor.js`:
``` js
  get() {
    // Stored specified value.
    const value = this.getPropertyValue(property);
    if (this._computed) {
      return this._getComputedValue(property, value);
    }
    return value;
  }
```

CSSStyleDeclaration:
``` js
  constructor() {
    // Options for computed style.
    this._computedStyleOpts = null;
  }

  /**
   * Returns the computed value for a given property.
   *
   * @private
   * @param {string} property - The property name.
   * @param {string} value - The specified value.
   * @returns {string} The computed value.
   */
  _getComputedValue(property, value) {
    if (!propertyDefinitions.has(property)) {
      return "";
    }
    const { inherited, initial, longhands } = propertyDefinitions.get(property);
    if (isGlobalKeyword(value)) {
      const { parentElement } = this._ownerNode;
      switch (value) {
        case "initial": {
          if (initial && !longhands) {
            return initial;
          }
          break;
        }
        case "inherit": {
          if (parentElement) {
            // Look for cached parent style.
            const cachedValue =
              this._global.getComputedStyle(parentElement).getPropertyValue(property);
            if (typeof cachedValue === "string") {
              return cachedValue;
            }
            // TODO: resolve inherited value.
          } else if (initial && !longhands) {
            return initial;
          } else {
            // TODO: resolve shorthands.
          }
          break;
        }
        case "unset": {
          if (inherited === "yes" && parentElement) {
            // Look for cached parent style.
            const cachedValue =
              this._global.getComputedStyle(parentElement).getPropertyValue(property);
            if (typeof cachedValue === "string") {
              return cachedValue;
            }
            // TODO: resolve inherited value.
          } else if (initial && !longhands) {
            return initial;
          } else {
            // TODO: resolve shorthands.
          }
          break;
        }
        case "revert-layer": {
          // TODO: https://drafts.csswg.org/css-cascade-5/#revert-layer
          break;
        }
        case "revert": {
          // TODO: https://drafts.csswg.org/css-cascade-5/#default
          break;
        }
        default:
      }
    } else {
      const opts = this._prepareComputedStyleOpts();
      if (hasVarFunc(value)) {
        // TODO: resolve css var()
      } else if (!longhands) {
        if (value) {
          // TODO: resolve computed value.
        } else {
          // TODO: resolve computed initial value.
        }
      } else {
        // TODO: resolve shorthands.
      }
    }
    return value;
  }

  /**
   * Prepares the options for resolving the computed values.
   * Options are used in cssColor functions e.g. resolveCalc(), resolveColor() etc.
   *
   * @private
   * @returns {object} An object containing the options for computed values.
   */
  _prepareComputedStyleOpts() {
    if (this._computedStyleOpts) {
      return this._computedStyleOpts;
    }
    const customProperty = {};
    const dimension = {
      rem: ?? // this.ownerNode.ownerDocument.documentElement.fontSize in pixels ?? 16,
      vh: this._global.innerHeight / 100,
      vw: this._global.innerWidth / 100
    };
    const opts = {
      baseURL: this.ownerNode.ownerDocument.baseURI,
      format: "computedValue",
    };
    for (let i = 0; i < this.length; i++) {
      const property = this.items(i);
      const value = this.getPropertyValue(property);
      if (property.startsWith("--")) {
        customProperty[property] = value;
      } else if (hasVarFunc(value)) {
        continue;
      } else if (property === "color") {
        // TODO: Set this._systemColors from helpers/systemColors.js
        const systemColors = ??;
        if (systemColors.has(value)) {
          const { light, dark } = systemColors.get(value);
          if (/dark/.test(this.getPropertyValue("color-scheme"))) {
            opts.currentColor = dark;
          } else {
            opts.currentColor = light;
          }
        } else {
          opts.currentColor = value;
        }
      } else if (property === "color-scheme") {
        opts.colorScheme = value;
      } else if (property === "font-size") {
        if (/\dpx$/.test(value)) {
          const parsedValue = parseFloat(value);
          dimension.em = parsedValue;
        } else {
          // TODO: Resolve from parent element font size in pixels.
        }
      }
    }
    opts.customProperty = customProperty;
    opts.dimension = dimension;
    this._computedStyleOpts = opts;
    return this._computedStyleOpts;
  }
```

asamuzaK:
Closing.

PR #4080, opened 2026-02-22
Export allSystemColors

asamuzaK:
Planned for future use.


domenic:
This seems reasonable, but I would prefer to not merge code until it has a use; otherwise it is dead code. If there is a PR on top of this that is ready to merge, I can merge this PR.

asamuzaK:
Fine with me.

domenic:
Should no longer be necessary now that everything's in one repo!

PR #4085, opened 2026-02-25
Use resolveProperty callback for lazy computed style resolution

domenic:
Lazy computed style resolution via a `resolveProperty` callback, replacing the hardcoded property table with metadata derived from the generated property definitions.


domenic:
Finally got this working. @asamuzaK do you want to take a look? I'd like to merge this and https://github.com/jsdom/jsdom/pull/4117 and do a patch release soon.

asamuzaK:
I'm thinking of going in the direction of #4119.

domenic:
Great! Maybe I should do a release without either first, to get the minor bug fixes released.

Regarding which approach is better, I am happy to let you take the lead. Maybe we can combine them. I think we can judge which is based based on:

- Amount of new passing tests
- Speedup per benchmarks (including new benchmarks added in this PR; maybe I should merge those benchmarks separately from this PR)
- Reduction in code complexity. (Right now this PR has an advantage because it completely removes `propertiesWithResolvedValueImplemented`. But I see #4119 only uses that to figure out colors, so, maybe borrowing some work from this PR, like the prepare script, can help close the gap.)

PR #4088, opened 2026-03-02
Skip non-style at-rules in getComputedStyle processing

### LINKED ISSUE #4083: Regression: `TypeError: Cannot destructure property 'value' of 'Specificity.max(...)' as it is undefined.`
### Node.js version

25.6.1

### jsdom version

28.1.0 / 3d35b2cacaf5b26af032533f83bc11344c395d17

### Minimal reproduction case

```js
function repro() {

  const file = `
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">
<head>
  <style type="text/css">
    @page {
      margin: 0;
    }
  </style>
</head>

<body>
</body>
</html>`;

  const dom = new jsdom.JSDOM(file, { contentType: "application/xhtml+xml" });
  const elem = dom.window.document.querySelector("body");

  // Error: TypeError: Cannot destructure property 'value' of 'Specificity.max(...)' as it is undefined.
  dom.window.getComputedStyle(elem);
}
```

Previous version of jsdom did not crash

### What is the problem?

Seemingly a combination of `contentType: "application/xhtml+xml"` and `@page`, triggered via `getComputedStyle`.

Git bisect says this issue exists since ce4c58fac40002e024d32f50d11c6e9c5deaff89 (28.1.0) and is not fixed in current main (3d35b2cacaf5b26af032533f83bc11344c395d17).

domenic:
Fixes #4083. Only pass CSSStyleRule instances to handleRule(), skipping @page and other non-style at-rules that don't apply to elements.

### DIFF
--- a/test/web-platform-tests/to-upstream/css/cssom/getComputedStyle-page-rule.xhtml
+++ b/test/web-platform-tests/to-upstream/css/cssom/getComputedStyle-page-rule.xhtml
@@ -2,6 +2,7 @@
 <head>
 <title>getComputedStyle does not crash with @page rules</title>
 <meta charset="utf-8" />
+<link rel="help" href="https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle" />
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 


domenic:
@codex can you fix the build error by pushing a commit to this branch which adds an appropriate `<link rel=help>`?

(First time trying this...)

PR #4092, opened 2026-03-04
Merge in `cssstyle` and replace `@acemir/cssom`

### LINKED ISSUE #690: Memory leak when stylesheets removed
Replacing/removing a DOM tree with a stylesheet with a call such as

```
head.innerHTML = "";
```

will result in a leak as the stylesheet is not removed from `Document.styleSheets`.  There's a TODO in `core.Document.prototype.__defineGetter__` for the `styleSheets` property that suggests a solution:

```
  // TODO: each style and link element should register its sheet on creation                                               
  // nad remove it on removal.                                                                                             
```

For a workaround, I'm iterating through `ownerDocument.styleSheets` and deleting its (array) properties, then setting `styleSheets._length=0`

### LINKED ISSUE #992: Implement `CSSStyleSheet` `ownerNode` and `href`
I know this project has a disclaimer about not supporting all of the CSSOM, but I encountered some errors while trying to get [HTML Inspector](https://github.com/philipwalton/html-inspector) to work with jsdom instead of PhatomJS, and I tracked it down to the lack of these two properties on the `CSSStyleSheet` object.

From what I can tell, the [following function in level2/style.js](https://github.com/tmpvar/jsdom/blob/2ff5747488ad4b518fcef97a026c82eab42a0a14/lib/jsdom/level2/style.js#L111-L119) is where this could/should be added, but I'm not familiar enough with the codebase to know for sure. 

``` js
function evaluateStylesheet(data, sheet, baseUrl) {
  // this is the element
  var newStyleSheet = cssom.parse(data);
  var spliceArgs = newStyleSheet.cssRules;
  spliceArgs.unshift(0, sheet.cssRules.length);
  Array.prototype.splice.apply(sheet.cssRules, spliceArgs);
  scanForImportRules.call(this, sheet.cssRules, baseUrl);

  // Add references to `this` and `baseUrl`.
  sheet.href = baseUrl;
  sheet.ownerNode = this;

  this.ownerDocument.styleSheets.push(sheet);
}
```

Anyway, this changed work for me locally, so hopefully it's as simple as that. Thanks for the consideration.

### LINKED ISSUE #2974: CSSMediaRule is missing insertRule function
### Basic info:

- **Node.js version:** 10.18.1
- **jsdom version:** 11.12.0

### Minimal reproduction case

```js

const dom = new JSDOM();
const styleElement = document.createElement('style');
styleElement.appendChild(document.createTextNode(''));
document.head.appendChild(styleElement);
const sheet = this.styleElement.sheet;
sheet.insertRule('@media only screen and (orientation: landscape) {}',0);
// This is the CSSMediaRule
sheet.cssRules[0].insertRule('.elem{background-color: red;}');
```
- Actual
-- Error: "insertRule is not a function"

- Expected
-- insertRule to work on a CSSMediaRule

See:
[CSSMediaRule MDN](https://developer.mozilla.org/en-US/docs/Web/API/CSSMediaRule) : 
>As a CSSConditionRule, and therefore both a CSSGroupingRule and a CSSRule, CSSMediaRule also implements the properties of these interfaces. It has the following specific property:

[W3C CSSMediaRule](https://www.w3.org/2003/01/dom2-javadoc/org/w3c/dom/css/CSSMediaRule.html):
image

### LINKED ISSUE #3766: Implement CSSStyleSheet replace() and replaceSync()
### Basic info:

- **Node.js version:** 22.6.0
- **jsdom version:** 24.1.1

It looks like the current implementation of `CSSStyleSheet` is missing a couple of methods that I am using to power automated tests for a web components library.

### Minimal reproduction case

```js
import {JSDOM} from 'jsdom'

const {window} = new JSOM('<!doctype html>')

const stylesheet = new window.CSSStyleSheet()
await stylesheet.replace('body { color: red; }')
```

I expect the methods `replace()` and `replaceSync()` to exist and behave as documented (https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet)

### LINKED ISSUE #3988: Nested @media in style sheet fails to parse
### Node.js version

v20.19.2

### jsdom version

27.2.0

### Minimal reproduction case

```js

const options = {};
const dom = new JSDOM(`
  <style>
    @media (min-width:0px) {
      @media (orientation: landscape){
        .css-1gak8h1-MuiToolbar-root-MuiTablePagination-toolbar {
          min-height:48px;
        }
      }
    }
  </style>
`, options);
```

This pattern is used by [Mui](https://mui.com/) an generally worked in the past, like `jsdom@27.0.1`.

### What is the problem?

This fails with `Could not parse CSS stylesheet`, the exact error is hidden.

domenic:
As discussed in https://github.com/jsdom/cssstyle/issues/318, over time cssstyle has become increasingly coupled with the internals of jsdom. Ongoing work on improving style resolution, and future work on moving `CSSStyleDeclaration` to webidl2js, will only increase this coupling. As such, we merge the cssstyle package into jsdom.

Doing so while using `@acemir/cssom` directly is difficult, since `@acemir/cssom` hard-codes the `cssstyle` package name, and won't properly integrate with a jsdom-internal `CSSStyleDeclaration`. Additionaly, `@acemir/cssom` is a package with a lot of historical baggage, including a custom build system, its own parser, an ES5-era codebase, and various nonstandard extensions. As such, this includes fresh reimplementations of all of the CSSOM classes, using webidl2js wrappers, and using the csstree parser instead of a custom parser, with extra work to match the latest specs (or document deviations from them).

Some specific improvements over `@acemir/cssom`:

* `CSSMediaRule`/`CSSSupportsRule` `matches` getters
* Proper media query parsing in `MediaList` (instead of just splitting on commas)
* `CSSKeyframeRule` `keyText` getter/setter with validation
* `CSSStyleRule` `selectorText` setter rejects invalid selectors
* `CSSStyleSheet` `replace()` properly defers to a microtask, and the "disallow modification flag" is checked in all methods.
* `StyleSheet` `ownerNode`, `href`, and `title` are set.
* No nonstandard extensions (e.g. `CSSDocumentRule`, `CSSHostRule`, `CSSValue`, `MatcherList`, various parser-related methods).

Fixes #690. Fixes #992. Fixes #2974. Fixes #3766. Fixes #3988.
 
This is not perfect. In particular, csstree does not have a spec-compliant serializer, and its parser needs some hacks to be spec-compliant. (See https://github.com/csstree/csstree/issues/52 and the `css-parser.js` file.) So the ideal strategy, of parsing, storing AST notes, and serializing, is not possible. Instead, we do our best to preserve the input as strings, and serialize it back out; this gives reasonable results (at least as good as `@acemir/cssom`). In the future it may be worth investigating other CSS parsers like PostCSS, `@adobe/css-tools`, or similar. For now, since cssstyle was using csstree, we do so in jsdom.

As part of this work, consolidate all CSS-related files into `lib/jsdom/living/css/`, moving several from `lib/jsdom/living/helpers/`. Many of the files from cssstyle were also renamed as part of the merge.

Many new to-upstream web platform tests are included, on an ad-hoc basis: they test things that came up in earlier drafts of this work, and are not an exhaustive test of all CSSOM functionality.

Note that apart from renames, the imported cssstyle code is mostly unchanged; in particular, `CSSStyleDeclaration` is still not using webidl2js.


lib/jsdom/living/helpers/style-rules.js
```diff
@@ -160,62 +158,63 @@ function prepareComputedStyleDeclaration(elementImpl, { styleCache }) {
...
-  return opts;
-}
-
 function applyStyleSheetRules(elementImpl, declaration) {
   if (!parsedDefaultStyleSheet) {
-    parsedDefaultStyleSheet = cssom.parse(defaultStyleSheet);
+    parsedDefaultStyleSheet = parseStylesheet(defaultStyleSheet, elementImpl._globalObject);
   }
 
+  const { implForWrapper } = require("../../../generated/idl/utils.js");
```
asamuzaK: Should declare above in L8.

lib/jsdom/living/helpers/stylesheets.js
```diff
@@ -63,31 +63,44 @@ exports.removeStylesheet = (sheet, elementImpl) => {
...
-    globalObject: elementImpl._globalObject,
-    ...parserOptions
-  }, err => {
-    if (elementImpl._ownerDocument._defaultView) {
-      const error = new Error("Could not parse CSS stylesheet", { cause: err });
-      error.sheetText = sheetText;
-      error.type = "css-parsing";
-
-      elementImpl._ownerDocument._defaultView._virtualConsole.emit("jsdomError", error);
+  const { implForWrapper } = require("../../../generated/idl/utils.js");
```
asamuzaK: Should declare above with wrapperForImpl

lib/jsdom/living/helpers/stylesheets.js
```diff
@@ -63,31 +63,44 @@ exports.removeStylesheet = (sheet, elementImpl) => {
...
+        const error = new Error("Could not parse CSS stylesheet", { cause: err });
+        error.sheetText = sheetText;
+        error.type = "css-parsing";
+
+        elementImpl._ownerDocument._defaultView._virtualConsole.emit("jsdomError", error);
+      }
     }
   });
 
+  const { wrapperForImpl } = require("../../../generated/idl/utils.js");
```
asamuzaK: Should declare above with implForWrapper

domenic: Fixed to not use wrappers, here and elsewhere.

lib/jsdom/living/cssom/MediaList-impl.js
```diff
@@ -0,0 +1,85 @@
...
+  get length() {
+    return this._list.length;
+  }
+
+  item(index) {
+    return this._list[index] || null;
+  }
+
+  appendMedium(medium) {
+    medium = medium.trim();
```
domenic: Not a real parse.

lib/jsdom/living/cssom/StyleSheet.webidl
```diff
@@ -0,0 +1,11 @@
+// https://drafts.csswg.org/cssom/#the-stylesheet-interface
+[Exposed=Window]
+interface StyleSheet {
+  readonly attribute CSSOMString type;
+  readonly attribute USVString? href;
+  // readonly attribute (Element or ProcessingInstruction)? ownerNode;
+  readonly attribute CSSStyleSheet? parentStyleSheet;
+  readonly attribute DOMString? title;
+  [SameObject, PutForwards=mediaText] readonly attribute MediaList media;
```
domenic: No backing implementation, always null?

domenic:
I've reviewed all files here except the new `css-parser.js`, which has a lot of room for improvement still.

lib/jsdom/living/css/CSSStyleSheet-impl.js
```diff
@@ -0,0 +1,109 @@
...
+    this.deleteRule(index);
+  }
+
+  replaceSync(text) {
+    if (!this._constructed || this._disallowModification) {
+      throw DOMException.create(this._globalObject, [
+        "Cannot call replaceSync on non-constructed CSSStyleSheet.",
+        "NotAllowedError"
+      ]);
+    }
```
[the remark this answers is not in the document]
domenic: @copilot apply changes based on [this feedback](https://github.com/jsdom/jsdom/pull/4092#discussion_r2930320748)

lib/jsdom/living/css/MediaList-impl.js
```diff
@@ -0,0 +1,141 @@
...
+const idlUtils = require("../../../generated/idl/utils.js");
+const csstree = require("./helpers/patched-csstree.js");
+
+// Parse a single medium string. Returns { normalized, canonical } or null.
+// - normalized: lowercased original with whitespace collapsed (for storage/display)
+// - canonical: csstree's serialization lowercased (for comparison)
+//
+// If csstree was spec-compliant, we would only need canonical. However, it is not:
+// https://github.com/csstree/csstree/issues/52. This is a compromise, and still not fully spec-complaint.
+function parseSingleMedium(text) {
```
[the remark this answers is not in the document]
domenic: @copilot apply changes based on [this feedback](https://github.com/jsdom/jsdom/pull/4092#discussion_r2930320704)

lib/jsdom/living/css/properties/fontFamily.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+  const values = parsers.splitValue(v, {
+    delimiter: ","
+  });
+  const parsedValues = [];
+  for (const val of values) {
+    if (!val) {
+      return "";
+    }
+    const value = parsers.parsePropertyValue(property, val, {
```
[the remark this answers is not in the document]
domenic: Going to ignore the properties/* for now since they're just ported from cssstyle.

domenic:
This is ready. I'll probably merge it tomorrow. Any reviews are welcome.

PR #4097, opened 2026-03-10
perf: lazy-init DOMSelector to avoid overhead on throwaway Documents

thypon:
DOMSelector initialization costs ~0.5ms per Document due to the Finder constructor registering 9 event listeners and nwsapi compiling regexps.

This is paid on every DOMParser.parseFromString() call since it creates a new Document each time. Libraries like DOMPurify call parseFromString on every sanitize() but never use querySelector/matches, making the DOMSelector initialization pure waste.

Convert the eager `this._domSelector = new DOMSelector(...)` in the Document constructor to a lazy getter that defers initialization until first access. This restores DOMPurify.sanitize() performance to v26 levels (~0.15ms vs ~2.6ms, a ~16x improvement).

All existing tests pass since the getter is transparent to consumers.

---

Since v27.1.0 (https://github.com/jsdom/jsdom/commit/a833763c2dd4bf1018217e38de3ae029e5b86fe7), _domSelector.clear() is called on every
attribute change, not just class changes. Combined with the lazy
DOMSelector getter from the previous commit, this forces full
DOMSelector initialization on the first attribute set during HTML
parsing — defeating the lazy init for any input with attributes.

Access __domSelector directly (the backing field) so that throwaway
Documents from DOMParser.parseFromString() never pay the ~0.5ms
DOMSelector construction cost. The clear() is a no-op when the
selector engine was never initialized (nothing cached to invalidate).

This closes the remaining 2x gap seen in jsdom >= 27.1.0 after the
lazy init fix, bringing DOMPurify.sanitize() to full v26 parity.

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -99,8 +99,13 @@ class ElementImpl extends NodeImpl {
       this._classList.attrModified();
     }
 
-    // clear domSelector cached results on attribute change
-    this._ownerDocument._domSelector.clear();
+    // Clear domSelector cached results on attribute change.
+    // Access __domSelector directly to avoid triggering lazy initialization —
+    // throwaway Documents (e.g. from DOMParser) that never use querySelector
+    // should not pay the DOMSelector construction cost.
+    if (this._ownerDocument.__domSelector) {
+      this._ownerDocument.__domSelector.clear();
+    }
 
     this._attrModifiedSlotableMixin(name, value, oldValue);
   }


domenic:
This is a great idea. I will work to tidy up the style a bit---I don't like the two-level `__domSelector` vs. `_domSelector`---and land this.

thypon:
Do you have any idea on how you would like this cleanup? I can fix that on my side too!

domenic:
It's fine, it would take me about as long to describe as to fix it up :).

Anyway, I tried to push to your branch, but I guess "allow edits from maintainers" is not on. So I posted the result as https://github.com/jsdom/jsdom/pull/4101. It should merge soon, and I'll try to do a release soon too.

PR #4109, opened 2026-03-14
Convert CSSStyleDeclaration to webidl2js

domenic:
Convert the last CSSOM class to use webidl2js wrappers, and introduce the spec's CSSStyleProperties subclass.

## Summary

- Add `CSSStyleDeclaration.webidl` and convert the impl class to use webidl2js conventions (private fields, `DOMException.create()`, `supportedPropertyIndices`)
- Add `CSSStyleProperties` as a generated webidl2js interface extending `CSSStyleDeclaration`, with per-CSS-property attributes
- Update all consumers (`CSSStyleRule`, `CSSKeyframeRule`, `CSSNestedDeclarations`, `CSSPageRule`, `CSSFontFaceRule`, `ElementCSSInlineStyle`) to create style objects via webidl2js wrappers, removing manual setter hacks
- Remove `prepareValue()` and `parseCSS()` helpers that are no longer needed


domenic:
@asamuzaK review welcome!

package.json
asamuzaK, line 68: IIRC, we can remove this?

asamuzaK (commented):
LGTM with nits.
Great job!

package.json
domenic: So, it is still used for test, and the `--continue-on-error` functionality in particular is not trivial to replicate.

I tried moving tests to wireit as well, and it works... OK. The different `test:*`s run in parallel, which is both good (faster) and bad (output is confusing). And there is no `:*` globbing so we have to duplicate everything.

For now I will keep `npm-run-all2`, but I am unsure what the best path is.

PR #4113, opened 2026-03-15
Version 29.0.0

domenic:


asamuzaK:
If possible, I would appreciate it if we could bump dom-selector to v7.0.3.

domenic:
Ah, bad timing! Well, it is included for any fresh installs by users, so there is no real difference; the package-lock.json is only used for CI.

asamuzaK:
That's okay. Never mind.

Issue #4112, opened 2026-03-02
Regression: unable to unset border-width


sophstad:
This works when `border` is set to a full border property like `1px solid black`, but not a keyword like `inherit`.

Code
```js
import { CSSStyleDeclaration } from "cssstyle"

const style = new CSSStyleDeclaration();
style.border = "inherit";
style.borderWidth = "unset";
```

Exception
```sh
node_modules/.pnpm/cssstyle@6.2.0/node_modules/cssstyle/lib/normalize.js:290
          shorthandObj[key] = parsedValue;
                            ^

TypeError: Cannot create property 'border-width' on string 'inherit'
    at replaceBorderShorthandValue (node_modules/.pnpm/cssstyle@6.2.0/node_modules/cssstyle/lib/normalize.js:290:29)
    at prepareBorderArrayValue (node_modules/.pnpm/cssstyle@6.2.0/node_modules/cssstyle/lib/normalize.js:707:31)
    at prepareBorderProperties (node_modules/.pnpm/cssstyle@6.2.0/node_modules/cssstyle/lib/normalize.js:935:5)
    at CSSStyleDeclaration._borderSetter (node_modules/.pnpm/cssstyle@6.2.0/node_modules/cssstyle/lib/CSSStyleDeclaration.js:449:30)
    at CSSStyleDeclaration.set (node_modules/.pnpm/cssstyle@6.2.0/node_modules/cssstyle/lib/properties/borderWidth.js:29:14)

Node.js v22.18.0
```

Seems related to jsdom/cssstyle#247. Thanks very much!

mvolz:
Getting a similar issue! In the wild parsing https://www.wired.com/2008/05/batman-iron-man/ did it. I can reproduce this with the latest version of jsdom- 29.0.0

```
TypeError: Cannot create property 'border-width' on string 'var(--border-width, 1px)'
    at replaceBorderShorthandValue (node_modules/cssstyle/lib/normalize.js:197:29)
    at exports.prepareBorderProperties (node_modules/cssstyle/lib/normalize.js:624:30)
    at CSSStyleDeclaration.value (node_modules/cssstyle/lib/CSSStyleDeclaration.js:456:32)
    at CSSStyleDeclaration.set (node_modules/cssstyle/lib/generated/properties.js:2345:14)
    at CSSStyleDeclaration.setProperty (node_modules/cssstyle/lib/CSSStyleDeclaration.js:373:20)
[...]

PR #4116, opened 2026-03-17
Return empty string if parsed shorthand is not an object

asamuzaK:
Fix https://github.com/jsdom/jsdom/issues/4112


lib/jsdom/living/css/helpers/shorthand-properties.js
```diff
@@ -277,6 +277,9 @@ function replaceBorderShorthandValue(value, shorthandValue) {
     {
       [borderFirstInitialKey]: borderFirstInitialValue
     };
+  if (shorthandObj && typeof shorthandObj !== "object") {
```
domenic, line 280: Probably `if (typeof shorthandObj === "string")` would be a bit clearer. But this is fine.

PR #4117, opened 2026-03-19
Handle global keywords in CSS shorthand property handlers

domenic:
Several shorthand property handlers (border, background, and their sub-shorthands) crashed or produced incorrect results when encountering CSS-wide keywords like "inherit" or "initial". The root cause is that `parse()` functions return strings for global keywords, but many callers assumed object/array returns.

- In `processBorderProperties`, handle global keywords uniformly at the top of the loop before any property-specific branching, fixing `cssText = "border-top: inherit"` (and all border side/line shorthands) which previously threw a TypeError swallowed by the cssText setter's catch block.
- In `matchesBorderShorthandValue`, return false for non-object parse results, fixing `border: inherit` followed by setting a longhand like `borderTopWidth`.
- In `prepareBorderStringValue` and `prepareBorderObjectValue`, guard `replacePositionValue` calls against global keyword inputs, fixing `border: inherit` followed by setting a sub-shorthand like `borderTop`.
- In `background.parse()`, return global keywords as strings (matching `border.parse()` behavior). Update the descriptor setter to propagate them to all longhands and the getter to verify longhand consistency before returning.
- In `prepareProperties`, skip `replaceBackgroundShorthand` for global keyword shorthands, and fix a pre-existing Map mutation-during-iteration bug by updating `parsedProperties` instead of `properties`.
- Mark `cssom-setProperty-shorthand.html` as passing.

@asamuzaK after merging https://github.com/jsdom/jsdom/commit/d589a8e9fb3c8ca41ee10fa93a36fe0d7e23ef08 I asked Claude if it could find any similar issues and this is what it found. WDYT?


asamuzaK:
LGTM

PR #4119, opened 2026-03-19
Resolve computed style value

asamuzaK:
This pull request improves how jsdom resolves computed style values for DOM elements. This includes updating the style calculation logic, which will improve the accuracy and performance of handling computed CSS properties within jsdom.

Fixes #3614, fixes #3339, fixes #3563, fixes #3971, fixes #3972.
Related #3984, #4085, #3615.

PR #4133, opened 2026-03-31
Create system-color-compute-dont-upstream.html

asamuzaK:
PoC for #4119


asamuzaK:
I'll leave it up to you whether we merge this or not.

domenic:
I guess let's just use the upstream system-color-compute.html to track our progress on this problem? If we think that test can be improved then we can send a pull request upstream.

(I hope AI will make syncing WPT with upstream easier... I should try it soon.)

PR #4144, opened 2026-04-10
Extract LRU cache factory function for standardization

asamuzaK:
WIP


domenic:
I know that this is still marked as draft, but my initial take is that this isn't worth it: there's no real reason the defaults should be the same across different parts of jsdom, and saving less than a single line of code (`new LRUCache({ max: 2048 })` -> `createCache()`) is not a big win.

Additionally, we want to add as few caches as possible, only when they're really needed. So a module that encourages more caches to be added is kind of counterproductive.

asamuzaK:
Closing.

PR #4151, opened 2026-04-26
Support ratio type

### LINKED ISSUE #4150: aspect-ratio CSS property is discarded when using decimal values
### Node.js version

v24.15.0

### jsdom version

29.0.1

### Minimal reproduction case

```js

const dom = new JSDOM(`<!DOCTYPE html><div id="my-div"></div>`);
const el = dom.window.document.getElementById("my-div");

// Test 1: The Slash
el.style.setProperty("aspect-ratio", "16 / 9");
console.log("With Slash:  ", el.outerHTML);

// Reset
el.style.removeProperty("aspect-ratio");

// Test 2: The Decimal
el.style.setProperty("aspect-ratio", "1.7777");
console.log("With Decimal:", el.outerHTML);
```

In modern browsers setting `aspect-ratio` to a decimal value like `1.7777` is perfectly valid. The browser retains the property, and `el.outerHTML` outputs `aspect-ratio: 1.7777;`.
JSFiddle Link: https://jsfiddle.net/Lqznjwdv/2/

### What is the problem?

I expected jsdom to retain valid decimal values for the `aspect-ratio` CSS property when set via `setProperty`, just as modern browsers do.
Instead, `jsdom` silently discards decimal values resulting in an empty string when accessing `.outerHTML`.
Note: Slash syntax for `aspect-ratio` (like "16/9") is retained in the `.outerHTML` output string as intended.

asamuzaK:
Fix #4150 

Note: This PR adds support for the generic `<ratio>` CSS data type, but does not fully implement `aspect-ratio`.

### DIFF
--- a/lib/jsdom/living/css/helpers/css-values.js
+++ b/lib/jsdom/living/css/helpers/css-values.js
@@ -27,6 +27,7 @@ const AST_TYPES = Object.freeze({
   IDENTIFIER: "Identifier",
   NUMBER: "Number",
   PERCENTAGE: "Percentage",
+  RATIO: "Ratio",
   STRING: "String",
   URL: "Url"
 });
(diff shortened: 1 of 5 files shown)


domenic (approved):
Great stuff, thanks!

Issue #4153, opened 2026-04-27
29.1.0 regressions: `border-radius` shorthand returns "" + crash on `<h*>` with inline gradient


pbakaus:
## Summary

Two CSS-related regressions appeared in `jsdom@29.1.0` (released 2026-04-27). Both reproduce against a fresh install with no other dependencies.

`29.0.0` does not exhibit either issue.

> **Edit (2026-04-28):** rewrote the regression-1 repro after [@asamuzaK's note](https://github.com/jsdom/jsdom/issues/4153#issuecomment-4339123414) — 29.0.0 never expanded `border-radius` shorthand to longhands, so `borderTopLeftRadius` was always `""` there. The actual regression is on the **shorthand read**, not the longhand. Also re-bisected regression 2 down to a much simpler trigger.

---

### Regression 1: `getComputedStyle().borderRadius` returns `""`

**Repro** (Node 24, jsdom 29.1.0):

```js
import { JSDOM } from 'jsdom';

const dom = new JSDOM(`
  <style>.x { width: 64px; height: 64px; border-radius: 12px; }</style>
  <div class="x"></div>
`);
const el = dom.window.document.querySelector('.x');
const s = dom.window.getComputedStyle(el);

console.log(s.borderRadius);                       // "" in 29.1.0, "12px" in 29.0.0
console.log(s.getPropertyValue('border-radius'));  // "" in 29.1.0, "12px" in 29.0.0
```

The shorthand reads as `""` regardless of the value (`12px`, `50%`, etc.) and regardless of where the rule lives — `<style>` block or inline `style="…"`:

```js
const dom2 = new JSDOM(`<div style="border-radius: 12px"></div>`);
const el2  = dom2.window.document.querySelector('div');
console.log(el2.style.borderRadius);                                 // "12px" — works
console.log(dom2.window.getComputedStyle(el2).borderRadius);         // "" in 29.1.0, "12px" in 29.0.0
```

So `el.style.*` direct reads still work; only `getComputedStyle().borderRadius` (and `getPropertyValue('border-radius')`) broke.

**Why this matters**: any consumer that reads `getComputedStyle(el).borderRadius` to check for a value (e.g. detecting circular avatars via `>= width/2`, or rounded-card heuristics via `> 0`) gets a silent empty/false. We hit this in a CSS anti-pattern detector and downgraded back to `29.0.0`.

(The longhands like `borderTopLeftRadius` weren't expanded by 29.0.0 either — that's not where the regression is.)

[...]

PR #4154, opened 2026-04-28
Exclude border-radius from border shorthand handler

asamuzaK:
Partially addresses the regression of `border-radius` in the first half of #4153.
Note that it does not fix `border-radius` longhands.

### DIFF
--- a/lib/jsdom/living/css/CSSStyleDeclaration-impl.js
+++ b/lib/jsdom/living/css/CSSStyleDeclaration-impl.js
@@ -392,7 +392,7 @@ class CSSStyleDeclarationImpl {
         return this.#resolvePositionShorthand(property);
       }
       default: {
-        if (property.startsWith("border")) {
+        if (property.startsWith("border") && !property.endsWith("radius")) {
           return this.#resolveBorderShorthands(property);
         }
         return value;
--- a/lib/jsdom/living/css/CSSStyleDeclaration-impl.js
+++ b/lib/jsdom/living/css/CSSStyleDeclaration-impl.js
@@ -391,10 +391,17 @@ class CSSStyleDeclarationImpl {
       case "padding": {
         return this.#resolvePositionShorthand(property);
       }
+      case "border":
+      case "border-top":
+      case "border-right":
+      case "border-bottom":
+      case "border-left":
+      case "border-width":
+      case "border-style":
+      case "border-color": {
+        return this.#resolveBorderShorthands(property);
+      }
       default: {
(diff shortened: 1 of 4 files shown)


lib/jsdom/living/css/CSSStyleDeclaration-impl.js
```diff
@@ -392,7 +392,7 @@ class CSSStyleDeclarationImpl {
         return this.#resolvePositionShorthand(property);
       }
       default: {
-        if (property.startsWith("border")) {
+        if (property.startsWith("border") && !property.endsWith("radius")) {
```
domenic: I think this is too fragile. It misses, e.g., `border-block`, `border-clip`, etc.

`#resolveBorderShorthands` only handles a specific set: `border`, `border-top`/right/bottom/left, and `border-width`, `border-style`, and `border-color`.

What about doing this instead?

```js
switch (property) {
  case "margin":
  case "padding":
    return this.#resolvePositionShorthand(property);
  case "border":
  case "border-top":
  case "border-right":
  case "border-bottom":
  case "border-left":
  case "border-width":
  case "border-style":
  case "border-color":
    return this.#resolveBorderShorthands(property);
  default:
    return value;
}
```

test/web-platform-tests/to-upstream/css/css-borders/border-radius-shorthand.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>border-radius shorthand</title>
```
domenic, line 3: Might be worth going through all the properties that start with `border-` (in the generated property definitions file) and adding test cases for those. A coding agent would probably do it quickly.

domenic (commented):
Thanks for jumping on this. I can merge this ASAP if you would prefer, but I think we should do better eventually, so I have left some suggestions.

test/web-platform-tests/to-upstream/css/css-borders/border-radius-shorthand.html
asamuzaK: How about doing that in a separate PR?

The CSS Backgrounds and Borders Module has been integrated into css-backgrounds.
Therefore, I think we need to consolidate the local wpt tests, which are currently scattered across css-backgrounds/, css-borders/, and cssom/, into a single directory.

At that time, how about adding border-related tests?

domenic: Sure, that sounds great to me.

domenic (approved):

PR #4155, opened 2026-04-29
Fix background-origin, background-clip handler in background shorthand

asamuzaK:
Fixes latter issue of #4153


domenic (approved):
The `background` property is so complicated, it took a long time for me to understand the spec enough to see why the test expectations were correct! Nice work.

PR #4159, opened 2026-05-02
Add handlers to convert values to pixels when calculating computed values

### LINKED ISSUE #3538: calc expressions with parentheses are ignored or treated as invalid CSS values
The CSS value spec [allows for](https://www.w3.org/TR/css-values-3/#calc-syntax) parentheses to be used to group sub-expressions in `calc()` in order to force operation order when the natural precedence isn't sufficient for the expression.

But when `getComputedStyle()` is used to retrieve the `calc()` expression, if it contains any pairs of parens, the reported value is `""` instead of the expected expression.

### Basic info:

- **Node.js version:** v19.7.0
- **jsdom version:** v21.1.1

### Minimal reproduction case

```js

[
  'calc(1px)',
  'calc((1px))',
  'calc(1px + 4px)',
  'calc(1px + (4px))',
  'calc(1px + 4px / 2)',
  'calc(1px + (7px - 3px) / 2)',
].forEach((testCase, idx) => {
  const dom = new JSDOM(`
    <!DOCTYPE html>
    <p style="width: ${testCase};">hello</p>
  `);
  const style = dom.window.getComputedStyle(
    dom.window.document.querySelector('p')
  );
  const widthStyle = style.getPropertyValue('width');
  console.log(
    'test case %s: width="%s" | expected="%s" // %s',
    idx,
    widthStyle,
    testCase,
    widthStyle === testCase ? '✅ PASS' : '⛔️ FAIL'
  );
});
```

Output:

```
test case 0: width="calc(1px)" | expected="calc(1px)" // ✅ PASS
test case 1: width="" | expected="calc((1px))" // ⛔️ FAIL
test case 2: width="calc(1px + 4px)" | expected="calc(1px + 4px)" // ✅ PASS
test case 3: width="" | expected="calc(1px + (4px))" // ⛔️ FAIL
test case 4: width="calc(1px + 4px / 2)" | expected="calc(1px + 4px / 2)" // ✅ PASS
test case 5: width="" | expected="calc(1px + (7px - 3px) / 2)" // ⛔️ FAIL
```

https://jsbin.com/zeximin/edit?html,js,console,output

The output in the browser is actually different on __all__ counts because the `calc` expressions are actually evaluated, but if you use DevTools to inspect the widths, the original `calc` expressions, as JSDom should report them, are all shown as valid CSS values.

```
test case 0: width="1px" | expected="calc(1px)" // ⛔️ FAIL
test case 1: width="1px" | expected="calc((1px))" // ⛔️ FAIL
test case 2: width="5px" | expected="calc(1px + 4px)" // ⛔️ FAIL
test case 3: width="5px" | expected="calc(1px + (4px))" // ⛔️ FAIL
test case 4: width="3px" | expected="calc(1px + 4px / 2)" // ⛔️ FAIL
test case 5: width="3px" | expected="calc(1px + (7px - 3px) / 2)" // ⛔️ FAIL
```

Here's test case 6 in Chrome (Version 112.0.5615.137 (Official Build) (arm64)) DevTools:
Screenshot 2023-04-26 at 12 33 03 PM

❗ But __I'm not implying jsdom needs to eval the expressions__. I'm just saying it should allow for parentheses within `calc()` expressions and return the expression, not an empty string as though the specified value was invalid. The CSS spec does allow for parentheses: https://www.w3.org/TR/css-values-3/#calc-syntax

I did a bit of digging and I think the problem may lie deeper here: https://github.com/rrweb-io/CSSOM/blob/402a2c7a813890c6c116244827f200cd6c04f0bb/lib/CSSValueExpression.js#L36 But that repo doesn't allow filing issues so I figured I would report this here.

asamuzaK:
### Description
This PR introduces the necessary logic to resolve relative and absolute CSS units (such as `em`, `rem`, `vh`, `vw`, `pt`, etc.) into pixel values when querying `getComputedStyle`. This enhancement aligns the CSSOM implementation more closely with standard browser behavior.
Fixes #3538

**Implementation Notes**
* **Extracted font-size logic:** Added a new module `font-sizes.js` to cleanly encapsulate the font-size mapping tables and the conversion scripts (`resolveEmInPixels`, `resolveFontSizeInPixels`, `resolveLengthInPixels`).
* **Absolute font-size mapping:** The [CSS Fonts 4 specification](https://drafts.csswg.org/css-fonts-4/#absolute-size-mapping) provides guidelines for scaling factors but explicitly notes that *"The user agent may fine-tune these values for different fonts or different types of display devices."* In practice, major browsers do not strictly adhere to the suggested factors for keywords like `large`, `small`, `x-small`, and `xx-small`. Therefore, this implementation aligns with observed browser behavior to ensure compatibility.
* **Resolved internal circular references:** Internal property setters (e.g., `_borderSetter`, `_flexBoxSetter`, `_positionLonghandSetter`) were prone to stack overflows when they eagerly called `this.getPropertyValue(property)`. To safely handle shorthand/longhand expansion without triggering infinite recursive calculations, these calls have been updated to directly read raw values using `this.#values.get(property)`.
* **Potential Breaking Change:** While this PR is fundamentally a spec-compliant fix, tests that rely on jsdom's previous behavior (e.g., expecting relative units like `em` to be returned uncomputed) are likely to fail. It might be better to treat this as a breaking change. However, note that percentage values for properties other than `font-size` (such as `width`) will be returned as `%`, maintaining the previous behavior.


test/web-platform-tests/to-upstream/css/css-fonts/font-size-computed.html
domenic, line 1: Can you add a similar test using inline styles? I don't think it will work with the current architecture.

E.g. some local testing shows the following bug:

```js
const el = document.body;

el.style.fontSize = "2em";
getComputedStyle(el).fontSize; // "64px", should be 2 * 16 = 32px

el.style.fontSize = "0.5em";
getComputedStyle(el).fontSize; // "4px", should be 0.5 * 16 = 8px

el.style.fontSize = "50%";
getComputedStyle(el).fontSize; // "4px, should be 0.5 * 16 = 8px
```

asamuzaK: Since we have below, resolved value should be 24px for 2em, not 32px, and 6px for 0.5em and 50%?
```
:root {
  font-size: 75%;
}
```

However, Chrome is returning 10px for 0.5em, and 50%.
I'm not sure why.

asamuzaK: It seems the test was hitting Chrome's minimum font size limit.
Changed value from 0.5em to 0.9em.

domenic:
This fixed inline `font-size` fix, but the same underlying issue still affects non-`font-size` properties that use `em`. Repro:

```js
// parent inline font-size:50% → computed 8px; child inherits → child's computed fs is also 8px
// so width:1em on the child should be 8px
<div style="font-size:50%"><div id=c style="width:1em"></div></div>
getComputedStyle(c).width // → "4px", expected "8px"

<div style="font-size:200%"><div id=c style="width:1em"></div></div>
// → "64px", expected "32px"
```

The same cases work when the parent's font-size comes from a stylesheet rather than `style="..."`.

The root cause is the same as before: `collectFontSizes` reads `node.style.getPropertyValue("font-size")` (inline only) at every ancestor, so an inherited inline value gets counted twice — once when discovered via inheritance from the child, once again as the parent's own inline value. The `font-size` branch you added in `#resolveLonghand` sidesteps this by reading the parent's already-computed font-size; for other properties, `options.dimension.em` still goes through the broken path.

Rather than mirror the same workaround for non-font-size properties, it might be worth retiring `collectFontSizes` and computing `em` directly from the cached computed declarations: for `font-size` itself, `em = parseFloat(parent.getPropertyValue("font-size"))` (what you already do); for everything else, `em = parseFloat(this.getPropertyValue("font-size"))` once the element's own font-size has resolved. That removes the inline/stylesheet asymmetry entirely and avoids the multi-walk overhead per `getComputedStyle` call. WDYT?

asamuzaK:
I see.
However, I'm a little worried about the possibility of circular references occurring if there's no cache.
I'll try implementing it again and see how it goes.

Converting to draft again.

asamuzaK:
Rewritten font size resolver.

domenic:
I think it is still not working. Can you try passing the following test?

```html
<!DOCTYPE html>
<meta charset="utf-8">
<title>Font-relative units on non-font-size properties resolve against the element's own font-size</title>
<link rel="help" href="https://drafts.csswg.org/css-values-4/#font-relative-lengths">

<!--
  Per CSS Values 4 §6.1.1: "the font-relative lengths refer to the font metrics
  of the element on which they are used. The exception is when they occur in
  the value of the font-size property itself, in which case they refer to the
  font metrics of the parent element."

  So for `em` on `width`, `padding`, etc., the unit must resolve against
  the element's own computed font-size, not its parent's.
-->

<div id="own-px"      style="font-size: 20px; width: 1em"></div>
<div id="own-px-pad"  style="font-size: 32px; padding: 1em"></div>
<div id="own-px-calc" style="font-size: 20px; width: calc(1em + 5px)"></div>

<div style="font-size: 30px">
  <div id="own-overrides-parent" style="font-size: 50px; width: 2em"></div>
</div>

<div style="font-size: 30px">
  <div id="inherits-from-parent" style="width: 2em"></div>
</div>

<script>
"use strict";

test(() => {
  const el = document.getElementById("own-px");
  assert_equals(getComputedStyle(el).width, "20px");
}, "width: 1em uses the element's own font-size, not the parent's");

test(() => {
  const el = document.getElementById("own-px-pad");
  assert_equals(getComputedStyle(el).paddingTop, "32px");
}, "padding: 1em uses the element's own font-size");

test(() => {
  const el = document.getElementById("own-px-calc");
  assert_equals(getComputedStyle(el).width, "25px");
}, "calc(1em + 5px) uses the element's own font-size");

test(() => {
  const el = document.getElementById("own-overrides-parent");
  // own font-size 50px overrides parent 30px; 2em = 100px, not 60px
  assert_equals(getComputedStyle(el).width, "100px");
}, "Own font-size override takes precedence over parent's for em");

test(() => {
  const el = document.getElementById("inherits-from-parent");
  // No own font-size, inherits 30px from parent; 2em = 60px
  assert_equals(getComputedStyle(el).width, "60px");
}, "When element has no own font-size, em resolves against inherited value");
</script>
```

asamuzaK:
Thank you for the thorough review.
I'm really impressed by your insights.
Hope I have addressed all the issues you pointed out.

domenic (approved):

PR #4160, opened 2026-05-05
WPT maintenance

asamuzaK:
WPT maintenance
* Update wpt to the latest
* Add css-borders and css-box
* Skip tentative tests


asamuzaK:
Regarding css/selectors/old-tests/**:
* We are currently skipping all of them at once.
  Is this fine, or should we handle each file individually?
* If skipping all is fine, we are currently using `fail-slow` as the reason for skipping.
  Should we instead add a new reason like `obsolete` to `resolveReason(reason)` in `expectations-utils.js`?

Regarding tentative tests:
* Do we need to run tentative tests?

domenic:
> * We are currently skipping all of them at once.
>       Is this fine, or should we handle each file individually?

It would be better to handle them individually.

These tests are not obsolete, or value-less. They are just old. They still provide valuable coverage, and [according to wpt.fyi](https://wpt.fyi/results/css/selectors/old-tests?label=experimental&label=master&aligned), browseres pass them.

> * Do we need to run tentative tests?

No. Tentative tests are for features that don't yet have a solid spec. jsdom almost always requires a spec for what it implements.

I think there might already be a runner-level exclusion for all tentative tests?

In rare cases there are tentative tests that are for features that only recently got a good spec. In that case the test writer has just not yet renamed the test to remove `.tentative`. In such cases it's best to fix the filename upstream.

asamuzaK:
PR ready.

test/web-platform-tests/to-run.yaml
```diff
@@ -813,7 +1033,14 @@ DOMTokenList-coverage-for-attributes.html: [fail, Several DOMTokenList attribute
 
 DIR: dom/nodes
```
domenic, line 1034: A lot of the new failures here are scary, especially since they are now blanket failures instead of subtest failures, so we lose coverage. So I will work on a follow-up PR to get as many passing as possible.

domenic (approved):
Thank you so much for this work.

Issue #4147, opened 2026-04-17
jsdom XPath public API throws uncaught TypeError for valid expression


oetr:
### Node.js version

v22.22.2, v24

### jsdom version

v29.0.2

### Minimal reproduction case

```js
const { JSDOM } = require("jsdom");

const dom = new JSDOM("<!doctype html><html><body><div>alpha<span>beta</span></div></body></html>");
const { document, XPathResult } = dom.window;

const result = document.evaluate("string(/)", document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null);
console.log(result.singleNodeValue);
```

### How does similar code behave in browsers?

In Chromium-based browsers, the call throws immediately from `document.evaluate()`. In jsdom, it does not fail the same way, which suggests missing validation of a non-node-set result against `FIRST_ORDERED_NODE_TYPE`.

### What is the problem?

Observed result:

```
TypeError: Cannot read properties of null (reading 'nodes')
    at get singleNodeValue (.../lib/jsdom/level3/xpath.js:1810:26)
```

The core issue is that jsdom does not properly validate the requested result type against the actual expression result type.Per the DOM Level 3 XPath specification, requesting FIRST_ORDERED_NODE_TYPE for a string-valued expression should raise a TYPE_ERR DOMException. Instead, jsdom omits that check and allows a null reference to propagate as a TypeError. This is a spec-compliance bug.

afonsojanu:
Just tried the exact repro from this issue against current `main` (v30.0.1) and it no longer reproduces the reported crash:

```js
document.evaluate("string(/)", document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null);
// TypeError: The result is not a node set, and therefore cannot be converted to the desired type.
//   thrown from evaluate() itself, not from inside singleNodeValue
```

That matches the "Expected behavior" described above (throws immediately from `document.evaluate()`, doesn't let a null reference leak into `singleNodeValue`). Looks like this got fixed by a673621c ("Validate XPath result type in document.evaluate()"), which also references this issue and adds a WPT-style test at `test/web-platform-tests/to-upstream/domxpath/evaluate-non-node-set-result-type.html` covering all six node-set result types against `string(/)`.

[...]

PR #4163, opened 2026-05-06
Validate XPath result type in document.evaluate()

dokson:
## Summary

- When `document.evaluate()` is called with a node-set `resultType` but the expression yields a non-node-set value (or `null`), it now raises `XPathException.TYPE_ERR` (code 52) at evaluation time, matching DOM Level 3 XPath. Previously the call returned an `XPathResult` with an unusable internal value, and a generic `TypeError` only surfaced later from `singleNodeValue` / `snapshotLength` / `iterateNext`.
- The previous validation rejected only non-`object` values, but `typeof null === "object"` so `null` slipped through. The check is now `value === null || typeof value !== "object" || !Array.isArray(value.nodes)`.
- This addresses the spec-compliance side of #4147. The underlying evaluator bug (`string(/)` returning `null` instead of a string) is intentionally **not** fixed in this PR and is left for a separate change — the comment at `lib/jsdom/level3/xpath.js:1754` already notes the XPath implementation should preferably be replaced wholesale.

Refs #4147

## Test plan

- [x] Added a WPT-format regression test at `test/web-platform-tests/to-upstream/domxpath/evaluate-non-node-set-result-type.html` covering all six node-set result types (iterator/snapshot ord+unord, `ANY_UNORDERED_NODE_TYPE`, `FIRST_ORDERED_NODE_TYPE`).
- [x] `npm run test:tuwpt -- --fgrep "evaluate-non-node-set-result-type" --reporter min` passes.
- [x] `npm run test:tuwpt -- --reporter min` — full to-upstream suite: 373 passing, no regressions.
- [x] `npm run test:api -- --reporter min` — 574 passing, no regressions.

### DIFF
--- a/lib/jsdom/level3/xpath.js
+++ b/lib/jsdom/level3/xpath.js
@@ -1737,10 +1737,9 @@ module.exports = core => {
             core.DOMException.NOT_SUPPORTED_ERR,
             'You must provide an XPath result type (0=any).');
       else if (XPathResult.ANY_TYPE !== type &&
-               (value === null || 'object' !== typeof value || !Array.isArray(value.nodes)))
-        throw new XPathException(
-            XPathException.TYPE_ERR,
-            'The expression cannot be converted to return the specified type.');
+               (value === null || 'object' !== typeof value))
+        throw new TypeError(
+            'The result is not a node set, and therefore cannot be converted to the desired type.');
       return new XPathResult(doc, value, type);
     }
   }
--- a/lib/jsdom/level3/xpath.js
+++ b/lib/jsdom/level3/xpath.js
@@ -1737,10 +1737,10 @@ module.exports = core => {
             core.DOMException.NOT_SUPPORTED_ERR,
             'You must provide an XPath result type (0=any).');
       else if (XPathResult.ANY_TYPE !== type &&
-               'object' !== typeof value)
+               (value === null || 'object' !== typeof value || !Array.isArray(value.nodes)))
         throw new XPathException(
             XPathException.TYPE_ERR,
-            'Value should be a node-set: ' + value);
+            'The expression cannot be converted to return the specified type.');
       return new XPathResult(doc, value, type);
     }
   }
--- a/test/web-platform-tests/to-upstream/domxpath/evaluate-non-node-set-result-type.html
+++ b/test/web-platform-tests/to-upstream/domxpath/evaluate-non-node-set-result-type.html
@@ -1,36 +1,27 @@
 <!DOCTYPE html>
-<title>Document.evaluate() with a node-set result type for a non-node-set expression throws</title>
+<title>document.evaluate() with a node-set result type for a non-node-set expression throws</title>
 <meta charset="utf-8">
 <script src="/resources/testharness.js"></script>
 <script src="/resources/testharnessreport.js"></script>
 <!-- https://github.com/jsdom/jsdom/issues/4147 -->
 
-<div id="testDiv">alpha<span>beta</span></div>
-
 <script>
 "use strict";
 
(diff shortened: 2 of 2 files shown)


domenic (commented):
Thanks for working on this. The test does not appear to match browser behavior, though: https://jsbin.com/mutemizelo/edit?html,output

dokson:
Thanks for the catch. You're right — real browsers throw a plain `TypeError`, not an `XPathException` with `code === 52`. Pushed 23242dc0 to:

- Throw `TypeError` from `XPathExpression.prototype.evaluate` instead of constructing an `XPathException` (`lib/jsdom/level3/xpath.js`).
- Update the WPT test to use `assert_throws_js(TypeError, ...)`, which is also the convention in the existing `domxpath/evaluator-constructor.html` upstream test.

Local `npm run test:tuwpt` and `npm run test:api` are still green.

test/web-platform-tests/to-upstream/domxpath/evaluate-non-node-set-result-type.html
```diff
@@ -0,0 +1,29 @@
...
+"use strict";
+
+const nodeSetTypes = [
+  ["UNORDERED_NODE_ITERATOR_TYPE", XPathResult.UNORDERED_NODE_ITERATOR_TYPE],
+  ["ORDERED_NODE_ITERATOR_TYPE", XPathResult.ORDERED_NODE_ITERATOR_TYPE],
+  ["UNORDERED_NODE_SNAPSHOT_TYPE", XPathResult.UNORDERED_NODE_SNAPSHOT_TYPE],
+  ["ORDERED_NODE_SNAPSHOT_TYPE", XPathResult.ORDERED_NODE_SNAPSHOT_TYPE],
+  ["ANY_UNORDERED_NODE_TYPE", XPathResult.ANY_UNORDERED_NODE_TYPE],
+  ["FIRST_ORDERED_NODE_TYPE", XPathResult.FIRST_ORDERED_NODE_TYPE]
+];
```
domenic, line 18: Please simplify this to just the names, since you can look up the values using `XPathResult[name]`.

test/web-platform-tests/to-upstream/domxpath/evaluate-non-node-set-result-type.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
+<title>Document.evaluate() with a node-set result type for a non-node-set expression throws</title>
```
domenic: `Document.evaluate()` is not a function. Probably you meant `document.evaluate()`.

lib/jsdom/level3/xpath.js
```diff
@@ -1737,10 +1737,9 @@ module.exports = core => {
             core.DOMException.NOT_SUPPORTED_ERR,
             'You must provide an XPath result type (0=any).');
       else if (XPathResult.ANY_TYPE !== type &&
-               'object' !== typeof value)
-        throw new XPathException(
-            XPathException.TYPE_ERR,
-            'Value should be a node-set: ' + value);
+               (value === null || 'object' !== typeof value || !Array.isArray(value.nodes)))
```
domenic: I don't see any test coverage for the new `!Array.isArray(value.nodes)` part of this.

dokson: dropping this one because it's unreachable...

test/web-platform-tests/to-upstream/domxpath/evaluate-non-node-set-result-type.html
dokson: yep, my fault

PR #4166, opened 2026-05-15
Update the processing of the function value

asamuzaK:
A fix for a bug found during the implementation of #4162


domenic (approved):
Nice work.

I hope to one day investigate switching to a more spec-compliant parser/serializer to solve a bunch of these issues at once. (Many more came up during https://github.com/jsdom/jsdom/pull/4092.) But I am not sure any are suitable. I researched it a while ago. https://chatgpt.com/share/6a0853e6-5c14-83aa-a862-ce78b8a5e53c

PR #4170, opened 2026-05-20
Bump Node.js floor and dependencies

domenic:
Raise the minimum supported Node.js version to exclude Node 20, and update all direct dependencies and transitive lockfile resolutions for the newer runtime floor.

Remove compatibility code that is no longer needed with the new Node.js and undici floors, including old undici handler adaptation, Node 20 ArrayBuffer.transfer fallbacks, direct minimatch usage for WPT expectations, and Node 20-only WPT skips.

The @asamuzakjp/dom-selector 8 universal-selector fast path calls the internal createTreeWalker implementation directly, bypassing the generated Web IDL default for its filter argument. Repeat that default in DocumentImpl so universal querySelectorAll() calls continue to return elements in tree order.


domenic:
Seems like the failures are in `dom/nodes/ParentNode-querySelector-All-dont-upstream.html` on the fragment case, probably due to the dom-selector upgrade...

asamuzaK:
Wouldn't it be better not to change the minimum support for v22 and v24?

domenic:
> Wouldn't it be better not to change the minimum support for v22 and v24?

It would be, but some of our dependencies / dev deps have higher minimums... Also some of them exclude v25, which I am not sure what to do about.

Edit: specifically, it's `npm-run-all2@9.0.0` that is forcing this.

I guess I am OK with it because I generally prefer people be running the latest versions of things anyway, and only support older things because I want to respect the semver contract.

asamuzaK:
I also agree with encouraging the use of the latest Node.js, but taking in this npm-run-all2 update feels too impactful considering where jsdom stands. Could we just stick with the current npm-run-all2 for now and maybe consider swapping to https://github.com/open-cli-tools/concurrently in a separate PR?

domenic:
I don't really understand "too impactful". Who is impacted?

asamuzaK:
To clarify, the impact I'm concerned about is on the broader ecosystem and the end users.

While it’s completely understandable to introduce breaking changes in a major version bump, jsdom raising its minimum Node.js requirement practically means that many dependent projects will be forced to do the same.

If this change were driven by an upgrade in jsdom's own core dependencies, it should be okay.
However, this is just about one of the devDependencies.
From the users' perspective, it will appear as though Node.js support was dropped without any substantial reason.

Consequently, we might end up with a counterproductive situation where, contrary to our intent, nobody adopts the new jsdom version, and we just face unnecessary backlash.
Users aren't quick to upgrade Node, and they don't care about our view that staying on the latest version is the best practice.
I'm raising this because it's a lesson I learned from when we (me) bumped parse5 to a new major version.

PR #4164, opened 2026-05-12
Optimize cache handling

asamuzaK:
This PR optimizes cache handling.

PR #4171, opened 2026-05-21
Replace lru-cache with generational-cache 2nd attempt

asamuzaK:
Cut branch from #4164 
Closes #4164, also closes #4165


asamuzaK:
Since we aren't facing any issues with lru-cache at the moment, I'll withdraw this PR for now.

PR #4175, opened 2026-06-01
Add Blob.stream

ThunderKey:
This adds [Blob.stream()](https://developer.mozilla.org/en-US/docs/Web/API/Blob/stream) to the `BlobImpl`.

The [ReadableStream](https://nodejs.org/api/webstreams.html#class-readablestream) was added to node in 16.5.0, so it matches the range in `package.json`.

Since I'm new to the project I don't know exactly how the `ReadableStream` should be handled. Is it ok like that, or should it be added to the `Window` globals as well? If so, should I just expose the one from `node:stream/web` directly, which would be a pragmatic approach, or should it be recreated fully? The latter would be quite a lot of effort.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -141,6 +141,8 @@ export default [
         promise_test: "readonly",
         promise_rejects: "readonly",
         promise_rejects_dom: "readonly",
+        promise_rejects_js: "readonly",
+        promise_rejects_exactly: "readonly",
         generate_tests: "readonly",
         setup: "readonly",
         done: "readonly",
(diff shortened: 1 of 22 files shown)


domenic:
Unfortunately yes, we need our own implementation of `ReadableStream`; using Node.js globals (which are singletons belonging to `globalThis`) does not work well with jsdom `Window`s (which are not `=== globalThis`).

ThunderKey:
That's a pity. I've created a minimal implementation with Claude Code for the `Blob.stream()`. I test it in a project of ours and tried to review it, but I'm not that well versed with the real implementation and this project in general.

If this is too sketchy for you, feel free to close this PR :)

PR #4178, opened 2026-06-05
Keep mixed length-percentage and calc() track lists in grid-template-columns

### LINKED ISSUE #4141: `grid-template-columns` silently rejects valid top-level track lists that mix bare `<length-percentage>` and `calc()`
### Node.js version

v24.8.0

### jsdom version

29 (and below)

### Minimal reproduction case

```js

const { window } = new JSDOM('<div id="a"></div>')
const el = window.document.getElementById('a')

for (const value of [
  '100px 1fr',
  '100px calc(100% - 100px)', // not working, returns ""
  'calc(100px) calc(100% - 100px)',
]) {
  el.style.gridTemplateColumns = ''
  el.style.gridTemplateColumns = value
  console.log(JSON.stringify(value), '->', JSON.stringify(el.style.gridTemplateColumns))
}
```

In the browser the 2nd value is parsed correctly:

```
"100px 1fr" -> "100px 1fr"
"100px calc(100% - 100px)" -> "100px calc(100% - 100px)"
"calc(100px) calc(100% - 100px)" -> "calc(100px) calc(100% - 100px)"
```

### What is the problem?

When a `grid-template-columns` track list mixes:

- a bare `<length-percentage>` token, such as `100px`, `60%`, or `6.25rem`
- and a top-level `calc(...)` token

jsdom drops the whole declaration.

This appears to be a parser bug. The same values are accepted by browsers, and the CSS Grid grammar allows the mix because `<track-breadth>` accepts `<length-percentage>`, and `calc()` resolves to `<length-percentage>`.

This seems specific to **top-level** mixed track lists.

These also work in jsdom:

```js
el.style.gridTemplateColumns = '40% 60%'
el.style.gridTemplateColumns = '6.25rem 50%'
el.style.gridTemplateColumns = 'calc(100%) calc(50%)'
el.style.gridTemplateColumns = 'repeat(2, minmax(calc(50% - 2rem), 50%))'
el.style.gridTemplateColumns = 'minmax(8rem, calc(100% - 8rem))'
```

But these do not:

```js
el.style.gridTemplateColumns = '100px calc(100% - 100px)'
el.style.gridTemplateColumns = '6.25rem calc((100% - 6.25rem) / 1)'
el.style.gridTemplateColumns = '6.25rem 60% calc((40% - 6.25rem) / 2) calc((40% - 6.25rem) / 2)'
```

I initially hit this through a test failure after upgrading to jsdom 29, but the underlying parser behavior already existed in 28.x. jsdom. v29 only made it more visible because `getComputedStyle()` now returns the CSS initial value (`none`) for unset `grid-template-columns`, where jsdom 28 returned `""`.

viraatdas:
Fixes #4141.

`grid-template-columns` (and the sibling grid track-list properties) silently rejected valid top-level track lists mixing a bare `<length-percentage>` with `calc()` (e.g. `100px calc(100% - 100px)`), because the value was reassembled from a name/value fallback that dropped `calc()` items. Regenerate each item from its parsed CSS node instead. Added a test covering mixed length/percentage/calc track lists.

### DIFF
--- a/lib/jsdom/living/css/helpers/css-values.js
+++ b/lib/jsdom/living/css/helpers/css-values.js
@@ -156,7 +156,7 @@ function resolveCalc(val, opt = { format: "specifiedValue" }) {
     } else if (itemType === AST_TYPES.STRING) {
       values.push(`"${itemValue}"`);
     } else {
-      values.push(itemName ?? itemValue);
+      values.push(cssTree.generate(item));
     }
   }
   const resolvedValue = values.join(" ");
--- a/test/to-port-to-wpts/level2/style.js
+++ b/test/to-port-to-wpts/level2/style.js
@@ -287,6 +287,29 @@ describe("level2/style", () => {
     window.getComputedStyle(window.document.body);
   });
 
+  specify("gridTemplateColumns mixed length + calc (GH-4141)", () => {
+    const document = (new JSDOM()).window.document;
+    const div = document.createElement("div");
+    const values = [
+      "100px 1fr",
+      "40% 60%",
+      "6.25rem 50%",
+      "calc(100px) calc(100% - 100px)",
+      "calc(100%) calc(50%)",
+      "repeat(2, minmax(calc(50% - 2rem), 50%))",
+      "minmax(8rem, calc(100% - 8rem))",
+      "100px calc(100% - 100px)",
+      "6.25rem calc((100% - 6.25rem) / 1)",
+      "6.25rem 60% calc((40% - 6.25rem) / 2) calc((40% - 6.25rem) / 2)"
+    ];
+
+    for (const value of values) {
+      div.style.gridTemplateColumns = "";
+      div.style.gridTemplateColumns = value;
+      assert.equal(div.style.gridTemplateColumns, value);
+    }
+  });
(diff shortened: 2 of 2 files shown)


viraatdas:
Closing this one. Digging into the CI failure, the root cause of #4141 is that `resolveCalc` (lib/jsdom/living/css/helpers/css-values.js) drops the unit on bare `<dimension>` tokens, so a track list like `100px calc(100% - 100px)` reconstructs as `100 calc(...)` and gets rejected.

Preserving the unit there does fix the grid case, but it regresses `box-shadow: 1px calc(2px + 2%)` (correctly invalid, since `%` is not allowed in a `<length>`). That rejection currently depends on the same unit-dropping behaviour, and the `<length>` vs `<length-percentage>` distinction needed to scope it lives in cssstyle rather than here, so it cannot be fixed correctly inside this single change.

The right fix belongs at the cssstyle layer (or by threading property context through value resolution). Closing rather than leaving a regression; happy to reopen with a proper cross-package fix.

domenic:
cssstyle is no longer a dependency of this package, so I'm not sure exactly what you're saying.

Issue #4183, opened 2026-07-09
Undici Security Vulnerability


FynniX:
Hello,
in our audit we have two vulnerabilities from jsdom dependency **undici**.

Would it be possible to update the dependency?

```
┌─────────────────────┬────────────────────────────────────────────────────────┐
│ high                │ undici vulnerable to TLS certificate validation bypass │
│                     │ via dropped requestTls in SOCKS5 ProxyAgent            │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Package             │ undici                                                 │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Vulnerable versions │ >=7.23.0 <7.28.0                                       │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Patched versions    │ >=7.28.0                                               │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Paths               │ .>jsdom>undici                                         │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ More info           │ https://github.com/advisories/GHSA-vmh5-mc38-953g      │
└─────────────────────┴────────────────────────────────────────────────────────┘
┌─────────────────────┬────────────────────────────────────────────────────────┐
│ high                │ undici WebSocket client vulnerable to denial of        │
│                     │ service via fragment count bypass                      │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Package             │ undici                                                 │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Vulnerable versions │ >=7.0.0 <7.28.0                                        │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Patched versions    │ >=7.28.0                                               │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Paths               │ .>jsdom>undici                                         │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ More info           │ https://github.com/advisories/GHSA-vxpw-j846-p89q      │
└─────────────────────┴────────────────────────────────────────────────────────┘
┌─────────────────────┬────────────────────────────────────────────────────────┐
│ high                │ undici vulnerable to cross-origin request routing via  │
│                     │ SOCKS5 proxy pool reuse                                │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Package             │ undici                                                 │
├─────────────────────┼────────────────────────────────────────────────────────┤
[...]

PR #4185, opened 2026-07-15
Update undici dependency version to 7.28.0

abdulahmad307:
As mentioned in an [existing issue](https://github.com/jsdom/jsdom/issues/4183) - undici has a security vulnerability between the following versions: `>= 7.23.0, < 7.28.0`. 

While `v7.28` _does_ get installed when running `npm install` - it would be best to lock the version to `7.28` and above in `package.json`

summary of vulnerability:
>When using Socks5ProxyAgent, undici reuses a single connection pool across different origins without verifying that the pool's origin matches the requested origin. All requests are dispatched through the pool connected to the first origin, regardless of the intended destination.

>This causes cross-origin request routing: credentials and request data intended for origin B are sent to origin A, responses from the wrong origin are trusted, and HTTPS requests may be silently downgraded to HTTP.

>Impacted users are applications that use Socks5ProxyAgent (directly or via setGlobalDispatcher) and make requests to more than one origin.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,7 @@
         "saxes": "^6.0.0",
         "symbol-tree": "^3.2.4",
         "tough-cookie": "^6.0.1",
-        "undici": "^7.25.0",
+        "undici": "^7.28.0",
         "w3c-xmlserializer": "^5.0.0",
         "webidl-conversions": "^8.0.1",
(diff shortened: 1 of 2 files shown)


domenic:
> While `v7.28` _does_ get installed when running `npm install`

This is fine, we don't need to bump the lockfile.

PR #4190, opened 2026-07-23
Fix WPT https-variant tests being served over http

colincasey:
Found a potential test runner bug while investigating a discrepancy with https://github.com/salesforce/tough-cookie/pull/534 which was initially attributed to the tough-cookie implementation. The actual issue seems to be that WPT variants tagged `wpt_flags=https` expect a secure context, but the jsdom runner served every variant over `http`. This PR attempts to route those variants to the `https` base instead.

Loading over `https` then broke the sync-XHR worker that `websockets/cookies/007.html` uses. It builds its own undici dispatcher and can't inherit the main thread's cert-skipping one, so requests to WPT's self-signed server failed on cert validation. I've marked this test as an expected failure and left the underlying limitation as follow-up.

There is also a change to pass `ignoreError` in the `api/cookies.js` secure cookie assertion to match how Document's cookie setter calls the store. This is a no-op today but it relates to https://github.com/salesforce/tough-cookie/pull/534 and will be needed once the secure check described in https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19#section-5.7-3.13.1 is reintroduced.

### DIFF
--- a/lib/jsdom/living/xhr/xhr-sync-worker.js
+++ b/lib/jsdom/living/xhr/xhr-sync-worker.js
@@ -17,6 +17,17 @@ const { parentPort } = require("node:worker_threads");
 const { JSDOM } = require("../../../..");
 const idlUtils = require("../../../generated/idl/utils");
 
+// The synchronous XHR worker runs in its own thread and cannot receive the
+// parent's undici dispatcher (an Agent is not structured-cloneable). It builds
+// its own XHR from `getGlobalDispatcher()`, so a parent that disabled TLS
+// verification (e.g. the WPT harness talking to a self-signed https server)
+// would otherwise see sync XHRs fail with a certificate error. Honor an opt-in
+// env var so the worker's global dispatcher matches; production is unaffected.
+if (process.env.JSDOM_SYNC_WORKER_INSECURE_TLS === "1") {
+  const { Agent, setGlobalDispatcher } = require("undici");
+  setGlobalDispatcher(new Agent({ connect: { rejectUnauthorized: false } }));
+}
+
 const dom = new JSDOM();
 const IDLE_TIMEOUT_MS = 5000;
 
--- a/lib/jsdom/living/xhr/xhr-sync-worker.js
+++ b/lib/jsdom/living/xhr/xhr-sync-worker.js
@@ -17,17 +17,6 @@ const { parentPort } = require("node:worker_threads");
 const { JSDOM } = require("../../../..");
 const idlUtils = require("../../../generated/idl/utils");
 
-// The synchronous XHR worker runs in its own thread and cannot receive the
-// parent's undici dispatcher (an Agent is not structured-cloneable). It builds
-// its own XHR from `getGlobalDispatcher()`, so a parent that disabled TLS
-// verification (e.g. the WPT harness talking to a self-signed https server)
-// would otherwise see sync XHRs fail with a certificate error. Honor an opt-in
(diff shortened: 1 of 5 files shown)


domenic:
Can you explain in more detail what problem this is solving? I don't see much point in running these tests with hacks to get them to run, since then they're not really testing what they should be testing.

colincasey:
Sure thing. The two WPT tests related to this PR are:
* https://github.com/web-platform-tests/wpt/blob/master/websockets/cookies/006.html - tests if a Secure cookie can cross into the WebSocket handshake based on transport security (wss=yes, ws=no)
* https://github.com/web-platform-tests/wpt/blob/master/websockets/cookies/007.html - tests if a handshake-response cookie is committed only after handshake completes

Test `006` is the primary issue. When the WPT variant `wss&wpt_flags=https` is executed by the jsdom runner it ends up running with the following setup:
* page origin = http://web-platform.test:PORT - this is where the `document.cookie` is set and the origin is **not** trustworthy
* websocket origin = wss://web-platform.test:PORT - this is where the handshake retrieves the cookie and the origin is trustworthy

This is an incorrect setup for this test as the page origin used by jsdom should be `https`. This is corrected by the change [here](https://github.com/colincasey/jsdom/blob/4626edb5c590da9c9ba7135edcc2c5fd50e8cc70/test/web-platform-tests/run-wpts.js#L25-L27). This change also fixes the set-time check for secure cookies that https://github.com/salesforce/tough-cookie/pull/534 reverted and which i'd like to reintroduce as this is a security concern. The only reason it passes today under the incorrect setup is because tough-cookie only performs the secure cookies check at retrieval-time.

Correcting the above issue raised a new problem with test `007`. For the sync-XHR worker, it builds it's own agent that doesn't inherit from the test runner's cert-skipping one. This is why the `JSDOM_SYNC_WORKER_INSECURE_TLS ` env var was introduced. An alternative to the hacky env var would be to add a failure expectation to `to-run.yaml‎`. E.g.; 
```yaml
cookies/007.html?wss&wpt_flags=https: [fail, sync XHR to self-signed https server fails cert validation in the worker]
```
But I didn't want to start just disabling tests without attempting to find a fix first 😅

domenic:
Sorry, I don't really feel like that answered my question. You just gave more detail on what you did. You did not explain why you did it.

I feel like such an answer should not take more than a sentence or two, if that helps constrain.

colincasey:
@domenic I misunderstood. My apologies.

I want to reintroduce a tough-cookie check that stops a secure cookie from being set in an insecure context and that check trips `websockets/cookies/006.html` from the WPT test suite, but only because the jsdom runner serves that test's "secure" variant over `http` instead of `https`. This PR fixes that runner misconfiguration so the test runs in the context it's written for and the secure cookie check can pass.

I've also reverted the hacky env var stuff in [7def5ae](https://github.com/jsdom/jsdom/pull/4190/commits/7def5ae6e98fe4454317e818c02c0a3eb33b29a8) so the PR is scoped to testing code only.

domenic:
Thanks so much, that helps a lot!

I opened https://github.com/jsdom/jsdom/pull/4199 as an alternative that also handles `.https` filename flags. Would you be able to take a look and tell me what you think?

colincasey:
@domenic #4199 looks good. It omits the `test/api/cookies.js { ignoreError: true }` change, which is actually cleaner as that was pre-emptive on my end. I can open up a new issue for that later if necessary.

PR #4191, opened 2026-07-24
Optimize WPT test file matching and filtering

asamuzaK:
Improve performance of web platform test execution by:

expectations-utils.js
- Separating exact path matches from glob pattern matches in expectation resolution, prioritizing exact matches to avoid unnecessary glob operations
- Using a Set for O(1) exact path lookups instead of O(n) array operations in checkExpectations

run-wpt.js
- Pre-filtering test files by directory in run-wpts.js to avoid repeated startsWith checks in the test loop


domenic:
Thanks very much for noticing this regression. I think https://github.com/jsdom/jsdom/pull/4192 manages to be even faster, with a bit less complexity.

PR #4195, opened 2026-07-27
Bump @asamuzakjp/dom-selector

### LINKED ISSUE #4187: Spurious `invalid` event from `:invalid` and `:valid` CSS selectors
### Node.js version

24.18.0

### jsdom version

29.1.1

### Minimal reproduction case

```js
import { JSDOM } from 'jsdom';

const dom = new JSDOM(`<!doctype html><html><head><style>
  input:invalid { background-color: red; }
</style></head><body>
  <form>
    <input id="x" type="number" min="0" value="-100">
  </form>
</body></html>`);

const input = document.getElementById('x');

let invalidEvents = 0;
input.addEventListener('invalid', () => {
  invalidEvents++;
});

console.log(`input.validity.valid = ${input.validity.valid} (expected false)`);

// Trigger 1: query the :invalid selector (nwsapi directly).
document.querySelectorAll('input:invalid');
console.log(`after querySelectorAll('input:invalid'): invalid events = ${invalidEvents}`);

// Trigger 2: compute styles, which evaluates the stylesheet's :invalid rule.
dom.window.getComputedStyle(input).backgroundColor;
console.log(`after getComputedStyle():                 invalid events = ${invalidEvents}`);

const buggy = invalidEvents > 0;
console.log(
  buggy
    ? `RESULT: BUG PRESENT — :invalid matching dispatched ${invalidEvents} spurious invalid event(s)`
    : `RESULT: OK — :invalid matching was side-effect-free`
);
```

https://codepen.io/joshkel/pen/RNKJLNo

```
input.validity.valid = false (expected false)
browser.html:21 after querySelectorAll('input:invalid'): invalid events = 0
browser.html:25 after getComputedStyle():                 invalid events = 0
browser.html:28 RESULT: OK — :invalid matching was side-effect-free
```

### What is the problem?

The implementation for selecting on the `:invalid` and `:valid` pseudo-classes calls `element.checkValidity()` to determine the patch. `checkValidity()` is not a pure query — [per the HTML standard it *fires* an `invalid` event](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-checkvalidity) at every invalid control when it returns `false`. As a result, merely matching an `:invalid` selector mutates the page by dispatching `invalid` events.

Selector matching is required to be side-effect-free. `:valid`/`:invalid` match on an element's validity state, which is observable without side effects via the `validity` getter — no method call, no events.

The bug appears to be around https://github.com/dperini/nwsapi/blob/c14d10c200dacbfe05125d5a1811f000f41f7083/src/nwsapi.js#L1369. If it would be better to report at https://github.com/asamuzaK/domSelector or https://github.com/dperini/nwsapi, I'm happy to do so.

Spec references:

- `:valid`/`:invalid` reflect validity *state* — https://html.spec.whatwg.org/multipage/semantics-other.html#selector-valid
- `checkValidity()` fires the `invalid` event — https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-checkvalidity

### LINKED ISSUE #4188: `querySelector('#some-id')` can fail when adding nodes to different documents
### Node.js version

v24.15.0

### jsdom version

29.1.1

### Minimal reproduction case

```js
const {JSDOM} = require("jsdom");
const {window} = new JSDOM();
const source = `<html><div id="test1"><div id="test2"></div></div></html>`;
const doc = new window.DOMParser().parseFromString(source, 'text/html');
const root = doc.documentElement;
const test1 = root.querySelector('#test1');
const test1Clone = test1.cloneNode(true);
console.log(
  'test1',
  test1Clone.querySelector('#test2'),
  test1Clone.querySelector('[id=test2]'),
  test1Clone.querySelectorAll('#test2')[0],
);
root.appendChild(test1Clone);
console.log(
  'test2',
  test1Clone.querySelector('#test2'),
  test1Clone.querySelector('[id=test2]'),
  test1Clone.querySelectorAll('#test2')[0],
);

```

All `querySelector` calls should return a non-null value.

### What is the problem?

`test1Clone.querySelector('#test2')` does not work and returns `null` in the second `console.log`. `test1Clone.querySelector('[id=test2]')` does work as expected in both logs. `querySelectorAll('#test2')[0]` also returns the expected value.

asamuzaK:
* Does not call internal method any more
  https://github.com/asamuzaK/domSelector/pull/296
* Can close #4187
  https://github.com/asamuzaK/domSelector/pull/284
* Can close #4188
  https://github.com/asamuzaK/domSelector/pull/293

### DIFF
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -848,8 +848,6 @@ class DocumentImpl extends NodeImpl {
     });
   }
 
-  // `@asamuzakjp/dom-selector` calls this method directly for universal selectors, bypassing the Web IDL wrapper
-  // that supplies the default value.
   createTreeWalker(root, whatToShow, filter = null) {
     return TreeWalker.createImpl(this._globalObject, [], { root, whatToShow, filter });
   }
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -848,7 +848,7 @@ class DocumentImpl extends NodeImpl {
     });
   }
 
-  createTreeWalker(root, whatToShow, filter = null) {
+  createTreeWalker(root, whatToShow, filter) {
     return TreeWalker.createImpl(this._globalObject, [], { root, whatToShow, filter });
   }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@asamuzakjp/css-color": "^6.0.5",
-        "@asamuzakjp/dom-selector": "^8.2.5",
+        "@asamuzakjp/dom-selector": "^8.3.0",
         "@bramus/specificity": "^2.4.2",
         "@csstools/css-syntax-patches-for-csstree": "^1.1.6",
         "@exodus/bytes": "^1.15.1",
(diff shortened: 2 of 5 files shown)


lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -848,8 +848,6 @@ class DocumentImpl extends NodeImpl {
     });
   }
 
-  // `@asamuzakjp/dom-selector` calls this method directly for universal selectors, bypassing the Web IDL wrapper
-  // that supplies the default value.
   createTreeWalker(root, whatToShow, filter = null) {
```
```suggestion
  createTreeWalker(root, whatToShow, filter) {
```

Since it is no longer necessary.

domenic (approved):
Nice, thank you. Will merge after the tweak, but we should not close those issues until we have regression tests for them.

asamuzaK:
Added tests.

PR #4202, opened 2026-07-31
Return the XML and XMLNS namespaces from locate a namespace

vojtisprime11:

The `Element` branch of [locate a namespace](https://dom.spec.whatwg.org/#locate-a-namespace) begins with two unconditional steps:

> 1. If *prefix* is "`xml`", then return the [XML namespace](https://infra.spec.whatwg.org/#xml-namespace).
> 2. If *prefix* is "`xmlns`", then return the [XMLNS namespace](https://infra.spec.whatwg.org/#xmlns-namespace).

`locateNamespace()` started at step 3, so those two prefixes were only answered when the element happened to carry a matching `xmlns:` declaration:

```js
const el = document.createElementNS('fooNamespace', 'prefix:elem');

el.lookupNamespaceURI('xml');
// before: null
// after:  'http://www.w3.org/XML/1998/namespace'

el.lookupNamespaceURI('xmlns');
// before: null
// after:  'http://www.w3.org/2000/xmlns/'
```

Both values match what Chrome and Firefox return.

The two steps live in the `Element` branch only, so `DocumentFragment`, `DocumentType` and a disconnected `Attr` keep returning null — `Node-lookupNamespaceURI.html` asserts exactly that, and a `Document` reaches the element branch through its document element, which is also what the test expects.

### Tests

This enables `dom/nodes/Node-lookupNamespaceURI.html`, previously listed as `[fail, Unknown]`. It has 4 failing subtests today ("Element should have XML namespace", "Element should have XMLNS namespace", "Element should have namespace with xmlns prefix", "Child element should have XMLNS namespace") and passes in full with this change.

`isDefaultNamespace()` and `lookupPrefix()` are unaffected: the former goes through `locate a namespace` with a null prefix, the latter through `locate a namespace prefix`.

I can't run the WPT server on this machine, so I ran the test through a static file server pointed at the vendored WPT checkout, reusing `run-single-wpt.js` and its `/resources/` interceptor. Before/after comparison on `dom/nodes/attributes.html`, `dom/nodes/Element-getAttributeNames.html` and `domparsing/XMLSerializer-serializeToString.html` shows no newly failing subtests. `test:api` (573 passing) and `test:to-port-to-wpts` (191 passing) are green, ESLint clean on the touched file.

### DIFF
--- a/lib/jsdom/living/node.js
+++ b/lib/jsdom/living/node.js
@@ -4,7 +4,7 @@ const NODE_TYPE = require("./node-type");
 
 const orderedSetParse = require("./helpers/ordered-set").parse;
 const { createElement } = require("./helpers/create-element");
-const { HTML_NS, XMLNS_NS } = require("./helpers/namespaces");
+const { HTML_NS, XML_NS, XMLNS_NS } = require("./helpers/namespaces");
 const { cloningSteps, domSymbolTree } = require("./helpers/internal-constants");
 const { asciiCaseInsensitiveMatch, asciiLowercase } = require("./helpers/strings");
 
@@ -270,6 +270,14 @@ exports.locateNamespacePrefix = (element, namespace) => {
 exports.locateNamespace = (node, prefix) => {
   switch (node.nodeType) {
     case NODE_TYPE.ELEMENT_NODE: {
+      if (prefix === "xml") {
+        return XML_NS;
+      }
+
+      if (prefix === "xmlns") {
+        return XMLNS_NS;
+      }
+
       if (node._namespaceURI !== null && node._prefix === prefix) {
         return node._namespaceURI;
       }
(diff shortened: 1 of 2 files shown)


vojtisprime11:
The `Tests (22.22.2)` job failed with `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`, at ~608s into `test:wpt`, right after the `encoding` directory finished and `fetch/api/headers` started:

```
✔ utf-32.html
fetch/api/headers
<--- Last few GCs --->
[...] Scavenge 4035.7 (4129.6) -> 4030.7 (4145.7) MB [...] allocation failure
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
ERROR: "test:wpt" exited with 134.
```

The other five test jobs plus `Tests with node-canvas` and `Lint` are green, and the change here is two early returns in `locateNamespace()` — it allocates nothing and does strictly less work than before, so it can't plausibly move the heap by 4 GB. Looks like the suite brushing against the heap limit on that particular Node version. A re-run of that job would confirm; I don't have permission to restart it myself.

Happy to rebase if that helps.

vojtisprime11:
Extra data point: #4203, which I opened after this one and which is a much larger change (a new named property getter on `Document`), just went green on all seven checks including `Tests (22.22.2)`. Same suite, same Node version, heavier change — so the heap limit hit here really does look ambient rather than related to these two early returns.

domenic (approved):
LGTM. I hope you are able to get a machine on which you can run jsdom tests soon.

PR #4204, opened 2026-08-01
Default BlobEvent timecode to NaN

domenic:
Default constructor-created `BlobEvent.timecode` values to `NaN`. This failure surfaced when [the BlobEvent constructor test was migrated into WPT](https://github.com/web-platform-tests/wpt/commit/90d9bd61d4ec48eeeaf9fea1c999bcdd68997304).

### DIFF
--- a/lib/jsdom/living/events/BlobEvent-impl.js
+++ b/lib/jsdom/living/events/BlobEvent-impl.js
@@ -6,7 +6,7 @@ class BlobEventImpl extends EventImpl {}
 // Cannot use the usual pattern since `data` is required.
 BlobEventImpl.defaultInit = {
   __proto__: null,
-  timecode: 0
+  timecode: NaN
 };
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


lib/jsdom/living/events/BlobEvent-impl.js
```diff
@@ -6,7 +6,7 @@ class BlobEventImpl extends EventImpl {}
 // Cannot use the usual pattern since `data` is required.
 BlobEventImpl.defaultInit = {
   __proto__: null,
-  timecode: 0
+  timecode: NaN
```
domenic, line 9: Maybe it can be omitted, and thus go through the `undefined` -> NaN path? The spec doesn't seem to have any explicit discussion of NaN.

domenic: Seems like not, hmm.

PR #4205, opened 2026-08-01
Replace lone surrogates before XML parsing

domenic:
Replace lone UTF-16 surrogates before XML parsing so they become U+FFFD instead of parser errors. This failure surfaced with [the upstream lone-surrogate parser coverage](https://github.com/web-platform-tests/wpt/commit/6308049a5f8d8f1e07d75d606a009af2415c5d81).


domenic (commented):
I wish this was clearer from the specs, but as https://html.spec.whatwg.org/multipage/xhtml.html#xml-parser says, there is actually no spec for XML parsing, so this is just a bunch of browsers copying each other. Will merge.

PR #4209, opened 2026-08-01
Return duplicate elements from XPath id()

domenic:
Return every matching element from XPath `id()`, including duplicate IDs, in document order. This failure surfaced with [the upstream XPath `id()` coverage](https://github.com/web-platform-tests/wpt/commit/3a6328d775f9ba515888f68a1ef34e95dc8db2fc).

### DIFF
--- a/lib/jsdom/level3/xpath.js
+++ b/lib/jsdom/level3/xpath.js
@@ -1351,13 +1351,12 @@ module.exports = core => {
         var a = idsString.split(/[ \t\r\n]+/g);
         ids = a;
       }
-      for (var i = 0; i < ids.length; ++i) {
-        var id = ids[i];
-        if (0 === id.length)
-          continue;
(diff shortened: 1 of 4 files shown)


lib/jsdom/level3/xpath.js
```diff
@@ -1351,13 +1351,12 @@ module.exports = core => {
...
         ids = a;
       }
-      for (var i = 0; i < ids.length; ++i) {
-        var id = ids[i];
-        if (0 === id.length)
-          continue;
-        var node = doc.getElementById(id);
-        if (null != node)
-          r.nodes.push(node);
+      var elements = doc.getElementsByTagName('*');
```
domenic: Is it just me or is this way less efficient? Maybe we shouldn't touch the xpath implementation in general.

domenic:
Not worth it.

PR #4210, opened 2026-08-01
Honor Blob charset in FileReader.readAsText

domenic:
Use the `Blob` MIME type’s charset when `FileReader.readAsText()` has no supported explicit encoding. This failure surfaced with [the upstream Blob-charset WPT](https://github.com/web-platform-tests/wpt/commit/2a91c2e71952bdceef8b2825ce3a0a0dc73aa588).

### DIFF
--- a/lib/jsdom/living/file-api/FileReader-impl.js
+++ b/lib/jsdom/living/file-api/FileReader-impl.js
@@ -41,7 +41,12 @@ class FileReaderImpl extends EventTargetImpl {
     this._readFile(file, "dataURL");
   }
   readAsText(file, encodingLabel) {
-    this._readFile(file, "text", labelToName(encodingLabel) || "UTF-8");
+    let encoding = labelToName(encodingLabel);
+    if (encoding === null) {
+      const type = MIMEType.parse(file.type);
+      encoding = type === null ? null : labelToName(type.parameters.get("charset"));
+    }
+    this._readFile(file, "text", encoding || "UTF-8");
   }
 
   abort() {
--- a/lib/jsdom/living/file-api/FileReader-impl.js
+++ b/lib/jsdom/living/file-api/FileReader-impl.js
@@ -40,13 +40,8 @@ class FileReaderImpl extends EventTargetImpl {
   readAsDataURL(file) {
     this._readFile(file, "dataURL");
   }
-  readAsText(file, encodingLabel) {
-    let encoding = labelToName(encodingLabel);
-    if (encoding === null) {
-      const type = MIMEType.parse(file.type);
-      encoding = type === null ? null : labelToName(type.parameters.get("charset"));
-    }
-    this._readFile(file, "text", encoding || "UTF-8");
+  readAsText(file, encodingName) {
+    this._readFile(file, "text", encodingName);
   }
 
   abort() {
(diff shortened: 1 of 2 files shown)


lib/jsdom/living/file-api/FileReader-impl.js
```diff
@@ -41,7 +41,12 @@ class FileReaderImpl extends EventTargetImpl {
     this._readFile(file, "dataURL");
   }
   readAsText(file, encodingLabel) {
-    this._readFile(file, "text", labelToName(encodingLabel) || "UTF-8");
+    let encoding = labelToName(encodingLabel);
+    if (encoding === null) {
+      const type = MIMEType.parse(file.type);
+      encoding = type === null ? null : labelToName(type.parameters.get("charset"));
```
domenic: Follow the spec more exactly here. Logic is correct but cosmetically this is off. https://w3c.github.io/FileAPI/#blob-package-data

lib/jsdom/living/file-api/FileReader-impl.js
```diff
@@ -40,8 +40,8 @@ class FileReaderImpl extends EventTargetImpl {
   readAsDataURL(file) {
     this._readFile(file, "dataURL");
   }
-  readAsText(file, encodingLabel) {
-    this._readFile(file, "text", labelToName(encodingLabel) || "UTF-8");
+  readAsText(file, encodingName) {
```
domenic: The actual spec argument name for the public API is `encoding`. `encodingName` is correct elsewhere per spec, but the spec is using it incorrectly.

Prepare a PR (do not send it without my approval) to w3c/FileAPI fixing all instances of encodingName -> encodingLabel. Comment here with the compare link when it's ready. Then update this PR to go back to encodingLabel.

lib/jsdom/living/file-api/FileReader-impl.js
```diff
@@ -132,7 +132,21 @@ class FileReaderImpl extends EventTargetImpl {
         break;
       }
       case "text": {
-        this.result = legacyHookDecode(file._bytes, encodingLabel);
+        let encoding = null;
+        if (encodingName !== undefined) {
+          encoding = labelToName(encodingName);
+        }
+        if (encoding === null) {
```
domenic, line 139: The spec has an "if mimeType is present", but mimeType is always passed. Similar to the above, prepare a PR (do not send) and link it to me, to remove the optionality and the unnecessary check.

domenic (commented):
Looking good with one code change plus a couple requests for upstream fixes.

lib/jsdom/living/file-api/FileReader-impl.js
domenic: Prepared the File API change without opening a PR: https://github.com/w3c/FileAPI/compare/main...codex/encoding-label

lib/jsdom/living/file-api/FileReader-impl.js
domenic: Added this to the same File API branch without opening a PR: https://github.com/w3c/FileAPI/compare/main...codex/encoding-label

domenic (commented):
LGTM, nice work. Upstream PR also sent.

PR #4212, opened 2026-08-01
Treat empty translate attributes as enabled

domenic:
Follow the `translate` attribute states explicitly so empty values enable translation, missing or invalid values inherit, and parentless elements default to translation enabled. This failure surfaced when [the upstream test was tightened to avoid a false positive](https://github.com/web-platform-tests/wpt/commit/2b0ab8e368b55bbdcbac614009327ef07af2a042).


lib/jsdom/living/nodes/HTMLElement-impl.js
```diff
@@ -46,7 +46,7 @@ class HTMLElementImpl extends ElementImpl {
     const translateAttr = this.getAttributeNS(null, "translate");
     const translateAttrString = asciiLowercase(translateAttr || "");
 
-    if (translateAttrString === "yes" || (translateAttr && translateAttrString === "")) {
+    if (translateAttrString === "yes" || (translateAttr !== null && translateAttrString === "")) {
```
domenic: This doesn't look like a correct change to me? If I'm reading https://html.spec.whatwg.org/multipage/dom.html#the-translate-attribute right, missing attribute or any invalid value should fall back to inheritance, not to true.

domenic (commented):
LGTM

PR #4213, opened 2026-08-01
Serialize CDATA sections as text in HTML

domenic:
Expose adopted XML `CDATASection` nodes to the HTML serializer as text so their content is preserved and escaped. This failure surfaced with [the upstream CDATA serialization WPT](https://github.com/web-platform-tests/wpt/commit/fe92d5e9ea136e2c97e75710af6189178faff5dd).


domenic (commented):
Nice, LGTM

PR #4214, opened 2026-08-01
Run post-connection scripts after batch insertion

domenic:
Implement the DOM insertion algorithm’s post-connection phase: complete tree insertion and mutation-record queuing, collect a static shadow-including node list, then run script and iframe connection side effects in tree order. This fixes sibling removal during `replaceWith()` plus related atomic-insertion, mutation-observer, removed-script, and dynamic-script-order cases. The regressions surfaced when WPT updated [sibling removal during `replaceWith()`](https://github.com/web-platform-tests/wpt/commit/39963a563a274169bab64bf12bf331781a2b9f45) and [the expected execution order after a script removes a later script](https://github.com/web-platform-tests/wpt/commit/2ba7ecc7fc312196891eeddb736501432f850e57).


domenic (commented):
This doesn't seem to follow https://dom.spec.whatwg.org/#concept-node-insert very well at all. I'm impressed with all the new test passes, but we shouldn't make this sort of change blindly. Can you change this to follow the spec more exactly?

domenic:

PR #4215, opened 2026-08-01
Handle detached TextDecoder input

domenic:
Treats input views detached during `TextDecoder.decode()` option conversion as empty input, matching Web IDL's byte-copy semantics; the failure surfaced because jsdom's WPT roll incorporated [web-platform-tests/wpt@cc6ea24](https://github.com/web-platform-tests/wpt/commit/cc6ea2479a37aea01fdeb26496e195898efeb98b), which added reentrant argument-conversion coverage, and [web-platform-tests/wpt@0dc50ad](https://github.com/web-platform-tests/wpt/commit/0dc50adc2e989134defaea66aacff377ab940543), which corrected the detached-buffer expectation to an empty byte sequence.

### DIFF
--- a/lib/jsdom/living/encoding/TextDecoder-impl.js
+++ b/lib/jsdom/living/encoding/TextDecoder-impl.js
@@ -8,6 +8,11 @@ exports.implementation = class TextDecoderImpl {
   }
 
   decode(input, options) {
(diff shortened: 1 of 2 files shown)


domenic (commented):
We should fix this holistically per the plan in https://github.com/jsdom/webidl-conversions/issues/29 instead, I think.

PR #4216, opened 2026-08-01
Ignore replaced children during insertion validation

domenic:
Aligns pre-insertion validity with the current DOM algorithm across pre-insertion, replacement, `replaceChildren()`, and range insertion. Each call now supplies a `Set` of children to exclude, replacement delegates to the shared validator with the replaced child excluded, and document validation examines the full preceding and following child order instead of only adjacent siblings.

This fixes valid document replacements that were rejected because the outgoing doctype or document element was still considered, as well as invalid element and doctype placements hidden by intervening comments. Focused to-upstream tests cover those exclusion and document-order cases, and the three corresponding upstream `Document.replaceChildren()` tests are now enabled.

The `Document.replaceChildren()` failures surfaced with [the upstream `replaceChildren()` coverage](https://github.com/web-platform-tests/wpt/commit/e3cad70090774540f1e8ee0137b3d1310a86dd8b).

### DIFF
--- a/lib/jsdom/living/nodes/Node-impl.js
+++ b/lib/jsdom/living/nodes/Node-impl.js
@@ -628,7 +628,7 @@ class NodeImpl extends EventTargetImpl {
   }
 
   // https://dom.spec.whatwg.org/#concept-node-ensure-pre-insertion-validity
-  _preInsertValidity(nodeImpl, childImpl, childrenToExclude = []) {
+  _preInsertValidity(nodeImpl, childImpl, childrenToExclude) {
     const { nodeType } = nodeImpl;
     const parentType = this.nodeType;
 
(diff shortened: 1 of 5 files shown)


lib/jsdom/living/nodes/Node-impl.js
```diff
@@ -628,7 +628,7 @@ class NodeImpl extends EventTargetImpl {
   }
 
   // https://dom.spec.whatwg.org/#concept-node-ensure-pre-insertion-validity
```
domenic, line 630: The spec has many more instances of "childrenToExclude" than just this one. Follow it more closely.

Ideally, add to-upstream test coverage for the delta between what you wrote here, and what the spec says.

Also, I suspect a Set would be better than an array.

domenic:

PR #4217, opened 2026-08-01
Invalidate XPath iterators after mutations

domenic:
Uses existing document mutation versions to invalidate iterator-style `XPathResult` objects and aligns `iterateNext()` wrong-type and invalid-state exceptions with platform behavior; the failure surfaced because jsdom's WPT roll incorporated [web-platform-tests/wpt@8f2316d](https://github.com/web-platform-tests/wpt/commit/8f2316d1e370116ba4124d08fd0b0dad928b9553), which added coverage for those cases.


domenic:
I guess this is self-contained enough that even though I don't really want to touch our xpath implementation, I'll merge it.

PR #4218, opened 2026-08-01
Preserve fragment owner document during insertion

domenic:
Align the node mutation algorithms with the DOM Standard: pre-insert now delegates directly to insertion, and insertion adopts each node from the extracted insertion list before adding it to the parent. As a result, inserting `template.content` adopts the moved children without changing the hosted `DocumentFragment`'s owner document.

The failure surfaced after [web-platform-tests/wpt@3f9d8f8](https://github.com/web-platform-tests/wpt/commit/3f9d8f8cd7134bc977db89cf3d48e4f8fe886142) added coverage for this distinction.

### DIFF
--- a/lib/jsdom/living/nodes/Node-impl.js
+++ b/lib/jsdom/living/nodes/Node-impl.js
@@ -770,7 +770,9 @@ class NodeImpl extends EventTargetImpl {
       referenceChildImpl = domSymbolTree.nextSibling(nodeImpl);
     }
 
-    this._ownerDocument._adoptNode(nodeImpl);
+    if (nodeImpl.nodeType !== NODE_TYPE.DOCUMENT_FRAGMENT_NODE) {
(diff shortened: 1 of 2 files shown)


lib/jsdom/living/nodes/Node-impl.js
```diff
@@ -770,7 +770,9 @@ class NodeImpl extends EventTargetImpl {
       referenceChildImpl = domSymbolTree.nextSibling(nodeImpl);
     }
 
-    this._ownerDocument._adoptNode(nodeImpl);
+    if (nodeImpl.nodeType !== NODE_TYPE.DOCUMENT_FRAGMENT_NODE) {
```
domenic: This seems to invent a step that doesn't exist in the spec. The spec change corresponding to this test was https://github.com/whatwg/dom/pull/1486/changes ; can you try implementing that instead?

domenic:

PR #4219, opened 2026-08-01
Complete documents created by DOMImplementation

domenic:
Default newly constructed documents to `readyState === "complete"`, as required for documents without a browsing context, while explicitly initializing Window documents to `loading` before parsing. This fixes documents returned by `createDocument()` and `createHTMLDocument()` and removes redundant per-caller complete overrides. The failures surfaced with [the upstream created-document readiness coverage](https://github.com/web-platform-tests/wpt/commit/a86c04befa0761e709a23c2f83e03fab50aa2329).


lib/jsdom/living/nodes/DOMImplementation-impl.js
```diff
@@ -40,7 +40,8 @@ class DOMImplementationImpl {
     const document = documents.createImpl(this._globalObject, {
       contentType,
       parsingMode: "xml",
-      encoding: "UTF-8"
+      encoding: "UTF-8",
+      readyState: "complete"
```
domenic: To follow the spec better, we should set the default to "complete", and only have certain callers reset it. See https://html.spec.whatwg.org/#current-document-readiness :

> For [Document](https://html.spec.whatwg.org/#document) objects created via the [create and initialize a Document object](https://html.spec.whatwg.org/#initialise-the-document-object) algorithm, this will be immediately reset to "loading" before any script can observe the value of [document.readyState](https://html.spec.whatwg.org/#dom-document-readystate). This default applies to other cases such as [initial about:blank](https://html.spec.whatwg.org/#is-initial-about:blank) [Document](https://html.spec.whatwg.org/#document)s or [Document](https://html.spec.whatwg.org/#document)s without a [browsing context](https://html.spec.whatwg.org/#concept-document-bc).

domenic (commented):
LGTM

PR #4221, opened 2026-08-01
Implement link fetchPriority reflection

domenic:
Implements `HTMLLinkElement.fetchPriority` reflection so [`attr-link-fetchpriority.html`](https://github.com/web-platform-tests/wpt/blob/e94af787e39d4161f01f0ef78d933f1103ee32b4/html/semantics/document-metadata/the-link-element/attr-link-fetchpriority.html), surfaced by [web-platform-tests/wpt@817b065e](https://github.com/web-platform-tests/wpt/commit/817b065e035c5624351472c9131c05eb94e2ea36) when the test moved into the rolled document-metadata directory, now passes.


domenic (commented):
I don't think we should implement just reflection, if we're not passing this along to our fetch implementation.

(Also, it's annoying to manually implement enumerated value reflection.)

Issue #3985, opened 2025-11-12
Performance regression in jsdom@27


Janpot:
### Node.js version

20.19.1

### jsdom version

27.1.0

### Minimal reproduction case

```json
// package.json
{
  "name": "jsdom-regression-repro",
  "version": "1.0.0",
  "type": "module",
  "scripts": {
    "start": "node compare.js"
  },
  "dependencies": {
    "jsdom-26": "npm:jsdom@26.0.0",
    "jsdom-27": "npm:jsdom@27.1.0"
  }
}
```

```tsx
// compare.js
/**
 * jsdom Performance Regression - Pure DOM Style Operations
 */

async function testJsdomVersion(jsdomPackage, versionLabel) {
  // Dynamically import the jsdom version
  const { JSDOM } = await import(jsdomPackage);

  // Initialize jsdom with empty HTML
  const dom = new JSDOM('<!DOCTYPE html><html><head></head><body></body></html>', {
    url: 'http://localhost',
    pretendToBeVisual: true,
  });
  const document = dom.window.document;

  // Test 1: innerHTML with simple elements (no styles)
  function testInnerHTMLSimple() {
    const container = document.createElement('div');
    document.body.appendChild(container);

    const start = performance.now();
    container.innerHTML = '<div>Hello</div>'.repeat(20);
    const duration = performance.now() - start;

    document.body.removeChild(container);
    return duration;
  }

  // Test 2: innerHTML with inline styles
  function testInnerHTMLStyled() {
    const container = document.createElement('div');
    document.body.appendChild(container);

    const start = performance.now();
    container.innerHTML =
      '<div style="color: blue; background-color: white; padding: 10px; border: 1px solid black;">Cell</div>'.repeat(
        20,
      );
    const duration = performance.now() - start;

    document.body.removeChild(container);
    return duration;
  }

  // Test 3: createElement + style property setting
  function testCreateElementWithStyles() {
    const container = document.createElement('div');
    document.body.appendChild(container);

    const start = performance.now();
    for (let i = 0; i < 20; i++) {
      const div = document.createElement('div');
      div.style.color = 'blue';
      div.style.backgroundColor = 'white';
      div.style.padding = '10px';
      div.style.border = '1px solid black';
      div.textContent = `Cell ${i}`;
      container.appendChild(div);
    }
    const duration = performance.now() - start;

    document.body.removeChild(container);
    return duration;
  }

  // Test 4: createElement + setAttribute('style')
[...]

Issue #3984, opened 2025-11-12
Performance regression in `getComputedStyle()`


Janpot:
### Node.js version

20.19.1

### jsdom version

27.1.0

### Minimal reproduction case

```json
// package.json
{
  "name": "jsdom-regression-repro",
  "version": "1.0.0",
  "type": "module",
  "scripts": {
    "start": "node compare.js"
  },
  "dependencies": {
    "jsdom-26": "npm:jsdom@26.0.0",
    "jsdom-27": "npm:jsdom@27.1.0"
  }
}
```

```html
<!-- jsdom-regression-repro.html -->
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8" />
    <title>jsdom Performance Regression Test</title>
  </head>
  <body>
    <div></div>
    <div>
      <div class="MuiDateRangeCalendar-root">
        <div class="MuiDateRangeCalendar-monthContainer">
          <div class="MuiPickersArrowSwitcher-root">
            <button
              class="MuiButtonBase-root MuiIconButton-root"
              tabindex="0"
              type="button"
              title="Previous month"
              aria-label="Previous month"
            >
              <svg
                class="MuiSvgIcon-root"
                focusable="false"
                aria-hidden="true"
                viewBox="0 0 24 24"
                data-testid="ArrowLeftIcon"
              >
                <path
                  d="M15.41 16.59L10.83 12l4.58-4.59L14 6l-6 6 6 6 1.41-1.41z"
                ></path>
              </svg>
            </button>
            <span class="MuiTypography-root" id="_r_0_-grid-0-label"
              >January 2019</span
            >
            <button
              class="MuiButtonBase-root MuiIconButton-root"
              tabindex="0"
              type="button"
              title="Next month"
              aria-label="Next month"
            >
              <svg
                class="MuiSvgIcon-root"
                focusable="false"
                aria-hidden="true"
                viewBox="0 0 24 24"
                data-testid="ArrowRightIcon"
              >
                <path
                  d="M8.59 16.59L13.17 12 8.59 7.41 10 6l6 6-6 6-1.41-1.41z"
                ></path>
              </svg>
            </button>
          </div>
          <div
            role="grid"
            aria-labelledby="_r_0_-grid-0-label"
            class="MuiDayCalendar-root"
          >
            <div role="row" class="MuiDayCalendar-header">
              <span
                class="MuiTypography-root"
                role="columnheader"
                aria-label="Sunday"
                >S</span
              >
[...]

PR #4223, opened 2026-08-04
Replace CSS implementations

cdoublev:
This pull request is intended to review the use of the CSS implementations from [`cdoublev/css`](https://github.com/cdoublev/css). It replaces almost all [`lib/jsdom/living/css/`](https://github.com/jsdom/jsdom/tree/main/lib/jsdom/living/css) and these dependencies:

  - `@asamuzakjp/css-color`
  - `@asamuzakjp/dom-selector`
  - `@bramus/specificity`
  - `@csstools/css-syntax-patches-for-csstree`
  - `css-tree`

I believe they are closer to the specifications and offer broader coverage of CSS features.

If this change is deemed worthwhile, the plan is to move `cdoublev/css` to an appropriate location, package it, and add it as a dependency.

---

The changes are detailed below. The first is the most significant and includes all the replacements mentioned above. The others can actually be moved to separate PRs.

<details>
  <summary><code>Window.getComputedStyle()</code> (resolving styles) - 0eaa70dc</summary><br>

  The current implementation returns a [`CSSStyleDeclaration`](https://drafts.csswg.org/cssom-1/#dom-window-getcomputedstyle) that is not live ([test](https://github.com/jsdom/jsdom/blob/main/test/web-platform-tests/to-upstream/css/cssom/getComputedStyle-live.html)). This liveness suggests resolving styles dynamically within `CSSStyleDeclaration` rather than providing them to it (as currently specified in CSSOM though, but see w3c/csswg-drafts#6144).

  These resolved styles are not cleared when required ([test](https://github.com/jsdom/jsdom/blob/main/test/web-platform-tests/to-upstream/css/css-cascade/revert-val-004.html), there are probably others). Caching `CSSStyleDeclaration` instances in a `WeakMap` that is replaced whenever resolved styles need to be cleared is not enough, and not very useful if they do not store them.

  If I am not mistaken, each declared value for the element is resolved and serialized before determining the "winning" one. This is inefficient: only the cascaded value (the one that wins the cascade) should be resolved and serialized.

  The `read-only` flag is not set on the `CSSStyleDeclaration`. Although this prevents [`.setProperty()`](https://drafts.csswg.org/cssom-1/#dom-cssstyledeclaration-setproperty) from throwing an error, using this method to store resolved values is also inefficient, since it will parse the resolved value.

  All selectors are also re-parsed when matched against the element, and eventually when calculating their specificity, something `@bramus/specificity` cannot do if they include `&` (bramus/specificity#20).

  Declarations ​​are not collected when nested within `@container`, `@scope`, `@supports`, and nested style rules.

  `cdoublev/css` fixes all of that. As for the downsides, it does no support:

  - resolving `<color-mix()>`, `<length>`, math functions
  - matching `:active`, `:defined`, `:hover`, `:focus`, `:focus-visible`, `:focus-within`, `:open`, `:state()`
  - custom element reactions

  This list includes only currently supported features, albeit often partially. I prefer not to add partial implementations of these features in `cdoublev/css`.

  Resolving `<color-mix()>` requires adding support for resolving all color values, and resolving relative color functions is non-trivial. Resolving `<length>` seems relatively simple, with the exception of container query units. Resolving math functions depends on resolving `<length>`.

  I have not found a solution for the custom element reactions that does not couple `cdoublev/css` to `jsdom`, as well as for matching many pseudo-classes. Solutions like subscribing to DOM events to track which element receives focus, are out of the question. I do not think the support for `:hover` is important, given that `jsdom` does not render the document.

  These would be the only regressions from this change.

  ---

  Below is the list of changes that, while not directly related, are necessary for resolving styles.

  **Expose the user and user agent style sheet**

  The user agent style sheet is now exposed by an internal property of `DocumentOrShadowRoot`. A user style sheet is created, and exposed in the same way. But resolved styles also depend on adopted (constructed) style sheets (`DocumentOrShadowRoot.adoptedStyleSheets`) and on transition and animation declarations. I also would like to find a solution to access all these styles that avoids coupling to `jsdom`. I do not yet know whether the best solution would be to pass (some of) them to the constructor of the computed `CSSStyleProperties`.

  `jsdom` should define its supported color schemes (see step 1 to [resolve a used color scheme](https://drafts.csswg.org/css-color-adjust-1/#used-color-scheme)), presumably `['light', 'dark']`, and it should allow its users to set their [preferred color scheme](https://drafts.csswg.org/css-color-adjust-1/#preferred-color-scheme) and the contents of their style sheet.

  **Fetching and removing a style sheet**

  Fetching `@import` style sheet depends on conditions declared in its prelude that are more appropriate to evaluate in `CSSImportRule`. It can fetch immediately without waiting for the parse of its parent style sheet to complete. It cannot access the `jsdom` resource queue (again, without coupling `cdoublev/css` to `jsdom`), so blocking DOM loading while the style sheet is being fetched required some plumbing that needs a careful review.

  `@import` is only allowed at the top-level and so is no longer searched in other rules when creating the style sheet of `<link>` or `<style>`.

  DOM and CSS should probably share all or part of the algorithms for fetching and processing `<link>` and `@import` style sheets, in order to avoid inconsistencies in character encoding detection, setting the internal properties to create the style sheet, etc. I do not know how to correctly set the "origin clean" flag. It is now set to `true` if the response and the document have the same origin. But they should have [CORS-same-origin](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#cors-same-origin) instead, which relies on `undici/fetch` correctly setting `response.type`. This is not the case at the moment (nodejs/undici#4740), unfortunately.

  There are also minor tweaks related to removing a style sheet, after replacing the implementation of `StyleSheetList` (`DocumentOrShadowRoot.styleSheets`).

  **`Element.style` attribute change steps**

  The implementation of the `Element.style` attribute change steps is delegated to `CSSStyleDeclaration`. This seems cleaner because they access the internal declaration list and use an internal entry point of the CSS parser. For his part, `jsdom` still checks that the name of the updated attribute is `style` (step 3) and that its value has changed (unspecified).

  However, I do not think the namespace of the updated attribute is being correctly checked, nor do I know how to check it: `element.setAttributeNS('http://www.w3.org/ns/unknown', 'style', 'color: red')` should not trigger an update of `element.style`. This requires follow-up.

  **Clearing the selector**

  After replacing the selector engine, the (new) selector cache needed to be cleared:

  - after modifying a `Node` contents (not only after modifying an attribute), for `:dir()` and `:lang()` (perhaps others)
  - after updating the `checked` and `indeterminate` properties (which are not reflected with an attribute), for `:checked` and `:indeterminate`
  - after setting `.setCustomValidity()`, for `:valid` and `:invalid`

  I do not know how related WPT tests did not failed before (ie. without clearing the cache).

  **Add a selector cache to `DocumentFragment`**

  It inherits from `ParentNode`, which defines `.querySelector()` and `.querySelectorAll()`, which depend on the existence of the selector cache.

  I do not know how no test failed before.

  ---

  **Benchmarks**

  **Note:** `npm run benchmark:compare` fails due to malformed JSON, which I have not investigated.

  <details>
  <summary><code>style/get-computed-style</code></summary><br>

    Before:

    ┌─────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
    │ Task name                           │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
    ├─────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
    │ flat (20 siblings)                  │ 33449 ± 0.83%    │ 32298 ± 283.00   │ 30696 ± 0.06%          │ 30962 ± 272            │ 29897   │
    │ deep (10-level nesting)             │ 1673.9 ± 0.67%   │ 1619.0 ± 34.00   │ 615098 ± 0.01%         │ 617665 ± 12852         │ 597392  │
    │ deep, all levels (10-level nesting) │ 16461 ± 1.92%    │ 15828 ± 144.00   │ 62970 ± 0.03%          │ 63179 ± 576            │ 60751   │
    └─────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘

    After:

    ┌─────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
    │ Task name                           │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
    ├─────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
    │ flat (20 siblings)                  │ 3978.3 ± 0.48%   │ 3863.0 ± 26.00   │ 256732 ± 0.02%         │ 258866 ± 1731          │ 251363  │
    │ deep (10-level nesting)             │ 227.92 ± 0.09%   │ 215.00 ± 3.00    │ 4552859 ± 0.01%        │ 4651163 ± 65819        │ 4387498 │
    │ deep, all levels (10-level nesting) │ 2082.7 ± 1.05%   │ 1959.0 ± 18.00   │ 505046 ± 0.02%         │ 510465 ± 4648          │ 480138  │
    └─────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘

  </details><details>
  <summary><code>style/get-computed-style-cold</code></summary><br>

    This benchmark hangs when using `cdoublev/css` and tasks are not executed concurrently (e459635e). This is concerning because the cause could be a memory leak, unless it is an issue in `tinybench`. Both tables result from running the benchmark concurrently.

    Before:

    ┌─────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
    │ Task name                           │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
    ├─────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
    │ flat (20 siblings)                  │ 1750787 ± 1.60%  │ 1664032 ± 13626  │ 584 ± 0.92%            │ 601 ± 5                │ 572     │
    │ deep (10-level nesting)             │ 111218 ± 2.29%   │ 102782 ± 733.50  │ 9597 ± 0.15%           │ 9729 ± 70              │ 8992    │
    │ deep, all levels (10-level nesting) │ 956587 ± 1.76%   │ 904269 ± 5698.0  │ 1080 ± 0.68%           │ 1106 ± 7               │ 1046    │
    └─────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘

    After:

    ┌─────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
    │ Task name                           │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
    ├─────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
    │ flat (20 siblings)                  │ 5179.3 ± 13.33%  │ 4514.5 ± 28.00   │ 210893 ± 4.29%         │ 221508 ± 1374          │ 64      │
    │ deep (10-level nesting)             │ 294.47 ± 33.13%  │ 232.00 ± 3.00    │ 4081061 ± 3.91%        │ 4310345 ± 56468        │ 64      │
    │ deep, all levels (10-level nesting) │ 2388.9 ± 5.74%   │ 2305.0 ± 15.00   │ 426875 ± 2.19%         │ 433839 ± 2821          │ 64      │
    └─────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘

  </details><details>
  <summary><code>style/get-computed-style-property-access</code></summary><br>

    This benchmark shows a huge regression. I think this stems primarily from the removal of the resolved style cache.

    Before:

    ┌──────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
    │ Task name            │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
    ├──────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
    │ few properties (4)   │ 7531.9 ± 0.77%   │ 7319.0 ± 61.00   │ 135955 ± 0.02%         │ 136631 ± 1148          │ 132769  │
    │ many properties (30) │ 33829 ± 1.04%    │ 32731 ± 134.00   │ 30420 ± 0.05%          │ 30552 ± 125            │ 29561   │
    └──────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘

    After:

    ┌──────────────────────┬──────────────────┬───────────────────┬────────────────────────┬────────────────────────┬─────────┐
    │ Task name            │ Latency avg (ns) │ Latency med (ns)  │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
    ├──────────────────────┼──────────────────┼───────────────────┼────────────────────────┼────────────────────────┼─────────┤
    │ few properties (4)   │ 1077894 ± 1.31%  │ 1008260 ± 33199   │ 953 ± 0.85%            │ 992 ± 33               │ 928     │
    │ many properties (30) │ 10404366 ± 0.59% │ 10426772 ± 176279 │ 96 ± 0.60%             │ 96 ± 2                 │ 97      │
    └──────────────────────┴──────────────────┴───────────────────┴────────────────────────┴────────────────────────┴─────────┘

  </details><details>
    <summary><code>selectors/complex-selectors</code></summary><br>

    I have concerns about the relevance of this test. It always use the same selector, for each interface. So adding a simple cache immediately showed a 10x improvement. I doubt that authors use the same selector multiple times in their style sheets.

    For `matches`, there is a huge performance regression when using `selectors` instead of `:matches(${selectors}$)` as the cache entry key. I did not investigated why.

    Before:

    ┌──────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
    │ Task name        │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
    ├──────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
    │ matches          │ 9116.2 ± 0.72%   │ 8901.0 ± 43.00   │ 111922 ± 0.02%         │ 112347 ± 545           │ 109695  │
    │ closest          │ 13522 ± 0.97%    │ 12946 ± 69.00    │ 76496 ± 0.05%          │ 77244 ± 410            │ 73952   │
    │ querySelector    │ 12467 ± 0.75%    │ 12110 ± 68.00    │ 82187 ± 0.03%          │ 82576 ± 466            │ 80209   │
    │ querySelectorAll │ 52123 ± 0.64%    │ 50974 ± 194.00   │ 19525 ± 0.06%          │ 19618 ± 75             │ 19186   │
    └──────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘

    After:

    ┌──────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
    │ Task name        │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
    ├──────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
    │ matches          │ 142.61 ± 1.03%   │ 133.00 ± 2.00    │ 7467736 ± 0.00%        │ 7518797 ± 111390       │ 7011912 │
    │ closest          │ 137.27 ± 0.94%   │ 128.00 ± 2.00    │ 7736049 ± 0.00%        │ 7812501 ± 120193       │ 7285096 │
    │ querySelector    │ 180.78 ± 0.36%   │ 178.00 ± 2.00    │ 5582542 ± 0.00%        │ 5617978 ± 63840        │ 5531735 │
    │ querySelectorAll │ 415.37 ± 7.67%   │ 376.00 ± 4.00    │ 2602159 ± 0.01%        │ 2659574 ± 28598        │ 2407471 │
    └──────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘

  </details><details>
  <summary><code>selectors/sizzle</code></summary><br>

    Before:

    ┌──────────────────┬───────────────────┬────────────────────┬────────────────────────┬────────────────────────┬─────────┐
    │ Task name        │ Latency avg (ns)  │ Latency med (ns)   │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
    ├──────────────────┼───────────────────┼────────────────────┼────────────────────────┼────────────────────────┼─────────┤
    │ querySelectorAll │ 147965108 ± 0.13% │ 147929435 ± 521075 │ 7 ± 0.13%              │ 7 ± 0                  │ 64      │
    │ querySelector    │ 191969193 ± 0.15% │ 191723559 ± 733036 │ 5 ± 0.15%              │ 5 ± 0                  │ 64      │
    └──────────────────┴───────────────────┴────────────────────┴────────────────────────┴────────────────────────┴─────────┘

    After:

    ┌──────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
    │ Task name        │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
    ├──────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
    │ querySelectorAll │ 5411810 ± 1.18%  │ 5527688 ± 394631 │ 186 ± 1.19%            │ 181 ± 12               │ 185     │
    │ querySelector    │ 5113124 ± 0.98%  │ 5327662 ± 193976 │ 197 ± 1.00%            │ 188 ± 7                │ 196     │
    └──────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘

  </details>

  ---

</details>

<details>
  <summary>Create a style sheet as specified - 6b4b0417</summary><br>

  The implementation of [*create a style sheet*](https://drafts.csswg.org/cssom/#create-a-css-style-sheet) now executes the implementation of [*add a CSS style sheet*](https://drafts.csswg.org/cssom/#add-a-css-style-sheet) as specified.

</details>

<details>
  <summary>Add a style sheet as specified - 0674ea7a</summary><br>

  The implementation of [*add a CSS style sheet*](https://drafts.csswg.org/cssom/#add-a-css-style-sheet) now sets the disabled and title properties as specified.

  However, this does not seem to be tested. And as reported in w3c/csswg-drafts#7573, the spec seems incorrect.

</details>

<details>
  <summary>Uncomment <code>@namespace</code> in user agent style sheet - 7fa5e1fa</summary><br>

  `@namespace` rules are now uncommented in the user agent style sheet, since it is supported by `cdoublev/css`.

</details>

<details>
  <summary><code>HTMLLinkElement.media attribute</code> change steps - 07446a3a</summary><br>

  The `media` attribute of `<link>` and `<style>` is now reflected on its `sheet.media.mediaText`, as [specified in CSSOM](https://drafts.csswg.org/cssom-1/#concept-css-style-sheet-media).

  WPT only has a [rendering test](https://github.com/web-platform-tests/wpt/blob/master/css/cssom/medialist-dynamic-001.html) for this requirement.

</details>

<details>
  <summary>Add support for resolving styles of a pseudo-element - f0e7d63b</summary><br>

  When a pseudo-element type is provided as the second argument, the current implementation returns the resolved styles of the element. Or if it starts with `::part` or `::slotted`, it throws an error rather than returning resolved styles with the empty string.

  This is now fixed.

</details>

<details>
  <summary>Run benchmark of cold computed style concurrently - e459635e</summary><br>

  As explained above, the `style/get-computed-style-cold` benchmark hangs without this change. This certainly requires follow-up.

</details>

---

While many WPT tests turned green, here are the ones that are now failing:

<details>
<summary><code>css/css-backgrounds/parsing/background-valid.html</code></summary><br>

  The test cases below do not expect [`background-position`](https://drafts.csswg.org/css-backgrounds-4/#background-position) and [`background-repeat`](https://drafts.csswg.org/css-backgrounds-4/#background-repeat) to be shorthands, and so do not expect their sub-properties to be set:

  - `"e.style['background'] = \"none\" should not set unrelated longhands"`
  - `"e.style['background'] = \"url(\\\"https://example.com/\\\") 1px 2px / 3px 4px space round local padding-box content-box, rgb(5, 6, 7) url(\\\"https://example.com/\\\") 1px 2px / 3px 4px space round local padding-box content-box\"`
  - `"e.style['background'] = \"border-box content-box\" should not set unrelated longhands"`
  - `"e.style['background'] = \"border-area\" should not set unrelated longhands"`
  - `"e.style['background'] = \"border-area text\" should not set unrelated longhands"`

  The test cases below expect `background-origin` to be set to `border-box` when `background-clip` is set to `border-area` or `border-area text`, which is not specified. The initial value of `background-origin` is `padding-box`:

  - `"e.style['background'] = \"border-area\" should set background-origin"`
  - `"e.style['background'] = \"border-area text\" should set background-origin"`

</details><details>
<summary><code>css/css-backgrounds/parsing/border-image-valid.html</code></summary><br>

  The failing test cases expect either all or none of the optional values of [`border-image`](https://drafts.csswg.org/css-borders-4/#border-image) to be omitted, but not just `<'border-image-width'>`. For example, `none 1 / 1 / 1px` and `1 / / 1px` are accepted, but not `1 / 1 / 1px`. This third option is definitely valid and my preference. I can be convinced to change that though.

</details><details>
<summary><code>css/css-box/parsing/margin-computed.html</code></summary><br>

  `cdoublev/css` does no support resolving relative `<length>`.

</details><details>
<summary><code>css/css-box/parsing/padding-computed.html</code></summary><br>

  `cdoublev/css` does no support resolving relative `<length>`.

</details><details>
<summary><code>css/css-cascade/at-scope-parsing.html</code></summary><br>

  The failing test cases have not been updated following the resolution in w3c/csswg-drafts#12959 to serialize all rules on a single line (except `@media`, keyframe and style rules). Fwiw, I think they should all serialize on a single line; backward-compatibility is not a problem here.

</details><details>
<summary><code>css/css-cascade/important-vs-inline-002.html</code></summary><br>

  `cdoublev/css` does no support resolving relative `<length>`.

</details><details>
<summary><code>css/css-cascade/parsing/layer.html</code></summary><br>

  Same as for `css/css-cascade/at-scope-parsing.html`.

</details><details>
<summary><code>css/css-cascade/parsing/supports-import-parsing.html</code></summary><br>

  This test expects [`<declaration>`](https://drafts.csswg.org/css-conditional-5/#typedef-declaration) to serialize with the original text input rather than the list of tokens.

  This is not specified and historically, only custom property values serialize with the original text input. Other [arbitrary values](https://drafts.csswg.org/css-syntax-3/#any-value) serialize with the list of tokens interleaved by a whitespace, following general [serialization](https://drafts.csswg.org/cssom-1/#serialize-a-css-value) [principles](https://drafts.csswg.org/css-values-4/#fuctional-serialization).

    > [...] joining space-separated tokens with a single space, and following each serialized comma, colon, semi-colon, or slash with a single space (U+0020 SPACE)

</details><details>
<summary><code>css/css-cascade/scope-cssom.html</code></summary><br>

  Same as for `css/css-cascade/at-scope-parsing.html`.

</details><details>
<summary><code>css/css-color/color-mix-currentcolor-visited-getcomputedstyle.html</code></summary><br>

  `cdoublev/css` does no support resolving `<color-mix()>`.

</details><details>
<summary><code>css/css-color/nested-color-mix-with-currentcolor.html</code></summary><br>

  `cdoublev/css` does no support resolving `<color-mix()>`.

</details><details>
<summary><code>css/css-color/relative-color-with-zoom.html</code></summary><br>

  `cdoublev/css` does no support resolving `<length>`.

</details><details>
<summary><code>css/css-nesting/invalid-inner-rules.html</code></summary><br>

  Same as for `css/css-cascade/at-scope-parsing.html`.

</details><details>
<summary><code>css/css-nesting/nested-declarations-cssom-whitespace.html</code></summary><br>

  Same as for `css/css-cascade/at-scope-parsing.html`.

</details><details>
<summary><code>css/css-nesting/serialize-group-rules-with-decls.html</code></summary><br>

  Same as for `css/css-cascade/at-scope-parsing.html`.

</details><details>
<summary><code>css/css-shadow/part/*</code></summary><br>

  `cdoublev/css` does not support resolving styles of elements in a shadow tree. This is relatively high on my TODO list though.

  I did not listed this in the list of missing support above, because I do not think it is currently supported in `jsdom` as well. I do not know why this test did not failed before though.

</details><details>
<summary><code>css/css-shadow/scope-pseudo-in-shadow.html</code></summary><br>

  Same as for `css/css-shadow/part/*`.

</details><details>
<summary><code>css/cssom/CSSGroupingRule-insertRule.html</code></summary><br>

  The test cases `"constraint violation"` and `"disallowed namespace rule"` expect a `HierarchyRequestError` to be thrown when trying to [insert](https://drafts.csswg.org/cssom-1/#insert-a-css-rule) `@import` or `@namespace` in a grouping rule like `@media`. `cdoublev/css` throws a `SyntaxError` instead.

    > 5. If `new rule` is a syntax error, throw a `SyntaxError` exception.
    > 6. If `new rule` cannot be inserted into list at **the zero-indexed position `index` due to constraints specified by CSS**, then throw a `HierarchyRequestError` exception.
    >
    > Note: For example, a CSS style sheet cannot contain an `@import` at-rule after a style rule.

  In my opinion, index-based constraints are subordinate to grammar-based constraints, which prevent adding `@import` and `@namespace` in a grouping rule, for which a `SyntaxError` is therefore more appropriate, like when you try to add `@top-left` in `@page` in the current version of Chrome.

  I opened web-platform-tests/wpt#56935 to make the case for this, but without success. I could change my opinion, given the right arguments.

</details><details>
<summary><code>css/cssom/CSSKeyframesRule.html</code></summary><br>

  The test case `"name, cssRules, appendRule, findRule, deleteRule"` expects [setting `CSSKeyframesRule.name`](https://drafts.csswg.org/css-animations-1/#dom-csskeyframesrule-name) with `'initial'` to be successful. `cdoublev/css` throws a `SyntaxError` instead. Nothing is currently specified (w3c/csswg-drafts#9364).

  `CSSKeyframesRule.name` reflects the [`<keyframe-name>`](https://drafts.csswg.org/css-animations-1/#typedef-keyframes-name) in the prelude of `@keyframes`. It is defined as a `<custom-ident>` excluding CSS-wide keywords like `initial`. I could be convinced to escape `initial` or to wrap it in a `<string>`, as resolved for `animation-name` (w3c/csswg-drafts/issues/2435), which references a `CSSKeyframesRule.name`.

</details><details>
<summary><code>css/cssom/cssimportrule.html</code></summary><br>

  The test case `"Values of CSSRule attributes"` expects `CSSImportRule.styleSheet` to be defined even if fetching failed. I opened w3c/csswg-drafts#13292 to confirm this is expected.

</details><details>
<summary><code>css/cssom/cssstyledeclaration-custom-properties.html</code></summary><br>

  This test expects the computed `CSSStyleProperties.length` to return the number of resolved property values for its owner element. It is equal to the addition of:

    - the number of standard longhand properties
    - the number of custom properties registered with `@property`
    - the number of unregistered custom properties declared on the element or an ancestor

  Resolving the first two is easy. Instead of looking for a way to resolve the third just to fix this test, I wonder if it would not be simpler to create a flat tree with resolved styles while loading the document, the way browsers do. Perhaps later. ^^

</details><details>
<summary><code>css/cssom/escape.html</code></summary><br>

  `"Various tests"` expects `CSS.escape()` to escape some code points (eg. U+0080) via [*serialize an identifier*](https://drafts.csswg.org/cssom-1/#serialize-an-identifier). This algorithm has not been updated following the change of the definition of [non-ASCII characters](https://drafts.csswg.org/css-syntax-3/#non-ascii-ident-code-point) (see this [comment](https://github.com/w3c/csswg-drafts/issues/7129#issuecomment-1148619956)). They do not need to be escaped.

  Note, however, that this change was intended to match custom property names, whose range of valid code points was subsequently relaxed (whatwg/dom#1079). Unfortunately, my [follow-up](https://github.com/w3c/csswg-drafts/issues/7129#issuecomment-4639799261) on this has gone unanswered.

</details><details>
<summary><code>css/cssom/cssimportrule-parent.html</code></summary><br>

  This test expects `CSSImportRule.styleSheet.parentStyleSheet` to be unset when removing `CSSImportRule`. This is unspecified, as reported in w3c/csswg-drafts#13291.

  I do not mind fix that without waiting for any feedback in this issue.

</details><details>
<summary><code>css/cssom/serialize-all-longhands.html</code></summary><br>

  Same as for `css/cssom/cssstyledeclaration-custom-properties.html`.

</details><details>
<summary><code>css/selectors/focus-in-focus-event-001.html</code></summary><br>

  `cdoublev/css` does not support matching `:focus-*`.

</details><details>
<summary><code>css/selectors/focus-in-focusin-event-001.html</code></summary><br>

  `cdoublev/css` does not support matching `:focus-*`.

</details><details>
<summary><code>css/selectors/focus-visible-014.html</code></summary><br>

  `cdoublev/css` does not support matching `:focus-*`.

</details><details>
<summary><code>css/selectors/invalidation/host-context-pseudo-class-in-has.html</code></summary><br>

  Same as for `css/css-shadow/part/*`.

</details><details>
<summary><code>css/selectors/invalidation/host-pseudo-class-in-has.html</code></summary><br>

  Same as for `css/css-shadow/part/*`.

</details><details>
<summary><code>css/selectors/is-where-basic.html</code></summary><br>

  The test case `":is() matches expected elements"` expects an empty `:is()` to be valid, which is not allowed by its grammar (w3c/csswg-drafts#11808).

</details><details>
<summary><code>css/selectors/old-tests/css3-modsel-22.xml</code></summary><br>

  This test expects `:lang(en-GB)` to match an element with `lang="en-gb-wa"` but [Selectors](https://drafts.csswg.org/selectors-4/#the-lang-pseudo) defines `en-gb-wa` as an invalid tag that do not match anything.

</details><details>
<summary><code>css/selectors/old-tests/css3-modsel-65.xml</code></summary><br>

  `cdoublev/css` does not support matching `:focus-*`.

</details><details>
<summary><code>css/selectors/selectors-attr-many-2.html</code></summary><br>

  This test expects an element to match a selector including 57 `<subclass-selector>`s, but [CSS Values & Units](https://drafts.csswg.org/css-values-4/#component-multipliers) requires UAs to *support at least 20 repetitions of the component*.

  Note that the user agent style sheet also includes selectors with more than 20 complex selectors.

</details><details>
<summary><code>custom-elements/attribute-changed-callback.html</code></summary><br>

  `cdoublev/css` does not support `CEReaction`.

</details><details>
<summary><code>dom/nodes/ParentNode-querySelector-escapes.html</code></summary><br>

  Same as for `css/cssom/escape.html`, basically.

  This test expects `#\u{2003}` and `#\u{A0}` to match an element with the corresponding `id`. But U+2003, U+00A0, U+12345, are no longer considered [non-ASCII code points](https://drafts.csswg.org/css-syntax-3/#non-ascii-ident-code-point), and are no longer accepted in a `<hash-token>` value.

</details><details>
<summary><code>html/semantics/forms/the-input-element/time-focus-dynamic-value-change.html</code></summary><br>

  `cdoublev/css` does not support matching `:focus-*`.

</details><details>
<summary><code>html/semantics/selectors/pseudo-classes/readwrite-readonly.html</code></summary><br>

  This test fails because `jsdom` does not implement `ElementContentEditable`. `cdoublev/css` depends on it for matching `:read-write` and `:read-only`.

</details><details>
<summary><code>html/semantics/selectors/pseudo-classes/required-optional-hidden.html</code></summary><br>

  This test expects `:optional` to match `<input type="hidden">`, but HTML requires the pseudo-class to only match elements to which the `required` attribute applies (whatwg/html#11273).

</details>

---

The same for WPT tests "to-upstream", but 4 of them have been updated to pass successfully:

<details>
<summary><code>css/css-backgrounds/background-shorthand.html</code></summary><br>

  Same as for `css/css-backgrounds/parsing/background-shorthand-serialization.html`.

</details><details>
<summary><code>css/css-borders/border-style-none.html</code></summary><br>

  This test expects declarations to serialize in a different order than what I believe is required by [*serialize a CSS declaration block*](https://drafts.csswg.org/cssom-1/#serialize-a-css-declaration-block) (which is rather complex).

  It sets `border-top` to `solid` (border style), `border-style` to `none`, and expects the declaration block to be `border-top-width: medium; border-top-color: currentcolor; border-style: none;`.

  Fundamentally, the serialization algorithm processes declarations in a loop and shorthands in a nested loop. The shorthands are sorted in the preferred order, which prioritize shorthand with the most sub-properties. Shorthand values are parsed and expanded to sub-property declarations sorted in canonical order as defined by the grammar of the shorthand. So `border-top: solid` expands to declarations for `border-*-width`, followed by those for `border-*-style`, then for `border-*-color`. And `border-style: none` expands to declarations for `border-*-style`, which overrides the previous ones.

  So it first tries with `border` but not all sub-properties are declared. Then with `border-top`, and serializes it. Then it processes the next remaining declaration, which is `border-right-style`. Etc. You will get `border-top: medium; border-right-style: none; border-bottom-style: none; border-left-style: none;`.

  I updated the test accordingly.

</details><details>
<summary><code>css/css-color/computed-color.html</code></summary><br>

  `cdoublev/css` does no support resolving `<color-mix()>`.

</details><details>
<summary><code>css/css-color/nested-color-mix-with-currentcolor.html</code></summary><br>

  `cdoublev/css` does no support resolving `<color-mix()>`.

</details><details>
<summary><code>css/css-display/display-initial.html</code></summary><br>

  The test case `"Revert-layer value of display"` expects an inline style declaration to have a different cascade layer than unlayered style declaration in a style sheet. This is currently unspecified (w3c/csswg-drafts#14091).

</details><details>
<summary><code>css/css-fonts/font-size-computed-keyword.html</code></summary><br>

  `cdoublev/css` does no support resolving relative `<length>`.

</details><details>
<summary><code>css/css-fonts/font-size-computed-no-child.html</code></summary><br>

  `cdoublev/css` does no support resolving relative `<length>`.

</details><details>
<summary><code>css/css-fonts/font-size-computed.html</code></summary><br>

  `cdoublev/css` does no support resolving relative `<length>`.

</details><details>
<summary><code>css/css-values/calc-parentheses.html</code></summary><br>

  `cdoublev/css` does no support resolving relative `<length>`.

</details><details>
<summary><code>css/css-values/font-relative-length.html</code></summary><br>

  `cdoublev/css` does no support resolving relative `<length>`.

</details><details>
<summary><code>css/cssom/CSSImportRule-insertRule-styleSheet-parentStyleSheet.html</code></summary><br>

  Similarly as `css/cssom/cssimportrule.html`, this test expects `CSSImportRule.styleSheet` to be defined synchronously, before its fetching completes. I asked a clarification in w3c/csswg-drafts#14099.

</details><details>
<summary><code>css/cssom/CSSImportRule-media-identity.html</code></summary><br>

  This test expects `CSSImportRule.media` and `CSSImportRule.styleSheet.media` to be the same object. I think the resolution in w3c/csswg-drafts#12063 aims at differentiating them, similarly as for `HTMLLinkElement.media` and `HTMLLinkElement.sheet.media`.

</details><details>
<summary><code>css/cssom/CSSKeyframesRule-appendRule-invalid.html</code></summary><br>

  This test expects `CSSKeyframesRule.appendRule()` to ignore its input if it is invalid. This is unspecified and inconsistent with `CSSStyleSheet.insertRule()` and `CSSGroupingRule.insertRule()`, which throw a `SyntaxError` (w3c/csswg-drafts#6972).

</details><details>
<summary><code>css/cssom/CSSKeyframesRule-appendRule-multiple.html</code></summary><br>

  Same as for `css/cssom/CSSKeyframesRule-appendRule-invalid.html`.

</details><details>
<summary><code>css/cssom/getComputedStyle-calc.html</code></summary><br>

  `cdoublev/css` does not support resolving math functions.

</details><details>
<summary><code>css/cssom/getComputedStyle-pseudo-shadow.html</code></summary><br>

  As noted earlier, CSSOM requires returning the empty string for the resolved value of `::part()`, rather than throwing an error.

  I updated the test accordingly.

</details><details>
<summary><code>css/cssom/style-background-position-longhands.html</code></summary><br>

  `cdoublev/css` does not support `background-position` as a shorthand (w3c/csswg-drafts#9253)

</details><details>
<summary><code>css/cssom/style-background-shorthand.html</code></summary><br>

  For the test case ```"`transparent` sets default longhand values"```: same as for `css/css-backgrounds/parsing/background-shorthand-serialization.html`.

  For the test cases `"Background shorthand sets longhand values"`, `"Update longhand sets background value with its updated value"`, `"Update longhand with var() sets empty background value"`: same as for `css/css-backgrounds/parsing/background-shorthand-serialization.html`.

  For the test case `"Background with var() sets empty longhand values"`: same as for `css/css-cascade/parsing/supports-import-parsing.html`. Serialization of arbitrary substitution containing value is unspecified (w3c/csswg-drafts#6484). Only custom property values need to be serialized with the text input, imo.

</details><details>
<summary><code>css/cssom/style-flex-shorthand.html</code></summary><br>

  The failing tests have not been updated following the resolution in w3c/csswg-drafts#5742.

</details><details>
<summary><code>css/cssom/style-set-cssText.html</code></summary><br>

  The test case `"sets only the valid properties for partially valid cssText"` expects a rule in a `style` attribute to be ignored. This follows a change in CSS Syntax that has no corresponding resolution and is not backward-compatible, as reported in w3c/csswg-drafts#11113.

</details><details>
<summary><code>css/cssom/style-set-property.html</code></summary><br>

  This test has not been updated following the resolution in w3c/csswg-drafts#9847 and the corresponding edits in CSS Variables:

    > The one exception to serialization being precisely as specified in the source file is that an empty custom property (such as `--foo:;`) must serialize with a single space as its value.

  I updated the test accordingly.

</details><details>
<summary><code>css/selectors/focus-within.html</code></summary><br>

  `cdoublev/css` does not support matching `:focus-*`.

</details><details>
<summary><code>html/editing/focus/focus-management/focus-blur.html</code></summary><br>

  `cdoublev/css` does not support matching `:focus-*`.

</details><details>
<summary><code>html/semantics/document-metadata/the-style-element/no-browsing-context.html</code></summary><br>

  This test expects a property value to resolve to its initial value for an element whose root has no navigable (formerly, a "browsing context"), but [CSSOM](https://drafts.csswg.org/cssom-1/#dom-window-getcomputedstyle) and [CSS Cascade](https://drafts.csswg.org/css-cascade-5/#value-stages) require it to be the empty string.

  I updated the test accordingly.

</details>

---

Other benchmarks:

<details>
<summary><code>api/constructor</code></summary><br>

  There is a moderate regression on all tasks. I suspect it comes from installing the `CSSStyleDeclaration` child classes with many properties as attributes. Relatedly, I noted that the WPT test `cssom-getPropertyValue-common-checks.html` pass successfully if subtests are run individually, otherwise it fails by exceeding the timeout.

  Before:

  
  ```
  ┌────────────────────────────────┬───────────────────┬───────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                      │ Latency avg (ns)  │ Latency med (ns)  │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├────────────────────────────────┼───────────────────┼───────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ new JSDOM() defaults           │ 2580311 ± 4.90%   │ 2116509 ± 104662  │ 431 ± 2.23%            │ 472 ± 24               │ 388     │
  │ new JSDOM() with many elements │ 10518460 ± 12.27% │ 10465089 ± 218464 │ 102 ± 3.26%            │ 96 ± 2                 │ 96      │
  └────────────────────────────────┴───────────────────┴───────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

  After:

  ```
  ┌────────────────────────────────┬──────────────────┬───────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                      │ Latency avg (ns) │ Latency med (ns)  │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├────────────────────────────────┼──────────────────┼───────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ new JSDOM() defaults           │ 3478032 ± 4.57%  │ 2872888 ± 141886  │ 314 ± 2.60%            │ 348 ± 18               │ 288     │
  │ new JSDOM() with many elements │ 14763280 ± 1.83% │ 15376509 ± 204561 │ 68 ± 1.94%             │ 65 ± 1                 │ 68      │
  └────────────────────────────────┴──────────────────┴───────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

</details><details>
<summary><code>css-supports/supports</code></summary><br>

  There is a huge regression on all tasks. The reason is that `cdoublev/css` does not cache parse results. I disagree with such optimization: it increases space complexity and no authors spam `CSS.supports()`.

  Before:

  ```
  ┌──────────────────────────────────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬──────────┐
  │ Task name                                                        │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples  │
  ├──────────────────────────────────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼──────────┤
  │ CSS.supports('color', 'red') [2 args]                            │ 4792.6 ± 0.43%   │ 4608.0 ± 66.00   │ 213617 ± 0.03%         │ 217014 ± 3110          │ 208655   │
  │ CSS.supports('unknown-prop', 'value') [2 args - false]           │ 80.55 ± 0.27%    │ 76.00 ± 2.00     │ 12889595 ± 0.00%       │ 13157895 ± 337382      │ 12413956 │
  │ CSS.supports('display: flex') [fast path]                        │ 5423.1 ± 9.60%   │ 4873.0 ± 51.00   │ 202016 ± 0.03%         │ 205212 ± 2170          │ 184397   │
  │ CSS.supports('--my-custom: red') [custom prop]                   │ 1260.6 ± 0.53%   │ 1231.0 ± 7.00    │ 806932 ± 0.01%         │ 812348 ± 4593          │ 793245   │
  │ CSS.supports('(display: grid) and (not (display: inline-grid))') │ 32031 ± 2.52%    │ 30164 ± 323.00   │ 32851 ± 0.06%          │ 33152 ± 357            │ 31220    │
  │ CSS.supports('color: ...') [cache miss simulation]               │ 23796 ± 1.27%    │ 22665 ± 363.00   │ 43668 ± 0.06%          │ 44121 ± 709            │ 42025    │
  │ Real-world: Modernizr-like startup batch (9 queries)             │ 61210 ± 3.92%    │ 53473 ± 599.50   │ 17786 ± 0.23%          │ 18701 ± 210            │ 16338    │
  │ Real-world: CSS-in-JS component list render (100 iterations)     │ 1511446 ± 1.12%  │ 1435684 ± 23481  │ 671 ± 0.77%            │ 697 ± 11               │ 662      │
  └──────────────────────────────────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴──────────┘
  ```

  After:

  ```
  ┌──────────────────────────────────────────────────────────────────┬──────────────────┬───────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                                                        │ Latency avg (ns) │ Latency med (ns)  │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├──────────────────────────────────────────────────────────────────┼──────────────────┼───────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ CSS.supports('color', 'red') [2 args]                            │ 754154 ± 0.68%   │ 721530 ± 7902.5   │ 1341 ± 0.50%           │ 1386 ± 15              │ 1326    │
  │ CSS.supports('unknown-prop', 'value') [2 args - false]           │ 469.96 ± 0.63%   │ 429.00 ± 16.00    │ 2255728 ± 0.01%        │ 2331002 ± 88850        │ 2127823 │
  │ CSS.supports('display: flex') [fast path]                        │ 279623 ± 2.74%   │ 265730 ± 1185.0   │ 3688 ± 0.31%           │ 3763 ± 17              │ 3577    │
  │ CSS.supports('--my-custom: red') [custom prop]                   │ 190479 ± 0.57%   │ 183539 ± 824.50   │ 5351 ± 0.25%           │ 5448 ± 25              │ 5250    │
  │ CSS.supports('(display: grid) and (not (display: inline-grid))') │ 748016 ± 0.72%   │ 714724 ± 3677.0   │ 1354 ± 0.52%           │ 1399 ± 7               │ 1337    │
  │ CSS.supports('color: ...') [cache miss simulation]               │ 1338461 ± 0.71%  │ 1282476 ± 11778   │ 753 ± 0.60%            │ 780 ± 7                │ 748     │
  │ Real-world: Modernizr-like startup batch (9 queries)             │ 3083710 ± 0.61%  │ 3002046 ± 77680   │ 325 ± 0.59%            │ 333 ± 9                │ 325     │
  │ Real-world: CSS-in-JS component list render (100 iterations)     │ 60917579 ± 0.10% │ 60853937 ± 108327 │ 16 ± 0.10%             │ 16 ± 0                 │ 64      │
  └──────────────────────────────────────────────────────────────────┴──────────────────┴───────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

</details><details>
<summary><code>dom/compare-document-position</code></summary><br>

  No significant change.

  Before:

  ```
  ┌────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name          │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ compare siblings   │ 178.78 ± 1.14%   │ 168.00 ± 2.00    │ 5903792 ± 0.00%        │ 5952381 ± 71715        │ 5593344 │
  │ compare descendant │ 20301 ± 1.07%    │ 19244 ± 351.00   │ 51222 ± 0.05%          │ 51964 ± 957            │ 49259   │
  │ compare ancestor   │ 22377 ± 0.85%    │ 22718 ± 240.00   │ 46360 ± 0.10%          │ 44018 ± 462            │ 44688   │
  └────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

  After:

  ```
  ┌────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name          │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ compare siblings   │ 189.95 ± 8.56%   │ 163.00 ± 2.00    │ 5996406 ± 0.01%        │ 6134969 ± 76211        │ 5264526 │
  │ compare descendant │ 22340 ± 0.58%    │ 22399 ± 379.00   │ 45560 ± 0.06%          │ 44645 ± 748            │ 44764   │
  │ compare ancestor   │ 21965 ± 0.52%    │ 22334 ± 313.00   │ 46391 ± 0.07%          │ 44775 ± 621            │ 45527   │
  └────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

</details><details>
<summary><code>dom/construction</code></summary><br>

  No significant change except for `DocumentFragment`. This stems from assigning it a selector cache.

  Before:

  ```
  ┌─────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                   │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├─────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ createElement               │ 982.11 ± 1.12%   │ 879.00 ± 36.00   │ 1109307 ± 0.02%        │ 1137656 ± 47178        │ 1018214 │
  │ createTextNode              │ 347.97 ± 0.77%   │ 321.00 ± 8.00    │ 3067998 ± 0.01%        │ 3115265 ± 79623        │ 2873841 │
  │ createComment               │ 390.02 ± 8.00%   │ 296.00 ± 7.00    │ 3193216 ± 0.02%        │ 3378378 ± 81829        │ 2563961 │
  │ createDocumentFragment      │ 319.00 ± 6.48%   │ 273.00 ± 6.00    │ 3547121 ± 0.01%        │ 3663004 ± 82315        │ 3134756 │
  │ createNodeIterator          │ 1459.8 ± 14.41%  │ 931.00 ± 140.00  │ 1098190 ± 0.05%        │ 1074114 ± 160454       │ 685040  │
  │ createEvent                 │ 856.89 ± 0.20%   │ 843.00 ± 5.00    │ 1178714 ± 0.01%        │ 1186240 ± 7078         │ 1167017 │
  │ createProcessingInstruction │ 408.16 ± 8.42%   │ 298.00 ± 8.00    │ 3119722 ± 0.02%        │ 3355705 ± 92571        │ 2537546 │
  └─────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

  After:

  ```
  ┌─────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                   │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├─────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ createElement               │ 937.41 ± 0.61%   │ 863.00 ± 36.00   │ 1129776 ± 0.02%        │ 1158749 ± 48981        │ 1066775 │
  │ createTextNode              │ 346.24 ± 2.88%   │ 315.00 ± 6.00    │ 3121997 ± 0.01%        │ 3174603 ± 61643        │ 2888199 │
  │ createComment               │ 386.17 ± 9.37%   │ 302.00 ± 7.00    │ 3144693 ± 0.02%        │ 3311258 ± 78572        │ 2589551 │
  │ createDocumentFragment      │ 49127 ± 7.22%    │ 46282 ± 1056.5   │ 21639 ± 0.07%          │ 21606 ± 491            │ 20356   │
  │ createNodeIterator          │ 1364.9 ± 12.62%  │ 925.00 ± 131.00  │ 1108674 ± 0.05%        │ 1081081 ± 151965       │ 739100  │
  │ createEvent                 │ 877.55 ± 1.79%   │ 842.00 ± 6.00    │ 1181500 ± 0.01%        │ 1187648 ± 8524         │ 1139531 │
  │ createProcessingInstruction │ 319.70 ± 0.72%   │ 295.00 ± 9.00    │ 3344738 ± 0.01%        │ 3389831 ± 100357       │ 3127891 │
  └─────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

</details><details>
<summary><code>dom/get-element-by-id</code></summary><br>

  There is no significant change except for `bulk insert 200 + getElementById`, which I did not investigated.

  Before:

  ```
  ┌───────────────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬──────────┐
  │ Task name                                     │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples  │
  ├───────────────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼──────────┤
  │ getElementById: unique ID among 1000          │ 76.62 ± 0.19%    │ 75.00 ± 2.00     │ 13264887 ± 0.00%       │ 13333334 ± 365296      │ 13050740 │
  │ getElementById: duplicate ID (10x) among 1000 │ 86.74 ± 16.33%   │ 72.00 ± 2.00     │ 13691088 ± 0.01%       │ 13888890 ± 396824      │ 11529346 │
  │ getElementById after mutation: 1000 elements  │ 26975 ± 7.77%    │ 25204 ± 1797.0   │ 39919 ± 0.09%          │ 39676 ± 2809           │ 37072    │
  │ bulk insert 200 + getElementById              │ 1106679 ± 16.83% │ 711660 ± 20244   │ 1239 ± 1.86%           │ 1405 ± 41              │ 904      │
  └───────────────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴──────────┘
  ```

  After:

  ```
  ┌───────────────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬──────────┐
  │ Task name                                     │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples  │
  ├───────────────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼──────────┤
  │ getElementById: unique ID among 1000          │ 73.62 ± 0.02%    │ 73.00 ± 2.00     │ 13664659 ± 0.00%       │ 13698631 ± 365297      │ 13582980 │
  │ getElementById: duplicate ID (10x) among 1000 │ 84.94 ± 1.28%    │ 84.00 ± 2.00     │ 11923278 ± 0.00%       │ 11904762 ± 290359      │ 11772450 │
  │ getElementById after mutation: 1000 elements  │ 23677 ± 0.82%    │ 23349 ± 1765.0   │ 43087 ± 0.09%          │ 42828 ± 3214           │ 42235    │
  │ bulk insert 200 + getElementById              │ 1926907 ± 2.55%  │ 1762761 ± 24220  │ 542 ± 1.27%            │ 567 ± 8                │ 519      │
  └───────────────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴──────────┘
  ```

</details><details>
<summary><code>dom/inner-html</code></summary><br>

  There is a regression for `tables`, which I did not investigated.

  Before:

  ```
  ┌───────────┬───────────────────┬──────────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name │ Latency avg (ns)  │ Latency med (ns)     │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├───────────┼───────────────────┼──────────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ tables    │ 894598982 ± 1.13% │ 879016449 ± 20130297 │ 1 ± 1.10%              │ 1 ± 0                  │ 64      │
  │ text      │ 22444768 ± 2.40%  │ 21544312 ± 758763    │ 45 ± 2.16%             │ 46 ± 2                 │ 64      │
  └───────────┴───────────────────┴──────────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

  After:

  ```
  ┌───────────┬────────────────────┬───────────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name │ Latency avg (ns)   │ Latency med (ns)      │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├───────────┼────────────────────┼───────────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ tables    │ 1345720273 ± 1.20% │ 1352249633 ± 52542325 │ 1 ± 1.20%              │ 1 ± 0                  │ 64      │
  │ text      │ 24944171 ± 12.62%  │ 21364593 ± 1043784    │ 44 ± 4.68%             │ 47 ± 2                 │ 64      │
  └───────────┴────────────────────┴───────────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

</details><details>
<summary><code>dom/named-properties</code></summary><br>

  There is a moderate regression, which I did not investigated.

  Before:

  ```
  ┌─────────────────────────────────────────────────────┬──────────────────┬───────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                                           │ Latency avg (ns) │ Latency med (ns)  │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├─────────────────────────────────────────────────────┼──────────────────┼───────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ setAttribute(): Remove a named property from window │ 6593643 ± 15.10% │ 6229591 ± 1438852 │ 178 ± 4.61%            │ 161 ± 34               │ 152     │
  └─────────────────────────────────────────────────────┴──────────────────┴───────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

  After:

  ```
  ┌─────────────────────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                                           │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├─────────────────────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ setAttribute(): Remove a named property from window │ 8283399 ± 1.89%  │ 7651091 ± 194571 │ 122 ± 1.84%            │ 131 ± 3                │ 121     │
  └─────────────────────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

</details><details>
<summary><code>dom/ranges</code></summary><br>

  No significant change except for `cloneContents`, which I did not investigated.

  Before:

  ```
  ┌────────────────────────────────────────────────┬──────────────────┬───────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                                      │ Latency avg (ns) │ Latency med (ns)  │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├────────────────────────────────────────────────┼──────────────────┼───────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ deleteContents: small span, small tree         │ 361365 ± 2.29%   │ 344510 ± 167002   │ 5260 ± 4.55%           │ 2903 ± 1198            │ 2768    │
  │ deleteContents: small span, large tree         │ 354377 ± 2.24%   │ 348065 ± 147396   │ 5239 ± 4.47%           │ 2873 ± 1069            │ 2823    │
  │ deleteContents: large span, large tree         │ 3315265 ± 4.42%  │ 3316973 ± 1020908 │ 368 ± 5.81%            │ 301 ± 80               │ 303     │
  │ extractContents: small span, large tree        │ 361792 ± 2.29%   │ 348717 ± 165280   │ 5111 ± 4.33%           │ 2868 ± 1145            │ 2765    │
  │ cloneContents: small span, large tree          │ 46873 ± 1.26%    │ 43935 ± 543.00    │ 22597 ± 0.09%          │ 22761 ± 281            │ 21335   │
  │ deleteContents: selectNodeContents             │ 1829300 ± 4.16%  │ 1741324 ± 685840  │ 749 ± 5.51%            │ 574 ± 196              │ 547     │
  │ deleteContents: partially contained boundaries │ 1264950 ± 3.39%  │ 1209121 ± 454947  │ 1052 ± 4.23%           │ 827 ± 278              │ 792     │
  └────────────────────────────────────────────────┴──────────────────┴───────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

  After:

  ```
  ┌────────────────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                                      │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├────────────────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ deleteContents: small span, small tree         │ 359283 ± 2.40%   │ 334678 ± 155020  │ 4813 ± 3.70%           │ 2988 ± 1195            │ 2784    │
  │ deleteContents: small span, large tree         │ 365583 ± 2.42%   │ 355704 ± 148400  │ 4690 ± 3.78%           │ 2811 ± 1023            │ 2737    │
  │ deleteContents: large span, large tree         │ 3562043 ± 3.73%  │ 3488799 ± 820121 │ 315 ± 4.25%            │ 287 ± 63               │ 282     │
  │ extractContents: small span, large tree        │ 396960 ± 2.28%   │ 374114 ± 141824  │ 3434 ± 2.36%           │ 2673 ± 913             │ 2520    │
  │ cloneContents: small span, large tree          │ 138845 ± 1.33%   │ 131065 ± 1264.0  │ 7502 ± 0.20%           │ 7630 ± 74              │ 7203    │
  │ deleteContents: selectNodeContents             │ 2010954 ± 3.67%  │ 2053481 ± 634736 │ 614 ± 4.48%            │ 487 ± 139              │ 498     │
  │ deleteContents: partially contained boundaries │ 1343334 ± 3.12%  │ 1278921 ± 406797 │ 917 ± 3.59%            │ 782 ± 225              │ 745     │
  └────────────────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

</details><details>
<summary><code>dom/root-node</code></summary><br>

  No significant change

  Before:

  ```
  ┌───────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬──────────┐
  │ Task name                             │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples  │
  ├───────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼──────────┤
  │ getRootNode: depth 200                │ 90.55 ± 1.60%    │ 71.00 ± 1.00     │ 13683166 ± 0.01%       │ 14084509 ± 201209      │ 11043383 │
  │ dispatchEvent (bubbling): depth 100   │ 148995 ± 1.39%   │ 147436 ± 535.00  │ 6763 ± 0.04%           │ 6783 ± 25              │ 6712     │
  │ isConnected: depth 200                │ 37.91 ± 0.07%    │ 37.00 ± 1.00     │ 26716812 ± 0.00%       │ 27027013 ± 750756      │ 26381043 │
  │ appendChild+removeChild: at depth 200 │ 38598 ± 1.49%    │ 36464 ± 579.00   │ 27074 ± 0.08%          │ 27424 ± 438            │ 25908    │
  └───────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴──────────┘
  ```

  After:

  ```
  ┌───────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬──────────┐
  │ Task name                             │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples  │
  ├───────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼──────────┤
  │ getRootNode: depth 200                │ 88.63 ± 1.95%    │ 73.00 ± 1.00     │ 13555520 ± 0.00%       │ 13698630 ± 190260      │ 11282402 │
  │ dispatchEvent (bubbling): depth 100   │ 158902 ± 12.65%  │ 144452 ± 1445.5  │ 6905 ± 0.06%           │ 6923 ± 69              │ 6294     │
  │ isConnected: depth 200                │ 35.31 ± 0.08%    │ 35.00 ± 1.00     │ 28772424 ± 0.00%       │ 28571449 ± 840319      │ 28318121 │
  │ appendChild+removeChild: at depth 200 │ 40133 ± 6.59%    │ 37027 ± 171.00   │ 26822 ± 0.07%          │ 27007 ± 125            │ 24918    │
  └───────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴──────────┘
  ```

</details><details>
<summary><code>dom/tree-modification</code></summary><br>

  There is a regression on all tasks, which I did not investigated.

  Before:

  ```
  ┌─────────────────────────────┬──────────────────┬────────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                   │ Latency avg (ns) │ Latency med (ns)   │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├─────────────────────────────┼──────────────────┼────────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ appendChild: no siblings    │ 2700.1 ± 0.81%   │ 2543.0 ± 72.00     │ 387387 ± 0.02%         │ 393236 ± 11131         │ 370359  │
  │ appendChild: many siblings  │ 1913351 ± 2.35%  │ 1678900 ± 16299    │ 550 ± 1.58%            │ 596 ± 6                │ 523     │
  │ appendChild: many parents   │ 19789062 ± 3.03% │ 18455007 ± 1214204 │ 51 ± 2.88%             │ 54 ± 4                 │ 64      │
  │ insertBefore: many siblings │ 2557860 ± 42.42% │ 1718860 ± 74413    │ 526 ± 1.90%            │ 582 ± 26               │ 391     │
  │ removeChild: no siblings    │ 3420.0 ± 0.73%   │ 3265.0 ± 64.00     │ 304408 ± 0.02%         │ 306279 ± 6026          │ 292398  │
  │ removeChild: many siblings  │ 2624730 ± 2.26%  │ 2302434 ± 27754    │ 396 ± 1.74%            │ 434 ± 5                │ 381     │
  │ removeChild: many parents   │ 1402324 ± 1.36%  │ 1313145 ± 9729.0   │ 730 ± 0.91%            │ 762 ± 6                │ 714     │
  └─────────────────────────────┴──────────────────┴────────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

  After:

  ```
  ┌─────────────────────────────┬──────────────────┬───────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                   │ Latency avg (ns) │ Latency med (ns)  │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├─────────────────────────────┼──────────────────┼───────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ appendChild: no siblings    │ 3906.6 ± 0.59%   │ 3786.0 ± 75.00    │ 262778 ± 0.02%         │ 264131 ± 5265          │ 255977  │
  │ appendChild: many siblings  │ 3003330 ± 1.12%  │ 2904191 ± 30782   │ 335 ± 0.81%            │ 344 ± 4                │ 333     │
  │ appendChild: many parents   │ 17554642 ± 1.31% │ 17525000 ± 179649 │ 57 ± 1.15%             │ 57 ± 1                 │ 64      │
  │ insertBefore: many siblings │ 3105421 ± 0.95%  │ 2989348 ± 41027   │ 324 ± 0.82%            │ 335 ± 5                │ 323     │
  │ removeChild: no siblings    │ 6276.4 ± 0.72%   │ 5937.0 ± 105.00   │ 164880 ± 0.04%         │ 168435 ± 2982          │ 159328  │
  │ removeChild: many siblings  │ 5384867 ± 1.62%  │ 5082618 ± 31878   │ 188 ± 1.33%            │ 197 ± 1                │ 186     │
  │ removeChild: many parents   │ 1624397 ± 1.19%  │ 1538801 ± 12104   │ 626 ± 0.86%            │ 650 ± 5                │ 616     │
  └─────────────────────────────┴──────────────────┴───────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

</details><details>
<summary><code>style/style-property</code></summary><br>

  There is a huge regression on all tasks, which I did not investigated.

  Before:

  ```
  ┌───────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                             │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├───────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ innerHTML: simple divs (no styles)    │ 170233 ± 1.16%   │ 160197 ± 1150.0  │ 6143 ± 0.25%           │ 6242 ± 45              │ 5875    │
  │ innerHTML: divs with inline styles    │ 2658439 ± 2.02%  │ 2410629 ± 40695  │ 388 ± 1.52%            │ 415 ± 7                │ 377     │
  │ createElement + style properties      │ 3186658 ± 1.63%  │ 2929996 ± 18582  │ 320 ± 1.37%            │ 341 ± 2                │ 314     │
  │ createElement + setAttribute('style') │ 2310978 ± 1.87%  │ 2105733 ± 17121  │ 446 ± 1.38%            │ 475 ± 4                │ 433     │
  │ createElement + style.cssText         │ 3077670 ± 16.06% │ 2216778 ± 35595  │ 390 ± 2.80%            │ 451 ± 7                │ 336     │
  └───────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

  After:

  ```
  ┌───────────────────────────────────────┬───────────────────┬─────────────────────┬────────────────────────┬────────────────────────┬─────────┐
  │ Task name                             │ Latency avg (ns)  │ Latency med (ns)    │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples │
  ├───────────────────────────────────────┼───────────────────┼─────────────────────┼────────────────────────┼────────────────────────┼─────────┤
  │ innerHTML: simple divs (no styles)    │ 399082 ± 0.75%    │ 385714 ± 1959.5     │ 2543 ± 0.32%           │ 2593 ± 13              │ 2506    │
  │ innerHTML: divs with inline styles    │ 146450667 ± 0.77% │ 146841885 ± 2605346 │ 7 ± 0.76%              │ 7 ± 0                  │ 64      │
  │ createElement + style properties      │ 150609308 ± 0.82% │ 150522448 ± 3501483 │ 7 ± 0.81%              │ 7 ± 0                  │ 64      │
  │ createElement + setAttribute('style') │ 144763409 ± 0.55% │ 144451686 ± 2467319 │ 7 ± 0.55%              │ 7 ± 0                  │ 64      │
  │ createElement + style.cssText         │ 149712884 ± 0.63% │ 149813763 ± 2469720 │ 7 ± 0.63%              │ 7 ± 0                  │ 64      │
  └───────────────────────────────────────┴───────────────────┴─────────────────────┴────────────────────────┴────────────────────────┴─────────┘
  ```

</details>


domenic:
This is really exciting to see! I know you have been working on this for a long time. Given how focused jsdom is on specification conformance, I am excited by your approach, which always seemed very aligned with implementing the specs as written.

That said, I'm hesitant about this change given the number of regressions it introduces. From what I can see we have roughly the following categories:

- Not supporting resolving `<length>` and `<color-mix()>`. jsdom was nowhere near perfect in its support for these, but the lack of partial support might be a regression for users, and it's hard to tell how seriously it would impact them. I know we have historically seem a good amount of bugs around simpler color resolution stuff; many of the to-upstream tests come from user-reported cases.

- Regressions due to lack of coupling with jsdom, notably on `:focus` and custom elements. I know this will be a serious issue for our users, given how many bug reports we've seen on these areas in the past.

  What we can do here depends on what you're willing to do. In a real browser, the style system is highly coupled to the rest of it, and so these types of features are trivial to implement. There is a spectrum of options here, from creating a stable generic adapter layer through which jsdom talks to a generic CSS implementation, all the way to merging `cdoublev/css` into the jsdom codebase. If I were to propose a concrete middle ground, it would be to have some sort of way for you to grab the jsdom impl objects, and document which underscored properties you need from them, and then we'd ensure to never break those without corresponding changes to `cdoublev/css`. But I am open to suggestions.

- A lot of cases where, per your analysis, the WPTs disagree with the spec, or disagree with your interpretation of some loose/unspecified parts of the spec, or the spec is terrible. In these cases, I would like to get the WPTs updated, or to align `cdoublev/css` with the WPTs. Otherwise, we decrease test coverage of these areas, and risk drifting away from browser behavior.

  I realize this is pretty annoying. If it helps, I think I still have WPT merge permissions, and can try to be a fallback reviewer if no CSSWG people respond within, say, 2 days.

  If there are cases where the WPTs disagree with 2+ browsers (especially in the to-upstream cases), then I am fine leaving the tests failing.

- Serious performance regressions, enough that some tests and benchmarks become unrunnable. Although I agree with you that caching `CSS.supports()` is not necessary, we've gotten complaints in the past about stuff like regressing `new JSDOM()` or the `style` property. And just generally slowing down tree mutation is pretty bad.

  Although in the past we've accepted some performance regressions when it brings more spec compliance, it's been pretty painful for our users. (See e.g. https://github.com/jsdom/jsdom/issues/3659 and the [v23.2.0](https://github.com/jsdom/jsdom/releases/tag/v23.2.0) / [v24.0.0](https://github.com/jsdom/jsdom/releases/tag/v24.0.0) churn, plus the more recent instances of https://github.com/jsdom/jsdom/issues/3984, https://github.com/jsdom/jsdom/issues/3985.) In such cases we need to do everything in our power to ensure the performance drops are really unavoidable.

  At the very least, I would suggest pointing some powerful coding agents at the regressions, and getting their analysis and suggestions for fixes. In my experience they're pretty good at this, but definitely need some steering in order to find the sweet spot in terms of performance optimizations that pay for themselves while not making the code overly complex. (Example, I tried to do https://github.com/jsdom/jsdom/pull/3668 by hand in 2024-01 and gave up, whereas https://github.com/jsdom/jsdom/pull/4065 in 2026-02 was more correct and thus more performant.)

So, although I am very enthusiastic about the direction and think this could become a much better foundation for CSS in jsdom, I do not think we can merge it in its current form without first addressing the above. I'm willing to help with that, e.g. on jsdom integration point brainstorming, WPT improvements, and performance investigation. What do you think?

cdoublev:
I appreciate your thoughtful answer, all the more so given you have been a long-standing key figure in Web standards, and that I have witnessed your expertise many times on GitHub.

So, the tasks that remain to be completed are:

  - [ ] (1) add support for resolving `<color-mix()>`
  - [ ] (2) add support for resolving `<length>` and math functions
  - [ ] (3) add support for matching `:focus-*`
  - [ ] (4) fix performance regressions
  - [ ] (5) open issues on w3c/csswg-drafts for tests based on unspecified requirements
  - [ ] (6) make PRs on web-platform-tests/wpt for tests that are no longer in sync with the specs

(2) seems easy to complete.

(4) piques my curiosity and motivates me, but I think you are clearly in the best position to figure it out quickly.

(5) does not motivate me all that much because I think these issues will remain unanswered. The same would apply to (6) without your privilege. I do not blame anyone for that.

(3) indeed requires further discussion.

Briefly, I think there are several advantages to keeping `cdoublev/css` separate. In particular, I am thinking of the [tools](https://csstree.github.io/docs/) provided by `css-tree`. I need to analyze more precisely which internal data would need to be exposed, to determine whether an adapter would be overly complex. If it would, then your middle-ground solution would probably be the appropriate one, as I do not think there will ever be another `jsdom`.

(1) is certainly the one that bothers me the most. I know how much work it would require, and just how complex it likely is. I read that you felt burned out by `jsdom` at times. I do not know if I feel the same way, but my motivation for taking on this task is practically nonexistent at the moment. And I do not want to return an unresolved value halfway through resolving mixed colors because I see that one of them includes a color component keyword deeply nested within a math function.

To be honest, I would have been surprised if you had *not* considered these regressions a no-go. I was also looking for some clarification about your strategy for adding new CSS features.

In my opinion, partially implementing a feature in order to quickly meet user requests (did anyone actually ask for adding support to resolve `color-mix()` values, though?) is the same philosophy that drove `cssstyle`. It was obviously a dead end, which is why I started `cdoublev/css`. I do not think that philosophy is appropriate for an open source project like `jsdom`.

So whether I will complete these tasks depend on that clarification, which, I believe you have already provided in your comment.

PR #4229, opened 2026-08-07
Fix fractional second handling in time strings

Jaybhade:

```js
const input = document.createElement("input");
input.type = "time";
input.value = "12:34:56.7";
input.valueAsNumber;   // Chrome: 45296700   jsdom: 45296007
```

`.7` means seven tenths of a second, but jsdom reads the digits after the `.` as a count of
milliseconds, so it lands 693ms away from where every browser does. `.04` becomes 4ms instead of
40ms. Only a three-digit fraction happens to come out right.

For `datetime-local` the same bug also corrupts the stored value, because sanitization round-trips
through the parser:

```js
const input = document.createElement("input");
input.type = "datetime-local";
input.value = "2026-01-01T12:34:56.7";
input.value;   // Chrome: "2026-01-01T12:34:56.7"   jsdom: "2026-01-01T12:34:56.007"
```

The element now holds a different instant than the page assigned to it, with no error anywhere.

### Spec

[Parse a time component](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#parse-a-time-component)
collects the digits and the `.` together and says to "interpret the resulting sequence as a
base-ten number (possibly with a fractional part). Set *second* to that number" — so the fraction
belongs to the second, not to a millisecond field.

The second half is serialization. A
[valid normalized local date and time string](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-normalised-local-date-and-time-string)
must express its time "as the shortest possible string for the given time (e.g. omitting the
seconds component entirely if the given time is zero seconds past the minute, omitting any trailing
zeros in the fractional seconds part, etc.)". `serializeNormalizedDateAndTime` reused
`serializeTime`, which always pads the fraction to three digits, so `.01` came back as `.010` and a
whole second as `.000`. `serializeTime` itself is unchanged — it still backs `type=time`, where a
fixed-width fraction is a perfectly valid time string and matches what Chrome emits from the
`valueAsNumber` setter.

### Verification

Both bugs already have upstream WPT tests, listed as expected failures in `to-run.yaml`; this
change enables them.

- `datetime-local-trailing-zeros.html`: 0/1 → 1/1
- `input-seconds-leading-zeroes.html`: 5/12 → 12/12
- `input-valueasnumber.html`: 63/64 before and after (the remaining `month` setter failure is
  unrelated and stays listed)

I ran those files directly against the built jsdom rather than through `npm run test:wpt`, since
the WPT server wants host entries I'd rather not add to this machine; `test:api` (573 passing) and
`test:to-port-to-wpts` were run normally and are green, as is `npm run lint`.

I also diffed jsdom against Chrome over 44 `time`/`datetime-local` values — valid, invalid, and
every fraction width — comparing both `value` and `valueAsNumber`. 18 of the 44 disagreed before
this change; all 44 agree after. The cases that changed:

| value assigned | type | jsdom before | jsdom after / Chrome |
| --- | --- | --- | --- |
| `12:34:56.7` | time | `45296007` | `45296700` |
| `12:34:56.07` | time | `45296007` | `45296070` |
| `12:34:56.10` | time | `45296010` | `45296100` |
| `00:00:00.5` | time | `5` | `500` |
| `2026-01-01T12:34:56.5` | datetime-local | `…T12:34:56.005` | `…T12:34:56.5` |
| `2026-01-01T12:34:56.010` | datetime-local | `…T12:34:56.010` | `…T12:34:56.01` |
| `2026-01-01T12:34:56` | datetime-local | `…T12:34:56.000` | `…T12:34:56` |

Invalid input is unaffected: `12:34:56.7891` and `12:34:56.` are still rejected, matching Chrome.


lib/jsdom/living/helpers/dates-and-times.js
```diff
@@ -172,8 +173,26 @@ function isValidLocalDateAndTimeString(str) {
 function isValidNormalizedLocalDateAndTimeString(str) {
   return parseLocalDateAndTimeString(str, true) !== null;
 }
+
+// A normalized local date and time string wants its time "expressed as the shortest possible string
+// for the given time", so unlike serializeTime this omits an all-zero fractional part and any
+// trailing zeros within one.
+function serializeShortestTime({ hour, minute, second, millisecond }) {
```
domenic: Instead of duplicating most of serializeTime, just fix serializeTime to always express the shortest string.

domenic (changes requested):
> I ran those files directly against the built jsdom rather than through npm run test:wpt, since
the WPT server wants host entries I'd rather not add to this machine

This is disappointing. I am considering a blanket plan on bot contributions where the bot is not willing to actually follow the Contributing.md guide.

Jaybhade:
Fair hit on the test setup — I've run the real harness this time. Results at the bottom.

**`serializeTime`.** Done in ee0f2d48: `serializeShortestTime` is gone, `serializeTime` itself always emits the shortest string, and `serializeNormalizedDateAndTime` is a one-liner over it again. −20/+7 against the previous revision.

That does change `type=time` output, so I measured what the engines produce for `input.valueAsNumber = ms` (Chromium 1234 and WebKit 2336 via Playwright; Firefox wouldn't launch headless here, so no Gecko column):

| `step` | ms | Chromium | WebKit | jsdom before | jsdom now |
| --- | --- | --- | --- | --- | --- |
| default | 45000000 | `12:30` | `12:30` | `12:30` | `12:30` |
| default | 45001000 | `12:30:01` | `12:30:01` | `12:30:01.000` | `12:30:01` |
| default | 45000500 | `12:30:00.500` | `12:30:00.5` | `12:30:00.500` | `12:30:00.5` |
| `1` | 45000000 | `12:30:00` | `12:30:00` | `12:30` | `12:30` |
| `0.001` | 45000000 | `12:30:00.000` | `12:30:00.0` | `12:30` | `12:30` |

The two engines disagree on the fraction — Chromium pads it to three digits, WebKit trims it — so shortest isn't picking a side against interop here; it matches WebKit for every default-`step` case and drops the `.000` jsdom used to emit. What neither engine does is stop at shortest: both widen the output to the `step` precision. That's step-aware serialization, a separate change, and I've left it alone — happy to follow up if you want it.

For `datetime-local` the two agree with each other and with "expressed as the shortest possible string", and jsdom now matches them on every case I probed, including the `2022-04-19T12:34:56.010` → `…56.01` that `datetime-local-trailing-zeros.html` asserts.

**Tests.** `npm run test:wpt`, full suite:

| | passing | pending | failing |
| --- | --- | --- | --- |
| this branch | 4578 | 2732 | 6 |
| `main`, same setup | 4577 | 2732 | 7 |

The 6 are a strict subset of main's 7, so nothing here regresses. They're all cross-origin/subdomain tests: `css/cssom/stylesheet-cross-origin-redirect-quirks.sub`, `dom/events/EventListener-addEventListener.sub`, `webmessaging/postMessage_crosssite.sub`, both `websockets/cookies` cross-origin cases, and `xhr/cookies.http`. Main's extra one is `encoding/unsupported-labels.window.html` timing out, which looks like a flake.

That's because of one thing I should disclose rather than bury: I don't have root on this machine to add the `web-platform.test` host entries, so the server is bound to `localhost` — `browser_host: "localhost"`, `check_subdomains: false`, and the `dns.lookup` guard in `wpt-server.js` reading `config.browser_host` instead of the literal string. Three uncommitted lines, reverted before pushing; the runner, the server and the tests are otherwise stock. Those 6 failures are that binding, which is why I ran main through the identical setup rather than just asserting it.

Narrowed to the directory this touches, `--fgrep the-input-element` is 78 passing / 10 pending / 0 failing, with both newly-unparked tests passing by name:

```
✔ datetime-local-trailing-zeros.html
✔ input-seconds-leading-zeroes.html
```

`test:api` (573) and the older mocha suites (191 / 557 / 876 / 162) are green.

domenic (approved):
Thanks for the work!

PR #4231, opened 2026-08-08
Add Element querySelectorAll test and bump dom-selector

### LINKED ISSUE #4227: Regression: element.querySelectorAll() returns an empty NodeList when the first selector matches the context element
### Node.js version

22.22.2

### jsdom version

30.0.1

### Minimal reproduction case

```js

const dom = new JSDOM(`
<div class="root">
  <table>
    <tbody>
      <tr id="row"></tr>
    </tbody>
  </table>
</div>
`);

const root = document.querySelector(".root");

console.log(root.querySelectorAll(".root table tbody tr").length);
// jsdom 29.1.1 -> 1
// jsdom 30.0.1 -> 0

console.log(root.querySelectorAll("table tbody tr").length);
// 1

console.log(document.querySelectorAll(".root table tbody tr").length);
// 1
```

I have not verified this behavior in browsers yet.

## Expected behavior

The query should return the `<tr>` element in all three cases.

## Actual behavior

In jsdom 30.0.1, the first query returns an empty `NodeList`.

## Additional context

This causes existing Vue Test Utils tests to fail.

For example:

```js
wrapper.findAll(".productDetail_v2_cmpnt .nav-item .nav-link");
```

returns no matches, while

```js
wrapper.findAll(".nav-item .nav-link");
```

works correctly.

Internally, `wrapper.findAll()` calls `element.querySelectorAll()`, so this appears to be a regression in jsdom rather than Vue Test Utils.

The issue is reproducible after upgrading from jsdom 29.1.1 to 30.0.1 without changing the application or test code.

asamuzaK:
Fix #4227 

Related: https://github.com/jsdom/jsdom/pull/4203

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@asamuzakjp/css-color": "^6.0.7",
-        "@asamuzakjp/dom-selector": "^8.3.0",
+        "@asamuzakjp/dom-selector": "^9.0.1",
         "@bramus/specificity": "^2.4.2",
         "@csstools/css-syntax-patches-for-csstree": "^1.1.7",
         "@exodus/bytes": "^1.15.1",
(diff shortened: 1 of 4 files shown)


domenic (approved):
Thank you as always!

PR #4232, opened 2026-08-10
Normalize computed font-weight values

### LINKED ISSUE #4186: Elements `font-weight` returned by `getComputedStyle` is not normalized
### Node.js version

20.20.0

### jsdom version

29.1.1

### Minimal reproduction case

```js
import { JSDOM, VirtualConsole } from "jsdom";

const { window } = new JSDOM(undefined, {
  virtualConsole: new VirtualConsole(),
});
const { document } = window;

div.style.fontWeight = "bold";

console.log(window.getComputedStyle(div).fontWeight);
```

https://jsfiddle.net/m9qL6tju/1/

### What is the problem?

Expected to output "700", instead JSDOM outputs "bold".

tianrking:
## Summary

- normalize the computed `font-weight` keywords `normal` and `bold` to their numeric values
- resolve `bolder` and `lighter` against the inherited weight using the CSS Fonts relative-weight table
- add a WPT-format regression covering keyword, numeric, initial, and relative weights

Fixes #4186.

## Testing

- `npm run lint -- lib/jsdom/living/css/CSSStyleDeclaration-impl.js lib/jsdom/living/css/helpers/computed-style.js test/web-platform-tests/to-upstream/css/css-fonts/font-weight-computed.html`
- `npm run test:api -- --reporter min` (573 passing)
- direct API regression assertions for 17 keyword and relative-weight cases


domenic:
@asamuzaK can you take a look to see if this matches how you'd implement it?

asamuzaK:
I think we should write property-specific logic in the individual files under `properties/*` when needed.
For this PR, `properties/fontWeight.js`.

The basic idea would be to update the getter like this:
``` javascript
get() {
  const value = this.getPropertyValue(property);
  if (this._computed) {
    // Do something here
    return computedValue;
  }
  return value;
}
```

It might be better to preprocess special keywords like `auto` or relative keywords like `bolder`/`lighter` in `computed-style.js`.
However, I haven't completely made up my mind about this yet.

WDYT?

asamuzaK:
Btw, jsdom doesn't have `css/css-fonts` tests enabled, but the WPT tests are here:
https://wpt.fyi/results/css/css-fonts/parsing/font-weight-computed.html

tianrking:
Thanks for the direction. Updated in c24d2f8: the font-weight resolver now lives in `properties/fontWeight.js`, exposed as a property-specific computed-value hook used by the longhand path. I used the hook instead of resolving only in the IDL getter so `getComputedStyle(...).fontWeight` and `.getPropertyValue(font-weight)` stay consistent. I also expanded the regression around the boundary values from the linked WPT and check both APIs. Validation: 66 production-API assertions passed, the changed WPT file passes WPT lint, targeted ESLint passes, and diff check passes. The API suite reached 572 passing with one unrelated existing `setTimeout` handle-release test timing out locally (including when rerun alone).

tianrking:
The first CI run after the refactor exposed one generic-path regression: custom properties do not have entries in the property descriptor table, so resolving --foo attempted to destructure an undefined descriptor. Fixed in cf2d3f8 by treating a missing descriptor as having no property-specific hook. A direct computed-style check for a custom property now passes alongside the font-weight boundary checks; targeted ESLint and diff check also pass.

domenic:
@asamuzaK WDYT of the latest commit's approach?

asamuzaK:
I think `resolveComputedValue` is versatile and a very smart solution.

However, since the current code only applies to resolve inherited properties, I think there is room to reconsider the naming, for example, changing it to `resolveInheritedComputedValue`.
Alternatively, if the plan is to keep `resolveComputedValue` as is and handle different cases within the function in the future, I think that is perfectly fine too.

As a minor request, please keep the following comment for future reference: `// TODO: Resolve special cases other than color.`

asamuzaK:
One additional point: since `helpers/font-weight.js` is likely rarely referenced by other files, I think moving its contents into `properties/fontWeight.js` would make the code's responsibilities clearer.

domenic:
> However, since the current code only applies to resolve inherited properties, I think there is room to reconsider the naming, for example, changing it to `resolveInheritedComputedValue`.

Since it also handles non-inherited transformations, e.g. normal -> 400 and bold -> 700, I think the current name is more correct?

> As a minor request, please keep the following comment for future reference: `// TODO: Resolve special cases other than color.`

Good catch, restored, thank you.

> One additional point: since `helpers/font-weight.js` is likely rarely referenced by other files, I think moving its contents into `properties/fontWeight.js` would make the code's responsibilities clearer.

Done.

My original thinking was that all the "resolve some CSS grammar" helpers should live in `helpers/`, like we have for font sizes already. It could be used in the future for `@font-face` descriptors. But, right now it is only used in one place, so the simpler architecture that you suggest is better.

PR #4235, opened 2026-08-12
Only track named window properties from the window's associated document

### LINKED ISSUE #4234: window-properties named-property tracker retains elements (and whole documents) from non-associated same-window documents
The named-property tracker in `lib/jsdom/living/window-properties.js` registers id/`name`-bearing elements from **any** document that shares a `Window` — not just the window's associated document. Entries are only removed by `elementDetached`, which never fires for a document that is simply discarded, so under a long-lived window every parsed side document containing an `id` or `name` attribute is retained forever.

### Impact

1. **Unbounded memory growth in servers.** The common server-side sanitizing setup — DOMPurify on one long-lived jsdom window (what `isomorphic-dompurify` does) — parses each input into a fresh document via the window's `document.implementation.createHTMLDocument()` / `DOMParser`. Every such document containing an element with an `id` (or a `name` on `img`/`form`/`iframe`/…) is pinned via `trackers` → per-window `Map<name, Set<Element>>` → `element._ownerDocument`. We measured **43.35 KiB retained per sanitize call** for id-bearing HTML vs 0.17 KiB for the same HTML without ids, and ~35 MiB/h growth on a production worker thread until the process was recycled (found via a heap-snapshot retainer walk: every leaked `Document` chained through the `trackers` WeakMap under the one persistent window).

2. **Spec-visible misbehavior.** Named window properties are defined to come from the Window's *associated Document* only ([HTML §7.2.2.3, "named objects of `Window` object `window` with the name `name`" are scoped to "`window`'s associated `Document`"](https://html.spec.whatwg.org/multipage/nav-history-apis.html#named-access-on-the-window-object)). In jsdom, elements of unrelated documents resolve as `window.<id>`:

### Repro

```js
const { JSDOM } = require("jsdom"); // 29.1.1

const { window } = new JSDOM("");
const side = window.document.implementation.createHTMLDocument("");
const el = side.createElement("div");
el.setAttribute("id", "sideElement");
side.body.appendChild(el);

console.log(window.sideElement); // HTMLDivElement — expected: undefined
// ...and `side` (the whole document tree) is now unreachable-code-wise garbage,
// yet strongly retained by the module-level `trackers` WeakMap for as long as
// `window` lives, because the WeakMap *value* holds the element.
```

For the leak shape: run the above in a loop with unique ids under one window and watch heap grow linearly; heap-snapshot retainers all point at `trackers` in `window-properties.js`.

### Suggested fix

In `elementAttached`, `elementDetached` and `elementAttributeModified`, bail out when the element's document is not the window's associated document, e.g.:

```js
const window = element._ownerDocument._globalObject;
if (!window) return;
if (idlUtils.implForWrapper(window._document) !== element._ownerDocument) return;
```

This restores the spec scoping and makes discarded parse documents collectable. We're running exactly this as a patch in production; happy to turn it into a PR if the approach looks right to you.

Environment: jsdom 29.1.1, Node 24/26, observed identically on Linux (ECS) and Windows.

Iaotle:
Fixes #4234.

`elementAttached`/`elementDetached`/`elementAttributeModified` in `lib/jsdom/living/window-properties.js` register id/`name`-bearing elements for every document that shares the window's global object — including side documents from `document.implementation.createHTMLDocument()` and `DOMParser`. Two consequences:

1. **Spec behavior**: [named access on the Window object](https://html.spec.whatwg.org/multipage/nav-history-apis.html#named-access-on-the-window-object) is scoped to the window's associated `Document`, so `window.<id>` currently resolving to elements of unrelated documents is incorrect.
2. **Unbounded retention**: tracker entries are only removed by `elementDetached`, which never fires for a discarded document, so under a long-lived window every parsed side document containing an `id`/`name` element is retained for the window's lifetime. This leaks unboundedly in the common server-side sanitizing setup (DOMPurify on one persistent window, as `isomorphic-dompurify` does) — we measured 43 KiB retained per sanitize call of id-bearing HTML and ~35 MiB/h on a production worker, with heap-snapshot retainer paths all going through the `trackers` WeakMap (details in #4234).

This adds the associated-document guard to all three hooks, plus regression tests: side-document elements are not exposed and not retained, elements adopted into the associated document still become exposed, and normal named access is unchanged.

Verified: `mocha test/to-port-to-wpts/window-named-properties.js` (new, 4 passing) and existing suites still pass locally; and against the patched build, the DOMPurify retention repro drops from 43.35 KiB/call to 0.19.

### DIFF
--- a/lib/jsdom/living/window-properties.js
+++ b/lib/jsdom/living/window-properties.js
@@ -68,6 +68,15 @@ exports.elementAttached = element => {
     return;
   }
 
+  // Only elements in the window's associated document contribute named window properties:
+  // https://html.spec.whatwg.org/multipage/nav-history-apis.html#named-access-on-the-window-object
+  // Tracking elements of other same-global documents would also retain them (and their
+  // whole documents) for the lifetime of the window, since elementDetached never fires
+  // for a document that is simply discarded. See #4234.
+  if (window._document && idlUtils.implForWrapper(window._document) !== element._ownerDocument) {
+    return;
+  }
+
   if (element.namespaceURI !== HTML_NS) {
     return;
   }
@@ -96,6 +105,15 @@ exports.elementDetached = element => {
     return;
   }
 
+  // Only elements in the window's associated document contribute named window properties:
+  // https://html.spec.whatwg.org/multipage/nav-history-apis.html#named-access-on-the-window-object
+  // Tracking elements of other same-global documents would also retain them (and their
+  // whole documents) for the lifetime of the window, since elementDetached never fires
+  // for a document that is simply discarded. See #4234.
+  if (window._document && idlUtils.implForWrapper(window._document) !== element._ownerDocument) {
+    return;
+  }
+
   if (element.namespaceURI !== HTML_NS) {
     return;
   }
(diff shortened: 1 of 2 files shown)


domenic:
Per Contributing.md, do not add new "to-port-to-wpts" tests.

PR #4237, opened 2026-08-15
perf(forms): Cache label associations by root

scttcper:
**Note**: This pr is from AI but I've iterated on it a bit and this change would improve a single expensive form test in [sentry](https://github.com/getsentry/sentry) by 35% [on this test file](https://github.com/getsentry/sentry/blob/c73856753969efc2e12f13363c4db17a3b80849c/static/gsAdmin/components/provisionSubscriptionAction.spec.tsx).

Every control's live `labels` NodeList currently walks the full DOM root and resolves labels independently. Reading labels for many controls after a mutation repeats that root-wide work once per control.

This shows up in Testing Library tests over larger forms like this:

```html
<form>
  <label for="field-0">Field 0</label>
  <input id="field-0">

  <!-- many more labeled inputs -->

  <label for="field-99">Field 99</label>
  <input id="field-99">
</form>
```

```js
screen.getByRole("textbox", { name: "Field 99" });
```

`getByRole` computes accessible names for the candidate textboxes. Those calculations read `input.labels`, so one query can make every control rescan the same root.

Build one label-to-control index per root version and share it across those live NodeLists. Tree and attribute mutations already update the root version; form-associated custom element upgrades now update it too because they can change implicit label ownership.

Adds live-list coverage for `for`/`id` changes and custom element upgrades, plus a benchmark that shows both the cold cost and the amortized case.

`node --run benchmark -- --suite forms/labels`, latency median:

| Task | Before | After | Change |
| --- | ---: | ---: | ---: |
| Read 1 of 100 controls after mutation | 0.563 ms | 0.571 ms | +1.5% |
| Read all 100 controls after mutation | 54.64 ms | 0.644 ms | 98.8% less time (84.8x) |

### DIFF
--- a/benchmark/forms/labels.js
+++ b/benchmark/forms/labels.js
@@ -5,35 +5,39 @@ const { JSDOM } = require("../..");
 module.exports = () => {
   const bench = new Bench();
 
-  const CONTROLS = 100;
-  const { document } = (new JSDOM()).window;
+  function addTask(controlCount, labelCount, readCount) {
+    const { document } = (new JSDOM()).window;
+    const inputs = [];
 
-  for (let i = 0; i < CONTROLS; i++) {
-    const label = document.createElement("label");
-    label.htmlFor = `control-${i}`;
-    label.textContent = `Control ${i}`;
+    for (let i = 0; i < controlCount; i++) {
+      if (i < labelCount) {
+        const label = document.createElement("label");
+        label.htmlFor = `control-${i}`;
+        label.textContent = `Control ${i}`;
+        document.body.append(label);
+      }
 
-    const input = document.createElement("input");
-    input.id = `control-${i}`;
+      const input = document.createElement("input");
+      input.id = `control-${i}`;
+      document.body.append(input);
+      inputs.push(input);
+    }
 
-    document.body.append(label, input);
+    bench.add(`read labels for ${readCount}/${controlCount} controls after mutation (${labelCount} labels)`, () => {
+      document.body.toggleAttribute("data-mutated");
+      let labels = 0;
+      for (let i = 0; i < readCount; i++) {
+        labels += inputs[i].labels.length;
+      }
+      return labels;
+    });
   }
 
-  const inputs = document.querySelectorAll("input");
-
-  bench.add(`read labels for 1 of ${CONTROLS} controls after mutation`, () => {
-    document.body.toggleAttribute("data-mutated");
-    return inputs[0].labels.length;
-  });
-
-  bench.add(`read labels for ${CONTROLS} controls after mutation`, () => {
-    document.body.toggleAttribute("data-mutated");
-    let labelCount = 0;
-    for (const input of inputs) {
-      labelCount += input.labels.length;
-    }
-    return labelCount;
-  });
+  addTask(100, 100, 1);
+  addTask(10, 10, 10);
+  addTask(100, 100, 100);
+  addTask(100, 10, 100);
+  addTask(100, 0, 100);
 
   return bench;
 };
--- a/benchmark/forms/labels.js
+++ b/benchmark/forms/labels.js
@@ -21,6 +21,11 @@ module.exports = () => {
 
   const inputs = document.querySelectorAll("input");
 
+  bench.add(`read labels for 1 of ${CONTROLS} controls after mutation`, () => {
+    document.body.toggleAttribute("data-mutated");
+    return inputs[0].labels.length;
+  });
+
   bench.add(`read labels for ${CONTROLS} controls after mutation`, () => {
     document.body.toggleAttribute("data-mutated");
     let labelCount = 0;
--- a/benchmark/forms/labels.js
+++ b/benchmark/forms/labels.js
@@ -1,0 +1,34 @@
+"use strict";
+const { Bench } = require("tinybench");
+const { JSDOM } = require("../..");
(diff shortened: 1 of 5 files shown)


domenic:
> **The following review was produced by ChatGPT 5.6 Sol Extra High, but then reviewed and sign off by @domenic. The prompt was as follows:**
>
>> Review [https://github.com/jsdom/jsdom/pull/4237](https://github.com/jsdom/jsdom/pull/4237) . Is this complexity worth the cost? Does it cause memory leaks? Are the benchmarks sound? Is the code located in a reasonable place? Is the invalidation logic a reasonable balance between correctness and code complexity?

Verdict: the caching strategy is worth pursuing, but I would not merge [PR #4237](https://github.com/jsdom/jsdom/pull/4237) as written. It has one correctness hole and introduces bounded but potentially indefinite memory retention.

## Findings

1. Blocking: custom-element invalidation happens too late and misses rollback.

The definition becomes observable before the constructor runs, but `_modified()` is called only after successful construction. The failure path clears the definition without invalidating anything ([custom-elements.js lines 83–133](https://github.com/jsdom/jsdom/blob/9569374aa38ef54404b05c5b104577337cc70715/lib/jsdom/living/helpers/custom-elements.js#L83-L133)).

I reproduced two incorrect cases:

- When the cache already exists, a successful form-associated constructor sees `[]` from `elementInternals.labels`, although its surrounding label already applies.
- When the constructor reads `elementInternals.labels` and then throws, the cache preserves the temporary association. A following input then incorrectly has no labels until an unrelated mutation occurs.

The latter works on the base revision and fails on this PR. Invalidation should occur when the form-associated definition becomes observable, and again if construction rolls it back. Tests should cover reads inside the constructor and a throwing constructor.

2. The cache introduces real, although bounded, memory retention.

The root retains the last `WeakMap` even after its version changes ([form-controls.js lines 144–170](https://github.com/jsdom/jsdom/blob/9569374aa38ef54404b05c5b104577337cc70715/lib/jsdom/living/helpers/form-controls.js#L144-L170)). Weak keys do not help when the associated control remains connected: the map value strongly holds its labels.

I verified this sequence:

- Read `inputA.labels`, building associations for the whole document.
- Remove a label belonging to still-connected `inputB`.
- Never read any `labels` property again.

The removed label is collectable on the base revision but remains reachable on the PR. Refreshing the cache releases it. Thus this is not an accumulating, unbounded leak, but one entire stale association snapshot can be retained indefinitely.

A good fix would be storing the map in the root’s existing `_memoizedQueries` storage, which `_clearMemoizedQueries()` already discards eagerly during `_modified()`, instead of keeping a versioned symbol property.

## Benchmarks

The central result is sound. I reproduced:

- 100 controls: approximately 60.0 ms → 0.74 ms, about 81× faster.
- One cold control: approximately 0.61 ms → 0.65 ms, about 7% slower in my run.

That closely supports the claimed 85× batch improvement. The exact “+1.5%” cold result is not stable enough to present confidently from a single result.

The benchmark itself ([labels.js lines 24–36](https://github.com/jsdom/jsdom/blob/9569374aa38ef54404b05c5b104577337cc70715/benchmark/forms/labels.js#L24-L36)) fairly measures the intended Testing Library pattern: one mutation followed by accessible-name calculations for many candidate controls. Mutation between iterations prevents accidental reuse.

Its limitations:

- It tests the ideal amortization case only.
- It lacks sparse/no-label documents and scaling across different document sizes.
- It does not measure a full Testing Library query.
- The stated 35% Sentry improvement has no command, sample count, or timing distribution.
- Building the cache is not actually linear: each explicit label’s `control` getter scans the root, so construction remains roughly O(labels × nodes). The PR merely pays that cost once instead of once per control.

I would retain this benchmark but add document-size and label-density variants. The linked Sentry result is valuable supporting evidence, just not independently reproducible from the PR.

## Placement and tradeoff

- `form-controls.js` is the right conceptual home for association construction.
- `benchmark/forms/labels.js` and the label tests are appropriately located.
- The custom-element upgrade helper is the right place to signal the non-DOM semantic change, but the timing needs correction.
- Piggybacking on the root version for all ordinary tree and attribute mutations is a reasonable correctness-versus-complexity tradeoff. It over-invalidates on unrelated mutations, but avoids a large, fragile list of relevant attributes and operations.

So: keep the shared root-level association idea, move its storage into eagerly cleared query-cache infrastructure, and fix upgrade/rollback invalidation. With those changes, the complexity is justified. The targeted label test passed; the additional constructor and GC probes exposed the issues above.

scttcper:
Addressed both blockers: moved the cache into `_memoizedQueries`, and fixed custom-element constructor/rollback invalidation. Added regression tests and confirmed removed labels are collectable.

Three-run median:

| Case | `main` | PR | Change |
|---|---:|---:|---:|
| Cold 1/100 | 0.607 ms | 0.628 ms | ~same |
| Small 10/10 | 0.082 ms | 0.014 ms | 6.0× faster |
| Dense 100/100 | 60.52 ms | 0.67 ms | 90.7× faster |
| Sparse 10/100 | 1.08 ms | 0.052 ms | 20.6× faster |
| No labels | 0.327 ms | 0.040 ms | 8.1× faster |

domenic (approved):
Solid stuff, thank you!

cmacdonnacha:
Hey @domenic , could we get this into the latest release?

PR #4241, opened 2026-08-18
Keep mixed length + calc() track lists in grid-template-columns

### LINKED ISSUE #4141: `grid-template-columns` silently rejects valid top-level track lists that mix bare `<length-percentage>` and `calc()`

rome-xi:
Fixes #4141

Setting `el.style.gridTemplateColumns = '100px calc(100% - 100px)'` currently reads back empty, even though browsers keep the declaration. `resolveCalc` rebuilt multi-token specified values from `name ?? value`, which drops the unit on Dimension and Percentage nodes, so the track list became `100 calc(...)` and failed the property match.

Serialize those tokens with the CSS tree generator so the unit is kept. Properties whose spec `percentages` field is N/A still reject math functions that contain `%`, so `box-shadow: 1px calc(2px + 2%)` stays invalid (that case is what closed #4178).

I hit this in a layout that sizes a sidebar with a bare length and gives the rest of the row to `calc(100% - ...)`.


domenic (changes requested):
Per Contributing.md, do not add new to-port-to-wpts.

domenic (changes requested):
The `cssTree.generate(item)` change looks good, but I don’t think the percentage validation is safe. A property’s “Percentages: N/A” field does not mean percentages are forbidden everywhere in its grammar. This patch rejects valid declarations such as:

```js
el.style.filter = "opacity(calc(50%))";
el.style.scale = "calc(50%)";
el.style.fillOpacity = "calc(50%)";
```

Current jsdom and Chromium accept all of these. Could the validation instead check whether a percentage is allowed at the math function’s actual syntactic position? We should preserve both the valid grid case and the existing rejection of `box-shadow: 1px calc(2px + 2%)`.

rome-xi:
Pushed a follow-up: generate() stays, percentages are only rejected where the grammar is <length>. opacity(calc(50%)), scale, fill-opacity, and the grid case stay valid; box-shadow: 1px calc(2px + 2%) still rejects.

PR #4245, opened 2026-08-20
Adopt an attribute into the document it is set on

Kjubikstronk:
Two places where an attribute kept pointing at the wrong node document.

"Set an attribute" is missing its adopt step, so `setNamedItem` with an attribute from another document left `attr.ownerDocument` on the old one. And `_adoptNode` set the node document of every shadow-including inclusive descendant but not of the attributes in each element's attribute list, which the adopt algorithm also asks for.

Removes two entries from `to-run.yaml`: `attributes-namednodemap-cross-document.window.html` and the `Node-mutation-adoptNode.html` sub-test.

Motivation: still working through the failures recorded in `to-run.yaml` rather than hitting this in an application.


lib/jsdom/living/attributes.js
```diff
@@ -225,6 +225,8 @@ exports.setAttribute = function (element, attr) {
     throw DOMException.create(element._globalObject, ["The attribute is in use.", "InUseAttributeError"]);
   }
 
+  element._ownerDocument._adoptNode(attr);
```
domenic: This does not appear in the spec.

Instead, the spec has steps which jsdom appears to be missing in https://dom.spec.whatwg.org/#concept-element-attributes-replace and https://dom.spec.whatwg.org/#concept-element-attributes-append .`

domenic (changes requested):

Kjubikstronk:
You're right, thanks — I had invented an adopt step in "set an attribute" instead of implementing what's actually written. Fixed.

"Append an attribute" step 3 and "replace an attribute" step 4 both say to set the attribute's node document to the element's node document, so it's a plain assignment in those two, not `_adoptNode`, and not in "set an attribute" at all. `_adoptNode` also does much more than that step asks for.

The other half of the change does appear in the adopt algorithm, so I've kept it:

> Otherwise, if inclusiveDescendant is an element: Set the node document of each attribute in inclusiveDescendant's attribute list to document.

`dom/`, `custom-elements`, `shadow-dom` and the API tests are all green.

domenic (approved):
Thank you, much better!

domenic:
@copilot resolve the merge conflicts in this pull request

PR #4246, opened 2026-08-20
Drop memoized queries when a node changes document

Kjubikstronk:
`getElementsByTagName` bakes the document's HTML-ness into the collection it returns, and per the spec a collection that already exists keeps behaving that way after its root moves to a document of the other kind. A *new* call has to reflect the new document, but the memoized collection was handed back instead, so it kept lowercasing.

Adoption now drops the memoized queries on the nodes it moves. Collections already handed out are untouched, which is what `Element-getElementsByTagName-change-document-HTMLNess.html` checks.

Note this touches the same loop in `_adoptNode` as #4245, so whichever lands second will need a trivial rebase.

Motivation: still working through the failures recorded in `to-run.yaml` rather than hitting this in an application.

### DIFF
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -1028,6 +1028,10 @@ class DocumentImpl extends NodeImpl {
     if (oldDocument !== newDocument) {
       for (const inclusiveDescendant of shadowIncludingInclusiveDescendantsIterator(node)) {
         inclusiveDescendant._ownerDocument = newDocument;
+
+        // Collections built before the move captured the old document's HTML-ness, and the spec
+        // keeps those working. A fresh call has to see the new document, so drop the memoized ones.
+        inclusiveDescendant._memoizedQueries = {};
       }
 
       for (const inclusiveDescendant of shadowIncludingInclusiveDescendantsIterator(node)) {
--- a/lib/jsdom/living/nodes/Document-impl.js
+++ b/lib/jsdom/living/nodes/Document-impl.js
@@ -1031,7 +1031,9 @@ class DocumentImpl extends NodeImpl {
 
         // Collections built before the move captured the old document's HTML-ness, and the spec
         // keeps those working. A fresh call has to see the new document, so drop the memoized ones.
-        inclusiveDescendant._memoizedQueries = {};
+        // Not _clearMemoizedQueries(), which walks up to the ancestors: it is this subtree that
+        // changed document, and the old parent's collections are unaffected by the move.
+        inclusiveDescendant._memoizedQueries = Object.create(null);
       }
 
       for (const inclusiveDescendant of shadowIncludingInclusiveDescendantsIterator(node)) {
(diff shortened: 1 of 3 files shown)


lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -1028,6 +1028,10 @@ class DocumentImpl extends NodeImpl {
     if (oldDocument !== newDocument) {
       for (const inclusiveDescendant of shadowIncludingInclusiveDescendantsIterator(node)) {
         inclusiveDescendant._ownerDocument = newDocument;
+
+        // Collections built before the move captured the old document's HTML-ness, and the spec
+        // keeps those working. A fresh call has to see the new document, so drop the memoized ones.
+        inclusiveDescendant._memoizedQueries = {};
```
```suggestion
        inclusiveDescendant._memoizedQueries = Object.create(null);
```

Bonus points if you also fix https://github.com/jsdom/jsdom/blob/b23f10d526fb9f688e4847819d14dadd705c2972/lib/jsdom/living/nodes/Node-impl.js#L254

lib/jsdom/living/nodes/Document-impl.js
```diff
@@ -1028,6 +1028,10 @@ class DocumentImpl extends NodeImpl {
     if (oldDocument !== newDocument) {
       for (const inclusiveDescendant of shadowIncludingInclusiveDescendantsIterator(node)) {
         inclusiveDescendant._ownerDocument = newDocument;
+
+        // Collections built before the move captured the old document's HTML-ness, and the spec
+        // keeps those working. A fresh call has to see the new document, so drop the memoized ones.
```
domenic, line 1033: Add a comment explaining why we don't use `_clearMemoizedQueries()`.

domenic (commented):
Nice work, although it could be better in a couple ways.

Kjubikstronk:
Adding the spec text this rests on, since I should have quoted it up front.

"The list of elements with qualified name qualifiedName for a node root" picks its filter once, from whether root's node document is an HTML document, and the collection keeps that filter. The caching allowance right below it is the part jsdom was over-reaching:

> When invoked with the same argument, and as long as root's node document's type has not changed, the same `HTMLCollection` object may be returned as returned by an earlier call.

`astMetaCache` aside, jsdom's `memoizeQuery` was returning the earlier collection after the document type *had* changed, which is what `Element-getElementsByTagName-change-document-HTMLNess.html` catches. Dropping the memoized queries during adopt keeps the reuse allowed by that sentence and stops it where the sentence stops.

Collections already handed out are deliberately left alone, which the same test checks in its first two assertions.

Kjubikstronk:
All three done — `Object.create(null)` in the adopt loop, the same in `_clearMemoizedQueries()` and in the `NodeImpl` constructor so the three sites agree, and a comment on why this doesn't go through `_clearMemoizedQueries()`: that walks up to the ancestors, and it's this subtree that changed document — the old parent's collections aren't affected by the move.

`dom/`, `html/dom`, `custom-elements`, `shadow-dom` and the API tests are all still green.

domenic (approved):

PR #4247, opened 2026-08-20
Absolutize lengths inside math functions containing percentages - close #4224

### LINKED ISSUE #4224: `getComputedStyle()` returns the specified value, not the computed value, for math functions containing a percentage
### Node.js version

v26.4.0

### jsdom version

30.0.1

### Minimal reproduction case

```js

// display:none keeps layout out of the picture, so getComputedStyle() returns the
// computed value rather than a used value.
const cases = [
  ["width", "calc(100% - 1rem)"],
  ["width", "calc(100% - 1em)"],
  ["margin-left", "calc(50% - 0.5rem)"],
  ["width", "min(20rem, 100%)"],
  ["width", "max(20rem, 100%)"],
  ["width", "clamp(10rem, 50%, 20rem)"],
  ["width", "calc(2rem + 8px)"], // control: fully reducible
  ["width", "50%"]               // control: bare percentage
];

const html = `<!doctype html><html><body>${cases
  .map(([prop, value], i) => `<div id="c${i}" style="display: none; ${prop}: ${value}"></div>`)
  .join("")}</body></html>`;

const { window } = new JSDOM(html);

cases.forEach(([prop, value], i) => {
  const el = window.document.getElementById(`c${i}`);
  const computed = window.getComputedStyle(el).getPropertyValue(prop);
  console.log(`${prop}: ${value}`.padEnd(34), "->", computed);
});
```

Chromium 151.0.7922.34, same markup. Chrome absolutizes the lengths inside the math
function and keeps the percentage; jsdom returns the declaration untouched:

| declared value (on `display: none`) | Chromium | jsdom 30.0.1 |
| --- | --- | --- |
| `width: calc(100% - 1rem)` | `"calc(100% - 16px)"` | `"calc(100% - 1rem)"` |
| `width: calc(100% - 1em)` | `"calc(100% - 16px)"` | `"calc(100% - 1em)"` |
| `margin-left: calc(50% - 0.5rem)` | `"calc(50% - 8px)"` | `"calc(50% - 0.5rem)"` |
| `width: min(20rem, 100%)` | `"min(320px, 100%)"` | `"min(20rem, 100%)"` |
| `width: max(20rem, 100%)` | `"max(320px, 100%)"` | `"max(20rem, 100%)"` |
| `width: clamp(10rem, 50%, 20rem)` | `"clamp(160px, 50%, 320px)"` | `"clamp(10rem, 50%, 20rem)"` |
| `width: calc(2rem + 8px)` (control) | `"40px"` | `"40px"` |
| `width: 50%` (control) | `"50%"` | `"50%"` |

### What is the problem?

When a math function can be reduced to a single length, jsdom 30 absolutizes it correctly
(`calc(2rem + 8px)` → `40px`, matching Chrome). When a percentage prevents full reduction,
the whole declaration is returned as specified, so font-relative units survive into the
computed value.

Per [CSS Values 4 § Math functions](https://drafts.csswg.org/css-values-4/#calc-computed-value),
the computed value of a math function is the expression with each of its terms computed —
lengths absolutize at computed-value time, and only the percentage has to remain for layout.
`calc(100% - 1rem)` should compute to `calc(100% - 16px)`, not stay `calc(100% - 1rem)`.

This looks like the other half of #4193. That issue was the crash on these same values, fixed
in 30.0.1 — thank you. As noted in [a comment there](https://github.com/jsdom/jsdom/issues/4193#issuecomment-5091866458),
`resolveCalc()` returns the value unreduced when a percentage is involved; with the crash
fixed, that unreduced value is now what surfaces from `getComputedStyle()`, still carrying its
`rem` / `em` / `ex` terms.

Practical impact is on anything comparing computed values against browser behavior — assertion
libraries that diff `getComputedStyle()` output, snapshot serializers, or code branching on a
computed length. Because the surviving units are font-relative, the value also depends on
inherited `font-size`, so the same declaration can serialize differently per subtree while
Chrome reports one absolutized result.

29.1.1 returned these unresolved too, so this is not a regression from 30 — it is the gap left
once the crash was fixed.

soroushm:
getComputedStyle() left math functions entirely unresolved whenever a percentage prevented full reduction, so font-relative units survived into the computed value. Lengths absolutize at computed-value time; only the percentage has to remain for layout, e.g. calc(100% - 1rem) now computes to calc(100% - 16px).

Fixes #4224.

### DIFF
--- a/lib/jsdom/living/css/CSSStyleDeclaration-impl.js
+++ b/lib/jsdom/living/css/CSSStyleDeclaration-impl.js
@@ -478,6 +478,18 @@ class CSSStyleDeclarationImpl {
       if (typeof resolvedValue === "number" && !Number.isNaN(resolvedValue)) {
         return `${Number(resolvedValue.toPrecision(6))}px`;
       }
+      // A percentage prevents the math function from being fully reduced, e.g.
+      // width: calc(100% - 1rem) computes to calc(100% - 16px). Only serialize the partially
+      // reduced string when the math function is the whole value, because reducing it
+      // lowercases the value, which would destroy case-sensitive tokens surrounding it, e.g.
+      // the line names in grid-template-rows: [Foo] minmax(1rem, calc(100% - 2rem)) [Bar].
+      if (
+        typeof resolvedValue === "string" && resolvedValue &&
+        Array.isArray(parsedValue) && parsedValue.length === 1 &&
+        parsedValue[0].type === cssValues.AST_TYPES.CALC
+      ) {
+        return resolvedValue;
+      }
     }
 
     // TODO: Resolve special cases other than color.
--- a/lib/jsdom/living/css/helpers/font-sizes.js
+++ b/lib/jsdom/living/css/helpers/font-sizes.js
@@ -5,6 +5,7 @@ const cssValues = require("./css-values");
 
 const FONT_SIZE_REGEXP = /^((?:\d+(?:\.\d+)?|\.\d+)(?:e[+-]?\d+)?)(%|[a-z]+)$/;
 const LENGTH_REGEXP = /^([+-]?(?:\d+(?:\.\d+)?|\.\d+)(?:e[+-]?\d+)?)(%|[a-z]*)$/;
+const PERCENTAGE_REGEXP = /([+-]?(?:\d+(?:\.\d+)?|\.\d+)(?:e[+-]?\d+)?)%/g;
 
 // Absolute font size mapping table.
 const absoluteFontSize = new Map([
@@ -89,6 +90,16 @@ function resolveFontSizeInPixels(elementImpl, size, root, parent) {
   return Number.NaN;
 }
 
+// Replaces the percentages in a font-size math function with their pixel equivalent.
+// Percentages in font-size are relative to the parent's font size, which is the em unit here.
+function replaceFontSizePercentages(size, em) {
+  if (cssValues.hasVarFunc(size)) {
+    return size;
+  }
+  const pxSize = em ?? absoluteFontSize.get("medium").px;
+  return size.replace(PERCENTAGE_REGEXP, (_, value) => `${value * relativeLength.get("%") * pxSize}px`);
+}
+
 function resolveLengthInPixels(elementImpl, size, dimension, isFontSize) {
   const { em, rem, vh, vw } = dimension;
   if (absoluteFontSize.has(size)) {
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -7,6 +7,7 @@ v8.log
 node_modules
 test/web-platform-tests/tuwpt-manifest.json
 npm-debug.log*
+.idea
```
domenic: Please add this to your user gitignore, not this project's gitignore.

domenic (changes requested):
The general approach makes sense: `resolveLengthInPixels()` already produces the partially reduced computed-value string, so allowing its caller to return that string is a reasonable way to handle a percentage that prevents complete reduction.

However, returning the string unconditionally exposes a regression for compound values containing case-sensitive tokens. The underlying `cssCalc()` implementation lowercases the entire input via `value.toLowerCase().trim()`, not only the math expression. For example:

```js
element.style.gridTemplateRows =
  "[Foo] minmax(1rem, calc(100% - 2rem)) [Bar]";

getComputedStyle(element).gridTemplateRows;
```

Chromium returns:

```text
[Foo] minmax(16px, calc(100% - 32px)) [Bar]
```

With this patch, jsdom returns:

```text
[foo] minmax(16px, calc(100% - 32px)) [bar]
```

So the line names are incorrectly lowercased. A minimal regression case could be added to the existing table:

```js
[
  "grid-template-rows",
  "[Foo] minmax(1rem, calc(100% - 2rem)) [Bar]",
  "[Foo] minmax(16px, calc(100% - 32px)) [Bar]"
]
```

One possible fix would be to return the calculated string only when the parsed declaration consists of a single top-level numeric math value. The caller already has `parsedValue`, and all the current `width`, margin, `min()`, `max()`, and `clamp()` cases should satisfy such a condition. Alternatively, `resolveCalc()` would need to preserve case-sensitive tokens outside the expression it is simplifying.

There is also a remaining `font-size` case. Percentages in `font-size` resolve at computed-value time, so they must not remain in the result. For example:

```html
<div id="parent" style="font-size: 20px">
  <div id="child" style="display: none; font-size: calc(100% + 1em)"></div>
</div>
```

```js
assert_equals(getComputedStyle(child).fontSize, "40px");
```

This patch instead produces `calc(100% + 20px)`. That problem predates this PR, but the new string-return path changes its serialization without fully computing it. It would be good either to handle this case or to exclude `font-size` from the new partial-string behavior until percentage resolution is supported.

soroushm:
Thank for taking time to review, both points addressed in this PR.

**Lowercasing of case-sensitive tokens.** Took the first suggestion: the partially reduced string is now returned only when the parsed declaration is a single top-level math value, i.e. `parsedValue` is one node of type `AST_TYPES.CALC`. For `[Foo] minmax(1rem, calc(100% - 2rem)) [Bar]` the single top-level node is the `minmax` `Function`, so it falls through to the specified value and the line names keep their case.All the `width`, margin, `min()`, `max()` and `clamp()` cases are single `Calc` nodes and are unaffected.

One note on the suggested test case: I couldn't assert the Chromium result `[Foo] minmax(16px, calc(100% - 32px)) [Bar]`, because jsdom doesn't absolutize lengths inside compound values at all today `grid-template-rows: 1rem` returns `1rem` as well, with or without this patch. Getting there needs the second option you mentioned (`resolveCalc()` preserving case-sensitive tokens and recursing into non-math functions), which felt like its own change. So the regression test asserts the part that is true in both engines that the line names survive with their case:

```js
test(() => {
  const div = document.getElementById("target");
  div.style.gridTemplateRows = "[Foo] minmax(1rem, calc(100% - 2rem)) [Bar]";
  const res = window.getComputedStyle(div).gridTemplateRows;
  div.style.removeProperty("grid-template-rows");
  assert_true(res.startsWith("[Foo] "), `"${res}" starts with the [Foo] line name`);
  assert_true(res.endsWith(" [Bar]"), `"${res}" ends with the [Bar] line name`);
}, "getComputedStyle() preserves the case of line names around a math function");
```

Happy to switch it to the exact expectation if you'd prefer I do the `resolveCalc()` work inthis PR instead.

**`font-size` percentages.** Handled rather than excluded. Percentages in `font-size` resolve against the parent's computed font size, which is already the `em` entry of the dimension passed to `resolveLengthInPixels()`, so they're substituted before the math function is reduced, the same resolution a plain percentage font size already gets in the non-`calc` branch just below. `font-size: calc(100% + 20px)` under a 20px parent now computes to `40px`.

Added to the test file:

```js
["calc(100% + 20px)", "40px"],
["max(1rem, 150%)", "30px"],
["calc(50% + 1rem)", "26px"]
```

Also checked by hand against Chromium: inherited chains (`html { font-size: calc(100% + 4px) }` -> `20px`, a nested `1em` below a computed `40px` parent -> `40px`, and `rem` following it), `calc(33% + 1px)` -> `7.6px`, `calc(-50% + 30px)` -> `20px`, `clamp(50%, 1rem, 200%)` -> `16px`, and values containing `var()`, which are left alone.

domenic (approved):
Thanks, seems good!

PR #4248, opened 2026-08-22
perf(css): Avoid unused getComputedStyle() calculations

scttcper:
`getComputedStyle()` currently calculates inherited color and font-size values for every property, even when reading something unrelated like `visibility`.

Only prepare that context for color and length properties. Everything still goes through `parsePropertyValue()` with `format: "computedValue"`.

Local benchmark reading `display`, `visibility`, `opacity`, and `position`:

| Before | After | Speedup |
|---:|---:|---:|
| 186k ops/s | 224k ops/s | 1.21x |


domenic (approved):
Seems pretty good; thank you!

PR #4249, opened 2026-08-24
Avoid rebuilding live collections after every mutation

scttcper:
Reading `childNodes` or `children` creates a live collection. Later DOM updates rebuild that collection after every mutation, even when it is not read between mutations.

The eager updates [date back to 2011](https://github.com/jsdom/jsdom/commit/421d9e5d65bc7b40b45ac3158b8dd6828265da5b), when live collections exposed materialized properties directly. The later [NodeList](https://github.com/jsdom/jsdom/commit/9c56380eb8ccdad4f9ee9d3736d03f0db07d18f9) and [HTMLCollection](https://github.com/jsdom/jsdom/commit/c2143cce551572f41361e75c9847b8615dcbfa54) WebIDL conversions kept them, but observable collection access now already refreshes stale data using the node version.

The benchmark starts with 100 items and replaces 25 of them. Three-run median:

| Collection | Access | `main` | PR | Change |
| --- | --- | ---: | ---: | ---: |
| `childNodes` | After updating the list | 0.251 ms | 0.166 ms | 1.51x faster |
| `childNodes` | After every mutation | 0.267 ms | 0.271 ms | ~same |
| `children` | After updating the list | 0.273 ms | 0.168 ms | 1.62x faster |
| `children` | After every mutation | 0.309 ms | 0.307 ms | ~same |


domenic (approved):
Really great find. However it introduced a memory leak, which I had Codex correct. (I also had it add tests for childNodes instead of just children, which required moving it to a different file, since childNodes belongs to Node instead of ParentNode.)

Let me know if you have any thoughts or comments. Otherwise I'll merge sometime in the next few days.

scttcper:
@domenic looks good! I tried doing a GC style test and it was flakey, glad to see you've figured something out.

PR #4250, opened 2026-08-24
perf(dom): Avoid unnecessary classList creation

scttcper:
While profiling class-name queries over large rendered trees, I found that `getElementsByClassName()` accesses `classList` on every descendant. This creates and retains a `DOMTokenList` even for elements without a `class` attribute.

Skip that work when the attribute-name cache proves an element cannot match. Existing `classList` instances and class-bearing elements continue through the current `DOMTokenList` matching path.

On the sparse first-read case below, this reduces materialized `classList`s from 1,000 to 100. Seven-run median:

| Case | `main` | PR | Change |
| --- | ---: | ---: | ---: |
| First read, no classes | 0.196 ms | 0.041 ms | 4.77x faster |
| First read, 10% classes | 0.214 ms | 0.081 ms | 2.64x faster |
| First read, all classes | 0.334 ms | 0.351 ms | 5% slower |
| Live refresh, 10% classes | 0.178 ms | 0.161 ms | 1.11x faster |
| Live refresh, no classes and 10 other attributes | 0.198 ms | 0.184 ms | 1.08x faster |

### DIFF
--- a/benchmark/dom/get-elements-by-class-name.js
+++ b/benchmark/dom/get-elements-by-class-name.js
@@ -1,0 +1,58 @@
+"use strict";
+const { Bench } = require("tinybench");
+const { JSDOM } = require("../..");
+
+module.exports = () => {
+  // Setting time to zero keeps fixture setup for the first-read cases from making the suite excessively slow.
+  const bench = new Bench({ time: 0, warmupTime: 0 });
+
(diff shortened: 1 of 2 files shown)


scttcper:
not convinced people render things with no classNames

PR #4251, opened 2026-08-24
Stop retaining nodes after MutationObserver.disconnect()

scttcper:
`MutationObserver`’s node list is specified to contain weak references. jsdom represents it as a strong array and currently retains every observed node after `disconnect()`.

Clear the array after removing the observer registrations which prevents keeping observed dom in memory.

Could move `MutationObserver` to use weakrefs but that would be a larger change.


domenic (approved):
Another good find. I went with the WeakRef approach after all, and added back a test.

Let me know if you have any thoughts; otherwise I'll merge sometime in the next few days.

scttcper:
mash it

PR #4252, opened 2026-08-24
Clear pending animation frame callbacks on window.close()

scttcper:
`window.close()` stops the animation frame interval but leaves pending callbacks and related state intact. Clear that state during teardown and prevent new callbacks from being registered after the window is closed.


domenic (approved):
The after-`close()` state is weird, but this is probably the best way to handle things.

PR #4254, opened 2026-08-26
Fire volumechange and ratechange from a queued task

christianaurichzm:
Setting `volume`, `muted`, `playbackRate`, or `defaultPlaybackRate` fires the event synchronously from inside the setter. The HTML Standard queues all four as media element tasks, so the event should not be observable before the assignment returns.

I hit this in a React 17 test suite, where setting `muted` on a `<video>` fired `volumechange` mid-commit and produced an `unstable_flushDiscreteUpdates` warning. The workaround that circulates for it is stubbing the `muted` setter on `HTMLMediaElement.prototype`, which disables `muted` for the whole suite.

`setImmediate` follows how `FileReader` already queues its tasks. `window.setTimeout` would additionally be cleared by `stopAllTimers()` on close, but it would not model `load()` removing pending tasks from the element's own task source, and it would make an internal operation depend on a script-facing API.

Upstream `event_volumechange.html` already covers the `volumechange` side, but jsdom does not currently run that test and there is no equivalent coverage for `ratechange`. The test added here focuses on the four modified setter paths and distinguishes synchronous dispatch from microtask dispatch. Full media element task-source semantics, including `load()` canceling pending tasks, require a per-element task source and are out of scope for this change.

### DIFF
--- a/lib/jsdom/living/nodes/HTMLMediaElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLMediaElement-impl.js
@@ -112,12 +112,17 @@ class HTMLMediaElementImpl extends HTMLElementImpl {
     notImplementedMethod(this._ownerDocument._defaultView, "HTMLMediaElement", "addTextTrack");
   }
 
+  // The HTML Standard queues these events as media element tasks, so they must not fire synchronously.
   _dispatchRateChange() {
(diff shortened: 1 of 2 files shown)


domenic (approved):
Reasonable enough, although I'm never sure what to do with these half-implemented elements.

PR #4255, opened 2026-08-27
Stop AbortSignal from retaining removed event listeners

scttcper:
Listeners added with `{ signal }` leave their abort cleanup registered after `removeEventListener()` or a `{ once: true }` listener runs. When the signal outlives those listeners, it keeps their targets and callbacks alive.

Store the cleanup on the listener record and remove both together. With 2,000 removed listeners and the signal kept alive, retained heap dropped from 41.8 MB to 0.1 MB.


domenic (approved):
Had to go a bit further, but great find in general.

PR #3475, opened 2022-12-29
Revert "Improve performance of appending <option> to <select>"

eps1lon:
Closes https://github.com/jsdom/jsdom/issues/3476

Reverts https://github.com/jsdom/jsdom/pull/3404 which broke `HTMLSelectElement.selectedOptions` when `HTMLOptionElement.selected` is set (revealed by broken `accname/name_file-label-embedded-select-manual.html` in https://github.com/eps1lon/dom-accessibility-api/pull/874

Browser:  HTMLOptionElement.selected
Pre 20.0.1: https://runkit.com/eps1lon/63ad844e952ef10008dda4f7
20.0.1: https://runkit.com/eps1lon/63ad82c22f199e000881c52e

## Test plan

- [x] test for expected behavior fails with https://github.com/jsdom/jsdom/pull/3404: https://github.com/jsdom/jsdom/actions/runs/3800475811/jobs/6463992433
- [x] test for expected behavior passes with https://github.com/jsdom/jsdom/pull/3404 reverted

PR #4257, opened 2026-08-28
Avoid repeated option scans while building selects

scttcper:
Each direct `<option>` insertion currently runs the select selectedness-setting algorithm. Once the first option is selected, later unselected options cannot change selectedness, but each insertion still scans the full options collection. This makes incremental construction quadratic.

Skip the reset when an unselected direct option is inserted after an already-selected first option. Selected options, optgroups, and other insertions continue using the full algorithm.

#3404 used cached selected-option state and was reverted in #3475. This version stores no selected-option state, and the regression test added after that revert now exercises the shortcut.

`benchmark/forms/select.js`, appending 2,000 options:

| `main` | This PR |
| ---: | ---: |
| 44.7 ms | 2.74 ms |

A 2,000-option React mount improved from 71.2 ms to 35.1 ms.

### DIFF
--- a/benchmark/forms/select.js
+++ b/benchmark/forms/select.js
@@ -1,0 +1,15 @@
+"use strict";
+const documentBench = require("../document-bench");
+
+module.exports = () => {
+  const { document, bench } = documentBench();
+
+  bench.add("append 2,000 options", () => {
+    const select = document.createElement("select");
+    for (let i = 0; i < 2000; i++) {
+      select.append(document.createElement("option"));
+    }
+  });
+
+  return bench;
+};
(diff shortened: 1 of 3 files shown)


domenic (changes requested):
Nice find. Please delete the test change, and then use `this.firstElementChild` instead of `domSymbolTree.firstChild(this)`.

We have a separate issue that our HTMLSelectElement implementation has significantly drifted from the spec, likely due to the spec's changes to support customizable select. But that'll have to be tackled another day...

domenic (approved):
Actually what am I thinking, I'll just have the AI fix my review comments and merge it.

PR #4258, opened 2026-08-28
Lazily allocate listener and live Range storage

scttcper:
Every `EventTarget` allocates an empty listener map, and every `Node` allocates an empty set used to track live Ranges. Most objects never use either API.

Both are opt-in bookkeeping on DOM base classes, so this uses `null` for the unused state and creates the collection when the first listener is added or the node becomes a Range boundary container. The Range set still contains weak references as introduced in #3996; only its allocation becomes lazy.

In a retained 2,000-row React/Testing Library tree, median heap usage across seven alternating fresh-process runs fell from 75.21 MB to 65.49 MB, a 12.92% reduction. In alternating runs of the existing `dom/construction` benchmark, median `document.createElement()` throughput increased from 1.37M to 1.50M operations per second. Existing event dispatch and Range mutation benchmarks showed no regression.


domenic (approved):
LGTM with minor tweaks. Also removed the test since I don't like testing implementation details to that extent.

scttcper:
@domenic good call on the test,  i keep trying to have it avoid that

PR #4262, opened 2026-08-31
Centralize memoized tree query caches

domenic:
Centralizes subtree-derived caches in a lazily allocated private aggregate on Node instances. This avoids making every node pay for an unused cache and prevents invalidated document named-property data from retaining removed subtrees.

Supersedes #4260.

### Benchmarks

Compared with `origin/main` at `6312bdee` using Node.js 26.3.1. Values are medians of five runs.

| Speed benchmark | `origin/main` | This PR | Change |
| --- | ---: | ---: | ---: |
| Mutation at depth 12 without query reads | 943,396 ops/s | 1,098,901 ops/s | +16% |
| 100 cached `getElementsByTagName()` reads | 242,718 ops/s | 317,460 ops/s | +31% |
| 100 cached reads across eight tag names | 230,415 ops/s | 293,255 ops/s | +27% |
| 100 cached `getElementsByTagNameNS(null, ...)` reads | 16,795 ops/s | 265,957 ops/s | 15.8x |
| Mutation followed by a query read | 653,595 ops/s | 714,286 ops/s | +9% |

The namespace case improves especially because the old generic memoizer declines to cache calls containing `null`.

Heap measurements used `--expose-gc` and five isolated processes per revision.

| Memory benchmark | `origin/main` | This PR | Change |
| --- | ---: | ---: | ---: |
| Heap for 100,000 retained, otherwise-unused elements | 214.6 MiB | 197.0 MiB | -8.2% |
| Heap retained after invalidating and removing a 20,000-element document named-property subtree | 62.0 MiB | 0.4 MiB | -99.3% |

### DIFF
--- a/benchmark/dom/memoized-queries.js
+++ b/benchmark/dom/memoized-queries.js
@@ -1,0 +1,57 @@
+"use strict";
+const documentBench = require("../document-bench");
+
+const DEPTH = 12;
+const READS = 100;
+const TAG_NAMES = ["a", "abbr", "address", "article", "aside", "b", "blockquote", "button"];
+
+module.exports = () => {
+  const { document, bench } = documentBench();
+  const root = document.createElement("div");
+  let target = root;
+
+  for (let i = 1; i < DEPTH; ++i) {
+    const child = document.createElement("div");
+    target.append(child);
+    target = child;
+  }
+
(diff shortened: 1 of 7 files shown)


scttcper:
good stuff!

PR #4263, opened 2026-09-01
Lazily allocate event handler storage

scttcper:
Every HTML and SVG element currently allocates an event-handler object and a Set, even when no `on*` property is used.

Leave the handler storage null until first use and use its keys to track installed listeners instead of keeping a separate Set. Apply the same registration tracking to the simple event accessors used by `AbortSignal`, `FileReader`, `WebSocket`, and `XMLHttpRequest`. This preserves handler reassignment and listener ordering.

With 20,000 retained elements and `onclick` on one in fourteen, median heap after GC fell from 43.32 MB to 36.70 MB (-15.3%). Speed was neutral.


domenic (commented):
Nice! Can you also fix `setupForSimpleEventAccessors`?

PR #4268, opened 2026-09-02
Compute border-width as an absolute length

### LINKED ISSUE #4267: getComputedStyle() returns 16px for the border-width of an element with no border
### Summary

`getComputedStyle().borderTopWidth` returns `16px` for an element with no border. It is resolving the `medium` keyword — the initial value of `border-width` — out of the **font-size** keyword table, where `medium` means 16px.

```js
const { window } = new JSDOM(`<div id="a"></div>`);
const div = window.document.getElementById("a");

window.getComputedStyle(div).borderTopWidth; // "16px", browsers say "0px"
```

### jsdom 30.0.1 vs Chrome 148

| declaration | Chrome | jsdom |
| --- | --- | --- |
| *(no border)* | `0px` | `16px` |
| `border-style: solid` | `3px` | `16px` |
| `border-style: solid; border-width: medium` | `3px` | `16px` |
| `border-style: solid; border-width: thin` | `1px` | `"thin"` |
| `border-style: solid; border-width: thick` | `5px` | `"thick"` |
| `border-style: none; border-width: 5px` | `0px` | `5px` |

Firefox agrees with Chrome on all six.

### Spec

[css-backgrounds § border-width](https://drafts.csswg.org/css-backgrounds/#the-border-width): the computed value is an absolute length, and 0 if the border style is `none` or `hidden`. jsdom's own generated property data already records this — `css-property-definitions.js` has `computedValue: "absolute length, snapped as a border width"` for `border-top-width`.

### Cause

`CSSStyleDeclaration-impl.js` sends every property with a `length` dimension type through `fontSizes.resolveLengthInPixels(..., isFontSize)`, and the first two branches of that function read the font-size keyword tables without consulting `isFontSize`:

```js
function resolveLengthInPixels(elementImpl, size, dimension, isFontSize) {
  const { em, rem, vh, vw } = dimension;
  if (absoluteFontSize.has(size)) {          // "medium" -> 16
    return absoluteFontSize.get(size).px;
  } else if (relativeFontSize.has(size)) {   // "smaller" / "larger"
```

`isFontSize` is already threaded through and already consulted further down for the `%` case. Separately, nothing maps the `<line-width>` keywords to lengths, and nothing applies the `border-style: none` rule.

This looks like fallout from the CSS implementation replacement (#4092 / #4223): 29.0.x returned `""` here and 29.1.1 returned `"medium"`, both of which parse to 0, so the wrong number only appears from 30.0.0 on.

### Real-world impact

A grid in our CRM frontend measures how wide each column's content needs to be, and adds what each container reserves to the right of it — `paddingRight + borderRightWidth` from `getComputedStyle`. On jsdom 30 every borderless element in a cell contributes a phantom 16px, so measurements came out 16px per nesting level too wide and the tests that cover it failed. Real browsers were fine; only the test environment was wrong.

PR follows.

Alberto-BaseNet:
Fixes #4267.

Every length property is resolved through `resolveLengthInPixels`, whose first two branches read the **font-size** keyword tables without consulting `isFontSize`. `medium` is the initial value of `border-width`, so a borderless element reported 16px of border. `isFontSize` now gates those branches — it was already a parameter and already consulted for the `%` case below.

On top of that, nothing mapped the `<line-width>` keywords to lengths; `helpers/line-widths.js` now does, using the values both Chrome and Firefox use. The zero-width rule from [css-backgrounds § border-width](https://drafts.csswg.org/css-backgrounds/#the-border-width) is a *used*-value rule, which browsers expose as the resolved value while the computed value — and so what a child inherits — stays the length. Computed declarations now keep the two apart: a property file can export a `getResolvedValue` hook, collected into a generated registry like `resolveComputedValue`, which `getPropertyValue()` applies on top of the computed value. Inheritance reads the computed value through `_getComputedPropertyValue()` and never sees the resolved one.

`border-width-cssom.html` and the four `border-*-width-composition.html` tests pass now, so they come off `to-run.yaml`. The `animations` expectation has to spell out what it still covers, because the first matching pattern wins and `animations/*` sorts before any narrower one. Those upstream tests cover the keywords, so the added to-upstream test covers the zero resolved width and that a child inherits the computed length.

Motivation: a grid in our CRM frontend sizes columns to their content, adding what each container reserves to the right of it — `paddingRight + borderRightWidth`. On jsdom 30 every borderless element inside a cell contributed a phantom 16px, so measurements came out too wide by 16px per nesting level. Browsers were fine; only the test environment was wrong.

Left alone: `outline-width` and `column-rule-width` take the same keywords and are still unresolved. They want the same treatment, but `outline-width` has its own zero-if-`none` rule, so it seemed better kept separate.

### DIFF
--- a/lib/jsdom/living/css/CSSStyleDeclaration-impl.js
+++ b/lib/jsdom/living/css/CSSStyleDeclaration-impl.js
@@ -6,6 +6,7 @@ const computedValueResolvers = require("../../../generated/css-property-computed
 const propertyDefinitions = require("../../../generated/css-property-definitions");
 const propertyDescriptors = require("../../../generated/css-property-descriptors");
 const propertyMetadata = require("../../../generated/css-property-metadata");
+const resolvedValueResolvers = require("../../../generated/css-property-resolved-value-resolvers");
 const { asciiLowercase } = require("../helpers/strings");
 const computedStyle = require("./helpers/computed-style");
 const cssValues = require("./helpers/css-values");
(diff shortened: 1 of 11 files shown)


domenic (changes requested):
`resolveBorderWidth()` applies the `none`/`hidden` rule too early. The computed width remains `5px`; only its resolved value becomes `0px`. This matters because inheritance transfers the computed value:

```css
#parent { border-top: 5px none }
#child { border-top-style: solid; border-top-width: inherit }
```

The parent should resolve to `0px`, but the child to `5px`; this patch returns `0px` for both. Please add this case as a WPT.

Could `line-widths.js` compute the width independently of `border-*-style`, with the style-dependent zeroing applied in a separate resolved-value step? In particular, inheritance must not retrieve the parent’s width through `getPropertyValue()`, since that exposes the resolved value.

Alberto-BaseNet:
Thanks @domenic, you're right. The none/hidden rule is now applied in a separate resolved-value step: property files can export a getResolvedValue hook that getPropertyValue() applies on top of the computed value, and inheritance reads the computed value through _getComputedPropertyValue() so it never sees the zero. line-widths.js only maps the keywords now. Added your inheritance case to the to-upstream test, plus a thick hidden parent and a child that keeps none; all pass in Chrome.

PR #4269, opened 2026-09-02
Reduce MutationObserver bookkeeping allocations

scttcper:
This removes several short-lived allocations from MutationObserver bookkeeping.

- Nodes share an empty registration list until the first `observe()`.
- Mutations walk the parent chain directly instead of creating an ancestor array. The first matching observer is stored directly, with the existing `Map` path used when another observer matches.
- Delivery replaces the record queue before calling the observer and uses the old queue directly instead of cloning it.

The `Map` fallback preserves deduplication and each observer's `oldValue` options. Replacing the record queue first keeps records created by the callback in the next delivery.

Testing Library's `waitFor()` normally observes one subtree, so the zero- and one-observer paths are common during DOM updates.

Median throughput from six runs:

| 100 attribute mutations | `main` | This PR | Change |
|---|---:|---:|---:|
| No observers | 31,915 ops/s | 34,042 ops/s | +6.4% |
| One observer | 21,858 ops/s | 23,857 ops/s | +8.8% |
| One observer on the target and an ancestor | 20,735 ops/s | 22,945 ops/s | +11.0% |
| Two observers | 16,388 ops/s | 16,575 ops/s | +1.6% |

Queue-and-delivery throughput improved 6.1–9.4% across batches of 1 to 1,000 records.

|---|---:|---:|---:|
| 30,000-element tree, retained heap | 52.12 MB | 51.17 MB | -1.84% |
| 2,000-row React mount, retained heap | 54.47 MB | 53.51 MB | -1.77% |

### DIFF
--- a/benchmark/dom/mutation-observer.js
+++ b/benchmark/dom/mutation-observer.js
@@ -2,11 +2,37 @@
 const { Bench } = require("tinybench");
 const { JSDOM } = require("../..");
 
-const RECORDS_PER_RUN = 1000;
+const MUTATIONS_PER_QUEUE_RUN = 100;
+const RECORDS_PER_DELIVERY_RUN = 1000;
 const recordsPerDeliveryValues = [1, 10, 100, 1000];
 
 module.exports = () => {
   const bench = new Bench();
+
+  addQueueBenchmark(bench, "queue 100 attribute records, no observers", () => []);
+  addQueueBenchmark(bench, "queue 100 attribute records, one observer", (window, ancestor, target) => {
+    const observer = new window.MutationObserver(() => {});
+    observer.observe(target, { attributes: true });
+    return [observer];
+  });
+  addQueueBenchmark(
+    bench,
+    "queue 100 attribute records, one observer on target and ancestor",
+    (window, ancestor, target) => {
+      const observer = new window.MutationObserver(() => {});
+      observer.observe(target, { attributes: true });
+      observer.observe(ancestor, { attributes: true, attributeOldValue: true, subtree: true });
+      return [observer];
+    }
+  );
+  addQueueBenchmark(bench, "queue 100 attribute records, two observers", (window, ancestor, target) => {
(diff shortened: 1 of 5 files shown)


scttcper:
starting to scrape the bottom of the list of ideas i had

domenic:
edit: welp, I meant to have it just include this stuff in the commit message to minimize my AI-generated comments, especially given how I appreciate you clearly filtering your output to be more useful. But I wasn't precise enough in my instructions, so here's some over-verbose explanation...

<details>
<summary>An explanation of the above extra commit</summary>

I pushed d49e47a4 as a small follow-up to the lazy registration-list optimization.

It preserves the per-node allocation saving, but represents an unallocated list with `null` instead of sharing a mutable empty array across every node. `observe()` allocates the target list on first use, and mutation queuing skips unallocated lists. This removes the shared-state invariant without changing the scalar observer/`Map` optimization or record-queue transfer.

I also added public coverage showing that observers registered on separate targets remain independent. Alternating MutationObserver benchmarks on Node.js 22, 24, and 26 showed no consistent throughput regression versus the shared-array version. The targeted WPT, the 577-test API suite, and lint all pass.
</details>

scttcper:
@domenic thanks for spending your time to look at these! Change looks good. I'm trying to keep them high enough value, running out of ideas and sol is starting to reward hack its way towards micro optimizations that don't hold up well. 

I'll try to open a few issues for things i might not get to like:
> Fix the removed-iframe registry leak. One hundred iframe append/remove cycles left 100 child Windows alive. Removing closed windows from windowsInSameOrigin reduced heap from 140.3 MB to 44.9 MB and prevents detached windows from receiving storage events. [Registry insertion](https://github.com/jsdom/jsdom/blob/126819028dc7134e4e8d2776d8694bf908af93e0/lib/jsdom/living/nodes/HTMLFrameElement-impl.js#L140-L153)

PR #4270, opened 2026-09-02
Lazily allocate MutationObserver registration lists

scttcper:
Every `Node` currently allocates an empty MutationObserver registration array, even though most nodes are never observed.

Share one empty array across unobserved nodes and replace it when `observe()` adds the first registration. The shared array is never modified, and the mutation loop remains unchanged.

|---|---:|---:|---:|
| 2,000-row React mount, retained heap | 54.47 MB | 53.51 MB | -1.76% |
| Create 100 elements with five attributes | 3,861 ops/s | 3,919 ops/s | +1.5% |

### DIFF
--- a/lib/jsdom/living/mutation-observer/MutationObserver-impl.js
+++ b/lib/jsdom/living/mutation-observer/MutationObserver-impl.js
@@ -61,10 +61,16 @@ class MutationObserverImpl {
 
       existingRegisteredObserver.options = options;
     } else {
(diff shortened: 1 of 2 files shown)


scttcper:
there was some other things to investigate in MutationObserver but none of them had clean solutions

scttcper:
Folded into #4269 so the two MutationObserver allocation changes can be reviewed and benchmarked together.

PR #4272, opened 2026-09-03
Add no-op setPointerCapture/releasePointerCapture/hasPointerCapture to Element

### LINKED ISSUE #4271: PointerEvent added in v27 without setPointerCapture/releasePointerCapture/hasPointerCapture — is that API planned?
### Node.js version

v24.19.0 (also reproduces on v20.14.0)

### jsdom version

30.0.1 (regression starts at 27.0.0, per the "Added ... PointerEvent" changelog entry)

### Minimal reproduction case

```js

const dom = new JSDOM(`<button id="b"></button>`);
const { window } = dom;
const button = window.document.getElementById("b");

console.log("PointerEvent supported:", typeof window.PointerEvent !== "undefined");
console.log("setPointerCapture exists:", typeof button.setPointerCapture);

// A real browser exposes setPointerCapture on every Element once Pointer
// Events are supported, and calling it here would just no-op (invalid
// pointerId) rather than throw.
button.setPointerCapture(1);
```

Running this with `node test.js` prints:

```
PointerEvent supported: true
setPointerCapture exists: undefined
```

then throws:

```
TypeError: button.setPointerCapture is not a function
```

In Chrome, Firefox, and Safari, `Element.prototype.setPointerCapture`, `releasePointerCapture`, and `hasPointerCapture` have existed on every `Element` for years (part of the Pointer Events spec, https://w3c.github.io/pointerevents/#dom-element-setpointercapture). Calling `setPointerCapture` with an invalid/unknown `pointerId` throws a `DOMException` ("InvalidPointerId"), not a `TypeError` — but the method itself is always present. Verifiable in any browser devtools console: `typeof document.body.setPointerCapture` returns `"function"`.

### What is the problem?

I expected that once `window.PointerEvent` is defined (as of jsdom 27, see the v27.0.0 changelog: "Added a variety of event constructors, even though we do not implement their associated specifications or ever fire them: ... PointerEvent ..."), the rest of the Pointer Events API surface that the same spec defines on `Element` — `setPointerCapture`, `releasePointerCapture`, `hasPointerCapture` — would also exist (even as no-ops), matching how every real browser behaves.

Instead, those three methods are entirely absent from `Element.prototype` in jsdom 27+. This means any code that does the common feature-detection `typeof PointerEvent !== "undefined"` to decide whether it's safe to use the Pointer Events API (a pattern that was previously a reliable way to detect "no real pointer event support, fall back to mouse events" in jsdom-based tests) now gets a false positive and crashes with `TypeError: ... .setPointerCapture is not a function` the moment it calls any of these methods.

**Context: where this surfaced.** I hit this through [react-aria](https://github.com/adobe/react-spectrum)'s `useMove()` hook, which has exactly this fallback baked in for test environments — `typeof PointerEvent === "undefined" ? wire up onMouseDown/onTouchStart : wire up onPointerDown`. That's not a bug in react-aria; it's a reasonable, long-standing pattern (see e.g. https://github.com/adobe/react-spectrum/issues/2871, where the react-aria team themselves suggested deleting `window.PointerEvent` as the workaround for a similar Safari-only pointer-events issue). It worked fine against jsdom < 27 and silently changed behavior after the bump, purely because jsdom now reports partial, inconsistent Pointer Events support (the constructor exists, the capture API doesn't) — not because anything in react-aria changed. I'd expect any library using this same feature-detection pattern to be similarly affected.

I understand from the v27 changelog that this isn't meant to be a full implementation, and I'm not asking for real hit-testing/capture semantics. My actual question: is adding `setPointerCapture`/`releasePointerCapture`/`hasPointerCapture` (even as no-ops) on the roadmap as part of eventually completing Pointer Events support, or is providing that API surface intentionally out of scope for the foreseeable future? If the latter, it'd be worth a callout in the docs/changelog that `typeof PointerEvent` is no longer a safe "full pointer events support" check post-v27, since it's a fairly established pattern that this change quietly invalidates.

Zeboch:
Every browser exposes `setPointerCapture`, `releasePointerCapture`, and `hasPointerCapture` on `Element` once Pointer Events are supported, and it's common to feature-detect Pointer Events support via `typeof window.PointerEvent !== "undefined"` before calling them (this is exactly what react-aria's `useMove()` hook does). Since jsdom added the `PointerEvent` constructor in v27 without this API surface, that established feature-detection pattern now leads straight to a `TypeError` in code that previously worked fine against jsdom.

This PR implements the actual [pointer capture algorithm](https://w3c.github.io/pointerevents/#pointer-capture) rather than a no-op: an active-pointers set populated by dispatched `pointerdown`/`pointerup`/`pointercancel`, a pending and an actual capture-target-override per `pointerId`, retargeting of subsequent pointer events to the capturing element, and `gotpointercapture`/`lostpointercapture` firing accordingly. `setPointerCapture`/`releasePointerCapture` throw the correct `DOMException`s (`NotFoundError` for an unmatched `pointerId`, `InvalidStateError` for a disconnected element).

jsdom has no hit-testing or real input devices, so active pointers are populated purely from script-dispatched pointer events, matching how jsdom's primary consumers (e.g. testing-library's `fireEvent`) actually drive pointer events in this environment. Pointer Lock interaction from the spec isn't modeled since jsdom doesn't implement that API. Capture on a disconnected element is released with a synchronous `lostpointercapture` at that element, rather than the spec's retarget-to-document-and-defer approach, for simpler, predictable behavior.

Almost every upstream web-platform-test for pointer capture drives input through `test_driver.Actions()`, which needs a real WebDriver-backed browser and can't run against jsdom, so this is verified with a jsdom-only test exercising the algorithm via script-dispatched `PointerEvent`s instead.

Fixes #4271

### DIFF
--- a/lib/jsdom/living/nodes/Element-impl.js
+++ b/lib/jsdom/living/nodes/Element-impl.js
@@ -541,6 +541,17 @@ class ElementImpl extends NodeImpl {
     return this.matches(selectors);
   }
 
+  // https://w3c.github.io/pointerevents/#dom-element-setpointercapture
+  // jsdom does not implement pointer capture or hit-testing, so these are no-ops provided for API
+  // compatibility with code that feature-detects Pointer Events support via `window.PointerEvent`.
+  setPointerCapture() {}
+
+  releasePointerCapture() {}
+
+  hasPointerCapture() {
+    return false;
+  }
+
   // https://html.spec.whatwg.org/#reflecting-content-attributes-in-idl-attributes
   _reflectGetTheElement() {
     return this;
--- a/lib/jsdom/living/nodes/Element.webidl
+++ b/lib/jsdom/living/nodes/Element.webidl
@@ -90,3 +90,10 @@ partial interface Element {
 };
 
 Element includes ARIAMixin;
+
+// https://w3c.github.io/pointerevents/#pointerevent-interface
+partial interface Element {
+  undefined setPointerCapture(long pointerId);
+  undefined releasePointerCapture(long pointerId);
+  boolean hasPointerCapture(long pointerId);
+};
--- a/test/web-platform-tests/to-upstream/jsdom-only/pointer-capture-noop.html
+++ b/test/web-platform-tests/to-upstream/jsdom-only/pointer-capture-noop.html
@@ -1,0 +1,37 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Element pointer capture methods are no-ops</title>
+<link rel="help" href="https://w3c.github.io/pointerevents/#dom-element-setpointercapture">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+
+<button></button>
+
+<script>
+"use strict";
+
+// jsdom does not implement pointer capture or hit-testing, so it does not track active pointers and cannot
+// meaningfully validate the pointerId. Real browsers throw a DOMException for an unknown pointerId
+// (https://w3c.github.io/pointerevents/#dom-element-setpointercapture); jsdom instead makes these methods
+// permanent no-ops, so that `typeof window.PointerEvent !== "undefined"` feature detection does not crash when
+// this API surface is used. This is jsdom-specific behavior and should not be upstreamed.
+test(() => {
+  const button = document.querySelector("button");
+
+  assert_equals(typeof button.setPointerCapture, "function");
+  assert_equals(typeof button.releasePointerCapture, "function");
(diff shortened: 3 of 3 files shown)


domenic:
Sorry, no-ops are not how we implement features in jsdom.

Zeboch:
Ok, but this is exactly what was done for `PointerEvent` until now. Just pushing this a step further to be more compatible with some elements.

I will dig a bit more to see if the implementation is doable in a coherent time 😉.

Zeboch:
Replaced the no-op with a real implementation of the pointer capture algorithm, following up on the feedback above. Couldn't reopen this PR, so continuing in #4273.

PR #4274, opened 2026-09-03
perf: Use the ID cache for exact ID attribute selectors

scttcper:
Testing Library resolves `aria-labelledby` references with scoped selectors such as:

```js
container.querySelector(`[id="${id}"]`);
```

jsdom currently sends these through DOMSelector and walks the query subtree, even though the document already maintains an ordered ID cache.

This uses `getElementById()` for the exact double-quoted, unescaped form. Scoped queries only use the cached element when it is a descendant and not the context itself; all other selectors and contexts continue through DOMSelector.

The public DOM benchmark models that `getByLabelText()` work: 24 controls use `aria-labelledby` to reference labels inside a roughly 1,000-element form, and each reference is resolved with the scoped selector above. These are medians from four alternating fresh-process pairs:

| Case | `main` | This PR |
|---|---:|---:|
| Unchanged tree | 8.56 ms | 0.0109 ms |
| After mutation | 8.56 ms | 0.0120 ms |

The new WPT covers scoped and duplicate matches, detached and shadow trees, escaped values, ID changes, adoption, and attribute namespace and casing rules.

### DIFF
--- a/benchmark/dom/query-selector-id.js
+++ b/benchmark/dom/query-selector-id.js
@@ -1,0 +1,57 @@
+"use strict";
+
+const { Bench } = require("tinybench");
+const { JSDOM } = require("../..");
+
+const ROW_COUNT = 24;
+const FILLER_ELEMENTS_PER_ROW = 38;
+
+module.exports = () => {
+  const bench = new Bench();
+  const { document } = (new JSDOM()).window;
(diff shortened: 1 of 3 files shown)


scttcper:
this belongs in dom-selector sorry

PR #4275, opened 2026-09-03
Avoid repeated root scans for form controls

scttcper:
Reading `form.elements`, calling `checkValidity()`, or constructing `FormData` can scan the complete root even when most of it cannot contribute controls.

Use `getElementById()` to resolve explicit `form="id"` owners. For forms without a non-empty ID, search only the form subtree. Identified forms keep the complete-root path for external controls and duplicate IDs. ID-cache updates only track null-namespace attributes so namespaced IDs cannot disrupt form associations.

The `form.elements` and `checkValidity()` rows include an unrelated DOM mutation before each call so the live collection must refresh. `FormData` rebuilds its control list on every construction.

| Workload, 100 controls and 10,000 outside elements | `main` | This PR | Speedup |
|---|---:|---:|---:|
| `form.elements`, no ID | 294.8 μs | 8.28 μs | 35.6× |
| `checkValidity()`, no ID | 343.7 μs | 50.6 μs | 6.8× |
| `new FormData(form)`, no ID | 315.9 μs | 33.4 μs | 9.5× |
| `form.elements`, external controls | 34.48 ms | 0.309 ms | 111.5× |

The benchmark also covers one-control forms and 0, 1,000, and 10,000 outside elements. Small forms were neutral, and the new subtree path was neutral for identified forms.

### DIFF
--- a/benchmark/forms/form-owner.js
+++ b/benchmark/forms/form-owner.js
@@ -4,6 +4,8 @@ const { Bench } = require("tinybench");
 const { JSDOM } = require("../..");
 
 const CONTROL_COUNT = 120;
+const FORM_CONTROL_COUNTS = [1, 100];
+const OUTSIDE_ELEMENT_COUNTS = [0, 1000, 10000];
 
 module.exports = () => {
   const bench = new Bench();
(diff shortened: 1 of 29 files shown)


domenic (approved):
I deduplicated the use-getElementById logic that has appeared in a couple places, which also fixed a bug with detached datalists. I then updated _attrModified to take the namespace, so that we could make the code cleaner in a lot of places. But that, of course, revealed that we'd had incorrect behavior for lots of attribute mutations, so I fixed that with a WPT.

Very happy with our combined work here!

PR #4276, opened 2026-09-04
Remove closed windows from storage event dispatch

scttcper:
This doesn't really affect me since i never render this sort of thing in jsdom.

Same-origin Windows are kept in a shared list for storage event dispatch. `window.close()` tears down the Window but leaves it in that list, retaining removed iframe Windows and continuing to target them for storage events.

Remove the Window from the list during `window.close()`, which covers both iframe removal and navigation.

| 100 iframe create/remove | Main | This PR |
|---|---:|---:|
| Retained heap | 131.85 MB | 41.48 MB |
| Removal time | 9.96 ms | 9.77 ms |

### DIFF
--- a/lib/jsdom/browser/Window.js
+++ b/lib/jsdom/browser/Window.js
@@ -834,6 +834,12 @@ function installOwnProperties(window, options) {
       window[i].close?.();
     }
 
+    const { windowsInSameOrigin } = window._currentOriginData;
+    const idx = windowsInSameOrigin.indexOf(window);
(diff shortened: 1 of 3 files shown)


domenic (approved):
Great catch, thank you! I wonder how many of jsdom's [historical memory issues](https://github.com/jsdom/jsdom/issues?q=is%3Aissue%20state%3Aopen%20label%3Amemory) are explained by something like this.

PR #4278, opened 2026-09-04
Use cached roots for event dispatch

scttcper:
Building an event path repeatedly walks to the same Document for each connected parent, even though jsdom already caches that root.

Use `_cachedRoot` for that case. Other roots keep the existing traversal.

Existing event-target benchmark, median time per dispatch (main → this PR):

| Depth | No listeners | Delegated listener | Listener on every node |
| ---: | ---: | ---: | ---: |
| 10 | 2.79 → 1.25 μs | 3.08 → 1.38 μs | 4.46 → 2.79 μs |
| 50 | 30.58 → 4.13 μs | 30.92 → 4.25 μs | 37.75 → 11.25 μs |
| 100 | 108.83 → 7.75 μs | 109.33 → 7.92 μs | 125.00 → 23.42 μs |

A Testing Library `fireEvent.click()` benchmark improved by 26-90% across depths 10-100.

### DIFF
--- a/benchmark/dom/event-target.js
+++ b/benchmark/dom/event-target.js
@@ -2,7 +2,7 @@
 const { Bench } = require("tinybench");
 const { JSDOM } = require("../..");
 
-const depths = [10, 25, 50, 100];
+const depths = [10, 50, 100];
 const listenerPatterns = ["none", "delegated", "every tree node"];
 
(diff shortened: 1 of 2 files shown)


domenic (approved):
Embarrassingly simple, and a big win; I love it!

PR #4279, opened 2026-09-04
Avoid custom element unused registry work and allocations

scttcper:
React tests that register custom elements can still render mostly ordinary HTML. Skip registry lookups when no definition can apply, and allocate custom-element reaction queues only when needed.

The registry guards reduced a 400-row ordinary UI mount benchmark from 12.60 ms to 10.95 ms with 100 registered definitions. Skipping the additional lookup when there is no `is` value reduced undefined-custom-element creation and reconnection time by another 28% and 35% with 500 definitions, compared with the guards alone.

Lazy reaction queues saved 325 KB (1.2%) of retained heap in a 2,000-row, 10,001-element form-like fixture. CPU measurements were roughly neutral across five alternating pairs.

### DIFF
--- a/benchmark/dom/custom-elements.js
+++ b/benchmark/dom/custom-elements.js
@@ -1,0 +1,125 @@
+"use strict";
+const { Bench } = require("tinybench");
+const { JSDOM } = require("../..");
+
+const ELEMENTS = 1000;
+const DEFINITIONS = 500;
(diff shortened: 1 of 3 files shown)


domenic (approved):
Cleaned up some of the nearby code that looked strange at the same time.

PR #4280, opened 2026-09-05
Avoid installing Window event handlers twice

scttcper:
Window construction installs 106 `on*` accessors, then immediately replaces them with conversion-aware accessors. Install the final accessors first so the remaining mixin only copies helper methods.

Median times for 1,000 environments across seven alternating runs:

| Benchmark | Before | After | Reduction |
|---|---:|---:|---:|
| Handler installation | 170 ms | 133 ms | 21.9% |
| Create and close | 1,698 ms | 1,638 ms | 3.5% |
| Create and close, `outside-only` | 2,129 ms | 2,070 ms | 2.7% |

Handler installation was timed separately; full create/close runs used uninstrumented code.

### DIFF
--- a/benchmark/api/constructor.js
+++ b/benchmark/api/constructor.js
@@ -21,5 +21,15 @@ module.exports = () => {
     new JSDOM(html);
   });
 
+  bench.add("new JSDOM() and close", () => {
+    const dom = new JSDOM();
+    dom.window.close();
+  });
+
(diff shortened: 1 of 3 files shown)


domenic (commented):
Your optimization was good but it was too confusing why it worked, relying on the skipping behavior of `mixin()` and using `GlobalEventHandlersImpl` only for its helpers. I had Codex do another round of tightening things up, while preserving the core idea.

PR #4281, opened 2026-09-05
Defer SymbolTree bookkeeping allocation

scttcper:
Defer SymbolTree bookkeeping until a node uses it. Attributes inherit from Node but usually don't need tree bookkeeping; initializing the symbol slot to null keeps the property in place without allocating the object.

For a React/Testing Library form with 2,000 rows and 28,000 attributes:

| Incremental mounted heap after GC | Before | After |
| --- | --- | --- |
| Retained heap | 45.91 MB | 42.77 MB (−6.84%) |

The memory reduction repeated in two rounds of seven alternating fresh-process pairs. Mount and rerender CPU were slightly lower; unmount was noisy, with no repeatable CPU regression.

### DIFF
--- a/lib/jsdom/living/nodes/Node-impl.js
+++ b/lib/jsdom/living/nodes/Node-impl.js
@@ -136,7 +136,7 @@ class NodeImpl extends EventTargetImpl {
   constructor(globalObject, args, privateData) {
     super(globalObject, args, privateData);
 
-    domSymbolTree.initialize(this);
+    this[domSymbolTree.symbol] = null;
 
     this._ownerDocument = privateData.ownerDocument;
 


domenic:
I wonder whether this should go in https://github.com/jsdom/js-symbol-tree instead, but then I looked at that library and was confronted with how old and old-style it is, and now I'm wondering about whether we should just merge it into the jsdom/jsdom codebase... Probably the right move is to just merge this as-is and shave that yak later.

(I'll do a batch of merging sometime in the next few days, but I'm trying to catch up on other errands today and maybe tomorrow. Commenting is all I'm allowing myself.)

scttcper:
@domenic no worries just trying out astra on some of these and burning my leftover tokens for the week. There's no rush, this one line change was pretty neat.

PR #4283, opened 2026-09-05
fix: keep document.hasFocus() true while focus moves within a document

### LINKED ISSUE #3794: document.hasFocus should not return false in a blur or focus event
### Basic info:

- **Node.js version:** any
- **jsdom version:** all

### Minimal reproduction case

```js

const {
  window: { document },
} = new JSDOM(`
  <input id="one" />
  <input id="two" />
`);

document.addEventListener(
  "blur",
  () => {
    console.log("blur", document.hasFocus());
  },
  {
    capture: true,
  }
);
document.addEventListener(
  "focus",
  () => console.log("focus", document.hasFocus()),
  {
    capture: true,
  }
);

document.getElementById("one").focus();
document.getElementById("two").focus();
```js

Result should be

```
focus true
blur true
focus true
```

But actually is

```
focus true
blur false
focus true
```

https://output.jsbin.com/tepozoj

Open browser console and press "test"

Tab between the inputs.  As long as you don't tab outside the document window into the browser Chrome then `document.hasFocus()` is always true even if the code is run in an event listener.

### Notes

I believe this issue is here:
https://github.com/jsdom/jsdom/blob/main/lib/jsdom/living/nodes/Document-impl.js#L276
https://github.com/jsdom/jsdom/blob/main/lib/jsdom/living/nodes/HTMLOrSVGElement-impl.js#L52

### PR DESCRIPTION
**Pull request by nirmal-shaji**
Fixes #3794.

`focus()` cleared `ownerDocument._lastFocusedElement` before dispatching the `blur`/`focusout` events on the previously focused element. Since `document.hasFocus()` is `Boolean(_lastFocusedElement)` (and `activeElement` derives from the same field), both reported the document had lost focus inside a `blur`/`focusout` handler when focus was merely moving between two elements of the same document:

```js
const { window: { document } } = new JSDOM(`<input id="one" /><input id="two" />`);
document.addEventListener("blur", () => console.log("blur", document.hasFocus()), { capture: true });
document.addEventListener("focus", () => console.log("focus", document.hasFocus()), { capture: true });
document.getElementById("one").focus();
document.getElementById("two").focus();
// before: focus true / blur false / focus true — after: focus true / blur true / focus true
```

Per the [focus update steps](https://html.spec.whatwg.org/multipage/interaction.html#focus-update-steps) the currently focused area only changes to the new target after the unfocusing steps run, so `_lastFocusedElement` should stay pointing at the previous element until `blur`/`focusout` have fired. `activeElement` now also correctly reports the element being blurred inside its own handler; the iframe path and `blur()` (a genuine focus drop) are unchanged.

Added a `test/api` regression test covering `hasFocus()` during the transfer, `activeElement` inside the blur handler, and `hasFocus()` becoming false after the sole focused element is blurred — it fails before and passes after. `npm run test:api` passes (580).

### DIFF
--- a/lib/jsdom/living/nodes/HTMLOrSVGElement-impl.js
+++ b/lib/jsdom/living/nodes/HTMLOrSVGElement-impl.js
@@ -52,8 +52,12 @@ class HTMLOrSVGElementImpl {
       return;
     }
 
-    ownerDocument._lastFocusedElement = null;
     if (previous) {
+      // Per the focus update steps, the currently focused area of the document is only changed to the
+      // new focus target after the unfocusing steps (the blur/focusout events) have run. Keeping
+      // _lastFocusedElement pointing at `previous` until then means that, while focus transfers
+      // between two elements of the same document, document.hasFocus() stays true and
+      // document.activeElement still reflects the element being blurred inside the blur handler.
       focusing.fireFocusEventWithTargetAdjustment("blur", previous, this);
       focusing.fireFocusEventWithTargetAdjustment("focusout", previous, this, { bubbles: true });
     } else {
--- a/test/api/focus-hasfocus.js
+++ b/test/api/focus-hasfocus.js
@@ -1,0 +1,54 @@
+"use strict";
+const assert = require("node:assert/strict");
+const { describe, it } = require("mocha-sugar-free");
+
+const { JSDOM } = require("../..");
+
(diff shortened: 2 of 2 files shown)


nirmal-shaji:
Closing this. The approach changes activeElement timing during blur and breaks a focus web-platform-test (focus-events.html). The hasFocus() bug is real; I'll open a corrected fix once it's verified against the full WPT suite. Sorry for the noise.

PR #4286, opened 2026-09-07
Cache serialized style declarations

scttcper:
Cache `cssText` until a property or priority changes, so style setters can reuse it when comparing and updating the style attribute.

Compared with main at 4a65f353 on Node 26.8.1:

| Benchmark | Before | After |
| --- | ---: | ---: |
| Style properties, 20 elements | 2.65 ms | 1.43 ms |
| Base UI hidden-input styles, 20 elements | 4.55 ms | 2.13 ms |
| Base UI form mount + unmount, run 1 | 3.15 ms | 2.45 ms |
| Base UI form mount + unmount, run 2 | 2.95 ms | 2.49 ms |

### DIFF
--- a/benchmark/style/style-property.js
+++ b/benchmark/style/style-property.js
@@ -42,5 +42,22 @@ module.exports = () => {
     }
   });
 
+  bench.add("createElement + visually hidden input styles", () => {
+    for (let i = 0; i < 20; i++) {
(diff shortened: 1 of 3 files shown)


domenic (approved):
I could swear we had something like this already, but apparently not! Nice one.

PR #4287, opened 2026-09-07
Reuse border definitions during serialization

scttcper:
Use the existing border property definitions to remove redundant longhands during serialization, instead of rebuilding property names on every call.

Base UI form mount + unmount, Node 26.8.1:

| Run | #4286 | + this change |
| --- | ---: | ---: |
| 1 | 2.42 ms | 2.20 ms |
| 2 | 2.55 ms | 2.29 ms |


domenic (approved):
I'm slightly skeptical that those small wins are statistically significant, but the code is simpler anyway, so why not.

PR #4288, opened 2026-09-07
Skip recursive traversal generators for leaves

scttcper:
Skip recursive generators for leaf nodes during DOM traversal. Found while profiling a Base UI form; subtree moves improved, but the full form showed no consistent gain.

100-field subtrees on Node 26.8.1:

| Subtree | Main | This change |
| --- | ---: | ---: |
| Detached, light DOM | 0.161 ms | 0.141 ms |
| Connected, light DOM | 0.232 ms | 0.200 ms |
| Detached, shadow DOM | 0.148 ms | 0.128 ms |
| Connected, shadow DOM | 0.192 ms | 0.162 ms |

### DIFF
--- a/benchmark/dom/tree-modification.js
+++ b/benchmark/dom/tree-modification.js
@@ -68,5 +68,27 @@ module.exports = () => {
     }
   });
 
+  for (const count of [1, 100]) {
+    for (const shadow of [false, true]) {
+      for (const connected of [false, true]) {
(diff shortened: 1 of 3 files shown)


domenic (approved):
I didn't love the original change since it felt like it was just special-casing the leaf node case to work around V8 generators not being fully optimized. Codex found another version that it claims is faster and feels less bad to me.

PR #4289, opened 2026-09-07
Keep attribute lookups correct after prefix changes

scttcper:
Fix stale attribute lookups when `setAttributeNodeNS()` replaces an attribute with a different prefix, preserving other attributes with the same name and their order.


domenic (approved):
Nice find!!

You may enjoy https://annevankesteren.nl/2014/08/attributes on this topic.

PR #4290, opened 2026-09-07
Cache repeated attribute query results

scttcper:
Cache matched elements for repeated `querySelectorAll()` calls with simple attribute selectors, such as `[data-testid]` and `[data-testid="target"]`. This avoids rescanning the subtree between repeated Testing Library lookups.

Reuse the existing query-cache invalidation for tree changes, attribute changes, and document adoption, while returning a fresh static NodeList on every call. Limit the cache to 16 selectors of at most 256 characters per root. Other selectors keep using the normal query path.

Testing Library `getByTestId()` benchmarks, with a mutation before each batch of lookups for different IDs. Both builds include our other pending PRs; this measures adding the cache to that build, not against main alone. Times are µs, median of three runs' mean times; lower is better.

| Rows | Lookups per mutation | Without cache | With cache | Speedup |
| ---: | ---: | ---: | ---: | ---: |
| 20 | 1 | 14.29 | 14.40 | 0.99× |
| 20 | 5 | 66.79 | 46.40 | 1.44× |
| 20 | 20 | 261.01 | 162.67 | 1.60× |
| 250 | 1 | 156.75 | 156.89 | 1.00× |
| 250 | 5 | 756.23 | 524.51 | 1.44× |
| 250 | 20 | 2997.37 | 1933.22 | 1.55× |

There is still a miss cost: a separate controlled benchmark with one raw query pair after each mutation went from 288.48 to 298.50 µs, about 3.5% slower. One run of the broader mutation benchmark also had latency spikes, so tail latency needs more investigation.

### DIFF
--- a/benchmark/dom/memoized-queries.js
+++ b/benchmark/dom/memoized-queries.js
@@ -53,5 +53,51 @@ module.exports = () => {
     queryRoot.getElementsByTagName("a");
   });
 
+  for (const size of [20, 250]) {
+    const container = document.createElement("div");
+    container.innerHTML = Array.from({ length: size }, (_, i) => {
+      return `<div data-testid="row-${i}"><span></span></div>`;
+    }).join("");
+    const last = container.lastElementChild;
+    const selectors = ["[data-testid]", `[data-testid="row-${size - 1}"]`, '[data-testid="missing"]'];
+    const options = {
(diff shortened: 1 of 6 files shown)


domenic:
Feels like this would belong in our domselector dependency?

PR #4292, opened 2026-09-07
Guard background parsing and normalize whitespace handling

### LINKED ISSUE #4285: Crash in replaceBackgroundShorthand when a url() token is not separated by whitespace
### Node.js version

v24.19.0

### jsdom version

30.0.1

### Minimal reproduction case

```js

// Throws. Any declaration after the `background` shorthand triggers it.
new JSDOM('<div style="background:url(a.png)no-repeat;color:red"></div>');

// Adding a space after url(a.png) avoids the crash.
```

Chrome 148 accepts the style and normalizes `background` to `url("a.png") no-repeat`.

### What is the problem?

Parsing the HTML throws:

```text
TypeError: Cannot read properties of undefined (reading 'length')
    at replaceBackgroundShorthand (jsdom/lib/jsdom/living/css/helpers/shorthand-properties.js:222:29)
    at prepareProperties (jsdom/lib/jsdom/living/css/helpers/shorthand-properties.js:1459:42)
    at set cssText (jsdom/lib/jsdom/living/css/CSSStyleDeclaration-impl.js:146:32)
    at HTMLElementImpl._attrModified (jsdom/lib/jsdom/living/nodes/HTMLElement-impl.js:175:26)
```

We found this while processing incoming email containing inline styles generated by Outlook.

asamuzaK:
Fix #4285 

Fixed 2 issues:
* Should not throw, but return an empty string if the CSS is invalid.
* Gracefully handle edge cases like missing whitespace (e.g., url()no-repeat).

### DIFF
--- a/lib/jsdom/living/css/properties/background.js
+++ b/lib/jsdom/living/css/properties/background.js
@@ -280,9 +280,11 @@ function parse(v) {
   } else if (parsers.hasCalcFunc(v)) {
     v = parsers.resolveCalc(v);
   }
-  if (/\)(?!\)|\s|,)/.test(v)) {
-    v = parsers.normalizeValue(v);
+  const prepared = parsers.prepareShorthandValue(v);
+  if (!prepared) {
+    return undefined;
   }
+  v = prepared;
   if (!parsers.isValidPropertyValue(property, v)) {
     return undefined;
   }
--- a/lib/jsdom/living/css/properties/background.js
+++ b/lib/jsdom/living/css/properties/background.js
@@ -280,12 +280,12 @@ function parse(v) {
   } else if (parsers.hasCalcFunc(v)) {
     v = parsers.resolveCalc(v);
   }
+  if (/\)(?!\)|\s|,)/.test(v)) {
+    v = parsers.normalizeValue(v);
+  }
   if (!parsers.isValidPropertyValue(property, v)) {
     return undefined;
   }
-  // Normalize whitespace after function closing parentheses ')'.
-  // Other boundaries like ',' and '/' are explicitly handled as delimiters.
-  v = v.replace(/\)(?!\)|\s|,)/g, ") ");
   const values = parsers.splitValue(v, {
     delimiter: ","
   });
--- a/lib/jsdom/living/css/properties/background.js
+++ b/lib/jsdom/living/css/properties/background.js
@@ -284,11 +284,10 @@ function parse(v) {
   if (!prepared) {
     return undefined;
   }
-  v = prepared;
-  if (!parsers.isValidPropertyValue(property, v)) {
+  if (!parsers.isValidPropertyValue(property, prepared)) {
     return undefined;
   }
-  const values = parsers.splitValue(v, {
+  const values = parsers.splitValue(prepared, {
     delimiter: ","
   });
   const bgValues = [];
--- a/lib/jsdom/living/css/properties/background.js
+++ b/lib/jsdom/living/css/properties/background.js
@@ -283,6 +283,9 @@ function parse(v) {
   if (!parsers.isValidPropertyValue(property, v)) {
     return undefined;
   }
+  // Normalize whitespace after function closing parentheses ')'.
+  // Other boundaries like ',' and '/' are explicitly handled as delimiters.
+  v = v.replace(/\)(?!\)|\s|,)/g, ") ");
   const values = parsers.splitValue(v, {
     delimiter: ","
   });
--- a/test/web-platform-tests/to-upstream/css/css-backgrounds/background-no-whitespace.html
+++ b/test/web-platform-tests/to-upstream/css/css-backgrounds/background-no-whitespace.html
@@ -28,7 +28,8 @@ test(() => {
   );
 
   assert_equals(
-    div.style.backgroundImage, 'url("a.png")',
+    div.style.backgroundImage,
+    'url("a.png")',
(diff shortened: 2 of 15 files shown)


test/web-platform-tests/to-upstream/css/css-backgrounds/background-no-whitespace.html
```diff
@@ -0,0 +1,40 @@
...
+  );
+
+  assert_equals(
+    div.style.color,
+    "red",
+    "The color property should be parsed and set despite the missing space in background."
+  );
+
+  assert_true(
+    div.style.backgroundImage.includes("a.png"),
```
domenic: Can we just make this an assert_equals?

lib/jsdom/living/css/properties/background.js
```diff
@@ -283,6 +283,9 @@ function parse(v) {
   if (!parsers.isValidPropertyValue(property, v)) {
     return undefined;
   }
+  // Normalize whitespace after function closing parentheses ')'.
+  // Other boundaries like ',' and '/' are explicitly handled as delimiters.
+  v = v.replace(/\)(?!\)|\s|,)/g, ") ");
```
domenic: This also matches `background: url("a)b.png") no-repeat`, thus changing it to `background: url("a) b.png") no-repeat`, which is pretty bad.

test/web-platform-tests/to-upstream/css/css-backgrounds/background-no-whitespace.html
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Background shorthand without whitespace</title>
+<link rel="help" href="https://drafts.csswg.org/css-backgrounds/#background">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<!-- regression test for https://github.com/jsdom/jsdom/issues/4285 -->
+
+<div id="div" style="background:url(a.png)no-repeat;color:red"></div>
```
domenic: Can we also test `url(a.png)no-repeat` and `url("a)b.png")no-repeat`?

domenic (commented):
ChatGPT says `border:rgb(0 0 0)solid 1px;color:red` crashes in a similar way...

asamuzaK:
I'll dig deeper.
Thanks for your quick review.

test/web-platform-tests/to-upstream/css/css-backgrounds/background-no-whitespace.html
asamuzaK: Done

lib/jsdom/living/css/properties/background.js
asamuzaK: Added normalizeValue() instead.

asamuzaK:
Re: background-no-whitespace.html
* Test 2 is an issue regarding error recovery when reaching the end of input with an escape character.
* Test 4 is an issue regarding how HTML entities should be handled during CSS parsing.

Since cases 2 and 4 aren't directly related to this issue, would it be okay to leave them unaddressed and keep them out of scope for this PR for now?
I'll do some further investigation and submit a follow-up PR later.

What do you think?

domenic:
Thanks for the work here! This direction looks better, but it has a few regressions.

The issue is that `normalizeValue()` still applies the same character-level regexp to css-tree's serialized output. css-tree can serialize a quoted URL containing `)` as an unquoted URL containing an escaped `)`, which the regular expression then mistakes for a function-closing delimiter. Thus, setting `style.background = 'url("a)b.png") no-repeat'` will fail, which is a regression compared to origin/main.

I'd encourage using as few regexps as possible (ideally zero), and instead operating on AST nodes where possible.

Normalization also happens before validation and can throw. On this branch, both `style.background = ")x"` and `style.border = ")x"` throw `SyntaxError`; origin/main correctly ignores these invalid assignments.

Minor: `processBorderProperties()` returns the empty string on its failure path, which is bad for a function that normally returns a `Map`.

> Since cases 2 and 4 aren't directly related to this issue, would it be okay to leave them unaddressed and keep them out of scope for this PR for now?

In general I am OK with leaving any cases unaddressed, as long as we avoid regressions. Incremental improvements are always welcome.

I had ChatGPT 5.6 Sol Extra High expand the tests. It came up with these: https://gist.github.com/domenic/277219f3608a49de59286b225055a2c6 . For this PR, I will be happy with any change that is a strict progression. That is, all subtests that currently pass on origin/main, need to keep passing.

asamuzaK:
Thank you for the detailed feedback and test cases.
I made a single adjustment to the test cases to align with the spec.
See https://drafts.csswg.org/css-cascade/#shorthand
See https://issues.chromium.org/issues/40423296 (a 10-year-old bug...)
> When values are omitted from a shorthand form, unless otherwise defined, each “missing” sub-property is assigned its initial value.

``` javascript
// Chrome returns "initial".
assert_equals(div.style.backgroundRepeat, "repeat", "backgroundRepeat");
```

domenic:
Pushed a couple more commits, one to fix not using the `/u` flag, and one to replace the hand-crafted tokenizer with css-tree's tokenizer. WDYT?

asamuzaK:
It's so much better than my pseudo-tokenization attempt.
I didn't think of looping through the children to handle it. 
Awesome work!

PR #4296, opened 2026-09-08
Avoid intermediate style attribute updates

scttcper:
Generally been avoiding css changes since i saw a big pr with a revamp of css coming. This one seemed interesting enough. Even if pendingStyleUpdate isn't very good using a set instead of a map here is pretty interesting.

Setting a shorthand can update and serialize the style attribute several times as its longhands change. This batches those changes within each property assignment and updates the attribute once with the completed declaration, avoiding intermediate mutation records and repeated serialization.

Also avoids allocating temporary property objects during `cssText` serialization, reading values and priorities after redundant properties have been removed. Adds mutation observer coverage for shorthand and longhand assignments, removals, unchanged values, and priority changes.

Results from `node --run benchmark -- --suite style/style-property --format markdown`, comparing main with this branch using the same dependencies. Each operation creates 20 elements; times are averaged across two runs with the order reversed for the second run.

| Benchmark | Before | After | Speedup |
|---|---:|---:|---:|
| `innerHTML`, no styles | 0.116 ms | 0.116 ms | 1.00× |
| `innerHTML`, inline styles | 1.949 ms | 1.934 ms | 1.01× |
| `createElement` + style properties | 0.767 ms | 0.277 ms | 2.77× |
| `createElement` + `setAttribute("style")` | 1.793 ms | 1.760 ms | 1.02× |
| `createElement` + `style.cssText` | 1.819 ms | 1.829 ms | 0.99× |

### DIFF
--- a/lib/jsdom/living/css/CSSStyleDeclaration-impl.js
+++ b/lib/jsdom/living/css/CSSStyleDeclaration-impl.js
@@ -32,6 +32,7 @@ class CSSStyleDeclarationImpl {
   parentRule;
   #ownerNode;
   #updating = false;
+  #pendingStyleUpdate = false;
 
   // Internal private fields.
   #computedValueOpts = new Map();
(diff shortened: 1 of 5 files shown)


domenic (approved):
Really nice; perf improvements, a simplification, and correctness improvements all at once.

I asked GPT-6 Astra High to stress test and it had the clever idea of also using custom element reactions. But all those tests passed to, so, I'll just add them to this PR and merge.

Issue #4291, opened 2026-09-07
Consider merging dom-symbol-tree into jsdom


domenic:
Some discussion in https://github.com/jsdom/jsdom/pull/4281. Now that we have AI, this should be pretty easy. We can aim it at creating a solution that gives the best performance and simplest code.

domenic:
**AI-generated: ChatGPT 5.6 Sol (Extra High).**

One lesson from #4288 is that the current jsdom–SymbolTree abstraction boundary makes it easy to turn a single traversal into per-node traversal machinery. The original idiomatic recursive generator created a generator and a SymbolTree child iterator for essentially every visited node.

An explicit-stack implementation was fastest, but made a simple preorder traversal significantly harder to read. The better compromise was to use SymbolTree's `treeIterator()` for each ordinary node tree and recurse only when crossing into a shadow root. That retained most of the performance improvement while expressing the DOM algorithm directly as a generator.

Ideally, a merged and redesigned tree layer would make the clean expression the fast one. Some implications:

- The tree layer should expose DOM-shaped traversals, especially shadow-including preorder, rather than requiring jsdom to splice generic child and light-tree iterators together. Making it aware of the host-to-shadow-root edge would let it provide shadow-including inclusive and exclusive traversal directly.
- Traversal state should be allocated once per traversal, or proportional to tree depth or nested shadow trees, rather than once per visited node.
- Mutation semantics should be explicit. `next()` on current `TreeIterator` instances computes the following node before returning the current one, while the old recursive helper captured the next sibling early but inspected the current node's descendants after yielding it. That mixed behavior is accidental. Separate stable/static, live, and snapshot traversal APIs might be appropriate.
- Iterable and visitor forms are both worth benchmarking. All nine current consumers of the shadow-including iterators exhaust them, so a visitor can avoid generator delegation and iterator-result overhead. Generators remain valuable when laziness or early termination matters.
- We should preserve #4281's lazy-bookkeeping win. Integrating tree storage into jsdom could make parent, sibling, child, root, and shadow relationships cheaper to access, but attributes and other nodes that never participate in a tree should not pay for full bookkeeping.

I would avoid mechanically copying the existing package into jsdom. The opportunity is to design DOM-specific primitives around the operations jsdom actually performs: light-tree and shadow-including preorder, ancestor and root queries, subtree moves, and mutation-aware traversal.

[...]

PR #4298, opened 2026-09-09
Avoid a second ancestor walk during node insertion

scttcper:
`isHostInclusiveAncestor` walks the ancestors, then looks up the root separately. Keep the root from the first walk so both checks share one traversal.

somewhat related to #4291


domenic (approved):
Excellent and simple!

PR #4300, opened 2026-09-10
Avoid reparsing cached computed styles

scttcper:
Copy cached computed-style declarations without resolving and reparsing every property.

Also fixes system-color shorthand resolution so first and repeated reads return resolved border colors.

For a 20-field React form mounting, updating three times, querying after each render, and unmounting:

| Queries | Before | After | Less elapsed time |
| --- | ---: | ---: | ---: |
| Role | 62.88 ms | 57.71 ms | 8.2% |
| Label + role + test ID | 69.25 ms | 63.56 ms | 8.2% |

jsdom's existing `style/get-computed-style-property-access` benchmark measures repeated computed-style calls followed by property reads. Using the same commits and five alternating process pairs:

| Property reads | Before | After | Less elapsed time |
| --- | ---: | ---: | ---: |
| 4 properties, including color | 5.763 µs | 3.913 µs | 32.1% |
| 4 properties without color or length resolution | 4.301 µs | 2.360 µs | 45.1% |
| 30 properties | 23.303 µs | 20.720 µs | 11.1% |

### DIFF
--- a/lib/jsdom/living/css/CSSStyleDeclaration-impl.js
+++ b/lib/jsdom/living/css/CSSStyleDeclaration-impl.js
@@ -46,6 +46,15 @@ class CSSStyleDeclarationImpl {
     this.#ownerNode = ownerNode || null;
   }
 
+  _copyDeclarationsFrom(source) {
+    for (const [property, value] of source.#values) {
+      this.#values.set(property, value);
+    }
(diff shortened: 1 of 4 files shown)


domenic (approved):
Nice nice. Added some more WPTs for the bugs you fixed, which Codex says were not 100% covered.

PR #4307, opened 2026-09-12
Parse numeric '0' as '0px' for flex-basis

### LINKED ISSUE #4305: CSS flex shorthand rejects a valid unitless zero basis
### Node.js version

24.15.0

### jsdom version

30.0.1

### Minimal reproduction case

```js

const dom = new JSDOM("<div id=panel></div>");
const panel = dom.window.document.querySelector("#panel");

panel.style.flex = "35 1 0";

console.log({
  node: process.version,
  flex: panel.style.flex,
  cssText: panel.style.cssText,
});
```

Chromium accepts the declaration and serializes it with an explicit zero length:

```js
const panel = document.createElement("div");
panel.style.flex = "35 1 0";
console.log(panel.style.flex, panel.style.cssText);
```

The browser retains the declaration as equivalent to `flex: 35 1 0px;`.

### What is the problem?

I expected jsdom to retain `flex: 35 1 0`, as Chromium and jsdom 26.1.0 do. Instead, jsdom 30.0.1 rejects the declaration and leaves both `style.flex` and `style.cssText` empty.

`flex: 35 1 0` is valid Flexbox shorthand because the `flex-basis` grammar permits a unitless zero. The relevant specification is the [CSS Flexible Box Layout Module](https://www.w3.org/TR/css-flexbox-1/).

I found related upstream work in [cssstyle#245](https://github.com/jsdom/cssstyle/pull/245), which fixes a flex-basis parsing case, and [jsdom#3933](https://github.com/jsdom/jsdom/pull/3933), which adds a regression test. Both pull requests are merged, but this exact reproduction still fails in jsdom 30.0.1. Could you confirm whether this is a separate case or an unreleased fix? I am happy to provide any additional information that would help.

asamuzaK:
Fix #4305

### DIFF
--- a/lib/jsdom/living/css/properties/flex.js
+++ b/lib/jsdom/living/css/properties/flex.js
@@ -136,6 +136,8 @@ function parse(v) {
           flex[flexBasis.property] = `${val3.value}${val3.unit}`;
         } else if (val3.type === AST_TYPES.PERCENTAGE) {
           flex[flexBasis.property] = `${val3.value}%`;
+        } else if (val3.type === AST_TYPES.NUMBER && val3.value === "0") {
+          flex[flexBasis.property] = `${val3.value}px`;
         } else {
           return undefined;
         }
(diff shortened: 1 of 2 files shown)


domenic (approved):
Fixed to cover other spellings of 0 (e.g. `0.0`, `-0`, etc.) and added more tests. WDYT?

asamuzaK:
Thanks!

PR #4308, opened 2026-09-14
Replace SymbolTree with a DOM-specific tree

### LINKED ISSUE #4291: Consider merging dom-symbol-tree into jsdom

scttcper:
This came out of profiling jsdom in my DOM benchmark suite.

Replace SymbolTree with a DOM-specific tree stored in `NodeImpl._links`. Keep link bookkeeping in the tree helper and expose traversal and tree comparisons as methods on nodes.

Allocate links only when needed, preserve sibling indexes before mutations, and calculate mutation indexes only when live Ranges need them. Walk shadow trees with an explicit stack while preserving behavior when callbacks change the tree.

Limit Range stringification and deletion to the selected content, and return immediately when stringifying a collapsed Range. Reuse common-ancestor results in Range operations. Skip descendants of nodes being deleted, and check the two partially contained children in `surroundContents()` instead of scanning unrelated content.

A few places we looked for inspiration:

- [Gecko’s shadow iterator](https://github.com/mozilla-firefox/firefox/blob/main/dom/base/ShadowIncludingTreeIterator.h): use a stack instead of recursive generators.
- [Gecko’s child-index cache](https://github.com/mozilla-firefox/firefox/blob/main/dom/base/nsINode.cpp): keep indexes for siblings before the mutation instead of invalidating them all.
- [WebKit’s traversal helpers](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/dom/NodeTraversal.h): walk directly through child, sibling, and parent pointers.
- [Blink’s compact sibling links](https://chromium.googlesource.com/chromium/src/+/main/third_party/blink/renderer/core/dom/container_node.h): tried smaller node records. Mutations were slower in JavaScript, so we kept a separate pointer to the last child.

Execution time measured at [39edbb73](https://github.com/jsdom/jsdom/commit/39edbb738f05a15c412afba5ae6f0a543fe2580e) versus the [SymbolTree base](https://github.com/jsdom/jsdom/commit/33e4fa72b9f44ae7b30e5b27c5b9120604b5045f), measured on Node 26.8.2 in two paired runs. Percentages combine both runs; cloning varied enough to show the observed range.

A live Range is a DOM `Range` whose start and end positions automatically update when the tree changes. Keeping those positions up to date adds work to mutations.

| Benchmark | Change |
|---|---:|
| Repeated middle insert/remove, without live Ranges | 94% less time |
| Repeated middle insert/remove, with a live Range | 93% less time |
| Random sibling reordering with a live Range | 61% less time |
| Adopt a 100-field subtree | 39% less time |
| Clone a 100-field subtree | Mixed: 17% less to 5% more time |
| Shallow clone | 7% less time |
| `customElements.upgrade()`, flat/deep/field-heavy trees | 42–64% less time |
| Element child/sibling getters, adjacent elements | 46–49% less time |
| Element child/sibling getters, mixed text, comments, and elements | 52–56% less time |
| Element child/sibling getters, no matching elements | 63–66% less time |
| Range common ancestor, deep branches and shared ancestors | 92–99.9% less time |
| `Range.deleteContents()`, fully contained deep subtree | 99.4% less time |
| `Range.surroundContents()`, adjacent text with 1,000 unrelated siblings | 99% less time |

Fixes #4291


domenic (commented):
This looks very promising. I haven't gone into full depth, but I think we could do better from a code readability perspective. To me this looks too much like the old shape.

For example:

- I don't think we need to use a symbol; we can just use `node._links` or similar.
- I don't like the `internal-constants.js` file or the various indirections like threading it through `shadow-dom.js`.
- At the very least, we should change helper calls from e.g. `domSymbolTree.index(childImpl)` to `treeHelpers.index(childImpl)` or `treeOps.index(childImpl)`. But I think it'd be nicer if we could do something like `childImpl._treeIndex()`.

I'd also prefer to avoid adding unit tests of internals in test/api. if we think they're really valuable, we can consider a new test/unit/ or something, but I'm hopeful the coverage isn't needed since WPTs cover everything exhaustively.

scttcper:
@domenic its a really interesting area! Looking around for inspiration that already exists was pretty neat.

Switched to _links, removed the import indirection, and moved index calculation onto NodeImpl. Removed the internal tests and kept the useful coverage in WPTs.

domenic:
I set Astra Extra High loose on this and it found more cleanups and performance improvements, especially in ranges.

I'm going to move github.com/jsdom/jsdom/pull/4308/commits/1feedc0cd62401ac6aceafbb194497606e840b2a to another PR, along with the same fix for processing instructions, then rebase this on that...

scttcper:
Also using astra xhigh

I ran the benchmarks before and after your commits on Node 26:

| Benchmark | Less time |
|---|---:|
| Common ancestor in a deeply nested tree | 99.9% |
| Delete a fully contained deep subtree | 97% |
| Surround text with 1,000 unrelated siblings | 97% |

Pushed a follow-up to skip unrelated subtrees in `toString()` and `deleteContents()` too.

PR #4313, opened 2026-09-15
Apply nested rules in imported style sheets

domenic:
Fix missing computed styles from nested imports and media rules in imported stylesheets, found during my script/resource-loading conformance audit.


domenic (commented):
LGTM

PR #4314, opened 2026-09-15
Preserve cancellation when reusing XMLHttpRequest

domenic:
Fix replacement XHR requests continuing to download after abort(), found during my script/resource-loading conformance audit.


test/api/xhr.js
domenic, line 1: Can this be replaced with a WPT, maybe with a Python server handler?

domenic (commented):
LGTM

PR #4315, opened 2026-09-15
Implement blob URLs

### LINKED ISSUE #1721: Add URL.createObjectURL and URL.revokeObjectURL
Hi, `jsdom` already supports the `URL` constructor, but not these 2 static methods, they are supported in most modern browsers.

[createObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL)
[revokeObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL)
[Can I use](http://caniuse.com/#feat=bloburls)

scttcper:
Add `URL.createObjectURL()` and `URL.revokeObjectURL()`, with blob URL loading through XMLHttpRequest, scripts, stylesheets, images, and iframes. This enables testing upload previews and generated content that use blob URLs.

Clean up URLs when their creating window closes while allowing in-flight loads to finish.

Closes #1721.


domenic:
Looking good overall, with one fix pushed.

The partial interface hack makes me too sad though so I'm going to try to add the appropriate webidl2js support.

scttcper:
@domenic i think there was a few places using that hack

PR #4316, opened 2026-09-15
Preserve currentScript across nested scripts

domenic:
Fix nested-script and shadow-tree `document.currentScript` behavior found during the classic-script lifecycle alignment.


domenic (commented):
LGTM

PR #4317, opened 2026-09-15
Honor script type and legacy event attributes

domenic:
Fix script attribute checks found during the classic-script lifecycle alignment, preventing scripts with unsupported attribute values from executing.


domenic (commented):
LGTM

PR #4318, opened 2026-09-15
Select storage event recipients at mutation time

domenic:
Fix late-created frames receiving earlier storage events, found during document-lifecycle alignment. This follows HTML and Firefox; Chromium still differs for localStorage, and WebKit for both storage types.


domenic (commented):
LGTM

PR #4319, opened 2026-09-15
Invalidate computed styles after CSS imports load

domenic:
Fix stale computed styles after CSS imports load, found during the resource-lifecycle alignment.


domenic (commented):
LGTM

PR #4323, opened 2026-09-16
Disable unused HTML reports in the WPT runner

domenic:
Disable unused WPT HTML reports to avoid the CI memory exhaustion encountered during the document-lifecycle work in #4321.


domenic:
Investigating why this is necessary... #4321 should not necessarily prevent `Document`s from being GCed, but apparently it does...

domenic:
GPT-6 Astra Extra High says that older Node.js versions can retain the whole JSDOM / Window / Document graph via some retention mechanism involving the Object / Array prototypes and the VM realm. It says https://github.com/v8/v8/commit/8901c793e4392dfea261bcbf9c679e2978847d4d on Node.js v26 fixes the problem.

Since this means jsdom isn't doing anything wrong, I'll proceed with #4321. But, I'll also merge this PR, since it's straightforwardly good to speed up our test runs and and reduce their memory.

PR #4324, opened 2026-09-16
Align classic script and resource lifecycles

domenic:
Continue jsdom's lifecycle conformance work by aligning classic-script processing and resource cancellation with HTML's document-loading model.


domenic:
The number of tests this moves to flaky makes me hesitate. It's a result of no longer serializing all resources into a single queue. Without parser-blocking script support, that might regress jsdom users.

I might hold off on merging this until we can get parser-blocking script work on top of it.

domenic:
**Posted by Claude Fable 5.1 High**

Reviewed as the base of the stack with #4356. I read the full diff, and ran lint, the API suite (626 passing), the to-upstream WPT suite (192 passing, 3 pending), and targeted upstream WPT runs for `dynamic-markup-insertion`, `the-script-element`, `custom-elements`, and `initial-empty-document`, all green on the stacked branch.

No blocking issues; this is mergeable as is. Two behavior changes worth a changelog line when this releases, since they will surprise some users even though both are spec-correct:

- `document.readyState` is now `"interactive"` synchronously after a script-free `new JSDOM()`, where main reported `"loading"`.
- Dynamically created external scripts default to `async`; `script.async = false` is now required for insertion-order execution.

One design note, not a request for change: already-fetched ordered dynamic scripts still execute after `window.stop()` while canceled predecessors release their slot in the execution order. A strict reading of the spec would have the canceled first script block the list forever. The commit message describes this deliberately, so I assume it is intentional.

Things I checked that are fine: `window.stop()` propagates to child documents and cancels their requests without firing completion events, link `href`/`rel` changes cancel in-flight stylesheet fetches, nested `@import` waits and cycle handling, script-blocking stylesheet eligibility limited to parser-created elements, and the load-delay accounting for images, frames, styles, and scripts.

PR #4325, opened 2026-09-16
Add focus fixup rule

### LINKED ISSUE #3029: activeElement can point to a disabled element

### LINKED ISSUE #3605: Support for `inert` HTMLElement property
Hi,

### Basic info:

- **Node.js version:** 18.8.0
- **jsdom version:** 22.1.0 

### Minimal reproduction case

[mini repro on stackblitz](https://stackblitz.com/edit/stackblitz-starters-gnny4b?file=test.js)

```js
 
const dom = new JSDOM(`
<body>
  <div></div>
</body>
`);

const div = document.querySelector("div");

div.inert = true;

// will print `<div></div>` instead of `<div inert></div>`
console.log(div.outerHTML);


```

When the `inert` property of a HTMLElement is set to true, then it will get an `inert` attribute.
[jsfiddle example](https://jsfiddle.net/r3fzjknh/)
HTML spec where this behavior is mentioned : https://html.spec.whatwg.org/multipage/interaction.html#dom-inert

Thank you very much

### LINKED ISSUE #3794: document.hasFocus should not return false in a blur or focus event

asamuzaK:
Fix #3029, fix #3794, fix #3605


asamuzaK:
todo:
* avoid new in method -> done
* <del>clear cache in domselector itself</del>
  domselector listens to only a very limited number of events, so explicitly clearing the cache from jsdom is safer.

asamuzaK:
Looks like the rebase failed, recreated https://github.com/jsdom/jsdom/pull/4345.

PR #4328, opened 2026-09-17
Roll web platform tests

domenic:
Roll WPT to include the tests upstreamed from jsdom in web-platform-tests/wpt#62514, remove their redundant local copies, and enable the complete upstream directories with triaged expectations while preserving existing regression coverage.


domenic:
Still tweaking GPT's judgement on how to do blanket exceptions... Will maybe try to get the process into a skill at the end of it all.

Issue #4344, opened 2026-09-18
relatedTarget on a focus event is the Document object (not <body> or null) after the previously-focused element is removed from the DOM — regression since 30.1.0


rmartins0:
Between jsdom 30.0.1 and 30.1.0, `event.relatedTarget` on a `focus` event can disagree with `document.activeElement` after the previously-focused element is removed from the DOM.

**Repro (standalone, no dependencies beyond jsdom):** https://github.com/rmartins0/jsdom-relatedtarget-repro

Sequence:
1. Focus an element.
2. Remove that element from the DOM without focusing anything else first.
3. `document.activeElement` correctly falls back to `<body>`.
4. Focus a new element. Its `focus` event's `relatedTarget` is the `Document` object — not `<body>` (what `activeElement` reported one line earlier) and not `null`.

```
document.activeElement: <body> (correct)
event.relatedTarget: Proxy(Document { location: [Getter/Setter] })
event.relatedTarget === document: true
typeof event.relatedTarget.focus: undefined
```

Confirmed via `npm install jsdom@30.0.1 --no-save && npm test` vs. `npm install jsdom@30.1.0 --no-save && npm test` in that repo — passes on 30.0.1, fails on 30.1.0 with identical code.

This looks like it was introduced by #4242 ("Avoid repeated event path work"), which reworked how `event.target`/`event.relatedTarget` get resolved during dispatch (a new `updateEventTargets()`/`effectiveTarget` mechanism, applied to the last path entry after the capturing+bubbling loop finishes). That PR states *"the target seen by listeners and the final state of the event still match the DOM Standard and browser behavior"* — this repro seems to be a case where that no longer holds, since a real browser never exposes `Document` as a `relatedTarget`.

We first noticed this via `@mui/material`'s `FocusTrap`, which assumes `relatedTarget` is always an `Element` or `null` (per spec) and calls `.focus()` on it — see mui/material-ui#49159 for that side of it.

scot:
This also makes jsdom fire `blur` on `window` for the same sequence, which is the part that breaks libraries that close popups on window blur (likely what #4342 hit). No dependencies:

```js
const { JSDOM } = require("jsdom");
const { window } = new JSDOM(`<button id="a">a</button><div id="b" tabindex="-1">b</div>`);
const { document } = window;
let windowBlurs = 0;
window.addEventListener("blur", () => windowBlurs++);
document.addEventListener("focusin", (e) =>
[...]

PR #4345, opened 2026-09-18
Add focus fixup rule (rebased)

asamuzaK:
Implement focus-fixup-rule


cjnoname:
Reviewed this against my #4346 (which this supersedes — see below). Checked out 020f1a8 and measured rather than just read, so here is what I found. Short version: your `blurTarget` change fixes everything I reported plus the two residuals I had explicitly left alone, and I have one timing question.

**What I verified**

- `focus-after-removal.html` (the file you took from #4346, byte-identical): **5/5 pass**.
- `focus-fixup.html`: **16/16 pass** — that closes the four gaps I had measured as still broken on my branch (`disabled`, `hidden`, `display: none`, removing `tabindex`).
- `npm run test:api`: **611 passing, 2 failing** — identical to the untouched baseline, so those two resource-loading failures are pre-existing and this introduces no regression.
- Our real-world suite (a shared React 19 + MUI 9 component library, ~4.5k tests on Vitest + jsdom, which is where we originally hit #4344): **4475 passing, 0 failing**, exactly matching my own patch. That includes the MUI `Dialog`/`FocusTrap` teardown sequence that was throwing `nodeToRestore.current.focus is not a function`.

**You fixed two things I had deliberately left out of #4346**

The frame path. I reported in #4346 that focusing into an iframe *after* the outer document's focused element was removed still fired a bogus `blur`/`focusout` at the outer `window`, and said I would not touch it without browser measurements. Your `frameBlurTarget` handling resolves it:

| scenario: focus+remove in outer → focus inner | #4346 | this PR |
|---|---|---|
| `blur` at outer `window` | fires | **gone** |
| `focusout` at outer `window` | fires | **gone** |

And `blur()` now parks the stand-in instead of clearing to `null`, so the two paths that both mean "the focused area is the viewport" finally agree — I had flagged this as out of scope:

```
            #4346                    this PR
blur():     hasFocus=false           hasFocus=true
remove():   hasFocus=true            hasFocus=true
```

**One question: the fixup runs in a microtask, and the test can't see the difference**

`Element-impl.js` schedules `focusFixupRule` via `queueMicrotask`, so attribute- and style-driven fixups are async while removal stays sync:

| trigger | sync | after 1 microtask | after 2×rAF |
|---|---|---|---|
| `el.disabled = true` | `INPUT#a` | `BODY` | `BODY` |
| `el.hidden = true` | `INPUT#a` | `BODY` | `BODY` |
| `el.style.display = "none"` | `INPUT#a` | `BODY` | `BODY` |
| `el.remove()` | **`BODY`** | `BODY` | `BODY` |

`focus-fixup.html` awaits two `requestAnimationFrame`s before asserting, so it passes either way — it can't distinguish a sync implementation from a deferred one, which means it won't catch a drift here after upstreaming. Two things I genuinely don't know: whether the spec intends one consistent moment for all fixup triggers (removal being sync and attributes being async looks internally inconsistent), and what engines actually do. I can't measure Chromium from this environment, so I'm asking rather than asserting — if browsers apply it synchronously, then `el.disabled = true; document.activeElement` reads differently here than in a browser, which is exactly the class of divergence that sent us on this hunt in the first place.

If the deferral is deliberate, it might be worth a subtest that asserts the state *immediately* after the trigger, so the intended timing is pinned rather than incidental.

**On #4346**

This is a strict superset of my change, so #4346 should not land alongside it — happy for it to be closed in favour of this, and I'll close it myself once this is merged (or right now if a maintainer prefers). Thanks for folding the tests in; no attribution needed on my side.

asamuzaK:
Regarding `focus-fixup.html`, I'm seeing inconsistent results in actual browsers.
 jsdom passes all of them.
It looks like these might be bugs in the browsers, WDYT?

Firefox: 3 failures.
* [Sync] focus-fixup-rule: When an iframe is removed from the DOM
  assert_array_equals: No focus-related events should be fired when the focused element is removed from the DOM (!isConnected). lengths differ, expected array [] length 0, got ["focus", "focus"] length 2
* [Async] focus-fixup-rule: When the hidden attribute is added to an iframe
  assert_not_equals: The element should no longer be the activeElement after the trigger is fired. got disallowed value Element node <iframe src="about:blank" title="iframe test 3" hidden=""...
* [Async] focus-fixup-rule: When display: none is applied to an iframe via CSS
  assert_not_equals: The element should no longer be the activeElement after the trigger is fired. got disallowed value Element node <iframe src="about:blank" title="iframe test 4" style="di...

Chrome: 2 Failures.
* [Sync] focus-fixup-rule: When the element is removed from the DOM
  assert_array_equals: No focus-related events should be fired when the focused element is removed from the DOM (!isConnected). lengths differ, expected array [] length 0, got ["blur", "focusout"] length 2
* [Sync] focus-fixup-rule: When an iframe is removed from the DOM
  assert_array_equals: No focus-related events should be fired when the focused element is removed from the DOM (!isConnected). lengths differ, expected array [] length 0, got ["blur", "focusout"] length 2

Related: https://issues.chromium.org/issues/40808086

cjnoname:
Agreed that the browser failures are browser bugs, and for Chrome there's hard confirmation. But first, **a correction to my previous comment**, and then some measurements suggesting this PR doesn't yet satisfy an existing WPT.

## Correction

I said the split between synchronous removal and asynchronous attribute fixups "looks internally inconsistent." **That was wrong** — the existing WPT `html/interaction/focus/processing-model/focus-fixup-rule-one-no-dialogs.html` (by @domenic) encodes exactly that split via its `expectSync` parameter: `remove()` is `expectSync = true`, every attribute/style trigger is `expectSync = false`. Your direction is right and my objection was unfounded. Sorry for the noise.

## Chrome: confirmed, and Chromium already owns it

Not a guess — from the Chrome Platform Status API for the entry referenced in #4344:

```
$ curl -s https://chromestatus.com/api/v0/features/5128696823545856
name:         "Stop sending blur events on element removal"
feature_type: "Feature removal"
chrome:       "In developer trial (Behind a flag)"
ff view:      "Shipped/Shipping"
safari view:  "Shipped/Shipping"
spec:         https://html.spec.whatwg.org/#dom-trees:event-blur
```

Their own summary: *"Currently Chromium-based browsers are **the only ones** that fire blur events when an element is removed from the DOM. This has recently been **clarified in the spec**."*

So both Chrome failures are a known Chromium deviation that Chromium has filed as a feature removal and is trialling behind a flag, with Gecko and WebKit already on the spec side. The spec text backs the assertion directly — the [removing steps](https://html.spec.whatwg.org/multipage/infrastructure.html#node-remove-focus-fixup) note says this "does not perform the unfocusing steps, focusing steps, or focus update steps, and thus no `blur` or `change` events are fired." Your expectation is correct; don't weaken it.

## Firefox: spec is on your side, with one caveat on wording

Same note covers the iframe-removal case, though strictly it enumerates `blur` and `change`, and Firefox is firing `focus` — that's ruled out by "does not perform the ... focusing steps" rather than by the list of event names, so it's an inference from the algorithm rather than a literal reading. The two `[Async]` iframe failures (`hidden`, `display: none` not fixing up) are plain focus-fixup-rule violations.

Worth noting Chrome and Firefox are **inconsistent with each other** here (Chrome fires `blur`/`focusout`, Firefox fires `focus`×2), so this can't be argued as a web-compat constraint on the spec — they're independent bugs. I searched Bugzilla for the iframe fixup cases and found nothing, but my queries were shallow, so please don't read that as "unreported."

## The part I'd look at before upstreaming: an existing WPT still fails

`focus-fixup-rule-one-no-dialogs.html` is still listed `[fail, Unknown]` in `to-run.yaml` and this PR doesn't remove that line. I ran its exact fixture and all eight cases against 020f1a8, dropping only the ResizeObserver ordering checks (jsdom has no ResizeObserver, and `new ResizeObserver` throws before the rAF assertions are reached, which is what currently masks this):

| case | trigger | sync assert | 1st-rAF assert | final |
|---|---|---|---|---|
| `#button3` | `remove()` (`expectSync`) | PASS | n/a | PASS |
| `#button1` | `disabled = true` | PASS | **FAIL** | PASS |
| `#button2` | `hidden = true` | PASS | **FAIL** | PASS |
| `#button4` | `fieldset1.disabled = true` | PASS | **FAIL** | PASS |
| `#div` | `removeAttribute("tabindex")` | PASS | **FAIL** | PASS |
| `#button6` | `style.visibility = "hidden"` | PASS | **FAIL** | PASS |
| `#button5` | `insertBefore(legend)` | PASS | PASS | **FAIL** |
| `#editable` | `contentEditable = false` | PASS | PASS | **FAIL** |

**Timing (five cases).** That WPT asserts `activeElement` has *not* changed yet inside the first `requestAnimationFrame`, with the comment "Fixup should run after animation frame callbacks, right before the end of *update the rendering*, so after resize observations." `queueMicrotask` lands far earlier, so `activeElement` is already `BODY` by the first rAF. The final state is right; only the moment is early. Whether jsdom can or should model "end of update the rendering" is a fair question — if the microtask is a deliberate approximation, it may be worth saying so in `to-run.yaml` rather than leaving the entry as `Unknown`.

**Two cases don't fix up at all**, and neither looks like something you introduced:

- `#editable` — `isFocusableAreaElement` returns true for *any* `contenteditable` attribute regardless of value (`helpers/focusing.js:66` checks only `hasAttributeNS`), so `contenteditable="false"` still counts as focusable. Pre-existing.
- `#button5` — the hook is on `_attrModified` (`Element-impl.js:99`), but this case changes focusability by *inserting* a `<legend>`, which flips `isDisabled` for the button without any attribute mutation on it. Would need the insertion steps to participate too.

## Test overlap

`focus-fixup.html` covers ground already covered by `focus-fixup-rule-one-no-dialogs.html` and `html/editing/focus/focus-management/focused-area-removal.html`. The existing one is stricter on timing (`expectSync` per case) while yours waits two rAFs and so can't distinguish sync from deferred; yours is stricter on events, which is the genuinely new coverage and the part that catches Chromium. Might be worth folding the event assertions into the existing files, or at least cross-linking, so reviewers upstream don't see three overlapping tests.

Repro script for the table is a plain Node file against `lib/api.js` — happy to post it if useful.

asamuzaK:
@cjnoname Thanks for the detailed research!
@domenic What are your thoughts on this?

domenic (changes requested):
Unfortunately this approach has the same problem as the 2024 attempt: https://github.com/jsdom/jsdom/pull/3762#issuecomment-2308217736 . It does not match the spec at all.

If you want to implement focus fixup, please implement it according to the spec. In particular, step 17 of https://html.spec.whatwg.org/#update-the-rendering , plus the HTML element removal steps: https://html.spec.whatwg.org/#node-remove-focus-fixup . There are no attribute changed steps on HTMLOrSVGElement in the spec.

I think it would also be much easier to review if you added `inert` in a separate PR, instead of together. Note that `inert` is on `HTMLElement`, not `Element`.

I also am not sure why there are so many changes to `focus()` and `blur()` in this PR. Maybe those are good bug fixes. But they are not related to focus fixup, and should be in a separate PR.

I'm sorry to give a negative review, as I know you have worked on this a few times over the years. But please follow the spec, as was requested previously, and split PRs into reviewable chunks.

asamuzaK:
Thanks for the review. I appreciate the feedback!
I'm going to mark this as a draft and rethink the implementation.

asamuzaK:
@domenic Thank you for pointing me in the right direction.

* Reverted all unrelated changes (like inert, focus(), and blur()) and removed the _attrModified attribute observation hook.
* Implemented the focus fixup rule according to the spec:
  * Sync removal steps: Implemented in Node-impl.js _removingSteps so it falls back to body immediately without firing events.
  * Async update the rendering steps: Hooked into Window.js after the animation frame callbacks are processed, executing a polling check via focusFixupRule.

Regarding Test Failures:
* Several `focus-fixup.html` test cases are failing in jsdom.
  They are because of the underlying isFocusableAreaElement bugs in jsdom (e.g., it ignores hidden/display:none for buttons due to an early return, and elImpl.style evaluates to undefined internally).
* Similarly, `focus-after-removal.html` and `hasfocus-focus-blur.html` are failing because of the pre-existing focus()/blur() bugs (e.g., firing blur events on body, and incorrectly reporting document.hasFocus() during internal focus shifts).

I kept this PR strictly to focus-fixup and those bug fixes should be addressed in the separate PRs.

domenic:
With #4351 merged, I suggest the following order:

- A PR narrowly focused on improving `isFocusableAreaElement()`. https://github.com/jsdom/jsdom/pull/4352 has a lot of that, but also some other stuff that doesn't match the spec by changing the `activeElement` getter algorithm to do a kind of lazy version of the focus fixup rule, which doesn't fire events and thus isn't correct. Maybe just narrow #4352 to that, with test updates to match that just call `focus()` on an appropriate candidate element?
- Then, we can add inert (https://github.com/jsdom/jsdom/pull/4350 + actual behavior changes).
- Then, this PR, with the focus fixup rule. The approach looks pretty good, although it should use the viewport (which jsdom codes as the Document) instead of the body where the spec says viewport, and we need to figure out a way to make it work even when user code has never called `requestAnimationFrame()`.

PR #4303, opened 2026-09-11
Align DOM and HTML node lifecycle steps

domenic:
Follow-up to reviewing #4244: align jsdom's node lifecycle with DOM and HTML, including preserving focus on removal, correcting radio-group membership and canceled activation, and updating script preparation, while retaining the current parser limitations.

PR #4346, opened 2026-09-18
Don't expose the viewport sentinel to focus events

### LINKED ISSUE #4344: relatedTarget on a focus event is the Document object (not <body> or null) after the previously-focused element is removed from the DOM — regression since 30.1.0

cjnoname:
Fixes #4344.

The removing steps park the `Document` itself in `_lastFocusedElement` as a stand-in for the viewport (af483ae, #4303), so that `activeElement` resolves to the body while `hasFocus()` stays true. `focus()` then treated that stand-in as a real element:

- it passed it to `fireFocusEventWithTargetAdjustment`, which retargets any `_defaultView` holder to the `Window`, turning the viewport stand-in into a real `blur`/`focusout` **at `window`**;
- it used it as the `relatedTarget` of the following `focus`/`focusin`, exposing the **`Document`** to author code.

Per spec neither should be reachable for a same-document move:

- [The removing steps](https://html.spec.whatwg.org/multipage/infrastructure.html#node-remove-focus-fixup) set the focused area to the viewport, and note this "does not perform the unfocusing steps, focusing steps, or focus update steps, and thus no `blur` or `change` events are fired".
- [Focus update steps](https://html.spec.whatwg.org/multipage/interaction.html#focus-update-steps) step 1 pops the shared trailing `Document` from both chains, so the old chain is just the viewport, and the Document→Window rule in step 2.2 never applies to it.
- The remaining branch of step 2.2: a viewport entry is neither an element nor a `Document`, so the blur event target is null and, per the note, no event is fired.
- Step 4.4: the related focus target is null unless the last old-chain entry is an `Element`, and the viewport is not.

### Before / after

```js
const { window } = new JSDOM(`<input id="a"><input id="b">`);
const { document } = window;
document.getElementById("a").focus();
document.getElementById("a").remove();
document.getElementById("b").focus();
```

| | Chromium 153 | jsdom 30.1.0 | this PR |
|---|---|---|---|
| `blur`/`focusout` at `window` | no | **yes** | no |
| `#b` focus `relatedTarget` | `null` | **`#document`** | `null` |
| `document.activeElement` after removal | `<body>` | `<body>` | `<body>` |
| `document.hasFocus()` after removal | `true` | `true` | `true` |

(Chromium additionally fires `blur` at `#a` at the *removal* step, which is the Chromium-only deviation tracked at chromestatus 5128696823545856, so that is left alone here.)

### Approach

The filter goes at the call site in `focus()` rather than inside `fireFocusEventWithTargetAdjustment`, because the Document→Window adjustment in that helper is deliberate and still needed for genuine cross-document moves — the frame branch continues to rely on it. Doing it at the call site is what lets the same-document case be told apart from a cross-document one; `hasFocus()` and the `activeElement` fallback that the sentinel exists for are untouched.

`blur()` still clears the slot to `null` rather than to the sentinel; that is existing behaviour and out of scope here.

### Tests

Adds `to-upstream/uievents/focusevents/focus-after-removal.html`. Against 30.1.0 three of the five subtests fail (spurious `window` `blur`/`focusout`, `Document` `relatedTarget`); the other two are controls that pass either way, covering the `activeElement`/`hasFocus()` fallback and a normal element-to-element move still reporting the other element as `relatedTarget`.

Reported downstream at mui/material-ui#49159, where MUI's `FocusTrap` stores `relatedTarget` as the node to restore focus to and calls `.focus()` on it while unmounting, throwing `nodeToRestore.current.focus is not a function` on the `Document`.

I could not run the WPT harness locally (it needs the `/etc/hosts` entries), so the new file has only been exercised directly against `resources/testharness.js`; CI will be the real check. `npm run test:api` shows no change (611 passing, same 2 unrelated resource-loading failures before and after).


cjnoname:
Scope note, to avoid overstating what this fixes and to flag the overlap with #4345.

This PR only makes the *removal* path stop producing events that cannot occur in a browser. It deliberately does not implement the focus fixup rule in general, so the other triggers for it are still unfixed on this branch (#3029):

| Trigger | Spec | jsdom, with this PR |
|---|---|---|
| `el.disabled = true` | focus moves to the viewport | `activeElement` stays `el` |
| `el.hidden = true` | focus moves to the viewport | `activeElement` stays `el` |
| `el.style.display = "none"` | focus moves to the viewport | `activeElement` stays `el` |
| `el.removeAttribute("tabindex")` | focus moves to the viewport | `activeElement` stays `el` |

Related, and also left alone here: `blur()` clears `_lastFocusedElement` to `null` while the removing steps set it to the Document, so the two paths that both mean "the focused area is the viewport" disagree about `hasFocus()`:

```
blur():   activeElement=BODY  hasFocus=false
remove(): activeElement=BODY  hasFocus=true
```

The underlying shape of all of this is that `_lastFocusedElement` carries two meanings (an element, or the Document standing in for the viewport), and this PR discriminates on that at the one call site that was getting it wrong rather than removing the overload. A full fix would model the focused area explicitly, or implement the focus update steps' chain computation properly — which is closer to what #4345 is doing.

So if #4345 lands first, this becomes redundant and should be closed; happy to rebase, narrow, or fold the test file into that PR instead, whichever is more useful. The reason I kept it separate is that #4345 targets #3029/#3794/#3605 and, as written, appears to relocate the bogus `window` blur to `<body>` rather than not firing it, so #4344 would remain partly open.

cjnoname:
One more data point while checking that this patch leaves cross-document focus alone: it does, but the same stand-in leak still exists on the frame path, and I left it alone deliberately.

With an iframe, focusing into the inner document *after* the outer document's focused element was removed still fires `blur`/`focusout` at the outer `window`:

```js
const dom = new JSDOM(`<input id="outer"><iframe id="f"></iframe>`, { pretendToBeVisual: true });
const d = dom.window.document;
const innerDoc = d.getElementById("f").contentDocument;
innerDoc.body.innerHTML = `<input id="inner">`;

const tmp = d.createElement("input");
d.body.append(tmp);
tmp.focus();
tmp.remove();                                  // outer's slot now holds the stand-in
innerDoc.getElementById("inner").focus();
// outer: blur    target=window  relatedTarget=null
// outer: focusout target=window  relatedTarget=null
```

That comes from the frame branch passing the outer document's slot straight through:

```js
const frameLastFocusedElement = frameElement.ownerDocument._lastFocusedElement;
focusing.fireFocusEventWithTargetAdjustment("blur", frameLastFocusedElement, null);
```

Reading the focus update steps the same way as in the commit message, this one should not fire either: the shared trailing `Document` is popped in step 1, so the old chain is just the outer viewport, and a viewport entry gives a null blur event target in step 2.2. Byte-identical output before and after my change, so it is pre-existing rather than something I introduced, and it is outside what #4344 reports (that repro is same-document).

I did not fix it here for two reasons: I have no browser measurement for this cross-document sequence to check against, and iframe focus is exactly the area where engines deviate from the spec in practice — I would rather not change it on a spec reading alone. It also looks like it belongs with the broader focus rework in #4345.

Happy to extend this PR with `frameLastFocusedElement === frameElement.ownerDocument ? null : frameLastFocusedElement` plus a test if you would prefer it fixed in one go, or to leave the frame path for #4345. Just say which.

asamuzaK:
@cjnoname Included your tests in https://github.com/jsdom/jsdom/pull/4345 and fixed the blur target logic.
WDYT?

cjnoname:
@asamuzaK Checked it out and measured — your version fixes everything this PR does, plus the two residuals I had explicitly left out of it (the bogus `window` blur on the frame path, and `blur()` disagreeing with the removal path about `hasFocus()`). Full numbers in https://github.com/jsdom/jsdom/pull/4345#issuecomment-5736515352, including that our ~4.5k-test MUI suite that originally hit #4344 is green on your branch, and that `test:api` matches baseline.

So: #4345 is a strict superset and should land instead of this. I left one open question there about the fixup running in a microtask for attribute changes while removal stays synchronous — not a blocker, and I can't measure a real browser from here to say which is right.

Keeping this open only so #4344 isn't left without a PR if #4345 stalls; I'll close it the moment yours is merged, or immediately if a maintainer would rather consolidate now. Thanks for folding the tests in.

domenic:
This PR seems promising, but it violates the project's [AI policy](https://github.com/jsdom/jsdom/blob/main/Contributing.md#ai-assisted-contributions) in several ways. The most important of which is that it gives me THREE huge AI-generated walls of text (the PR description, and two followups), which I just do not have the patience to read.

So, let me close this. If you are interested in fixing this issue, please follow the guidelines and produce concise PR descriptions that respect the time of human maintainers, and properly identify AI-authored code using the `Co-authored-by` footer that this project requires.

PR #4348, opened 2026-09-18
Preserve document focus during element transitions

### LINKED ISSUE #4347: 30.1.0 regression: removing the focused element makes the next focus() fire a window-targeted blur (breaks Radix Select close-on-blur in tests)
### Basic info

- **Node.js version:** v24.21.0
- **jsdom version:** 30.1.0 (30.0.1 is not affected)

### Minimal reproduction

```js
import { JSDOM } from 'jsdom'

const dom = new JSDOM('<button id="a">a</button><button id="b">b</button>')
const { window } = dom
const doc = window.document

window.addEventListener('blur', (e) => {
    console.log(
        `window blur: relatedTarget=${e.relatedTarget ? e.relatedTarget.tagName : null}, ` +
        `document.hasFocus()=${doc.hasFocus()}`,
    )
})

const a = doc.getElementById('a')
a.focus()
a.remove() // remove the element while it is focused
doc.getElementById('b').focus()
```

Output with jsdom **30.1.0**:

```
window blur: relatedTarget=BUTTON, document.hasFocus()=false
```

Output with jsdom **30.0.1**: no output — no `blur` reaches the window.

### Expected behavior

A focus transition *within* the document must not blur the window: `window` `blur` is how a page learns that it lost focus (e.g. the user switched windows/tabs). Here jsdom 30.1.0 fires `blur` **at the `window`**, with `relatedTarget` set to the element that is gaining focus.

### What happens internally

The new `_removingSteps()` in `Node-impl.js` (30.1.0) represents the viewport after removal by setting `_ownerDocument._lastFocusedElement = this._ownerDocument`. Its comment states the intent correctly:

> Represent the viewport with the Document so that activeElement resolves to the body (or document element) while hasFocus() remains true. Removal does not run unfocusing steps or fire blur/change events.

But the next `focus()` in `HTMLOrSVGElement-impl.js` treats any truthy `_lastFocusedElement` as a previously focused element:

```js
const previous = ownerDocument._lastFocusedElement;
...
if (previous) {
    focusing.fireFocusEventWithTargetAdjustment("blur", previous, this);
    ...
}
```

With `previous` being the Document, `fireFocusEventWithTargetAdjustment` retargets the event to the **window** (a Document has `_defaultView`) and fires a window `blur` — with `relatedTarget` = the newly focused element — which is exactly what removal was documented not to do. `document.hasFocus()` also reads `false` for the duration of that dispatch.

### Real-world impact

Anything that closes on `window` blur to detect "the user switched away" misfires in test environments. Concretely, **Radix UI `Select` attaches `window.addEventListener("blur", close)`**: with React Testing Library (whose `cleanup()` unmounts the focused tree) + user-event, every test that removes a focused element makes the *next* test's focus change fire a window `blur` — so clicking a Radix Select trigger opens it and it immediately closes (open → close). Upgrading jsdom 30.0.1 → 30.1.0 turned 45 of our admin tests red. We currently work around it in the test setup:

```js
window.addEventListener('blur', (event) => {
    if (event.target === event.currentTarget && event.relatedTarget !== null) {
        event.stopImmediatePropagation()
    }
}, { capture: true })
```

The symptom also looks timing-dependent in larger React trees (the blur can land after listeners attach), which makes tests flake inconsistently.

Happy to send a PR if the desired direction is clear — e.g. treating `previous === ownerDocument` inside `focus()` as "no `blur`/`focusout` to fire" (the Document is a viewport representation, not a focus event target), or giving the viewport its own state that the focus-transition path does not read as a previously focused element.

scttcper:
Addresses the Radix Select regression reported in #4347, where removing a focused element during test cleanup makes the next focus change close the menu.

Keep the document focused through element blur and focus changes, and avoid spurious window blur events when focus returns from the viewport to an element in the same document.

Fixes #4347.


scttcper:
#4345 will probably fix this

domenic:
Feel free to reopen as a more targeted fix than #4325.

PR #4349, opened 2026-09-18
fix(focus): Ignore tabindex on disabled controls and hidden inputs

scttcper:
`focus()` could move focus to a disabled form control or hidden input if it had `tabindex`.

Keeps those elements unfocusable as required by the [HTML focus rules](https://html.spec.whatwg.org/multipage/interaction.html#focusable-area), while preserving the first-legend exception for disabled fieldsets.


scttcper:
Closed #4348 since #4345 covers that fix. Keeping this one open because #4345 still lets disabled controls and hidden inputs receive focus when they have a tabindex.

Ran this PR's tests against #4345 and 14 of the 18 cases still fail.

domenic (approved):
Nice, thank you!

PR #4350, opened 2026-09-19
Implement inert global attribute

asamuzaK:
Split from #4345 

WPTs in `/inert/` currently fail.
This is expected because:
* jsdom does not yet implement the focus-fixup-rule to asynchronously remove focus when an element becomes inert.
* isFocusableAreaElement has pre-existing bugs that prevent it from correctly recognizing inert ancestors.


lib/jsdom/living/nodes/HTMLElement-impl.js
```diff
@@ -197,6 +197,18 @@ class HTMLElementImpl extends ElementImpl {
   get offsetHeight() {
     return 0;
   }
+
+  get inert() {
```
domenic: These are not necessary, since they are generated by [Reflect].

test/web-platform-tests/to-upstream/html/dom/reflection-inert-dont-upstream.html
domenic, line 1: This test is good, but it should be renamed to a dont-upstream test with a reference to https://github.com/web-platform-tests/wpt/pull/62822, similar to https://github.com/jsdom/jsdom/blob/main/test/web-platform-tests/to-upstream/html/dom/reflection-img-srcset-dont-upstream.html

domenic (commented):
We shouldn't merge this until it has an impact on the actual focusing algorithm, since otherwise it breaks people using the inert polyfill or similar.

So I guess this will probably need to be on top of https://github.com/jsdom/jsdom/pull/4352 . I will review that next.

test/web-platform-tests/to-upstream/html/dom/reflection-inert-dont-upstream.html
asamuzaK: Renamed and added a link to /html/dom/reflection.js.
Correct me if I'm wrong.

PR #4351, opened 2026-09-19
Add focus update steps

### LINKED ISSUE #4344: relatedTarget on a focus event is the Document object (not <body> or null) after the previously-focused element is removed from the DOM — regression since 30.1.0

### LINKED ISSUE #4347: 30.1.0 regression: removing the focused element makes the next focus() fire a window-targeted blur (breaks Radix Select close-on-blur in tests)

asamuzaK:
Refactor focus handling into shared focusing/unfocusing helpers that follow the HTML focus update algorithm.

Fixes #4344, fixes #4347


lib/jsdom/living/helpers/focusing.js
```diff
@@ -103,3 +103,101 @@ exports.fireFocusEventWithTargetAdjustment = (name, target, relatedTarget, { bub
 
   target._dispatch(event);
 };
+
+// https://html.spec.whatwg.org/multipage/interaction.html#focusing-steps
```
domenic: This doesn't seem to match the spec very much. It's missing the two optional arguments, and step 1, 2, 3. The final getSelection bit also is not in the spec, although maybe it is missing from the spec.

lib/jsdom/living/helpers/focusing.js
```diff
@@ -103,3 +103,101 @@ exports.fireFocusEventWithTargetAdjustment = (name, target, relatedTarget, { bub
...
+
+  if (document.getSelection) {
+    document.getSelection().collapse(target, 0);
+  }
+};
+
+// https://html.spec.whatwg.org/multipage/interaction.html#unfocusing-steps
+exports.unfocusingSteps = target => {
+  const document = target._ownerDocument;
+  if (document._lastFocusedElement !== target) {
```
domenic: This doesn't seem quite the same as step 5 of the spec.

Step 6 is also missing, and the getSelection step is not in the spec again.

domenic:
This was closer to matching the spec, but did not quite match in a few ways. I had GPT-6 Astra Extra High work on aligning them, with a long commit message explaining what it did. WDYT?

asamuzaK:
Looks good.

PR #4352, opened 2026-09-20
Update focused area handling

asamuzaK:
* <del>Add activeElement and  getTheFocusedAreaOfTheDocument helpers.
  Update DocumentOrShadowRootImpl to delegate activeElement lookup to the new helper.</del>
* Update isFocusableAreaElement handling

Note that the attribute and style checks are currently simplified.


lib/jsdom/living/helpers/focusing.js
```diff
@@ -19,15 +19,35 @@ exports.isFocusableAreaElement = elImpl => {
...
+  }
 
-    if (!elImpl.isConnected) {
-      return false;
+  let current = elImpl;
+  while (current) {
+    if (current.nodeType === NODE_TYPE.ELEMENT_NODE) {
+      // TODO: jsdom does not yet have a comprehensive internal concept of "inertness".
+      // Once #4350 (inert) is fully integrated, this might need refinement.
+      if (current.hasAttributeNS(null, "hidden") || current.hasAttributeNS(null, "inert")) {
```
domenic: Please remove this line:

- We don't implement inert as a feature yet, so we should not include part of the support for it here
- There is no part of the focus spec that checks for the hidden attribute specifically. The hidden attribute's only impact is the UA stylesheet's default `display: none`. Please add a test that `<input hidden style="display:block">` is focusable, which matches browsers but fails this PR until you delete this line.

lib/jsdom/living/helpers/focusing.js
```diff
@@ -19,15 +19,35 @@ exports.isFocusableAreaElement = elImpl => {
...
+        return false;
+      }
+      // TODO: This only checks inline styles. A full check would require computed styles.
+      if (current.style && (current.style.display === "none" || current.style.visibility === "hidden")) {
```
domenic: Why not check computed styles?

lib/jsdom/living/helpers/focusing.js
```diff
@@ -19,15 +19,35 @@ exports.isFocusableAreaElement = elImpl => {
...
+        return false;
+      }
+    }
+    const root = current.getRootNode();
+    if (root && root.nodeType === 11 && root.host) {
```
domenic: Please use the NODE_TYPE constants

lib/jsdom/living/helpers/focusing.js
domenic: The visibility handling here is incorrect. In the following, the input is focusable in browsers, but not in this PR:

```html
  <div style="visibility: hidden">
    <input style="visibility: visible">
  </div>
```

Please add a test.

lib/jsdom/living/helpers/focusing.js
```diff
@@ -19,15 +19,35 @@ exports.isFocusableAreaElement = elImpl => {
...
+      // Once #4350 (inert) is fully integrated, this might need refinement.
+      if (current.hasAttributeNS(null, "hidden") || current.hasAttributeNS(null, "inert")) {
+        return false;
+      }
+      // TODO: This only checks inline styles. A full check would require computed styles.
+      if (current.style && (current.style.display === "none" || current.style.visibility === "hidden")) {
+        return false;
+      }
+    }
+    const root = current.getRootNode();
```
domenic: Skipping directly to the root isn't correct here.

```html
<div id="host"></div>
<script>
  const shadow = document.getElementById("host").attachShadow({ mode: "open" });
  shadow.innerHTML = `
    <div style="display: none">
      <input>
    </div>
  `;

  const input = shadow.querySelector("input");
  input.focus();

  console.log(shadow.activeElement === input);
  // Expected: false. This PR: true.
</script>
```

However the simple correction of just

```js
current = isShadowRoot(current) ? current.host : current.parentNode;
```

isn't quite right either, as it traverses the shadow-including ancestors, whereas what the spec requires is [*flat tree*](https://drafts.csswg.org/css-shadow-1/#flattening) traversal:

```html
<div id="host">
  <input id="target">
</div>
<script>
  document.getElementById("host").attachShadow({ mode: "open" }).innerHTML = `
    <div style="display: none">
      <slot></slot>
    </div>
  `;

  const input = document.getElementById("target");
  input.focus();

  console.log(document.activeElement === input);
  // Expected: false.
</script>
```

I suggest adding the following to `shadow-dom.js` and using it here:

```js
// https://drafts.csswg.org/css-shadow/#flat-tree
// Returns inclusive element ancestors, nearest first, or null if the element
// is absent from its document's flat tree.
function flatTreeInclusiveAncestors(element) {
  const root = element._ownerDocument.documentElement;
  const ancestors = [];

  for (let current = element; current !== null; current = flatTreeParentElement(current)) {
    ancestors.push(current);
    if (current === root) {
      return ancestors;
    }
  }

  return null;
}

function flatTreeParentElement(element) {
  const parent = element.parentNode;

  if (isShadowRoot(parent)) {
    return parent.host;
  }

  if (parent === null || parent.nodeType !== NODE_TYPE.ELEMENT_NODE) {
    return null;
  }

  if (parent._shadowRoot) {
    // Unassigned light children are excluded. `findSlot()` includes closed
    // roots, unlike the public `assignedSlot` getter.
    return findSlot(element);
  }

  if (isSlot(parent) &&
      isShadowRoot(parent.getRootNode()) &&
      parent._assignedNodes.length > 0) {
    // Assigned nodes replace the slot's fallback children.
    return null;
  }

  return parent;
}
```

domenic (changes requested):
I appreciate this focused PR; it is definitely the right direction.

asamuzaK:
I think I've addressed most of the feedback.

PR #4354, opened 2026-09-21
Preserve CSS priorities across declaration updates

domenic:
Fix CSS priority corruption found while reviewing #4343, preserving existing declarations after invalid updates and honoring requested priorities when updating longhands after important shorthands.


domenic:
This is a version of #4343 that is more performant and also handles this case, which the PR missed:

```js
style.setProperty("color", "red", "important");
style.setProperty("color", "rgb(calc(1px) 0 0)");
style.getPropertyPriority("color");
// jsdom: ""
// Chromium, Firefox, WebKit: "important"
```

@asamuzaK, WDYT?

Adding a second parameter to the `set` functions is a bit strange, and probably we should change their shape to no longer be property descriptors at some point. But it seems OK for now...

asamuzaK:
At first glance, if having a priority set on the shorthand doesn't cause any regressions (referring to the deleted code here), it should be fine, right?
``` javascript
const priority =
  !this._priorities.get(shorthand) && this._priorities.has(property) ? this._priorities.get(property) : "";
```

domenic:
Good call. I added some more tests for that, and fixed the bugs it found in the padding/margin and flex helpers.

PR #4356, opened 2026-09-21
Pause document parsing for blocking scripts

domenic:
Bring classic-script parsing closer to HTML's lifecycle requirements, including blocking resources and incremental `document.write()`, on top of #4324.


domenic:
**Posted by Claude Fable 5.1 High**

Reviewed on top of #4324. I read the full diff, and ran lint, the API suite (626 passing), the to-upstream WPT suite (192 passing, 3 pending), and targeted upstream WPT runs for `dynamic-markup-insertion` (126 passing), `the-script-element` (210 passing), `custom-elements` (176 passing), and `initial-empty-document`, all green. I also probed edge cases by hand: `includeNodeLocations` offsets across chunk boundaries, `>` inside inline script text and text nodes, nested `document.write()` ordering across a pending external script, `close()` followed by `write()` while blocked, and `window.stop()` from a `connectedCallback`. All behaved correctly.

Correctness looks solid. I have one issue I would fix before landing, plus a few minor items.

## Script-free construction is about 2x slower

The report acknowledges an unresolved constructor slowdown. It is larger for realistic documents than the 1,000-element benchmark suggests:

| Workload | main | this PR |
| --- | ---: | ---: |
| `api/constructor` "many elements" (median) | 9.3 ms | 12.3 ms |
| 380 KB document, 5,000 nested blocks | 157 ms | 329 ms |
| 580 KB text-heavy document | 73 ms | 131 ms |

The cause is the per-`>` chunking in `#pump` in `lib/jsdom/browser/parser/html.js` (and the XML equivalent). Isolating parse5 with its default tree adapter on the same 380 KB input: one `write()` takes 16 ms, chunking at every `>` takes 148 ms, and chunking only at `</script` candidates takes 16 ms. So essentially the whole regression is the chunk granularity, not the parser-session machinery.

The `>` boundary does two jobs, and both have cheaper alternatives that stay within public parse5 APIs:

1. Script checkpoints only need the chunk to end at a script end tag. Splitting after `</script` followed by whitespace, `/`, or `>` (case-insensitive) is sufficient. A false positive inside a comment, attribute value, or double-escaped script data is just an extra harmless boundary. The one parse-error corner is `</script foo="a>b">`, which a small quote-aware scan for the tag end would cover.
2. Stopping mid-chunk on abort (custom element constructors, `connectedCallback`, or virtual console and interceptor callbacks calling `window.stop()`) does not need chunking. The abort listener already holds the stream, and `stream.parser.tokenizer.pause()` is public in both packages' typings. Calling it from the abort path makes the tokenizer loop exit after the current token, which is exactly the "at most one reaction per chunk" property the `>` boundary currently provides.

This is a trade-off call, but the constructor is jsdom's hottest path and I would not land a 2x regression on it when a same-API alternative appears to exist.

## Minor

- `document.open()` during an active parse cancels unrelated fetches. `_startParsingSession` calls `_abort()`, which aborts the document-wide fetch controller. I confirmed an in-flight `XMLHttpRequest` goes silent (stays at readyState 1, no events) when `document.open()` runs while the parser is blocked on a script. The spec only replaces the parser. Aborting just the parsing session would fix it, but the pending blocking script's fetch is tied to the document signal rather than the session, so decoupling takes a bit more work. Low priority given the scenario.
- Trailing text buffering is a real regression for a common pattern. `document.write("<p id=w>hello")` now leaves the paragraph's text empty until another token arrives, where main returned it immediately. The README documents this and the parse5 flush PR exists, but it may be worth waiting for parse5 before this ships in a release, since old `document.write()` reliance is exactly the code that uses jsdom this way.
- Frame parse errors that are not XML syntax errors leave the parent's load delayed. In `fetchFrame`, `reportParsingError` rethrows non-syntax errors inside the `catch`, so `parsingSession.finish()` never runs and the parent's load delay from `_startParsingSession` is never released. The same function reached from the async `onError` path in `XMLDocumentParser` would become an unhandled rejection. Hard to trigger today since customized built-in constructor errors are now caught, but the session should probably be aborted rather than left dangling.

PR #4367, opened 2026-09-22
Release closed windows held by pending interceptors

domenic:
Allow closed jsdom windows to be collected while an application-owned `requestInterceptor()` promise remains pending.


domenic:
Hmm I somehow didn't realize this overlapped with #4324. Probably shouldn't merge it until that is done.

